PR #20, opened 2016-12-17
Added Chinese Translation

felixqu:
Added simplified Chinese Translation


kspearrin:
Awesome. Mind doing the translations for the store COPY and CAPTIONS as well? https://github.com/bitwarden/mobile/tree/master/store

PR #30, opened 2016-12-30
Create french CAPTIONS.md

Primokorn:


kspearrin:
Thanks for this. Not sure what happened, but you have two PRs open for COPY and CAPTIONS each. Which should I accept? Are they any different?

King-Tut-Tut:
@kspearrin They're for google and apple, aren't they? So they are all valid.

kspearrin:
Ah yes, forgot.

PR #39, opened 2017-02-11
Swedish translation additional strings update

King-Tut-Tut:


kspearrin:
Can you update the COPY.ms file as well with the new short description we discussed?

PR #55, opened 2017-03-30
Fixes #8 Disable Google Analytics

### LINKED ISSUE #8: Disable Google Analytics Setting
We use google analytics to to better learn how the app is being used by users so that improvements can be made. Some users to not want to be tracked in this way. Add an option in settings to allow a user to opt-out of of google analytics.

Shan11812:
fixes #8

### DIFF
--- a/src/App/Pages/Settings/SettingsPage.cs
+++ b/src/App/Pages/Settings/SettingsPage.cs
@@ -40,6 +40,7 @@ namespace Bit.App.Pages
         private ExtendedTextCell TwoStepCell { get; set; }
         private ExtendedTextCell ChangeMasterPasswordCell { get; set; }
         private ExtendedTextCell ChangeEmailCell { get; set; }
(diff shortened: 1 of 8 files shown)


src/App/Resources/AppResources.resx
```diff
@@ -302,6 +302,9 @@
     <value>Ok</value>
     <comment>Acknowledgement.</comment>
   </data>
+  <data name="OptOutOfGA" xml:space="preserve">
```
kspearrin: Change the key to `DisableGA`.

src/App/Pages/Settings/SettingsPage.cs
```diff
@@ -140,6 +147,7 @@ private void Init()
             var otherSection = new TableSection(AppResources.Other)
             {
                 AboutCell,
+                AnalyticsCell,
```
kspearrin: You can make it first. Before `AboutCell`.

PR #74, opened 2017-05-15
Correct a typo

Igetin:


src/App/Resources/AppResources.fi.resx
```diff
@@ -792,4 +792,4 @@
   <data name="Beta" xml:space="preserve">
     <value>Beeta</value>
   </data>
-</root>
\ No newline at end of file
+</root>
```
doanbahai: +</root>

src/App/Resources/AppResources.fi.resx
```diff
@@ -792,4 +792,4 @@
   <data name="Beta" xml:space="preserve">
     <value>Beeta</value>
   </data>
-</root>
```
doanbahai: +< root>

doanbahai: +< root>

PR #168, opened 2017-11-12
TokenService.cs: Check if key exists before deleting it

alistair23:
To avoid errors in Task<ApiResult<TokenResponse>> when logging in on
UWP apps ensure that we check that they key exists before we delete the
2FA key token.


src/App/Services/TokenService.cs
```diff
@@ -152,7 +152,10 @@ public void SetTwoFactorToken(string email, string token)
             }
             else
             {
-                _secureStorage.Delete(key);
+                if(_secureStorage.Contains(key))
```
kspearrin: This should be combined with the else above as `else if`

alistair23:
I fixed your comments in V2.

PR #169, opened 2017-11-12
UWP/Assets: Update to use the Bitwarden logo

alistair23:
This updates the existing UWP images to use the Bitwarden logo. This isn't perfect (the image backgrounds aren't transparent and they should probably not take up the entire frame) but it is better then nothing.


alistair23:
Updated with transparent colours and reduced the size of the logos a little bit.

src/UWP/Package.appxmanifest
```diff
@@ -3,8 +3,8 @@
   <Identity Name="fb84989a-4cf4-440f-a76d-ef3506d9c9ba" Publisher="CN=Hicham" Version="1.0.0.0" />
   <mp:PhoneIdentity PhoneProductId="fb84989a-4cf4-440f-a76d-ef3506d9c9ba" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
   <Properties>
-    <DisplayName>UWP</DisplayName>
-    <PublisherDisplayName>Hicham</PublisherDisplayName>
+    <DisplayName>Bitwarden</DisplayName>
+    <PublisherDisplayName>Alistair</PublisherDisplayName>
```
kspearrin: Please change this to `8bit Solutions LLC`, which is what the app would be published under.

alistair23:
Fixed!

PR #188, opened 2017-11-26
Updates images for the UWP login page

alistair23:
Update the login and cancel images for the UWP login page.

### DIFF
--- a/src/App/Pages/EnvironmentPage.cs
+++ b/src/App/Pages/EnvironmentPage.cs
@@ -111,7 +111,7 @@ namespace Bit.App.Pages
                 Content = StackLayout
             };
 
-            var toolbarItem = new ToolbarItem(AppResources.Save, null, async () => await SaveAsync(),
+            var toolbarItem = new ToolbarItem(AppResources.Save, "folder.png", async () => await SaveAsync(),
                 ToolbarItemOrder.Default, 0);
 
             if(Device.RuntimePlatform == Device.iOS || Device.RuntimePlatform == Device.Windows)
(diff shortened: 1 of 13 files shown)


kspearrin:
I am a bit confused by this PR. Why was this needed? This seems to break the existing iOS layout... ?

alistair23:
On UWP apps the text isn't shown on the title bar unless you expand it. That means you only see the pictures. Without anything set it isn't clear there are buttons.

I didn't realise it'd break iOS. Is setting the image or is changing the cancel button layout the commit that breaks it?

alistair23:
This is what it looks like before this PR
nologo-collapsed
and this is after this PR
logo-collapsed

alistair23:
and this is after clicking the ...
logo-expanded

kumarharsh:
Is there a timeline wrt when it'll be released to the store?

alistair23:
I just updated the pull request for all the buttons that I could see.

This greatly improves the toolbar with UWP. I didn't test the other OSes though (I don't have a nice way to). If this breaks others we can add Windows if statements around them.

kspearrin:
Unfortunately this adds buttons everywhere for the other OSes as well, which we do not want. The best way to handle this would be to create a custom Toolbar control that applies the if condition to add or remove the icons based on OS.

alistair23:
Yep, I had an idea for this.

I think I'll add something to AppResources with device specific strings. So there can be AppResources.LogInImage which is NULL for all platforms besides UWP where it's a string.

I'll close this PR and open up a new one when I implement it.

kspearrin:
AppResources is handled in the common `App` project, so that will not work. Just create a subclassed ToolbarItem in which you can handle the logic of show/hide image basic on OS.

alistair23:
Ok, I'll do that then.

alistair23:
Actually, does it make more sense to have helper functions like Helpers.GetEmptyTableSectionTitle() that handles the image types?

That way there can be one Helpers.LoginImage() function which can be called for all of them.

PR #193, opened 2017-12-02
Login page fixes

alistair23:
This series provides a lot of fixes for Windows 10 and Windows Phone 10 UWP platforms.

This sets the toolbar options to have images (otherwise they look blank on the default view). See https://github.com/bitwarden/mobile/pull/188 for what the difference is.

This also fixes up the layout to correctly separate the login boxes (they used to be touching) and ensures that the reset password link can be seen by the user.

It fixes up some missing images.

Finally it allows the back button to appear when running on Windows 10. Previously it would not appear and there would be no obvious way to progress backwards. This allows us to remove the cancel button.

I'm getting the hang of this now, so if this is merged I can keep working on the other pages.

This is what it looks like now:
Phone
windowsphone-uwp-login
Desktop
windows10-uwp-login

### DIFF
--- a/src/App/Pages/LoginPage.cs
+++ b/src/App/Pages/LoginPage.cs
@@ -48,9 +48,9 @@ namespace Bit.App.Pages
                 Windows: new Thickness(10, 8));
 
             PasswordCell = new FormEntryCell(AppResources.MasterPassword, isPassword: true,
-                useLabelAsPlaceholder: true, imageSource: "lock", containerPadding: padding);
+                useLabelAsPlaceholder: true, imageSource: "lock.png", containerPadding: padding);
             EmailCell = new FormEntryCell(AppResources.EmailAddress, nextElement: PasswordCell.Entry,
-                entryKeyboard: Keyboard.Email, useLabelAsPlaceholder: true, imageSource: "envelope",
+                entryKeyboard: Keyboard.Email, useLabelAsPlaceholder: true, imageSource: "envelope.png",
                 containerPadding: padding);
 
             var lastLoginEmail = _settings.GetValueOrDefault(Constants.LastLoginEmail, string.Empty);
--- a/src/App/Pages/LoginPage.cs
+++ b/src/App/Pages/LoginPage.cs
@@ -45,7 +45,7 @@ namespace Bit.App.Pages
             var padding = Helpers.OnPlatform(
                 iOS: new Thickness(15, 20),
                 Android: new Thickness(15, 8),
-                WinPhone: new Thickness(15, 20));
+                Windows: new Thickness(10, 8));
 
             PasswordCell = new FormEntryCell(AppResources.MasterPassword, isPassword: true,
                 useLabelAsPlaceholder: true, imageSource: "lock", containerPadding: padding);
@@ -97,7 +97,7 @@ namespace Bit.App.Pages
             var layout = new StackLayout
             {
                 Children = { table, forgotPasswordButton },
-                Spacing = Helpers.OnPlatform(iOS: 0, Android: 10, WinPhone: 0)
+                Spacing = Helpers.OnPlatform(iOS: 0, Android: 10, Windows: 0)
             };
 
             var scrollView = new ScrollView { Content = layout };


src/App/Utilities/Extentions.cs
```diff
@@ -55,7 +55,14 @@ public static void FocusWithDelay(this View view, int delay = 1000, bool forceDe
...
         public static async Task PushForDeviceAsync(this INavigation navigation, Page page)
         {
-            await navigation.PushModalAsync(new ExtendedNavigationPage(page), true);
+            if (Device.RuntimePlatform == Device.Windows)
+            {
+                await navigation.PushAsync(page, true);
+            }
+            else
+            {
+                await navigation.PushModalAsync(new ExtendedNavigationPage(page), true);
```
kspearrin: Does UWP  not work with modals or something? I thought people were suppose to use the back button, like Android?

src/App/Utilities/Helpers.cs
```diff
@@ -73,6 +73,17 @@ public static string GetEmptyTableSectionTitle()
 
             return " ";
         }
+
+        public static string LoginImage()
```
kspearrin: I'd rather not have a helper for every type of image. Just create one method and pass the icon image string to it via parameter.

src/App/Pages/LoginPage.cs
```diff
@@ -112,14 +112,20 @@ private void Init()
...
-            var loginToolbarItem = new ToolbarItem(AppResources.LogIn, null, async () =>
+            var loginToolbarItem = new ToolbarItem(AppResources.LogIn, Helpers.LoginImage(), async () =>
             {
                 await LogIn();
             }, ToolbarItemOrder.Default, 0);
 
             ToolbarItems.Add(loginToolbarItem);
             Title = AppResources.Bitwarden;
-            Content = scrollView;
+            if (Device.RuntimePlatform == Device.Windows)
```
kspearrin: Why was this needed?

src/App/Pages/LoginPage.cs
```diff
@@ -67,7 +67,7 @@ private void Init()
 
             var table = new ExtendedTableView
             {
-                Intent = TableIntent.Settings,
+                Intent = TableIntent.Form,
```
kspearrin: Not sure how this will affect others.

alistair23: This isn't required, it just seemed to make more sense then settings.

src/App/Pages/LoginPage.cs
alistair23: The scroll ends up being really weird without this. Instead of everything fitting in a single page (which it could) it gets expanded to scroll. The register page is even worse. Just run the UWP app and you will see the issue

src/App/Utilities/Extentions.cs
alistair23: They do use the back button. It comes up in the desktop version when there is something to go back to (the settings app does this). It only comes up with PushAsync though

src/App/Utilities/Helpers.cs
alistair23: I'm not completely sure what you mean

src/App/Pages/LoginPage.cs
kspearrin: This should be reverted.

src/App/Pages/LoginPage.cs
kspearrin: We likely need to do some work on the custom renderers for TableView like we have for the other platforms.

src/App/Utilities/Extentions.cs
kspearrin: I think we need a close button then like iOS. There are many pages that use modal that will be affected.

src/App/Pages/LoginPage.cs
alistair23: Ok, I can remove the patch

src/App/Pages/LoginPage.cs
alistair23: Yeah, I think that is the problem. I looked at the vault last night and that display is all messed up.

Where is a good place to start?

src/App/Utilities/Extentions.cs
alistair23: Do you mean for other pages that pop up or for this page?

alistair23:
Ok, I dropped it down to just two patches, that aren't controversial.

I have some work in progress on the other issues and setting the custom renderer. It's here, but I don't think it's fully working yet: https://github.com/alistair23/mobile/tree/alistair/uwp-fixes

src/App/Utilities/Extentions.cs
alistair23: Looking at the Xamarin forms documentation this shouldn't be a modal push. Modals are for things that can't be navigated away from (hence the no back button). See here: https://developer.xamarin.com/guides/xamarin-forms/application-fundamentals/navigation/modal/

Would it be better to swap to only using modal sometimes? Either add an extra argument to specify non-modal or create a new function?

PR #200, opened 2017-12-07
App: LoginPage.cs: Use a logo for the login button

alistair23:
This adds a helper function which we can use to set logos for Windows
platforms to ensure the user can see the UWP toolbar actions. This won't
have an effect on non-Windows platforms.

### DIFF
--- a/src/App/Utilities/Helpers.cs
+++ b/src/App/Utilities/Helpers.cs
@@ -73,6 +73,17 @@ namespace Bit.App.Utilities
 
             return " ";
         }
+
+        public static string ToolbarImage(string image)
+        {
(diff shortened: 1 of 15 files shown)


src/App/Utilities/Helpers.cs
```diff
@@ -73,6 +73,17 @@ public static string GetEmptyTableSectionTitle()
 
             return " ";
         }
+
+        public static string TaskBarImage(string image)
```
kspearrin: Wouldn't "ToolbarImage" be more appropriate?

alistair23: Fixed.

alistair23:
Updated with more icons

alistair23:
Any thoughts on this?

PR #205, opened 2017-12-13
Improve the toast notifcations on Windows

alistair23:


src/UWP/UWP.csproj
```diff
@@ -145,7 +145,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Acr.UserDialogs">
-      <Version>6.3.10</Version>
+      <Version>6.5.1</Version>
```
kspearrin: We can't only update UWP here. We should update all projects. I'll look into that and report back

alistair23: Ok, I don't actually need the update for UWP to work, so I'll update this PR dropping this patch.

kspearrin:
We should probably shift this to the IDeviceAction service on each platform. I'd like to eventually get rid of UserDialogs all together so this is probably a good first step.

alistair23:
Ok. I'm not completely sure what you are looking for with that, so I might leave that up to you. I also have no way to test the other devices. Maybe I can revisit this in the future

PR #206, opened 2017-12-13
Pages: Settings: Disable fingerprint on Windows

alistair23:
Looking at the documentation here:
https://github.com/smstuebe/xamarin-fingerprint/blob/master/README.md
the GetAvailabilityAsync() and IsAvailableAsync() don't work on Windows.
To avoid always seeing the fingerprint option just diable it on Windows
as at the moment it's untested.

### DIFF
--- a/src/App/Pages/Settings/SettingsPage.cs
+++ b/src/App/Pages/Settings/SettingsPage.cs
@@ -91,7 +91,8 @@ namespace Bit.App.Pages
                 TwoStepCell
             };
 
-            if((await _fingerprint.GetAvailabilityAsync()) == FingerprintAvailability.Available)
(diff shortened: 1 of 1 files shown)


kspearrin:
Where on the Fingerprint plugin page do you see that it doesn't work with Windows?

alistair23:
Ah! So if I turn fingerprint login on then I can use Windows Hello to login. On my surface it asks for my account pin and on my phone it uses the Iris scanner.

So we just need to change some of the pharses to say "Windows Hello" instead of fingerprint.

alistair23:
How do you generate all of the translations? Does something like "Windows Hello" (https://support.microsoft.com/en-us/help/17215/windows-10-what-is-hello) need a translation?

alistair23:
Ok, that looks to be pretty easy. Good forward thinking there

PR #210, opened 2017-12-14
AppResources: Rename the English App resources file to use en

alistair23:
If the English app resource files aren't using the standard language
format we see GetString() returning NULL on UWP. To fix this rename the
files to include the language identifier.

More information here:
https://support.microsoft.com/en-us/help/839861/a-system-resources-missingmanifestresourceexception-exception-occurs-w


kspearrin:
I'm worried about how this is going to affect our translation platform, crowdin, since we are naming the source file that all translations are based on. I opened a support ticket with them to find out what affect this may have. Waiting for their reply.

kspearrin:
I fixed this with https://github.com/bitwarden/mobile/commit/c180422e8bf6d00b0f21459330de0692112c0bef#diff-720b57400b4182024f69ec43d93e33fbR7

No need for the file rename.

alistair23:
Cool, thanks for fixing that

PR #212, opened 2017-12-15
Add support for Windows Hello

alistair23:
Windows Hello is currently supported in the UWP app, but all of the
strings just point to using fingerprint. Windows Hello instead will
adjust based on what the user has avaliable and registered with the
Windows OS. To reflect that it isn't just fingerprints update the
strings when on UWP.

### DIFF
--- a/src/App/Pages/Lock/LockFingerprintPage.cs
+++ b/src/App/Pages/Lock/LockFingerprintPage.cs
@@ -7,6 +7,7 @@ using XLabs.Ioc;
 using Plugin.Fingerprint.Abstractions;
 using Plugin.Settings.Abstractions;
 using Bit.App.Abstractions;
+using Bit.App.Utilities;
 
 namespace Bit.App.Pages
 {
@@ -32,9 +33,23 @@ namespace Bit.App.Pages
 
         public void Init()
         {
+            var biometricIcon = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
+                        Android: "fingerprint.png",
+                        Windows: "smile.png");
+            var biometricText = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? AppResources.UseFaceIDToUnlock : AppResources.UseFingerprintToUnlock,
+                        Android: AppResources.UseFingerprintToUnlock,
+                        Windows: AppResources.UseWindowsHelloToUnlock);
+            var biometricTitle = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? AppResources.VerifyFaceID : AppResources.VerifyFingerprint,
+                        Android: AppResources.VerifyFingerprint,
+                        Windows: AppResources.VerifyWindowsHello);
+
+
             var fingerprintIcon = new ExtendedButton
             {
-                Image = _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
+                Image = biometricIcon,
                 BackgroundColor = Color.Transparent,
                 Command = new Command(async () => await CheckFingerprintAsync()),
                 VerticalOptions = LayoutOptions.CenterAndExpand,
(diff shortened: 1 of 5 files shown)


alistair23:
I know you auto generate new AppResources, but I don't know how to generate them. Do you get them from CrowdIn or do you generate them in VS?
Would you mind generating two for Windows Hello?

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
 
         public void Init()
         {
+            var image_helper = Helpers.OnPlatform(
```
kspearrin: This variable should use `camelCase` and also be named something more appropriate, such as `biometricIcon`.

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
...
+                        iOS: _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
+                        Android: _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
```
kspearrin: Android does not have faceid support, so just leave this as `fingerprint.png` only.

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
...
+                        Windows: "smile.png");
+            var text_helper = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? AppResources.UseFaceIDToUnlock : AppResources.UseFingerprintToUnlock,
+                        Android: _deviceInfoService.HasFaceIdSupport ? AppResources.UseFaceIDToUnlock : AppResources.UseFingerprintToUnlock,
+                        Windows: "Unlock with Windows Hello");
+            var title_helper = Helpers.OnPlatform(
```
kspearrin: This variable should use `camelCase` and also be named something more appropriate, such as `biometricTitle`.

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
 
         public void Init()
         {
+            var image_helper = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
+                        Android: _deviceInfoService.HasFaceIdSupport ? "smile.png" : "fingerprint.png",
+                        Windows: "smile.png");
+            var text_helper = Helpers.OnPlatform(
```
kspearrin: This variable should use `camelCase` and also be named something more appropriate, such as `biometricText`.

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
...
+                        iOS: _deviceInfoService.HasFaceIdSupport ? AppResources.UseFaceIDToUnlock : AppResources.UseFingerprintToUnlock,
+                        Android: _deviceInfoService.HasFaceIdSupport ? AppResources.UseFaceIDToUnlock : AppResources.UseFingerprintToUnlock,
```
kspearrin: same, no face id available on android

src/App/Pages/Lock/LockFingerprintPage.cs
```diff
@@ -32,9 +33,23 @@ public LockFingerprintPage(bool checkFingerprintImmediately)
...
+                        Windows: "Unlock with Windows Hello");
+            var title_helper = Helpers.OnPlatform(
+                        iOS: _deviceInfoService.HasFaceIdSupport ? AppResources.VerifyFaceID : AppResources.VerifyFingerprint,
+                        Android: _deviceInfoService.HasFaceIdSupport ? AppResources.VerifyFaceID : AppResources.VerifyFingerprint,
```
kspearrin: same, no ffaceid here. just fingerprint.

kspearrin:
If you open AppResources.resx in VS you should see the interface for adding new language strings.

alistair23:
Thanks for the feedback. I didn't realise that was all that is required to add strings. Updating now

src/App/Pages/Lock/LockFingerprintPage.cs
alistair23: Fixed all of these

kspearrin:
Is all biometic access on these UWP devices done with Windows Hello?

alistair23:
AFAIK it is. I don't have a fingerprint device to test, but from what I can tell Windows Hello is how you should be accessing all this stuff. It basically gets the user to verify they are the logged in user, using whatever Windows normally uses.

alistair23:
My Phone Iris scanner and Surface Pro both use Windows Hello

PR #215, opened 2017-12-17
Vault: Don't specify page options for ActivityLoading

alistair23:
When specifying page options the loader doesn't appear properly on UWP
applications. I also couldn't see the options documented here:
https://developer.xamarin.com/api/type/Xamarin.Forms.ActivityIndicator/

### DIFF
--- a/src/App/Pages/Vault/VaultAutofillListCiphersPage.cs
+++ b/src/App/Pages/Vault/VaultAutofillListCiphersPage.cs
@@ -115,9 +115,7 @@ namespace Bit.App.Pages
 
             LoadingIndicator = new ActivityIndicator
             {
-                IsRunning = true,
-                VerticalOptions = LayoutOptions.CenterAndExpand,
-                HorizontalOptions = LayoutOptions.Center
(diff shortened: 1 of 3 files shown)


kspearrin:
We still need these options for the other platforms, where they do actually work.

alistair23:
Shouldn't it default to being in the center? Looking at the pictures on that page it looks like these options don't change the position.

If they are actually changing the position on the other platforms I can wrap the changes in a conditional. Would you mind checking though?

kspearrin:
I'm not sure if that is the default or not. I would have to run on those platforms and check. The screenshots on the dev pages show that, but I'm not sure what settings were actually used in those.

kspearrin:
Seems to work the same on Android.

kspearrin:
Nevermind. Seems it broke after merging this. I fixed in https://github.com/bitwarden/mobile/commit/6fe5e89ecc5961cd8d3a943d9e7238dcd4b4d949

alistair23:
Thanks for fixing that. Looks good to me.

PR #240, opened 2018-01-04
App: Controls: Improve the vault view images

alistair23:
Make the buttons smaller and ensure the background is transparent.

The spacing between the buttons is still too large, but I'm not sure how to fix that.


src/App/Controls/LabeledValueCell.cs
```diff
@@ -114,6 +114,23 @@ public class LabeledValueCell : ExtendedViewCell
...
             }
+            else if (Device.RuntimePlatform == Device.UWP)
+            {
+                buttonStackLayout.Spacing = 0;
+
+                if (Button1 != null)
+                {
+                    Button1.BackgroundColor = Color.Transparent;
+                    Button1.BorderWidth = 0;
+                    Button1.Scale = 0.5;
```
kspearrin: I don't think we want to scale the button. we just need to use the proper sized images as discussed earlier

alistair23: I was a little worried how that would look on high res screens, as UWP doesn't have dynamic image selection like Android and iOS (at least not with Xamarin), but I'm not sure how well scale will work either on high res screens.

alistair23:
Fixed!

src/UWP.Images/UWP.Images.projitems
```diff
@@ -9,8 +9,6 @@
     <Import_RootNamespace>UWP.Images</Import_RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Content Include="$(MSBuildThisFileDirectory)android.png" />
```
kspearrin: I believe there are cases when these could be invoked still

alistair23: Shouldn't it have given an error during build then?

kspearrin: It would be a runtime error if the image was trying to be displayed and the file does not exist.

alistair23: Ok, I can add it back. Just the Android one?

kspearrin: Android and apple are both used in cases when the URI uses their app's scheme

alistair23: Ok, I'll add those back in.

alistair23:
Updated!

alistair23:
Updated again with correct spacing

PR #243, opened 2018-01-05
App: Pages: Add a null check in OnSelectTemplate()

alistair23:
To avoid accessing a null poiter add a null check in OnSelectTemplate().

### DIFF
--- a/src/App/Pages/Vault/VaultListGroupingsPage.cs
+++ b/src/App/Pages/Vault/VaultListGroupingsPage.cs
@@ -350,6 +350,10 @@ namespace Bit.App.Pages
 
             protected override DataTemplate OnSelectTemplate(object item, BindableObject container)
             {
+                if(item == null)
+                {
+                    return null;
+                }
                 return ((GroupingOrCipher)item).Cipher == null ? GroupingTemplate : CipherTemplate;
             }
         }


src/App/Pages/Vault/VaultListGroupingsPage.cs
```diff
@@ -350,6 +350,10 @@ public GroupingOrCipherDataTemplateSelector(VaultListGroupingsPage page)
 
             protected override DataTemplate OnSelectTemplate(object item, BindableObject container)
             {
+                if (item == null)
```
kspearrin: How could this occur?

alistair23: I see it with the UWP app, which has bugs with ListView and it ends up getting here. 

I also see it done here: https://blog.xamarin.com/customizing-list-view-cells-xamarin-forms-datatemplateselector/

kspearrin: Weird. So what happens to the cell that ends up not getting a template?

alistair23: One of the UWP bugs results in your clicks being offset. So when you click on an element it triggers but not on an element (null). So nothing happens instead of a crash.

alistair23:
Updated with correct spacing as well

PR #248, opened 2018-01-09
Add YubiKey keyboard support

alistair23:


src/App/Pages/LoginTwoFactorPage.cs
```diff
@@ -232,17 +232,23 @@ private void Init()
...
-                    Source = "yubikey",
+                    Source = "yubikey.png",
                     VerticalOptions = LayoutOptions.Start,
                     HorizontalOptions = LayoutOptions.Center,
                     WidthRequest = 266,
                     HeightRequest = 160,
                     Margin = new Thickness(0, 0, 0, 25)
                 };
 
+                TokenCell = new FormEntryCell(AppResources.YubiKeyTextEnter, useLabelAsPlaceholder: true,
```
kspearrin: Other platforms do not need this cell since they listen to NFC.

Also, no need for the placeholder string. UI should follow how it is in the browser extension.

image

src/App/Pages/LoginTwoFactorPage.cs
```diff
@@ -254,6 +260,7 @@ private void Init()
 
                 table.WrappingStackLayout = () => layout;
                 scrollView.Content = layout;
+                ToolbarItems.Add(continueToolbarItem);
```
kspearrin: Same

src/App/Pages/LoginTwoFactorPage.cs
```diff
@@ -453,7 +460,7 @@ protected override void OnDisappearing()
                             break;
                         case TwoFactorProviderType.YubiKey:
                             var nfcKey = p.Value.ContainsKey("Nfc") && (bool)p.Value["Nfc"];
-                            if(!_deviceInfoService.NfcEnabled || !nfcKey)
+                            if((!_deviceInfoService.NfcEnabled || !nfcKey) && (Device.RuntimePlatform != Device.UWP))
```
kspearrin: You dont need parens around your UWP check. Also dont you need some kind of "Has USB Port" check?

src/App/Pages/LoginTwoFactorPage.cs
alistair23: It is also possible to use the USB entry on Android, if you don't have NFC support, which is why I was adding the option. I will update the interface though.

alistair23: https://www.yubico.com/support/knowledge-base/categories/articles/can-yubikey-4c-plugged-directly-android-phones-tablets-usb-c-ports/

src/App/Pages/LoginTwoFactorPage.cs
alistair23: GitHub wants me to comment here as well. Same as above, USB entry does work on Android so I think it makes sense to leave it in. If you really want I can remove it on Android though.

alistair23:
I updated it, I tried to detect USB support, but it seems really hard to do. USB is a mess. This is my dev branch: https://github.com/alistair23/bytewarden-windows/tree/alistair/yubikey-wip

kspearrin:
Yes, but UWP is for mobile devices too. Do all windows mobile devices support USB?

alistair23:
This app doesn't support Windows for phones, so that isn't a huge issue.

I was thinking about this more though. So this app does support Xbox and maybe some future phone thing that Microsoft does. So there is merit in checking for USB. In saying that it is really hard to do. For example on my Surface Pro 2 the built-in camera is connected via USB, so the check will be a false positive there. I don't think it's unreasonable to assume that a device has a USB port, that isn't too unrealistic. If the user can't use it then they can change the second factor option.

alistair23:
It also occurred to me that the Xbox One does has USB ports and I'm pretty sure it supports USB keyboards. So there really isn't much that doesn't support USB input.

kspearrin:
Wait, so UWP apps don't run on any type of mobile phones?

alistair23:
There is no Windows Phone OS that runs with .NET standard 2.0. The latest and last Windows Phone OS is the fall creators update, which doesn't run .NET standard 2.0. This is why I had to create a fork, because Microsoft dropped the platform.

alistair23:
Unless some unannounced OS comes up this can't support a Windows Phone

kspearrin:
But aren't there mobile phones that run windows 10? If not, I am struggling to understand why we should even support UWP through the mobile project going forward. We will have a desktop app in a month or two.

alistair23:
There are no mobile phones that run a Windows OS that bitwarden can run on.  UWP is good for Windows S and Xbox as well as regular Windows.

alistair23:
My fork can run on Windows Phones though

src/App/Resources/AppResources.resx
```diff
@@ -883,7 +883,7 @@
     <comment>For 2FA</comment>
   </data>
   <data name="YubiKeyInstruction" xml:space="preserve">
-    <value>Hold your YubiKey NEO against the back of the device to continue.</value>
+    <value>Hold your YubiKey NEO against the back of the device to continue. You can also use the YubiKey keyboard functionality to enter the code.</value>
```
kspearrin: Should read: 

> To continue, hold your YubiKey NEO against the back of the device or insert your YubiKey into your device's USB port, then touch its button.

src/App/Pages/LoginTwoFactorPage.cs
```diff
@@ -232,17 +232,22 @@ private void Init()
...
-                    Source = "yubikey",
+                    Source = "yubikey.png",
                     VerticalOptions = LayoutOptions.Start,
                     HorizontalOptions = LayoutOptions.Center,
                     WidthRequest = 266,
                     HeightRequest = 160,
                     Margin = new Thickness(0, 0, 0, 25)
                 };
 
+                TokenCell = new FormEntryCell(AppResources.YubiKeyTextEnter, imageSource: "lock");
```
kspearrin: `AppResources.YubiKeyTextEnter` This text can go away to be consistent with the browser extension. It is blank.

kspearrin: Also, is this field auto-focused?

alistair23: Ok, fixed

src/App/Resources/AppResources.resx
alistair23: Updated

alistair23:
All updated again

kumarharsh:
Kinda bummed that phones won't be supported ☹ But it'd be a huge help if XBox does get the app - as inputting auto-generated passwords with a controller are a major PITA!

Also, just to be clear regarding Windows support:

1. The official Bitwarden app can't run on Windows 10 Mobile (which is stuck on Creators Update more or less).
2. There is a fork maintained by @alistair23 which *does* run on Windows 10 Mobile.
3. The official Bitwarden app can run on Windows 10 S and XBox.
4. The official Bitwarden app *will* run on any future phone-like devices which run Windows 10/Andromeda.

Am I right in these assumptions?

alistair23:
1. Correct
2. Correct. This does run on Xbox, although the Xbox has no copy/paste support so it isn't super useful.
3. At the moment yes.
4. In theory yes

PR #287, opened 2018-02-27
App/Pages/Vault: Maintain Vault List Position

alistair23:
When popping back to the vault list after selecting a cipher the
ListView resets the view back to the top. Let's ensure we maintain our
position in the list after the pop.

### DIFF
--- a/src/App/Pages/Vault/VaultListGroupingsPage.cs
+++ b/src/App/Pages/Vault/VaultListGroupingsPage.cs
@@ -57,6 +57,7 @@ namespace Bit.App.Pages
         private SearchToolBarItem SearchItem { get; set; }
         public ContentView ContentView { get; set; }
         public Fab Fab { get; set; }
+        private static Object lastTappedItem { get; set; }
 
         private void Init()
         {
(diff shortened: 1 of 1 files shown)


src/App/Pages/Vault/VaultListGroupingsPage.cs
```diff
@@ -148,7 +149,14 @@ private void Init()
...
             SearchItem?.InitEvents();
 
-            _filterResultsCancellationTokenSource = FetchAndLoadVault();
+            if(lastTappedItem != null)
+            {
+                ListView.ScrollTo(lastTappedItem, ScrollToPosition.Center, false);
+            }
+            else
+            {
+                _filterResultsCancellationTokenSource = FetchAndLoadVault();
```
kspearrin: Won't this now cause the list to never update when changes are made on the add/edit page and you go back?

alistair23: Ah, I didn't think of editing it...

I couldn't figure out how to get the ScrollTo() call to work in the FetchAndLoadVault() function. Even if I call it when the content is set to the ListView the refresh still happens afterwards and jumps to the top.

I'll keep digging, somehow the ScrollTo() needs to be called after FetchAndLoadVault() returns.

kspearrin: This must only be an issue on UWP because the position is remembered correctly on ios and android. Maybe it's because we use modals on those and you arn't on UWP?

alistair23: Hmmm... But we need not modals to get the back button

PR #328, opened 2018-06-16
Card Security Code masked (bitwarden/desktop issue 78)

neilsb:
By default shows Card Security Code masked in the same way as Passwords. (bitwarden/desktop issue 78)

Only tested on Android (not iOS devices available) but when setting input field to password, the keyboard type of "Numeric" isn't honoured and it shows standard keyboard with numbers at the top.  Tested this is new project and the same thing happens.


neilsb:
Hmm - just spotted a mistake in that I think....

neilsb:
OK - Updated to use it's own Property for storing revealed/hidden Card Code.  Shouldn't have caused any problems as Login and Card Code should not appear on the same screen - but updated to be more _correct_.

PR #360, opened 2018-08-20
Use BaseUrl when opening web vault if defined

BestPig:
In the tab tools when you have a self hosted bitwarden environment and click on the web vault.
The application open the link "https://vault.bitwarden.com" instead of the defined BaseUrl.

This commit fix the link if a BaseUrl is defined.

### DIFF
--- a/src/App/Pages/Tools/ToolsPage.cs
+++ b/src/App/Pages/Tools/ToolsPage.cs
@@ -158,7 +158,15 @@ namespace Bit.App.Pages
         private void WebCell_Tapped(object sender, EventArgs e)
         {
             _googleAnalyticsService.TrackAppEvent("OpenedTool", "Web");
-            Device.OpenUri(new Uri("https://vault.bitwarden.com"));
+            var appSettings = Resolver.Resolve<IAppSettingsService>();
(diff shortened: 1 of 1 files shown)


src/App/Pages/Tools/ToolsPage.cs
```diff
@@ -158,7 +158,15 @@ private void ExtensionCell_Tapped(object sender, EventArgs e)
         private void WebCell_Tapped(object sender, EventArgs e)
         {
             _googleAnalyticsService.TrackAppEvent("OpenedTool", "Web");
-            Device.OpenUri(new Uri("https://vault.bitwarden.com"));
+            var appSettings = Resolver.Resolve<IAppSettingsService>();
+            if (!string.IsNullOrWhiteSpace(appSettings.BaseUrl))
+            {
+                Device.OpenUri(new Uri($"{appSettings.BaseUrl}"));
```
kspearrin: No need for the `$"{` . Just pass `appSettings.BaseUrl`

BestPig:
I removed it ;)

kspearrin:
Thanks.

Issue #346, opened 2018-07-26
There is a problem connecting to the server - Self-signed Cert - Android


Joshndroid:
So finally got myself up and running today.
Self-hosted windows docker on windows server 2016.
Utilised the self-signed cert option during install.

Have smtp settings finally sorted.
Can connect locally from pc web browser and using the firefox extention. All is good.

Install the android app from the store, configure it up the same as the firefox extension, cannot connect to the server.

Pulled the certificate.crt from bwdata from device placed on android sd card and add to certificates for both 'VPN and apps' and 'wifi' still no connection. I have checked permissions and granted the app everything (only storage and camera listed on android oreo 8.1

I'm at a loss at what could be causing this?

I've googled and read both here and reddit and can't seem to find an answer.

I have read that you may not have an answer, can i somehow force a new cert using let's encrypt? or am i going to have to hard delete bwdata and re-install it again?

Cheers.

kspearrin:
There was a bug with the generated self-signed cert that was fixed in the upcoming version here: https://github.com/bitwarden/core/pull/327

You could try generating a new self-signed cert of your own, replacing what was generated in ./bwdata/ssl

Joshndroid:
Thank you for your reply!

So i followed your advice... 
Using ubuntu machine generated a new set of certificat.crt and private.key... 

i transferred them into the certificate location in bwdata/ssl/self/bitwardendoman/

The browser recognized a new certificate not yet trusted (so i know it worked).

I added the new certificate.crt to /sdcard of android device
navigated to security -> encryption&credentials -> install from storage
From there found certificate and added it twice (once for VPN & apps & another for Wifi - just in case).

The app configured with my local domain and port...... No dice. still no connection to the server
Navigate using firefox browser to the local domain, instant connection && no requirement to trust a new certificate (as it would appear that the device has now trusted it).

Still not sure what the heck is going on.
Any ideas for me to try literally everything else is working just fine just cant connect from the most valuable aspect of the whole thing... the android app :(

kspearrin:
Android app has some additional cert trust requirements. Are you able to check your website here:
https://www.digicert.com/help/

Joshndroid:
I have tried again remade my entire bitwarden instance with a new installation ID
Browsers & browser extensions work.

Android app does not, even when trusted.

I am only running the instance in a local IP address.

There needs to be a check box or an option to decrease the harsh certificate requirements for the android app specifically. I see others are having a similar issue that seem to be similar #209

[...]

PR #420, opened 2018-12-01
Honor user CA store when targeting Android Nougat SDK and later

dNetGuru:
See: https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html

Fixes: #346

### DIFF
--- a/src/Android/Android.csproj
+++ b/src/Android/Android.csproj
@@ -1021,5 +1021,8 @@
   <ItemGroup>
     <AndroidResource Include="Resources\drawable-xxxhdpi\upload2.png" />
   </ItemGroup>
+  <ItemGroup>
+    <AndroidResource Include="Resources\xml\network_security_config.xml" />
+  </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
 </Project>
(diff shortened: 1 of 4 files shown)


kspearrin:
According to the examples in that blog, it seems we probably want to exclude bitwarden.com endpoints for this, since those should never require custom CA trust.

dNetGuru:
Yes, I didn't add that since I don't have a complete list of all the domains this app contacts.

If everything is hosted under `*.bitwarden.com` then I'll just add an exception `bitwarden.com` and all subdomains and push it in a bit.

kspearrin:
Just add api.bitwarden.com and identity.bitwarden.com. I can add any others.

dNetGuru:
It seems like I can specify `includeSubdomains="true"` to include all sub-domains which would be a better approach. If you're using self-signed CAs for any other subdomain, you could explicitly exclude it by adding another `domain-config` entry.

I was reading further into the docs (here: https://developer.android.com/training/articles/security-config) and I found out that you can opt out of all clear-text traffic which is always a good idea, I also added to the commit.

Let me know if you want me to do the following as well:
There seems to be an option to enable certificate pinning (which could protect the users even in a case of a CA compromise) in the `network_security_config` area but I noticed that you're using Cloudflare SSL which you wouldn't be managing and I'm not sure how often they rotate/regenerate them.
The next best thing would be to specifically specify `COMODO ECC Certification Authority`'s certificate as the only trusted CA for `*.bitwarden.com` since they issue all the Cloudflare certificates and this way the users are protected from a possible compromise of all the other CAs.

kspearrin:
No need for pinning. Thanks.

PR #424, opened 2018-12-09
Colorized passwords

aksdb:
Same pattern as the jslib/desktop implementation.
Since I didn't see any theming (so far), I just used "default" colors.

### DIFF
--- a/src/App/Models/Page/PasswordGeneratorPageModel.cs
+++ b/src/App/Models/Page/PasswordGeneratorPageModel.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
+using Bit.App.Utilities;
+using Xamarin.Forms;
 
 namespace Bit.App.Models.Page
 {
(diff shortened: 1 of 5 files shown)


kspearrin:
Thanks for the PR. It seems that (at least on my test device here) the formatted password now makes the password generation page show a noticeable lag when generating new passwords. For example, try moving the length slider around to generate a lot of passwords.

aksdb:
I will look into this. At least in the android emulator it didn't seem slower than before (the emulator itself is probably not a very good measure for performance, though :-/ ), but I will do a few more before-after-comparisons and play around with other formatting options (if there are any in Xamarin.Forms).

kspearrin:
Looking at the code more, I am not really sure where any performance gains could be made.

FYI: I am testing on a Nexus 5X.

kspearrin:
Reverted on password generator page for now. https://github.com/bitwarden/mobile/commit/ee9ecfbee9c1a03a4a35f2da9a46d96fa23bf6af

PR #449, opened 2019-01-22
Handle exception when adding attachment that does not exist on file system.

ShirokaiLon:
Bitwarden crashes when trying to add a file that does not exist on the filesystem as an attachment.

The Android 'Documents' app seems to keep a cache of downloaded files despite those files having been deleted off the file system. This cases Bitwarden to crash when the user selects them.

I figure that presenting some error to the user when this happens makes sense, but beyond the generic "An error has occurred" message, I couldn't find a suitable resource.

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -203,12 +203,19 @@ namespace Bit.Android
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


kspearrin:
Thanks

PR #576, opened 2019-08-11
Add autofill support for Bromite

xcffl:
Bromite is a Chromium fork plus ad blocking and enhanced privacy
https://www.bromite.org/


kspearrin:
bromite is already in this list.

xcffl:
My mistake. I just found that Bitwarden can't recognize addresses from Bromite correctly in some cases and thought it's not in the list.

PR #577, opened 2019-08-12
fix: Remove doubled 'You you'

CanRau:
Not sure this is the right way and it seems I might've messed up line endings/breaks?
Anyway I thought the easiest to file that issue would be to include a possible fix via a PR.
Feel free to reject it. I thought it might easier to PR instead of using Crowdin in this case.

By the way, Thanks a lot for this amazing software, it's the most pleasing password manager I came across so far!!!


kspearrin:
Thanks for contributing @CanRau . Any localization updates should happen in Crowdin so that they can be synced back to the repo. See https://github.com/bitwarden/mobile/blob/master/CONTRIBUTING.md#localization-l10n

CanRau:
Okay then. I read the contributing note yet thought because it's spreading
across so many sites I thought Crowdin could be a hassle.
Yet I understand that it's important to follow the established route 🙏👍

Kyle Spearrin <notifications@github.com> schrieb am Mo., 12. Aug. 2019,
06:11:

PR #657, opened 2019-11-22
Switched Logout with Unlock

HLSiira:
Attempted to switch the logout button with the unlock button on the app to follow better UX flow. Hopefully will help prevent users (me) from clicking logout (The currently much bigger button below the pin entry) instead of clicing unlock (The currently smaller harder to reach button).

### DIFF
--- a/src/App/Pages/Accounts/LockPage.xaml
+++ b/src/App/Pages/Accounts/LockPage.xaml
@@ -20,7 +20,7 @@
     </ContentPage.Resources>
 
     <ContentPage.ToolbarItems>
-        <ToolbarItem Text="{u:I18n Unlock}" Clicked="Unlock_Clicked" />
+        <ToolbarItem Text="{u:I18n LogOut}" Clicked="LogOut_Clicked" />        
     </ContentPage.ToolbarItems>
 
     <ScrollView>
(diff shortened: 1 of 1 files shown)


kspearrin:
For the Bitwarden app, the primary action button for whatever is occurring on screen is always in the top right corner. I think in this case, maybe we just need to make the logout button less inviting to click.  Maybe remove its button background color so that it is just plain text?

HLSiira:
It makes sense for the save buttons to be at the top on the other screens because those save buttons affect the entire forms and more often than not, those forms extend past the screen height, so the save button needs to be visible at all times, and when editing a password, there is only one 'submit' button, which is the save button. There isn't a large inviting button at the bottom that says delete.

In the unlock screen however, it doesn't usually extend past the screen height, and so all controls can be on the same page. Just like paper, flow starts from the top and moves down, and naturally, (and in my opinion) more commonly, the submit button is the largest, bottom-right most button on any form/input.

This is only my opinion so hopefully I'm not coming off as rude or combative. I'm really tired of hitting logout instead of unlock because Bitwarden is the only app to have the continue/next screen button be hidden in a corner.

bassarf:
I agree that the unlock button should be more inviting than logout.
I also understand the primary action on the top thing.
The logout button could also be on the top, next to the unlock button.
Similar to the save and cancel arrangement you often see in apps.

@HLSiira
Although I hate comments in the tone of: "why don't you just do X or Y". Everyone has his own workflow.
But I believe you can just press the "enter" key on your keyboard to unlock after you entered the pin. Maybe this helps.

PR #662, opened 2019-11-30
WIP: Implement U2F TwoFactor Authorization (on Android)

mastacheata:
NOTE: This is a work in progress, just opened the pull request in order to properly share the idea and ask for help,

I started working on FIDO-U2F 2nd Factor stuff yesterday, but only got to a point where it will immediately return from the intent without even trying to ask for a token. Instead I get an ErrorResponse with ERROR_OTHER as the Response Code.

I've so far identified 2 possible causes, but am not sure about that as there is no more information in the error than "Other".

Cause 1 woud be that I have the appId wrong. I tried to get that from the challenge issued by the Bitwarden server, but that might be a mistake as a JSON URL is not actually an appId. Instead I should probably pick one of the IDs from that JSON file (in this case, the Android App identifier).
If that is the case, should I just hardcode the ID, grab the one from the remote JSON or is there a way for the Android part of the app to know it's own ID?

The other thing that might be wrong is how I deserialize and then convert to the requested byte array for the keys from a string. In the debugger the string representation of the keys list changes multiple times between the string from the JSON response, the byte array and the "RegisteredKeys" object.

In the end I somehow need to retry the login with the signed challenge / the response from the U2F key. I have no idea how to do that, All the other 2-factor challenges work without sending a challenge to the 3rd party and are either time or account based as it seems.

I guess I'll figure that out as soon as I got the correct signed challenge / the response from the U2F token.

### DIFF
--- a/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
+++ b/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
@@ -7,6 +7,7 @@ using Bit.Core.Abstractions;
 using Bit.Core.Enums;
 using Bit.Core.Utilities;
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Xamarin.Forms;
 
@@ -19,6 +20,7 @@ namespace Bit.App.Pages
         private readonly IPushNotificationService _pushNotificationService;
         private readonly IStorageService _storageService;
         private readonly ILockService _lockService;
+        private readonly ICipherService _cipherService;
         private readonly IDeviceActionService _deviceActionService;
         private readonly GroupingsPageViewModel _vm;
         private readonly string _pageName;
@@ -37,6 +39,7 @@ namespace Bit.App.Pages
             _pushNotificationService = ServiceContainer.Resolve<IPushNotificationService>("pushNotificationService");
             _storageService = ServiceContainer.Resolve<IStorageService>("storageService");
             _lockService = ServiceContainer.Resolve<ILockService>("lockService");
+            _cipherService = ServiceContainer.Resolve<ICipherService>("cipherService");
             _deviceActionService = ServiceContainer.Resolve<IDeviceActionService>("deviceActionService");
             _vm = BindingContext as GroupingsPageViewModel;
             _vm.Page = this;
@@ -97,7 +100,7 @@ namespace Bit.App.Pages
             var migratedFromV1 = await _storageService.GetAsync<bool?>(Constants.MigratedFromV1);
             await LoadOnAppearedAsync(_mainLayout, false, async () =>
             {
-                if(!_syncService.SyncInProgress)
+                if(!_syncService.SyncInProgress || (await _cipherService.GetAllAsync()).Any())
                 {
                     try
                     {


src/Android/MainActivity.cs
```diff
@@ -278,6 +297,40 @@ private void ListenYubiKey(bool listen)
             }
         }
 
+        private async Task<bool> ListenU2F(Dictionary<string, object> data)
+        {
+            var appId = Uri.Parse((string)data["appId"]);
+            var challenge = Encoding.ASCII.GetBytes((string) data["challenge"]);
```
kspearrin: I believe that challenge is a "url safe" base 64 encoded string. See here for a method that can decode that: https://github.com/bitwarden/mobile/blob/master/src/Core/Services/TokenService.cs#L203

Might want to mvoe that method into a static core helper.

src/Android/MainActivity.cs
```diff
@@ -278,6 +297,40 @@ private void ListenYubiKey(bool listen)
...
+            var keys = ((JArray) data["keys"]).ToObject<List<Dictionary<string, string>>>();
+            var registeredKeys = new List<RegisteredKey>();
+            foreach (var key in keys)
+            {
+                var keyHandle = new KeyHandle(
+                    Encoding.ASCII.GetBytes(key["keyHandle"]),
```
kspearrin: Might be the same, a "url safe" base 64 encoded string

src/Android/MainActivity.cs
mastacheata: Thanks for the hint, will try that in the next few days.

I'll clean that up once I got everything up and running. I just noticed that I'll need at least 2 entrypoints to the U2F functionality when I started and before that I just picked up where you left comments about implementing U2F sometime in the future.

mastacheata: Do you by any chance know how the U2F appId must be formatted? I fear that this is my actual problem here. The Bitwarden Server returns a link to the https://vault.bitwarden.com/app-id.json and from the documentation of U2F I read, that's not a valid app identifier. Instead it seems like this just tells me which appIds will be accepted by the server when I open that reference.

I think I might need to insert the Android App ID in the appId field, but I'm not sure how to obtain that yet.

kspearrin: The app id should be `https://vault.bitwarden.com/app-id.json`.

If that doesn't work, then maybe try `android:apk-key-hash:dUGFzUzf3lmHSLBDBIv+WaFyZMI`, which is defined in that JSON file for the Bitwarden android app.

src/Android/MainActivity.cs
nerrixDE: any updates?

mastacheata: No, I haven't found any documentation on how the Original Android API is supposed to work with other providers than Google itself. There's also no Open Source projects using either the Xamarin Library or the original Android Java classes to access third-party U2F/FIDO2 providers. I'm pretty much stuck not knowing what I might do wrong/what format the Android API expects.

nerrixDE: I am not really in this topic, but why doesn't https://github.com/android/security-samples/tree/master/Fido helps / is an option?

mastacheata:
Sadly that sample only covers how to get the authorization request from Google / for the Google account stored in the phone. The example doesn't show how to create that request because the Android API for the Google Account provides a ready-made object. It appears like noone has ever used the native U2F/FIDO APIs except for the Google account. There's lots of examples how to use U2F in the browser, even on mobile, but there's nothing about the native Android API.😟




Benedikt Bauer, M.Eng.







On Fri, Feb 21, 2020 at 6:44 PM +0100, "Nerrix" <notifications@github.com> wrote:












@nerrixDE commented on this pull request.





In src/Android/MainActivity.cs:
> @@ -278,6 +297,40 @@ private void ListenYubiKey(bool listen)
             }
         }
 
+        private async Task<bool> ListenU2F(Dictionary<string, object> data)
+        {
+            var appId = Uri.Parse((string)data["appId"]);
+            var challenge = Encoding.ASCII.GetBytes((string) data["challenge"]);
+            var keys = ((JArray) data["keys"]).ToObject<List<Dictionary<string, string>>>();
+            var registeredKeys = new List<RegisteredKey>();
+            foreach (var key in keys)
+            {
+                var keyHandle = new KeyHandle(
+                    Encoding.ASCII.GetBytes(key["keyHandle"]),



I am not really in this topic, but why doesn't https://github.com/android/security-samples/tree/master/Fido helps / is an option?

src/Android/MainActivity.cs
nerrixDE: Took a look at this? https://fidoalliance.org/specs/fido-uaf-v1.2-rd-20171128/fido-uaf-client-api-transport-v1.2-rd-20171128.html#org.fidoalliance.intent.fido_operation-intent

mKoonrad:
@mastacheata 1Password recently released their Android app (version 7.4) with security key support. It looks like they've found any solution.

mastacheata:
> @mastacheata 1Password recently released their Android app (version 7.4) with security key support. It looks like they've found any solution.

1password isn't open source and most likely not written in Xamarin.
The "easy" solution is to hand over the authentication to a web browser / web view. On Android webauthn is supported by Firefox and Chrome. That would even work on ios as long as the brave browser is installed.

nerrixDE:
Not having U2F support on android is a major issue for me and makes me really consider switching to 1password away from Bitwarden.
It is definitly possible to implement and well, it might require a bit more work than just implementing a library by adding 10 lines of code. 
But I really wish I don't have to wait for this until the next standard is released....

mastacheata:
I'm fine with the biometric authentication on my phone, but would definitely like to have U2F/FIDO2 as well, but I'm a web developer by trade and have close to zero knowledge of mobile development.
The Xamarin library is basically just a wrapper around the native Google Android APIs, but even those have barely any documentation.

Pretty much all the documentation you can find on both U2F and FIDO2 are strongly linked to Web Authentication through the browser. On Android you can easily get both to work in Chrome and Firefox and it even works on iOS if you use the Brave browser (no idea how they managed to include that feature as it's just like all other browsers only a wrapper around the included Safari on iOS).

If you find any documentation/examples about how to do FIDO2/U2F/Webauthn on Android in a native application and NOT using the Google account, please let me know and I'll try to pick up my work on this feature again. Otherwise, we'll have to wait and see if Kyle (the author/lead developer of Bitwarden) can find some time to look into that or find someone with more knowledge...

rdslw:
For all waiting: 1password currently supports 2FA on android.

mastacheata:
> 
> 
> For all waiting: 1password currently supports 2FA on android.

mKonrad already noticed that 3 posts above.
It's not gonna help with this issue, though, as 1password is not open source and we can't learn from how they did it. (And even if we could, it's probably a native Java Application and not taking the detour through Xamarin like Bitwarden does)

nerrixDE:
@Bitwarden-Staff How is the priority? You really should prioritize it, as consumers **want** it. Some consider switching.

schlomie:
I have not dug into the code, so I have absolutely no idea if this would prove helpful, but the OnlyKey guys have done some work with communicating with their usb token over U2F on android.

https://github.com/trustcrypto/Android-OnlyKey-U2F

arianvp:
Is it really important for it to be U2f?

FIDO2 is backwards-compatible wit U2F and the android APIs for FIDO2 Closely model the welldocumented `webauthn` spec for browsers. I'm currently experimenting with a minimal app to get things working on kotlin.

However this might require changes on bitwarden server. Is that realistic?

arianvp:
> I started working on FIDO-U2F 2nd Factor stuff yesterday, but only got to a point where it will immediately return from the intent without even trying to ask for a token

@mastacheata was this in the emulator by any chance? I have the same issue when trying out  the FIDO2 API.  I think FIDO2 stuff in the emulator is just bugged

mastacheata:
> > I started working on FIDO-U2F 2nd Factor stuff yesterday, but only got to a point where it will immediately return from the intent without even trying to ask for a token
> 
> @mastacheata was this in the emulator by any chance? I have the same issue when trying out the FIDO2 API using Google's own examples. I think FIDO2 stuff in the emulator is just bugged

Sadly, no. This was debugging on a physical device. (Pixel 3 and HTC10)

Maybe it's an option to ditch the native approach and instead use a WebView? Both Chrome and Firefox (only the stable version, not the Preview yet) work perfectly fine with WebAuthn / FIDO2 / FIDO-U2F.

> FIDO2 is backwards-compatible wit U2F
That's optional and up to the server to decide that.

> Is it really important for it to be U2f?
There are by far more U2F tokens than there are FIDO2 ones around. (At least if you exclude the Android software-token where your device can act as security token for your PC or another phone)
That's why I think U2F-support is important.

Whether you do that via the Webauthn/FIDO2 API or the U2F API doesn't matter, I think.

nerrixDE (approved):
any updates?

Cryonis:
Any update on this? This would be a major feature that would help convert a few to Bitwarden Premium

kibartas:
Updates? Is there still the same issue with this?

mastacheata:
Yeah, I haven't done any work on this as you can see and only opened this PR to look for support from other developers. Sadly Bitwarden doesn't seem to attract many developers and is pretty much developed solely by the company itself.

kibartas:
That's unfortunate then. It'd be interesting to have a crack at it, still. Some time has passed maybe it'd be more clear how to implement FIDO U2F now than back when you did it

PR #680, opened 2019-12-26
Add Autofill support for Tor browser

proletarius101:


kspearrin:
Dow e also need support in autofill accessibility service?

PR #686, opened 2020-01-01
Change password generator to use ColoredPassword

ShirokaiLon:
Passwords displayed in the password generator are not colour coded like they are on other platforms. This pull request changes the password generator password to be colour coded.

https://community.bitwarden.com/t/color-coded-password-not-working-on-android/9594

### DIFF
--- a/src/App/Utilities/PasswordFormatter.cs
+++ b/src/App/Utilities/PasswordFormatter.cs
@@ -25,10 +25,14 @@ namespace Bit.App.Utilities
         {
             if(password == null)
             {
-                return "";
+                return string.Empty;
             }
 
-            var result = "";
+            // First two digits of returned hex code contains the alpha,
+            // which is not supported in HTML color, so we need to cut those out.
+            var numberColor = $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordNumberColor"]).ToHex().Substring(2)}\">";
+            var specialColor = $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordSpecialColor"]).ToHex().Substring(2)}\">";
+            var result = string.Empty;
 
             // Start with an otherwise uncovered case so we will definitely enter the "something changed"
             // state.
@@ -55,7 +59,7 @@ namespace Bit.App.Utilities
                 if(charType != currentType)
                 {
                     // Close off previous span.
-                    if (currentType != CharType.None)
+                    if (currentType != CharType.None || currentType != CharType.Normal)
                     {
                         result += "</span>";
                     }
@@ -67,14 +71,11 @@ namespace Bit.App.Utilities
                     switch(currentType)
                     {
                         // Apply color style to span.
-                        // First two digits of returned hex code contains the alpha,
-                        // which is not supported in HTML color,
-                        // so we need to cut those out.
                         case CharType.Number:
-                            result += $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordNumberColor"]).ToHex().Substring(2)}\">";
+                            result += numberColor;
                             break;
                         case CharType.Special:
-                            result += $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordSpecialColor"]).ToHex().Substring(2)}\">";
+                            result += specialColor;
                             break;
                     }
                 }
@@ -82,7 +83,7 @@ namespace Bit.App.Utilities
             }
 
             // Close off last span.
-            if (currentType != CharType.Normal)
+            if (currentType != CharType.None || currentType != CharType.Normal)
             {
                 result += "</span>";
             }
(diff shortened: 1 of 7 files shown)


kspearrin:
This was intentionally not implemented on this page because it causes performance issues. I don't have a solution at this time.

ShirokaiLon:
I see the issue and belive I've found a solution. Using a HTML formatted string seems to have significantly better performance than FormattedString.

I see no noticable performance impact now.

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -21,17 +21,15 @@ private enum CharType
...
             {
-                return new FormattedString();
+                return "";
             }
 
-            var result = new FormattedString();
-            // Start off with an empty span to prevent possible NPEs. Due to the way the state machine
-            // works, this will actually always be replaced by a new span anyway.
-            var currentSpan = new Span();
+            var result = "";
```
kspearrin: use `string.Empty`

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -21,17 +21,15 @@ private enum CharType
             Special
         }
 
-        public static FormattedString FormatPassword(string password)
+        public static string FormatPassword(string password)
         {
             if(password == null)
             {
-                return new FormattedString();
+                return "";
```
kspearrin: use `string.Empty`

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -56,24 +54,39 @@ public static FormattedString FormatPassword(string password)
...
                     // default value.
                     switch(currentType)
                     {
+                        // Apply color style to span.
+                        // First two digits of returned hex code contains the alpha,
+                        // which is not supported in HTML color,
+                        // so we need to cut those out.
                         case CharType.Number:
-                            currentSpan.TextColor = (Color)Application.Current.Resources["PasswordNumberColor"];
+                            result += $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordNumberColor"]).ToHex().Substring(2)}\">";
```
kspearrin: Let's capture the hex codes as strings in variables outside of the for loop and then just use them here.

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -56,24 +54,39 @@ public static FormattedString FormatPassword(string password)
...
                             break;
                         case CharType.Special:
-                            currentSpan.TextColor = (Color)Application.Current.Resources["PasswordSpecialColor"];
+                            result += $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordSpecialColor"]).ToHex().Substring(2)}\">";
```
kspearrin: Let's capture the hex codes as strings in variables outside of the for loop and then just use them here.

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -56,24 +54,39 @@ public static FormattedString FormatPassword(string password)
                 // If the char type changed, build a new span to append the text to.
                 if(charType != currentType)
                 {
-                    currentSpan = new Span();
-                    result.Spans.Add(currentSpan);
+                    // Close off previous span.
+                    if (currentType != CharType.None)
```
kspearrin: Shouldn't this be `currentType != CharType.None || currentType != CharType.Normal`?

src/App/Utilities/PasswordFormatter.cs
```diff
@@ -56,24 +54,39 @@ public static FormattedString FormatPassword(string password)
...
+                            result += $"<span style=\"color:#{((Color)Application.Current.Resources["PasswordSpecialColor"]).ToHex().Substring(2)}\">";
                             break;
                     }
                 }
-                currentSpan.Text += c;
+                result += c;
             }
+
+            // Close off last span.
+            if (currentType != CharType.Normal)
```
kspearrin: Shouldn't this be the same as above? `currentType != CharType.None || currentType != CharType.Normal`?

kspearrin:
Awesome! We can also apply this to the passwords on the password history and password generator history page.

ShirokaiLon:
Good suggestions :+1:

I'll have those things fixed, and will apply it to the history pages tomorrow.

ShirokaiLon:
Used a converter class for the history pages. Looks clean to me, but not sure if there's a better way.

src/App/Utilities/PasswordFormatter.cs
ShirokaiLon: Indeed. Except && instead of ||

src/App/Utilities/PasswordFormatter.cs
ShirokaiLon: Indeed. Except && instead of ||

PR #687, opened 2020-01-01
Better Logout / Login Buttons

Max-We:
First off I swapped the Logout and Login Button on the Lock-Page to a more intuitive position. This idea was based on a suggestion I saw on Reddit (sorry forgot username).
Also I changed the Title of the Lock Page to something more fitting in German, as the old Title is too large to fit on the screen (cut off with "...").
Lastly I moved the Logout Button from the Settings into its Toolbar, which makes it a lot quicker to reach and adds consistency with the Lock page.

p.s. This is my first open-source commit and the first time I had to to with Xamarin. Please let me know if there are things I should pay attention to in the future, thank you! I hope you like my changes!


src/App/Resources/AppResources.de.resx
```diff
@@ -363,7 +363,7 @@
     <value>Fingerabdruck überprüfen</value>
   </data>
   <data name="VerifyMasterPassword" xml:space="preserve">
-    <value>Masterpasswort überprüfen</value>
```
ShirokaiLon: Bitwarden uses Crowdin for localisation, it would be better to submit this change there: https://github.com/bitwarden/mobile/blob/master/CONTRIBUTING.md

HLSiira:
Hey, I tried to do the same thing a while back I think just glancing at the title.

https://github.com/bitwarden/mobile/pull/657

eliykat:
It looks like these are substantially the same changes that were made by PR #1073, which has been merged. Please let me know if I'm mistaken.

PR #689, opened 2020-01-03
Fix for notification pendingIntent uri being set to browser package name

mportune-bw:
* AccessibilityHelpers.cs:  If AccessibilityNodeInfo.FindAccessibilityNodeInfosByViewId(..) returns null when the source package is a supported browser, return a null uri to prevent overwriting the existing notification's pendingIntent uri extra with the brower's packageName.

* AccessibilityService.cs: Added null uri checks as it is now possible for AccessibilityHelpers.getUri(..) to return a null uri when the Accessibility Service is misbehaving.


kspearrin:
Per slack chat, let's try to figure out why this `null` condition is occurring in the first place. This seems like a workaround to possibly a more low-level issue with which accessibility events we are processing. If we can stop that event in the execution cycle, we might not need this workaround at all.

PR #697, opened 2020-01-10
UI support in app settings for handling overlay permission requirement in Accessibility Service implementation

mportune-bw:
UI support in app settings for handling overlay permission requirement in Accessibility Service implementation


src/App/Pages/Settings/OptionsPage.xaml.cs
```diff
@@ -24,7 +24,7 @@ public OptionsPage()
             if(Device.RuntimePlatform == Device.Android)
             {
                 ToolbarItems.RemoveAt(0);
-                _vm.ShowAndroidAccessibilitySettings = true;
+                _vm.ShowAndroidAccessibilitySettings = false;
```
kspearrin: We need to come back and properly clean this up in a follow up.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -295,7 +295,8 @@ public void BuildList()
                 autofillItems.Add(new SettingsPageListItem
                 {
                     Name = AppResources.AutofillAccessibilityService,
-                    SubLabel = _deviceActionService.AutofillAccessibilityServiceRunning() ?
+                    SubLabel = (_deviceActionService.AutofillAccessibilityServiceRunning()
```
kspearrin: I would make this into a variable above the `.Add()` so that we can keep the shorthand operator clean and fix the weird indenting:

```
var accessibilityEnabled = _deviceActionService.AutofillAccessibilityServiceRunning() &&
    _deviceActionService.AutofillAccessibilityOverlayPermitted();
```

src/App/Resources/AppResources.resx
```diff
@@ -1588,6 +1588,24 @@
     <value>Use Biometrics To Unlock</value>
   </data>
   <data name="AccessibilityOverlayPermissionAlert" xml:space="preserve">
-    <value>Bitwarden requires additional accessibility configuration. Open the app for more info.</value>
+    <value>Bitwarden needs attention - See "Auto-fill Accessibility Service" from Bitwarden Settings</value>
+  </data>
+  <data name="BitwardenAutofillServiceOverlayPermission" xml:space="preserve">
```
kspearrin: What's the difference in this one vs `BitwardenAutofillServiceStep3`? Seem like almost the same thing.

mportune-bw: This one is for the toast and directs the user to the app settings for more detail (only 3.5 seconds to get a point across).  Step3 is more verbose since it's displayed on the settings screen.

PR #709, opened 2020-01-27
Accessibility fixes

mportune-bw:
Fixes to accessibility overlay anchor view tracking and node tree parsing.  Restricted calls to 'RootInActiveWindow' until absolutely necessary as it can be expensive (this was the culprit behind chromium based browser performance with view tracking).

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -422,7 +422,7 @@ namespace Bit.Droid.Accessibility
 
         public static bool IsStatusBarExpanded(IEnumerable<AccessibilityWindowInfo> windows)
         {
-            if(windows != null && windows.Count() > 0)
+            if(windows != null && windows.Any())
             {
                 var isSystemWindowsOnly = true;
                 foreach(var window in windows)
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -120,7 +120,7 @@ namespace Bit.Droid.Accessibility
                 if(addressNode != null)
                 {
                     uri = ExtractUri(uri, addressNode, browser);
-                    addressNode.Dispose();
+                    addressNode.Recycle();
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -350,38 +356,95 @@ public static WindowManagerLayoutParams GetOverlayLayoutParams()
...
                 {
-                    point = GetOverlayAnchorPosition(root, node);
-                    break;
+                    return new Point(-1, -1); ;
+                }
+
+                // node.VisibleToUser doesn't always give us exactly what we want, so attempt to tighten up the range
+                // of visibility
+                var rootNodeHeight = GetNodeHeight(root);
+                int limitLowY = 0;
```
kspearrin: var some of this

src/Android/Accessibility/AccessibilityService.cs
```diff
@@ -234,53 +243,80 @@ private void OverlayPromptToAutofill(AccessibilityNodeInfo root, AccessibilityEv
...
+            if(_overlayAnchorObserverRunning)
+            {
+                return;
             }
+            _overlayAnchorObserverRunning = true;
 
-            _anchorViewHash = e.Source.GetHashCode();
-            _lastAnchorX = anchorPosition.X;
-            _lastAnchorY = anchorPosition.Y;
+            _overlayAnchorObserverRunnable = new Java.Lang.Runnable(() =>
```
kspearrin: voodoo

src/Android/Accessibility/AccessibilityService.cs
```diff
@@ -54,106 +57,95 @@ public override void OnAccessibilityEvent(AccessibilityEvent e)
...
-                    case EventTypes.ViewScrolled:
+                        if(e.Source == null || e.PackageName == BitwardenPackage)
+                        {
+                            CancelOverlayPrompt();
+                            e.Recycle();
+                            break;
+                        }
+
+                        root = RootInActiveWindow;
+                        if(root == null)
```
kspearrin: I would just combine this with the check below, like it was originally.

Also, do we need this kind of check in the other switch case below? I don't see us doing this check there.

```
if(root == null || root.PackageName != e.PackageName)
                        {
                            e.Recycle();
                            break;
                        }

```

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -416,6 +420,24 @@ public static WindowManagerLayoutParams GetOverlayLayoutParams()
             return point;
         }
 
+        public static bool IsStatusBarExpanded(IEnumerable<AccessibilityWindowInfo> windows)
+        {
+            if(windows != null && windows.Count() > 0)
```
kspearrin: Use .Any() instead of Count() > 0. It can be more performant depending on the underlying type of the IEnumerable.

PR #713, opened 2020-01-29
Support for multiple browser UriViewIds when extracting a uri

mportune-bw:
Support for multiple browser UriViewIds when extracting a uri


kspearrin:
I think you can skip the contains check and just always split it?

PR #729, opened 2020-02-12
In-app vault export support

mportune-bw:
Support for vault export within the app.  Users can save the resulting json or csv file to local storage or to their cloud storage service of choice (via Storage Access Framework in Android - existing behavior with iOS "open file" flow is used since it already supports saving to local or cloud storage)

### DIFF
--- a/src/App/Pages/Settings/ExportVaultPage.xaml
+++ b/src/App/Pages/Settings/ExportVaultPage.xaml
@@ -1,5 +1,5 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
-<pages:BaseContentPage 
+﻿<?xml version="1.0" encoding="utf-8"?>
+<pages:BaseContentPage
     xmlns="http://xamarin.com/schemas/2014/forms"
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     x:Class="Bit.App.Pages.ExportVaultPage"
@@ -21,7 +21,7 @@
     </ContentPage.Resources>
 
     <ContentPage.ToolbarItems>
-        <ToolbarItem Text="{u:I18n Close}" Clicked="Close_Clicked"/>
+        <ToolbarItem Text="{u:I18n Close}" Clicked="Close_Clicked" />
     </ContentPage.ToolbarItems>
 
     <ScrollView>
@@ -47,30 +47,30 @@
                         <ColumnDefinition Width="Auto" />
                     </Grid.ColumnDefinitions>
                     <Label
-                            Text="{u:I18n MasterPassword}"
-                            StyleClass="box-label"
-                            Grid.Row="0"
-                            Grid.Column="0" />
+                        Text="{u:I18n MasterPassword}"
+                        StyleClass="box-label"
+                        Grid.Row="0"
+                        Grid.Column="0" />
                     <controls:MonoEntry
-                            x:Name="_masterPassword"
-                            Text="{Binding MasterPassword}"
-                            StyleClass="box-value"
-                            IsSpellCheckEnabled="False"
-                            IsTextPredictionEnabled="False"
-                            IsPassword="{Binding ShowPassword, Converter={StaticResource inverseBool}}"
-                            Grid.Row="1"
-                            Grid.Column="0"
-                            ReturnType="Go"
-                            ReturnCommand="{Binding ExportVaultCommand}" />
+                        x:Name="_masterPassword"
+                        Text="{Binding MasterPassword}"
+                        StyleClass="box-value"
+                        IsSpellCheckEnabled="False"
+                        IsTextPredictionEnabled="False"
+                        IsPassword="{Binding ShowPassword, Converter={StaticResource inverseBool}}"
+                        Grid.Row="1"
+                        Grid.Column="0"
+                        ReturnType="Go"
+                        ReturnCommand="{Binding ExportVaultCommand}" />
                     <controls:FaButton
-                            StyleClass="box-row-button, box-row-button-platform"
-                            Text="{Binding ShowPasswordIcon}"
-                            Command="{Binding TogglePasswordCommand}"
-                            Grid.Row="0"
-                            Grid.Column="1"
-                            Grid.RowSpan="2"
-                            AutomationProperties.IsInAccessibleTree="True"
-                            AutomationProperties.Name="{u:I18n ToggleVisibility}" />
+                        StyleClass="box-row-button, box-row-button-platform"
+                        Text="{Binding ShowPasswordIcon}"
+                        Command="{Binding TogglePasswordCommand}"
+                        Grid.Row="0"
+                        Grid.Column="1"
+                        Grid.RowSpan="2"
+                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.Name="{u:I18n ToggleVisibility}" />
                 </Grid>
                 <Label
                     Text="{u:I18n ExportVaultMasterPasswordDescription}"
@@ -80,7 +80,9 @@
                     Margin="0, 20">
                     <Label.FormattedText>
                         <FormattedString>
-                            <Span Text="{Binding Converter={StaticResource toUpper}, ConverterParameter={u:I18n Warning}}" FontAttributes="Bold"/>
+                            <Span
+                                Text="{Binding Converter={StaticResource toUpper}, ConverterParameter={u:I18n Warning}}"
+                                FontAttributes="Bold" />
                             <Span Text=": " FontAttributes="Bold" />
                             <Span Text="{u:I18n ExportVaultWarning}" />
                         </FormattedString>
@@ -88,9 +90,9 @@
                 </Label>
                 <StackLayout Spacing="20">
                     <Button Text="{u:I18n ExportVault}"
-                        Clicked="ExportVault_Clicked"
-                        HorizontalOptions="Fill"
-                        VerticalOptions="End"></Button>
+                            Clicked="ExportVault_Clicked"
+                            HorizontalOptions="Fill"
+                            VerticalOptions="End" />
                 </StackLayout>
             </StackLayout>
         </StackLayout>
--- a/src/App/Pages/Settings/ExportVaultPage.xaml
+++ b/src/App/Pages/Settings/ExportVaultPage.xaml
@@ -1,0 +1,99 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<pages:BaseContentPage 
+    xmlns="http://xamarin.com/schemas/2014/forms"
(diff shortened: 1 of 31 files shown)


src/App/Pages/Settings/ExportVaultPage.xaml
```diff
@@ -0,0 +1,99 @@
...
+                </Grid>
+                <Label
+                    Text="{u:I18n ExportVaultMasterPasswordDescription}"
+                    StyleClass="box-footer-label, box-footer-label-switch" />
+                <Label
+                    StyleClass="box-footer-label, box-footer-label-switch"
+                    Margin="0, 20">
+                    <Label.FormattedText>
+                        <FormattedString>
+                            <Span Text="{Binding Converter={StaticResource toUpper}, ConverterParameter={u:I18n Warning}}" FontAttributes="Bold"/>
```
kspearrin: Is this over the 120 line limit size? Break it into multiple lines?

src/App/Pages/Settings/ExportVaultPageViewModel.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+                catch(Exception ex)
+                {
+                    ClearResult();
+                    System.Diagnostics.Debug.WriteLine(">>> {0}: {1}", ex.GetType(), ex.StackTrace);
+                }
+            }
+            else
+            {
+                _platformUtilsService.ShowToast("error", _i18nService.T("AnErrorHasOccurred"),
```
kspearrin: Should probably be `ShowDialogAsync` too

src/App/Pages/Settings/ExportVaultPageViewModel.cs
```diff
@@ -0,0 +1,144 @@
...
+        public async void SaveFileSelected(string contentUri, string filename)
+        {
+            if(_deviceActionService.SaveFile(_exportResult, null, filename ?? _defaultFilename, contentUri))
+            {
+                ClearResult();
+                _platformUtilsService.ShowToast("success", null, _i18nService.T("ExportVaultSuccess"));
+                return;
+            }
+
+            ClearResult();
```
kspearrin: Looks like you call `ClearResult();` for both conditions, so just add it before the above `if`.

src/App/Pages/Settings/ExportVaultPageViewModel.cs
```diff
@@ -0,0 +1,144 @@
...
+            if(storedKeyHash != null && keyHash != null && storedKeyHash == keyHash)
+            {
+                try
+                {
+                    var data = _exportService.GetExport(FileFormatOptions[FileFormatSelectedIndex].Key);
+                    var fileFormat = FileFormatOptions[FileFormatSelectedIndex].Key;
+                    _defaultFilename = _exportService.GetFileName(null, fileFormat);
+                    _exportResult = Encoding.ASCII.GetBytes(data.Result);
+
+                    if(!_deviceActionService.SaveFile(_exportResult, null, _defaultFilename, null))
```
kspearrin: I am a little confused about why we have `_deviceActionService.SaveFile` here and in the `SaveFileSelected` method below. are there two paths for saving a file? I see the code-behind file calls into `SaveFileSelected` as well when a message is received.

src/Core/Models/Export/Cipher.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        public string OrganizationId { get; set; }
+        public string FolderId { get; set; }
+        public CipherType Type { get; set; }
+        public string Name { get; set; }
+        public string Notes { get; set; }
+        public bool Favorite { get; set; }
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
```
kspearrin: Are these ignored in the other apps when null too?

src/Core/Models/Export/Cipher.cs
```diff
@@ -0,0 +1,103 @@
...
+        public Cipher(CipherView obj)
+        {
+            OrganizationId = obj.OrganizationId;
+            FolderId = obj.FolderId;
+            Type = obj.Type;
+            Name = obj.Name;
+            Notes = obj.Notes;
+            Favorite = obj.Favorite;
+
+            if(obj.Fields != null)
```
kspearrin: You could skip the null check and just do:

```
Fields = obj.Fields?.Select(f => new Field(f)).ToList();
```

src/Core/Models/Export/Cipher.cs
```diff
@@ -0,0 +1,103 @@
...
+                view.OrganizationId = req.OrganizationId;
+            }
+
+            view.Name = req.Name;
+            view.Notes = req.Notes;
+            view.Favorite = req.Favorite;
+
+            if(req.Fields != null)
+            {
+                view.Fields = req.Fields.Select(f => Field.ToView(f)).ToList();
```
kspearrin: Same as above, can skip null check:

```
view.Fields = req.Fields?.Select(f => Field.ToView(f)).ToList();
```

src/Core/Models/Export/Login.cs
```diff
@@ -0,0 +1,51 @@
...
+        public Login(LoginView obj)
+        {
+            if(obj == null)
+            {
+                return;
+            }
+
+            if(obj.Uris != null)
+            {
+                Uris = obj.Uris.Select(u => new LoginUri(u)).ToList();
```
kspearrin: same above before, can skip null check with `?.`

src/Core/Models/Export/Login.cs
```diff
@@ -0,0 +1,51 @@
...
+        public static LoginView ToView(Login req, LoginView view = null)
+        {
+            if(view == null)
+            {
+                view = new LoginView();
+            }
+
+            if(req.Uris != null)
+            {
+                view.Uris = req.Uris.Select(u => LoginUri.ToView(u)).ToList();
```
kspearrin: same above before, can skip null check with `?.`

src/Core/Models/Export/Login.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Bit.Core.Models.Export
+{
+    public class Login
+    {
+        public Login() { }
+
+        public Login(LoginView obj)
+        {
+            if(obj == null)
```
kspearrin: Remove all of these `if(obj == null)` checks in the export models. This is a hack that we have to do in TS because they do not allow for ctor overloads. Our empty ctor will allow us to do what TS will not.

src/Core/Services/ExportService.cs
```diff
@@ -0,0 +1,194 @@
...
+
+                    if(c.OrganizationId != null)
+                    {
+                        continue;
+                    }
+
+                    var cipher = new ExportCipher();
+                    cipher.Folder = c.FolderId != null && foldersMap.ContainsKey(c.FolderId)
+                        ? foldersMap[c.FolderId].Name
+                        : null;
```
kspearrin: I would move the else condition up 1 line, but that's just my formatting preference

src/Core/Services/ExportService.cs
```diff
@@ -0,0 +1,194 @@
...
+            cipher.Name = c.Name;
+            cipher.Notes = c.Notes;
+            cipher.Fields = null;
+            // Login props
+            cipher.LoginUris = null;
+            cipher.LoginUsername = null;
+            cipher.LoginPassword = null;
+            cipher.LoginTotp = null;
+
+            if(c.Fields?.Any() ?? false)
```
kspearrin: Don't need to check `Any()`. A null check will suffice.

src/Core/Services/ExportService.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            switch(c.Type)
+            {
+                case CipherType.Login:
+                    cipher.Type = "login";
+                    cipher.LoginUsername = c.Login.Username;
+                    cipher.LoginPassword = c.Login.Password;
+                    cipher.LoginTotp = c.Login.Totp;
+
+                    if(c.Login.Uris?.Any() ?? false)
```
kspearrin: same, just need the null check

src/App/Pages/Settings/ExportVaultPageViewModel.cs
mportune-bw: `InvalidMasterPassword` is also shown on line 88 - would you prefer that be a dialog as well?

src/App/Pages/Settings/ExportVaultPageViewModel.cs
mportune-bw: The values being cleared are required in the `if` statement

src/App/Pages/Settings/ExportVaultPageViewModel.cs
mportune-bw: TLDR: It's a continuation of the save flow, specific to Android.

`_deviceActionService.SaveFile` starts the process.  For iOS it simply follows the existing `_deviceActionService.OpenFile` flow since that gives us what we want (options for saving locally as well as cloud services).  On Android however new logic was required to get the same options (via storage access framework).  Once the user selects a save destination, the uri & filename is returned via `MainActivity.OnActivityResult` and forwarded to `ExportVaultPageViewModel.SaveFileSelected` via `_messagingService.Send("selectSaveFileResult")`.

src/Core/Models/Export/Cipher.cs
mportune-bw: I used the web vault export as the gold standard to mimic (so not sure about other apps)

src/App/Pages/Settings/ExportVaultPageViewModel.cs
kspearrin: But it's called for both conditions, right?

mportune-bw: Correct

src/App/Pages/Settings/ExportVaultPageViewModel.cs
kspearrin: Yes

src/App/Pages/Settings/ExportVaultPageViewModel.cs
kspearrin: So couldnt you do?

```
            
            ClearResult();
            if(_deviceActionService.SaveFile(_exportResult, null, filename ?? _defaultFilename, contentUri))
            {
                _platformUtilsService.ShowToast("success", null, _i18nService.T("ExportVaultSuccess"));
                return;
            }
            await _platformUtilsService.ShowDialogAsync(_i18nService.T("ExportVaultFailure"));
```

mportune-bw: Ignoring per clarification in chat

PR #736, opened 2020-02-19
Beginning of policy support

mportune-bw:
Added models & service layer to support policies

### DIFF
--- a/src/Core/Services/PolicyService.cs
+++ b/src/Core/Services/PolicyService.cs
@@ -1,0 +1,78 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Bit.Core.Abstractions;
+using Bit.Core.Enums;
+using Bit.Core.Models.Data;
+using Bit.Core.Models.Domain;
+
+namespace Bit.Core.Services
+{
+    public class PolicyService : IPolicyService
+    {
+        private readonly IStorageService _storageService;
+        private readonly IUserService _userService;
+
+        private List<Policy> _policyCache;
+
+        public PolicyService(
+            IStorageService storageService,
+            IUserService userService)
+        {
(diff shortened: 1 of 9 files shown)


src/Core/Services/PolicyService.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            if(_policyCache == null)
+            {
+                var userId = await _userService.GetUserIdAsync();
+                var policies =
+                    await _storageService.GetAsync<Dictionary<string, PolicyData>>(Keys.PoliciesPrefix + userId);
+                var response = new List<Policy>();
+                foreach(var id in policies)
+                {
+                    if(policies.GetType().GetProperty(id.Key) != null)
```
kspearrin: This was a weird javascript thing. No need to do this in c#. You could replace this whole for loop with a simple:

```
var response = policies.Select(p => new Policy(policies[p.id]));
```

src/Core/Services/PolicyService.cs
```diff
@@ -0,0 +1,78 @@
...
+            else
+            {
+                return _policyCache;
+            }
+        }
+
+        public async Task Replace(Dictionary<string, PolicyData> policies)
+        {
+            var userId = await _userService.GetUserIdAsync();
+            await _storageService.SaveAsync(Keys.PoliciesPrefix + userId, policies);
```
kspearrin: Use string formatter like other services.

src/Core/Services/PolicyService.cs
```diff
@@ -0,0 +1,78 @@
...
+            _policyCache = null;
+        }
+
+        public async Task Clear(string userId)
+        {
+            await _storageService.RemoveAsync(Keys.PoliciesPrefix + userId);
```
kspearrin: Use string formatter like other services.

PR #738, opened 2020-02-22
Fix runtime exception.

oldium:
In the callback the processing does not go in the main UI thread, so we
need to switch there. Otherwise on Android this throws
Android.Util.AndroidRuntimeException with detail “Only the original thread
that created a view hierarchy can touch its views”.

Discovered by trying to login with Duo as a two-factor login type.


oldium:
I checked it once more and also calls outside and it looks like that the whole function shall be called from the UI thread. I will check which callback switched from UI thread to thread pool.

oldium:
Better fix in #739.

PR #739, opened 2020-02-23
Make callbacks from UI thread.

oldium:
The code expects to be called form UI thread, but it looks like JavaScript
callbacks are not. Switch to UI thread when invoking a callback.

This is a better fix than #738, because this fixes the whole callback thread issue.


oldium:
Please review the threading also in iOS, I have not checked that (no iOS device).

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,7 +208,7 @@ public async Task SubmitAsync()
                 _broadcasterService.Unsubscribe(nameof(TwoFactorPage));
                 var disableFavicon = await _storageService.GetAsync<bool?>(Constants.DisableFaviconKey);
                 await _stateService.SaveAsync(Constants.DisableFaviconKey, disableFavicon.GetValueOrDefault());
-                Device.BeginInvokeOnMainThread(() => { Application.Current.MainPage = new TabsPage(); });
+                Application.Current.MainPage = new TabsPage();
```
kspearrin: Let's just add it to the `await SubmitAsync()` call from the `page.DuoWebView.RegisterAction` callback above.

PR #741, opened 2020-02-27
Password generator policy enforcement

mportune-bw:
Enforcement of password generator policies.  Handling of UI widgets on generator screen could use some polish but I'd like a review for this so far.  (I don't see a way to comment on a review request, just a PR)

### DIFF
--- a/src/App/Pages/Generator/GeneratorPage.xaml
+++ b/src/App/Pages/Generator/GeneratorPage.xaml
@@ -152,6 +152,8 @@
                             HorizontalOptions="StartAndExpand" />
                         <Switch
                             IsToggled="{Binding Uppercase}"
+                            IsEnabled="{Binding EnforcedPolicyOptions.UseUppercase,
+                                Converter={StaticResource inverseBool}}"
                             StyleClass="box-value"
                             HorizontalOptions="End" />
                     </StackLayout>
@@ -163,6 +165,8 @@
                             HorizontalOptions="StartAndExpand" />
                         <Switch
                             IsToggled="{Binding Lowercase}"
+                            IsEnabled="{Binding EnforcedPolicyOptions.UseLowercase,
+                                Converter={StaticResource inverseBool}}"
                             StyleClass="box-value"
                             HorizontalOptions="End" />
                     </StackLayout>
@@ -174,6 +178,8 @@
                             HorizontalOptions="StartAndExpand" />
                         <Switch
                             IsToggled="{Binding Number}"
+                            IsEnabled="{Binding EnforcedPolicyOptions.UseNumbers,
+                                Converter={StaticResource inverseBool}}"
                             StyleClass="box-value"
                             HorizontalOptions="End" />
                     </StackLayout>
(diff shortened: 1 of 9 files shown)


kspearrin:
Did you format these files? I’m seeing a lot of inconsistencies on the if statement blocks.

mportune-bw:
> Did you format these files? I’m seeing a lot of inconsistencies on the if statement blocks.

Fixed, forgot to hit the button.

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -255,7 +259,129 @@ public async Task<PasswordGenerationOptions> GetOptionsAsync()
...
+                }
+
+                if(enforcedOptions == null)
+                {
+                    enforcedOptions = new PasswordGeneratorPolicyOptions();
+                }
+
+                var currentPolicyMinLength = currentPolicy.Data["minLength"];
+                if(currentPolicyMinLength != null &&
+                   Convert.ToInt32(currentPolicyMinLength) > enforcedOptions.MinLength)
```
kspearrin: I think just casting to int would be enough here, and a little cleaner. `(int)currentPolicyMinLength`.

`Convert` is more for when _changing_ from one type to another, like converting a string `"1"` to an int `1`.

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -255,7 +259,129 @@ public async Task<PasswordGenerationOptions> GetOptionsAsync()
...
+                {
+                    enforcedOptions.MinLength = Convert.ToInt32(currentPolicyMinLength);
+                }
+
+                var currentPolicyUseUpper = currentPolicy.Data["useUpper"];
+                if(currentPolicyUseUpper != null && Convert.ToBoolean(currentPolicyUseUpper))
```
kspearrin: Same with these booleans. They are already booleans, just happen to be represented as an object.

src/Core/Services/PasswordGenerationService.cs
mportune-bw: The object was coming back as an Int64 and throwing InvalidCastException.  Additional research shows double-unboxing will also work so I'll use that approach.

src/Core/Services/PasswordGenerationService.cs
mportune-bw: This was done for consistency - will change to simple cast

src/App/Pages/Generator/GeneratorPage.xaml
```diff
@@ -152,8 +175,7 @@
                             HorizontalOptions="StartAndExpand" />
                         <Switch
                             IsToggled="{Binding Uppercase}"
-                            IsEnabled="{Binding EnforcedPolicyOptions.UseUppercase,
-                                Converter={StaticResource inverseBool}}"
+                            IsEnabled="{Binding IsUppercaseSwitchEnabled}"
```
kspearrin: @mportune-bw What prompted this change? This is how I thought you would have had to have done it fromt he beginning when I questioned it the first time.

mportune-bw: Hm I thought there was a scenario where this was necessary but now I'm wondering if I just confused myself with the styling of disabled vs off switch states.  I've reverted it locally and everything is still working as it should.  I'll submit a PR shortly.

PR #743, opened 2020-02-28
Fixed password color and alignment on iOS password generator

mportune-bw:
Some key takeaways:

- One character of the alpha hex code was still in the hex color.  Android ignored the invalid char and was able to establish the intended hex color, whereas iOS threw the whole thing out as invalid, hence the color-less passwords.  Also had to add another span for the "normal" color so it respects what "normal" is based on the selected app theme.  I actually can't figure out how Android is doing this automatically.

- Changing the label to `HorizontalOptions="CenterAndExpand"` mimics the behavior of `wrap_content` in Android, so the label itself is only as wide as the text within.  This "fixes" the issue of iOS ignoring `HorizontalTextAlignment="Center"` when `TextType="Html"` is used, but only if the password is on a single line.  Once the password starts wrapping, the left-justification is visible again.

- Another issue that I haven't figured out how to solve:  `LineBreakMode="CharacterWrap"` is not working as expected with `TextType="Html"`.  It seems to be wrapping on "words" resulting in violent shifting of the screen controls as the wrapping index is re-evaluated for each new password.  (This is extra apparent when you download the app store version of bitwarden which properly wraps on characters, resulting in a much smoother experience when adjusting the length slider)

### DIFF
--- a/src/App/Pages/Generator/GeneratorPage.xaml
+++ b/src/App/Pages/Generator/GeneratorPage.xaml
@@ -45,6 +45,7 @@
                     Margin="0, 20"
                     StyleClass="text-lg"
                     HorizontalTextAlignment="Center"
+                    HorizontalOptions="CenterAndExpand"
                     LineBreakMode="CharacterWrap" />
                 <Button Text="{u:I18n RegeneratePassword}"
                         HorizontalOptions="FillAndExpand"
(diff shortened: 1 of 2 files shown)


kspearrin:
Whoops, I probably should have just accepted.

PR #751, opened 2020-03-02
New Android attachment handling to support saving or opening attachments

mportune-bw:
Introduced new attachment handling flow for Android:  Click download icon -> perform download -> If the file can be opened by the system, present user with "open or save" prompt -> If the file can't be opened by the system, go directly into the save flow.  This effectively means we can save any type of attachment + optionally open/view directly if supported.

### DIFF
--- a/src/App/Pages/Vault/ViewPageViewModel.cs
+++ b/src/App/Pages/Vault/ViewPageViewModel.cs
@@ -459,25 +459,15 @@ namespace Bit.App.Pages
 
         public async void PromptOpenOrSave(byte[] data, AttachmentView attachment)
         {
-            var options = new List<KeyValuePair<string, string>>
-            {
-                new KeyValuePair<string, string>("open", AppResources.Open),
-                new KeyValuePair<string, string>("save", AppResources.Save)
-            };
             var selection = await Page.DisplayActionSheet(attachment.FileName, AppResources.Cancel, null,
-                options.Select(o => o.Value).ToArray());
-            if(selection != null && selection != AppResources.Cancel)
+                AppResources.Open, AppResources.Save);
+            if(selection == AppResources.Open)
             {
-                switch(options.FirstOrDefault(f => f.Value == selection).Key)
-                {
-                    case "open":
-                        OpenAttachment(data, attachment);
-                        break;
-
-                    case "save":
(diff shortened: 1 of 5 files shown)


src/App/Pages/Vault/ViewPageViewModel.cs
```diff
@@ -433,6 +457,69 @@ private async void DownloadAttachmentAsync(AttachmentView attachment)
...
+                {
+                    case "open":
+                        OpenAttachment(data, attachment);
+                        break;
+
+                    case "save":
+                        SaveAttachment(data, attachment);
+                        break;
+                }
+            }
```
kspearrin: I think you are basing this off of one of my more dynamic/complex examples of `DisplayActionSheet`. We can simplify this here. Just pass `AppResources.Open` and `AppResources.Save` and compare that in `selection` result. 

```suggestion
            var selection = await Page.DisplayActionSheet(attachment.FileName, AppResources.Cancel, null,
                AppResources.Open, AppResources.Save);
            if(selection == AppResources.Open)
            {
                OpenAttachment(data, attachment);
            }
            else if(selection == AppResources.Save)
            {
                SaveAttachment(data, attachment);
            }
```

src/App/Resources/AppResources.resx
```diff
@@ -1615,4 +1615,14 @@
   <data name="PasswordGeneratorPolicyInEffect" xml:space="preserve">
     <value>One or more organization policies are affecting your generator settings</value>
   </data>
+  <data name="Open" xml:space="preserve">
+    <value>Open</value>
+    <comment>Button text for an open operation (verb).</comment>
+  </data>
+  <data name="UnableToSaveAttachment" xml:space="preserve">
+    <value>There was a problem saving this attachment.  If the problem persists, you'll need to save it from the web vault.</value>
```
kspearrin:
```suggestion
    <value>There was a problem saving this attachment. If the problem persists, you can save it from the web vault.</value>
```

src/App/Pages/Vault/ViewPageViewModel.cs
mportune-bw: Ah, yes that's much better.

PR #761, opened 2020-03-09
Fixed issue where multiple threads were attempting to modify search result list

mportune-bw:
- Perform search result list update on main thread to force thread safety on the underlying ObservableCollection.  This fixes the occasional `ArgumentOutOfRangeException` as well as the strange behavior where hitting the IME "enter" key after entering text would clear the results and state none were found even though they exist.

- Changed existing search execution delay for keystrokes to 200ms to match web

- Fixed race condition where non-delayed search (keyboard "enter") jumped the queue and the prior delayed search (text change) would still modify the search results, contributing to the "no results found" problem (we're now returning when `OperationCanceledException` is thrown)

### DIFF
--- a/src/App/Pages/Vault/CiphersPage.xaml.cs
+++ b/src/App/Pages/Vault/CiphersPage.xaml.cs
@@ -75,7 +75,7 @@ namespace Bit.App.Pages
             {
                 return;
             }
-            _vm.Search(e.NewTextValue, 500);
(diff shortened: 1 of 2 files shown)


kspearrin:
Isn’t there an autofill listing page too where we need this fix?

mportune-bw:
> Isn’t there an autofill listing page too where we need this fix?

Ah so there is.  One moment, please...

PR #767, opened 2020-03-12
Null check policies from SyncResponse before parsing

mportune-bw:
Fixes sync issue introduced with policies combined with org-less accounts


src/Core/Services/SyncService.cs
```diff
@@ -365,7 +365,8 @@ private async Task SyncSettingsAsync(string userId, DomainsResponse response)
 
         private async Task SyncPolicies(List<PolicyResponse> response)
         {
-            var policies = response.ToDictionary(p => p.Id, p => new PolicyData(p));
+            var policies = response?.ToDictionary(p => p.Id, p => new PolicyData(p)) ??
+                           new Dictionary<string, PolicyData>();
```
kspearrin: indented formatting should just have this 4 spaces over

```suggestion
                new Dictionary<string, PolicyData>();
```

PR #772, opened 2020-03-13
Enforce Passphrase Policy

vincentsalucci:
## Code Changes
- **GeneratorPage.xaml**: Added `IsEnabled` values for `Capitalize` and `IncludeNumber` switches depending on the `EnforcedPolicyOptions` returned
- **GeneratorPageViewModel.cs**: Updated boolean property to use newly added `InEffect()` function
- **PasswordGeneratorPolicyOptions.cs**: Added new fields for tracking the passphrase policy. Created helper method to clear up a long conditional in the view model.
- **PasswordGenerationService.cs**: Passphrase options are now enforced during `PasswordGeneratorPolicyOptions` objection creation.  User options are now updated with passphrase options during retrieval. Normalize options will make sure the floor is set with `MinNumberWords` variable.

### Android
0-android-passphrase-enforced
### iOS
0-ios-passphrase-enforced

### DIFF
--- a/src/Core/Models/Domain/PasswordGeneratorPolicyOptions.cs
+++ b/src/Core/Models/Domain/PasswordGeneratorPolicyOptions.cs
@@ -16,17 +16,17 @@
         
         public bool InEffect()
         {
-            return DefaultType != ""
-                   || MinLength > 0
-                   || NumberCount > 0
-                   || SpecialCount > 0
-                   || UseUppercase
-                   || UseLowercase
-                   || UseNumbers
-                   || UseSpecial
-                   || MinNumberOfWords > 0
-                   || Capitalize
-                   || IncludeNumber;
+            return DefaultType != "" || 
+                   MinLength > 0 ||
+                   NumberCount > 0 ||
+                   SpecialCount > 0 ||
+                   UseUppercase ||
+                   UseLowercase ||
+                   UseNumbers ||
+                   UseSpecial ||
+                   MinNumberOfWords > 0 ||
+                   Capitalize ||
+                   IncludeNumber;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Core/Models/Domain/PasswordGeneratorPolicyOptions.cs
```diff
@@ -2,12 +2,31 @@
...
         public int NumberCount { get; set; }
         public bool UseSpecial { get; set; }
         public int SpecialCount { get; set; }
+        public int MinNumberOfWords { get; set; }
+        public bool Capitalize { get; set; }
+        public bool IncludeNumber { get; set; }
+        
+        public bool InEffect()
+        {
+            return DefaultType != ""
```
kspearrin: Formatting done elsewhere in the codebase is to include `||`'s and `&&`'s at the end of the line. For example:

```
return DefaultType != "" ||
    Condition1 ||
    Condition2 ||
    ...
```

vincentsalucci: Updated!

src/Core/Models/Domain/PasswordGeneratorPolicyOptions.cs
```diff
@@ -2,12 +2,31 @@
...
         public bool UseSpecial { get; set; }
         public int SpecialCount { get; set; }
+        public int MinNumberOfWords { get; set; }
+        public bool Capitalize { get; set; }
+        public bool IncludeNumber { get; set; }
+        
+        public bool InEffect()
+        {
+            return DefaultType != "" || 
+                   MinLength > 0 ||
```
kspearrin: Indent spacing is off here now. Should be 4 spaces. Try formatting?

PR #783, opened 2020-03-18
Sanitize Password Length

vincentsalucci:
## Code Changes
- **PasswordGenerationService.cs**: Created `SanitizePasswordLength` method that can be called with UI changes or for password generation only.  Maintains existing workflows but will now properly update the UI if a minimum length condition is met.

### DIFF
--- a/src/Core/Services/PasswordGenerationService.cs
+++ b/src/Core/Services/PasswordGenerationService.cs
@@ -649,7 +649,7 @@ namespace Bit.Core.Services
             wordList[index] = wordList[index] + num;
         }
 
-        private static void SanitizePasswordLength(PasswordGenerationOptions options, bool forGeneration)
+        private void SanitizePasswordLength(PasswordGenerationOptions options, bool forGeneration)
         {
             var minUppercaseCalc = 0;
             var minLowercaseCalc = 0;
@@ -659,7 +659,8 @@ namespace Bit.Core.Services
             if(options.Uppercase.GetValueOrDefault() && options.MinUppercase.GetValueOrDefault() <= 0)
             {
                 minUppercaseCalc = 1;
-            } else if(!options.Uppercase.GetValueOrDefault())
+            } 
+            else if(!options.Uppercase.GetValueOrDefault())
             {
                 minUppercaseCalc = 0;
             }
@@ -667,7 +668,8 @@ namespace Bit.Core.Services
             if(options.Lowercase.GetValueOrDefault() && options.MinLowercase.GetValueOrDefault() <= 0) 
             {
                 minLowercaseCalc = 1;
-            } else if(!options.Lowercase.GetValueOrDefault())
+            }
+            else if(!options.Lowercase.GetValueOrDefault())
             {
                 minLowercaseCalc = 0;
             }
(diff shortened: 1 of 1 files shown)


src/Core/Services/PasswordGenerationService.cs
```diff
@@ -691,5 +648,56 @@ private async Task AppendRandomNumberToRandomWordAsync(List<string> wordList)
...
+        private static void SanitizePasswordLength(PasswordGenerationOptions options, bool forGeneration)
+        {
+            var minUppercaseCalc = 0;
+            var minLowercaseCalc = 0;
+            var minNumberCalc = options.MinNumber;
+            var minSpecialCalc = options.MinNumber;
+
+            if (options.Uppercase.GetValueOrDefault() && options.MinUppercase.GetValueOrDefault() <= 0) {
+                minUppercaseCalc = 1;
+            } else if (!options.Uppercase.GetValueOrDefault()) {
```
kspearrin: formatting on c# is different than JS. should be new line.

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -691,5 +648,56 @@ private async Task AppendRandomNumberToRandomWordAsync(List<string> wordList)
...
         }
+
+        private static void SanitizePasswordLength(PasswordGenerationOptions options, bool forGeneration)
+        {
+            var minUppercaseCalc = 0;
+            var minLowercaseCalc = 0;
+            var minNumberCalc = options.MinNumber;
+            var minSpecialCalc = options.MinNumber;
+
+            if (options.Uppercase.GetValueOrDefault() && options.MinUppercase.GetValueOrDefault() <= 0) {
```
kspearrin: formatting on c# is different than JS. should be no whitespace between `if(`

vincentsalucci:
@kspearrin - IDE suggested making the `SanitizePasswordLength` method `static`.  Did some research and didn't see any reason not to do it. What are your thoughts on that and when to make a method static?

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -691,5 +648,67 @@ private async Task AppendRandomNumberToRandomWordAsync(List<string> wordList)
...
+        {
+            var minUppercaseCalc = 0;
+            var minLowercaseCalc = 0;
+            var minNumberCalc = options.MinNumber;
+            var minSpecialCalc = options.MinNumber;
+
+            if(options.Uppercase.GetValueOrDefault() && options.MinUppercase.GetValueOrDefault() <= 0)
+            {
+                minUppercaseCalc = 1;
+            } else if(!options.Uppercase.GetValueOrDefault())
```
kspearrin: else if should be on a new line. Is formatter not fixing that?

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -691,5 +648,67 @@ private async Task AppendRandomNumberToRandomWordAsync(List<string> wordList)
             var num = await _cryptoService.RandomNumberAsync(0, 9);
             wordList[index] = wordList[index] + num;
         }
+
+        private static void SanitizePasswordLength(PasswordGenerationOptions options, bool forGeneration)
```
kspearrin: let's not make it static. we havent been doing that with others.

PR #788, opened 2020-03-20
Project lib updates and migration

mportune-bw:
**Summary**:

Update to all 3rd party libs used in the project.  Please make sure you're using the latest stable Visual Studio (just recently updated) as well as the latest Xamarin.Android and Xamarin.iOS SDKs.  If using Rider, make sure it's referencing the Xamarin SDKs from Visual Studio (the built-in versions aren't recent enough)

Occasionally when installing to Android the following error may occur: 

`The "CopyGeneratedJavaResourceClasses" task was not given a value for the required parameter "SourceTopDirectory".`

Simply re-run the install process and it should proceed normally.

**Additional Details**:

Minimum supported Android version now `5.0` (SDK `21`)
   - Removed `MultiDex` support as Android 5.0+ runtime supports multiple dex natively

`Xamarin.Forms`: Updated to `4.5.0.356` (from `4.4.0.991265`)

`Xamarin.Essentials`: Updated to `1.5.1` (from `1.3.1`)

`Xamarin.Android.Support*`: Updated to all appropriate `Xamarin.AndroidX*` and `Xamarin.Google*` libs
   - Updated all relevant imports

`Xamarin.Firebase.Messaging`: Updated to `71.1740.0` (from `60.1142.1`)

`Portable.BouncyCastle`: Updated to `1.8.6` (from `1.8.5.2`)

`LiteDB`: Updated to `5.0.4` (from `4.1.4`) which required these modifications:
   - Changed `_collection` type to `ILiteCollection` (from `LiteCollection`)
   - Added `Upgrade=true;` to DB constructor as the DB file format was changed
   - Changed collection `.Delete` to `.DeleteMany` to support the existing expression

`Plugin.Fingerprint`: Updated to `2.1.1` (from `1.4.9`) which required these modifications:
   - Removed `CustomFingerprintDialogFragment.cs` as it's no longer necessary
   - Added second `text` arg in `AuthenticationRequestConfiguration` constructor as the usage now varies between platforms

`ZXing.Net.Mobile.Forms`: Updated to `2.4.1` (from `2.1.47`)

`ZXing.Net.Mobile`: Added `2.4.1` as it's now required alongside `ZXing.Net.Mobile.Forms`)

`Refractored.FloatingActionButtonForms`: Removed as it's not compatible with Forms 4.5 & AndroidX
   - Replaced with visibly identical custom implementation, no lib required


src/App/App.csproj
```diff
@@ -14,12 +14,12 @@
...
-    <PackageReference Include="Plugin.Fingerprint" Version="1.4.9" />
-    <PackageReference Include="Refractored.FloatingActionButtonForms" Version="2.1.0" />
-    <PackageReference Include="Xamarin.Essentials" Version="1.3.1" />
+    <PackageReference Include="Plugin.Fingerprint" Version="2.1.1" />
+    <PackageReference Include="Xamarin.Essentials" Version="1.5.1" />
     <PackageReference Include="Xamarin.FFImageLoading.Forms" Version="2.4.11.982" />
-    <PackageReference Include="Xamarin.Forms" Version="4.4.0.991265" />
-    <PackageReference Include="ZXing.Net.Mobile.Forms" Version="2.1.47" />
+    <PackageReference Include="Xamarin.Forms" Version="4.5.0.356" />
+    <PackageReference Include="ZXing.Net.Mobile" Version="2.4.1" />
```
kspearrin: I dont think you need to specify `ZXing.Net.Mobile`. It should get pulled in by `ZXing.Net.Mobile.Forms`

src/App/Services/MobilePlatformUtilsService.cs
```diff
@@ -218,7 +218,7 @@ public async Task<bool> SupportsBiometricAsync()
                         var supportsFace = await _deviceActionService.SupportsFaceBiometricAsync();
                         text = supportsFace ? AppResources.FaceIDDirection : AppResources.FingerprintDirection;
                     }
-                    var fingerprintRequest = new AuthenticationRequestConfiguration(text)
+                    var fingerprintRequest = new AuthenticationRequestConfiguration(text, text)
```
kspearrin: What's the second param for now?

src/App/Styles/Base.xaml
```diff
@@ -97,13 +96,19 @@
...
                 Value="{StaticResource DisabledIconColor}" />
     </Style>
-    <Style TargetType="fab:FloatingActionButtonView">
-        <Setter Property="ColorNormal"
-                Value="{StaticResource FabColor}" />
-        <Setter Property="ColorPressed"
-                Value="{StaticResource FabPressedColor}" />
-        <Setter Property="ColorRipple"
-                Value="{StaticResource FabPressedColor}" />
+    <Style x:Key="floating-action-button" TargetType="Button">
```
kspearrin: let's call the style `btn-fab` to be similar to other button style naming conventions.

src/Core/Services/LiteDbStorageService.cs
```diff
@@ -68,7 +68,7 @@ public async Task SaveAsync<T>(string key, T obj)
         public async Task RemoveAsync(string key)
         {
             await InitAsync();
-            _collection.Delete(i => i.Id == key);
+            _collection.DeleteMany(i => i.Id == key);
```
kspearrin: Did Delete go away in v5? This is an Id delete, so should only ever be one record deleted with it.

src/Core/Services/PasswordGenerationService.cs
```diff
@@ -51,52 +51,7 @@ public async Task<string> GeneratePasswordAsync(PasswordGenerationOptions option
...
-            {
-                options.Length = 10;
-            }
-            var minLength = options.MinSpecial.GetValueOrDefault() + options.MinLowercase.GetValueOrDefault() +
-                options.MinNumber.GetValueOrDefault() + options.MinUppercase.GetValueOrDefault();
-            if(options.Length < minLength)
-            {
-                options.Length = minLength;
-            }
+            SanitizePasswordLength(options, true);
```
kspearrin: This looks like vince's stuff from a previous commit. What happened here?

src/App/App.csproj
mportune-bw: It should but for some reason isn't.  I found several instances of others who had to do the same thing with 2.4.1.

src/Core/Services/LiteDbStorageService.cs
mportune-bw: Delete no longer accepts the lambda expression but DeleteMany does.  I'll need to convert the expression into a `LiteDB.BsonValue` to use Delete.

src/Core/Services/PasswordGenerationService.cs
mportune-bw: Good question - I've been rebasing but maybe something got screwy - let me try again.

src/Core/Services/LiteDbStorageService.cs
mportune-bw: Rethinking this.. If I'm reading the expression correctly it can technically return more than one value (even if there is only one match).  Seems like they changed the method name to be more explicit, but the old `Delete` was performing the same action that `DeleteMany` is now.  How do you want to narrow down the correct ID to pass to the new `Delete`?

src/Core/Services/PasswordGenerationService.cs
mportune-bw: Found the problem locally but now it's complaining about the branches diverging and won't let me push until I pull - which will re-introduce the problem.  Not sure how to fix this properly so I may just start fresh.

src/App/Services/MobilePlatformUtilsService.cs
mportune-bw: The params are now `title` and `reason` (was originally only `region`).  I'm not sure why they did this because from what I can tell one is used with Android and the other with iOS, yet both serve the same purpose.

src/Core/Services/LiteDbStorageService.cs
kspearrin: `DeleteMany` is fine.

src/App/App.csproj
kspearrin: Ok

mportune-bw:
Replaced with PR #789

PR #789, opened 2020-03-20
Project lib updates and migration

mportune-bw:
**Summary**:

Update to all 3rd party libs used in the project.  Please make sure you're using the latest stable Visual Studio (just recently updated) as well as the latest Xamarin.Android and Xamarin.iOS SDKs.  If using Rider, make sure it's referencing the Xamarin SDKs from Visual Studio (the built-in versions aren't recent enough)

Occasionally when installing to Android the following error may occur: 

`The "CopyGeneratedJavaResourceClasses" task was not given a value for the required parameter "SourceTopDirectory".`

Simply re-run the install process and it should proceed normally.

**Additional Details**:

Minimum supported Android version now `5.0` (SDK `21`)
   - Removed `MultiDex` support as Android 5.0+ runtime supports multiple dex natively

`Xamarin.Forms`: Updated to `4.5.0.356` (from `4.4.0.991265`)

`Xamarin.Essentials`: Updated to `1.5.1` (from `1.3.1`)

`Xamarin.Android.Support*`: Updated to all appropriate `Xamarin.AndroidX*` and `Xamarin.Google*` libs
   - Updated all relevant imports

`Xamarin.Firebase.Messaging`: Updated to `71.1740.0` (from `60.1142.1`)

`Portable.BouncyCastle`: Updated to `1.8.6` (from `1.8.5.2`)

`LiteDB`: Updated to `5.0.4` (from `4.1.4`) which required these modifications:
   - Changed `_collection` type to `ILiteCollection` (from `LiteCollection`)
   - Added `Upgrade=true;` to DB constructor as the DB file format was changed
   - Changed collection `.Delete` to `.DeleteMany` to support the existing expression

`Plugin.Fingerprint`: Updated to `2.1.1` (from `1.4.9`) which required these modifications:
   - Removed `CustomFingerprintDialogFragment.cs` as it's no longer necessary
   - Added second `text` arg in `AuthenticationRequestConfiguration` constructor as the usage now varies between platforms

`ZXing.Net.Mobile.Forms`: Updated to `2.4.1` (from `2.1.47`)

`ZXing.Net.Mobile`: Added `2.4.1` as it's now required alongside `ZXing.Net.Mobile.Forms`)

`Refractored.FloatingActionButtonForms`: Removed as it's not compatible with Forms 4.5 & AndroidX
   - Replaced with visibly identical custom implementation, no lib required


src/Android/Android.csproj
```diff
@@ -168,7 +163,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\App\App.csproj">
-      <Project>{9F1742A7-7D03-4BB3-8FCD-41BC3002B00A}</Project>
+      <Project>{EE44C6A1-2A85-45FE-8D9B-BF1D5F88809C}</Project>
```
kspearrin: I am confused about why this changed. Isn't this the project's ID?

mportune-bw: VSmac did that when it restructured the other parts of this file after the move to Forms 4.5.  Some searching shows it's apparently just VSmac being VSmac - I'll test it with the original value...

mportune-bw: Didn't seem to phase anything, reverted.

PR #795, opened 2020-03-25
Added Quick Settings tile for triggering accessibility autofill

mportune-bw:
The user can now add a "scan & fill" quick setting tile on Android 7.0+ via the Android's tile edit/config mechanism.  It uses accessibility to perform the autofill and can be used in addition to, or in lieu of, the accessibility overlay (by leaving the overlay permission disabled they can safely use the tile only - the missing-overlay-permission toast only appears if the tile isn't enabled).

Fun stuff:

- You can only collapse the quick settings dropdown by launching an activity or opening a dialog.  Once I accepted defeat here I resorted to using a dialog for the "you need to enable accessibility" error, and an activity to trigger the broadcast message to the accessibility service to perform a field search.
- Without waiting 250ms after receiving the trigger message, the window parser would routinely only see the system UI which we ignore by default.  Can't beat a good nap.

### DIFF
--- a/src/Android/Accessibility/AccessibilityActivity.cs
+++ b/src/Android/Accessibility/AccessibilityActivity.cs
@@ -82,7 +82,7 @@ namespace Bit.Droid.Accessibility
 
         private void HandleIntent(Intent callingIntent, int requestCode)
         {
-            if(callingIntent?.GetBooleanExtra("autofillTileClicked", false) == true)
+            if(callingIntent?.GetBooleanExtra("autofillTileClicked", false) ?? false)
             {
                 Intent.RemoveExtra("autofillTileClicked");
                 var messagingService = ServiceContainer.Resolve<IMessagingService>("messagingService");
--- a/src/Android/Accessibility/AccessibilityActivity.cs
+++ b/src/Android/Accessibility/AccessibilityActivity.cs
@@ -4,6 +4,8 @@ using Android.OS;
 using Android.Runtime;
 using Android.Views;
 using System;
+using Bit.Core.Abstractions;
+using Bit.Core.Utilities;
 
 namespace Bit.Droid.Accessibility
 {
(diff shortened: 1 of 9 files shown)


src/Android/Accessibility/AccessibilityActivity.cs
```diff
@@ -78,6 +80,21 @@ protected override void OnActivityResult(int requestCode, [GeneratedEnum] Result
             Finish();
         }
 
+        private void HandleIntent(Intent callingIntent, int requestCode)
+        {
+            if(callingIntent?.GetBooleanExtra("autofillTileClicked", false) == true)
```
kspearrin: Do we need the `=== true` here? It also seems like you are doing a null check with `?.` here, so I would write it as:

```
callingIntent?.GetBooleanExtra("autofillTileClicked", false) ?? false
```

src/Android/Tiles/AutofillTileService.cs
```diff
@@ -0,0 +1,80 @@
...
+    [Service(Permission = Android.Manifest.Permission.BindQuickSettingsTile, Label = "@string/ScanAndFill",
+        Icon = "@drawable/shield")]
+    [IntentFilter(new string[] { ActionQsTile })]
+    [Register("com.x8bit.bitwarden.AutofillTileService")]
+    public class AutofillTileService : TileService
+    {
+        public override void OnTileAdded()
+        {
+            base.OnTileAdded();
+            AccessibilityHelpers.IsAutofillTileAdded = true;
```
kspearrin: Does this only work whenever you first turn on the tile? Is this function called again when the device restarts, for example?

src/Android/Accessibility/AccessibilityService.cs
```diff
@@ -48,6 +49,50 @@ public class AccessibilityService : Android.AccessibilityServices.AccessibilityS
...
+            AccessibilityHelpers.IsAccessibilityBroadcastReady = true;
+        }
+
+        public override void OnDestroy()
+        {
+            AccessibilityHelpers.IsAccessibilityBroadcastReady = false;
+            _broadcasterService.Unsubscribe(nameof(AccessibilityService));
+        }
+        
+        private void OnAutofillTileClick()
```
kspearrin: private functions should generally be at the bottom under the main public ones.

src/Android/Tiles/AutofillTileService.cs
mportune-bw: I thought it did but I fooled myself apparently.  Persisting in storage now.

PR #810, opened 2020-04-01
Conversion of HockeyApp to AppCenter for crash reporting

mportune-bw:
Validated working for Android & iOS in AppCenter portal, including metadata as attachment to match hockey's implementation.


src/iOS.Autofill/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.AppCenter" version="3.1.0" targetFramework="xamarinios10" />
```
kspearrin, line 3: Not sure why this happened. This is how old versions of nuget defined packages. These should be paclage references in the csproj file, like all others.

src/iOS.Extension/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
```
kspearrin, line 2: Not sure why this happened. This is how old versions of nuget defined packages. These should be paclage references in the csproj file, like all others.

src/iOS.Autofill/packages.config
mportune-bw: Didn't realize that was an older style - assumed it was something about the package itself (not sure why it happened either, I'd blame it on Rider but I've added other nuget packages without issue).  Changed it by hand and it seems happy.

PR #822, opened 2020-04-06
Add support for Vivaldi snapshot browser

mportune-bw:
Add autofill (accessibility) support for Vivaldi snapshot browser

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1169444489336079/1170066913744691) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -77,6 +77,7 @@ namespace Bit.Droid.Accessibility
             new Browser("jp.co.fenrir.android.sleipnir_black", "url_text"),
             new Browser("jp.co.fenrir.android.sleipnir_test", "url_text"),
             new Browser("com.vivaldi.browser", "url_bar"),
+            new Browser("com.vivaldi.browser.snapshot", "url_bar"),
             new Browser("com.feedback.browser.wjbrowser", "addressbar_url"),
         }.ToDictionary(n => n.PackageName);
 
--- a/src/Android/Autofill/AutofillHelpers.cs
+++ b/src/Android/Autofill/AutofillHelpers.cs
@@ -62,6 +62,7 @@ namespace Bit.Droid.Autofill
             "org.mozilla.rocket",
             "org.torproject.torbrowser",
             "com.vivaldi.browser",
+            "com.vivaldi.browser.snapshot",
         };
 
         // The URLs are blacklisted from autofilling


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -77,6 +77,7 @@ public static class AccessibilityHelpers
             new Browser("jp.co.fenrir.android.sleipnir_black", "url_text"),
             new Browser("jp.co.fenrir.android.sleipnir_test", "url_text"),
             new Browser("com.vivaldi.browser", "url_bar"),
+            new Browser("com.vivaldi.browser.snapshot", "url_bar"),
```
kspearrin, line 80: Do we also need to add this to Autofill Helpers CompatBrowsers?

mportune-bw: It doesn't appear to make a difference (neither vivaldi seems to work with autofill, at least for me).  I can add it anyway if you'd like.

kspearrin: Might as well add it

contribucious:
_Hello,_

@mportune-bw Is there a specific reason for not adding **com.vivaldi.browser.snapshot** to [autofillservice.xml](https://github.com/bitwarden/mobile/blob/master/src/Android/Resources/xml/autofillservice.xml) too? As was the case with **com.vivaldi.browser** when [it was added](https://github.com/bitwarden/mobile/commit/f657edf195d4770c610cf2cd03b88a5f00d5d033)?

@kspearrin More generally, is [autofillservice.xml](https://github.com/bitwarden/mobile/blob/master/src/Android/Resources/xml/autofillservice.xml) still used? If yes, why so many "package names" missing when compared to [AutofillHelpers.cs](https://github.com/bitwarden/mobile/blob/master/src/Android/Autofill/AutofillHelpers.cs)? Namely:

```
# OK, these are special cases (because listed as working natively) ...
com.duckduckgo.mobile.android
org.mozilla.focus
org.mozilla.klar

# ... but concerning these cases? (working using the compatibility shim)
com.android.browser
com.opera.touch
com.qwant.liberty
com.vivaldi.browser.snapshot
org.torproject.torbrowser
```

_Thanks._

kspearrin:
@mportune-bw Can you evaluate if we still need to be using autofillservice.xml definitions?

mportune-bw:
Looks like yes, if my understanding of the autofill service compatibility mode is correct.  #834 brings it all back together.  (Thanks for the pointer @contribucious )

contribucious:
@mportune-bw No problem and thank you! :+1:

contribucious:
@kspearrin / @mportune-bw :arrow_down: 

___
> @mportune-bw Can you evaluate if we still need to be using autofillservice.xml definitions?
___

To maintain a list of browsers in compatibility mode, considering that Android requires the [use of this dedicated XML file](https://developer.android.com/reference/android/service/autofill/AutofillService#compatibility-mode) _anyway_, why not rather reason in the opposite direction therefore, @kspearrin?

## Knowing that ...

1. The **data uniqueness** is important (easier maintenance, no desynchronization between two same data sets possible / thus no more warning comment to put for future contributors, ...);
2. We **already have a perfect file** which is: standardized (yay), dedicated (yay) and in a file format adapted to data (fluctuating data);

... and that you want to manage this data set from CompatBrowsers in your code, why not just populate it from this XML file, therefore? Below ... code for you on this subject! Feel free to do whatever you want with it, then!

## Code

### Basic illustration
:arrow_forward: https://dotnetfiddle.net/RSDPuP (code will auto run)

### Implementation (simple version: _.FirstAttribute_ used)
:arrow_forward: https://dotnetfiddle.net/5LffXY (code will auto run)

### Implementation (improved version: tag/attribute names used)
:arrow_forward: https://dotnetfiddle.net/jANQX4 (code will auto run)

PR #825, opened 2020-04-09
Additional accessibility tweaks

mportune-bw:
- Removed dependency on MainActivity for acquiring resource context from the `GetSystemResourceDimenPx` helper method as the "current" activity may not be ours (confirmed via invalid cast exceptions in crash logs).  Now using the resource context from our own service.
- Added some smarts to detect the presence of the autofill service prompt and back off if it's visible (we want the autofill service to take priority over accessibility - in most cases this happens automatically but in situations where it doesn't, this should help).
- Moved time-based overload protection (as well as new autofill detection) in front of overlay permission warning toast.  I believe the false positives (negatives?) returned by `Settings.CanDrawOverlays(...)` may be caused by excessive pounding on the method by accessibility events.  This appears to have stopped the erroneous toasts in places I was able to reproduce it, but additional user feedback is needed as I can't reproduce it in all the places reported.

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using Android.Content;
-using Android.Content.Res;
 using Android.Graphics;
 using Android.OS;
 using Android.Provider;
@@ -11,7 +10,6 @@ using Android.Views.Accessibility;
 using Android.Widget;
 using Bit.App.Resources;
 using Bit.Core;
-using Plugin.CurrentActivity;
 
 namespace Bit.Droid.Accessibility
 {
@@ -368,8 +366,8 @@ namespace Bit.Droid.Accessibility
             return layoutParams;
         }
 
-        public static Point GetOverlayAnchorPosition(AccessibilityNodeInfo anchorView, int overlayViewHeight, 
-            bool isOverlayAboveAnchor)
+        public static Point GetOverlayAnchorPosition(AccessibilityService service, AccessibilityNodeInfo anchorView, 
+            int overlayViewHeight, bool isOverlayAboveAnchor)
         {
             var anchorViewRect = new Rect();
             anchorView.GetBoundsInScreen(anchorViewRect);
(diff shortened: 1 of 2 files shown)


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -501,6 +501,21 @@ public static int GetInputMethodHeight(IEnumerable<AccessibilityWindowInfo> wind
...
+            {
+                foreach (var window in windows)
+                {
+                    if (window.Title?.ToLower().Contains("autofill") ?? false)
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
```
cscharf, line 517: can condense/simplify to:
`return windows?.Any(w => w.Title?.ToLower()?.Contains("autofill") ?? false) ?? false;`

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -515,24 +530,23 @@ public static int GetNodeHeight(AccessibilityNodeInfo node)
...
+        private static int GetSystemResourceDimenPx(AccessibilityService service, string resName)
         {
-            var activity = (MainActivity)CrossCurrentActivity.Current.Activity;
             var barHeight = 0;
-            var resourceId = activity.Resources.GetIdentifier(resName, "dimen", "android");
+            var resourceId = service.Resources.GetIdentifier(resName, "dimen", "android");
             if (resourceId > 0)
             {
-                barHeight = activity.Resources.GetDimensionPixelSize(resourceId);
+                barHeight = service.Resources.GetDimensionPixelSize(resourceId);
```
cscharf, line 549: To save allocation on the heap you can simply return the result of this call, and below the fall-back would be to return zero explicitly (vs. using a variable for a single possible assignment).

src/Android/Accessibility/AccessibilityHelpers.cs
mportune-bw: Still wrapping my head around Linq :)  I think I follow this one but can you confirm if the `?` after `ToLower()` is actually necessary?  I would assume if `Title` was null we would be returning false before getting to `ToLower()`.

src/Android/Accessibility/AccessibilityHelpers.cs
mportune-bw: Done

src/Android/Accessibility/AccessibilityHelpers.cs
cscharf: You're correct, the `?.` operator does short-circuit after the first null instance, in this case `?.Title`; so you could remove the second one after `ToLower()` (just a bad habit, but ToLower() would not also return `null` if we've already exited from title, just odd because `string` is not a value-type in .net but is a special snowflake type).

src/Android/Accessibility/AccessibilityService.cs
```diff
@@ -249,6 +246,12 @@ private void CancelOverlayPrompt()
 
         private void OverlayPromptToAutofill(AccessibilityNodeInfo root, AccessibilityEvent e)
         {
+            if (Java.Lang.JavaSystem.CurrentTimeMillis() - _lastAutoFillTime < 1000 ||
+                AccessibilityHelpers.IsAutofillServicePromptVisible(Windows))
+            {
+                return;
+            }
```
kspearrin, line 253: Nice hack!

PR #835, opened 2020-04-15
[Auto-Logout] Update Service layer logic

vincentsalucci:
## Code Changes
- **LockAlarmReceiver.cs**: Updated call to use refactored name `CheckVaultTimeoutAsync`
- **App.xaml.cs**: Updated Constants key reference to `VaultTimeoutKey`
- **SettingsPageViewModel.cs**: Updated Constants key reference to `VaultTimeoutKey`.  Added placeholder for new `SetVaultTimeoutOptionsAsync` method.
- **MobileStorageService.cs**: Added new action key to `preferenceStorageKeys`.
- **AppHelper.cs**: Updated Constants key reference to `VaultTimeoutKey`.  Added default value for `VaultTimeoutActionKey` and set it to `lock`
- **ITokenService.cs**: Added `ToggleTokensAsync` method reference
- **IVaultTimeoutService.cs**: Name refactors `CheckVaultTimeoutAsync` and `SetVaultTimeoutActionsAsync`. Added `LogOutAsync`.
- **Constants.cs**: Refactored `LockOptionKey` to `VaultTimeoutKey`. Added `VaultTimeoutActionKey`
- **CryptoService.cs**: Updated Constants key reference to `VaultTimeoutKey`
- **TokenService.cs**: Updated logic for whether or not to save token/refresh token based on the vault timeout and action saved. Added `ToggleTokensAsync` method which will properly clear out the tokens if necessary.
- **VaultTimeoutService.cs**: Added `TokenService` dependency.  Added `loggedOutCallback` dependency. Added logic for log out actions.
- **ServiceContainer.cs**: Updated init for `VaultTimeoutService`. Added log out actions through `messengerService`.
- **AppDelegate.cs**: Updated call to use refactored name `CheckVaultTimeoutAsync`

### DIFF
--- a/src/Core/Services/TokenService.cs
+++ b/src/Core/Services/TokenService.cs
@@ -35,6 +35,13 @@ namespace Bit.Core.Services
         {
             _token = token;
             _decodedToken = null;
+
+            if (await SkipTokenStorage())
+            {
+                // If we have a vault timeout and the action is log out, don't store token
+                return;
+            }
+            
             await _storageService.SaveAsync(Keys_AccessToken, token);
         }
 
@@ -51,6 +58,13 @@ namespace Bit.Core.Services
         public async Task SetRefreshTokenAsync(string refreshToken)
         {
             _refreshToken = refreshToken;
+
+            if (await SkipTokenStorage())
+            {
+                // If we have a vault timeout and the action is log out, don't store token
+                return;
+            }
+            
             await _storageService.SaveAsync(Keys_RefreshToken, refreshToken);
         }
 
@@ -62,6 +76,24 @@ namespace Bit.Core.Services
             }
             _refreshToken = await _storageService.GetAsync<string>(Keys_RefreshToken);
             return _refreshToken;
+        }
+
+        public async Task ToggleTokensAsync()
+        {
+            var token = await GetTokenAsync();
+            var refreshToken = await GetRefreshTokenAsync();
+            var timeout = await _storageService.GetAsync<int?>(Constants.VaultTimeoutKey);
+            var action = await _storageService.GetAsync<string>(Constants.VaultTimeoutActionKey);
+            if ((timeout != null || timeout == 0) && action == "logOut")
+            {
+                await ClearTokenAsync();
+                _token = token;
+                _refreshToken = refreshToken;
+                return;
+            }
+
+            await SetTokenAsync(token);
+            await SetRefreshTokenAsync(refreshToken);
         }
 
         public async Task SetTwoFactorTokenAsync(string token, string email)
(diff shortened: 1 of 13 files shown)


src/Core/Services/TokenService.cs
```diff
@@ -64,6 +78,24 @@ public async Task<string> GetRefreshTokenAsync()
             return _refreshToken;
         }
 
+        public async Task ToggleTokensAsync()
+        {
+            var token = await GetTokenAsync();
+            var refreshToken = await GetRefreshTokenAsync();
+            var timeout = await _storageService.GetAsync<int?>(Constants.VaultTimeoutKey);
+            var action = await _storageService.GetAsync<string>(Constants.VaultTimeoutActionKey);
+            if ((timeout != null || timeout == 0) && action == "logOut")
```
cscharf, line 87: Is this logic intended? if `timeout != null` wouldn't zero (`0`) be included in that test? Also, should use `timeout.HasValue` instead of `!= null` (IMO)

vincentsalucci: Great point - thanks for brining it up! I mentioned on my jslib PR regarding this exact check. I do agree with your suggestions. If we do make the change, I can use the private method `SkipTokenStorage` instead of the explicit checks here.

vincentsalucci: Updated PR.

src/Core/Services/TokenService.cs
```diff
@@ -64,6 +78,24 @@ public async Task<string> GetRefreshTokenAsync()
             return _refreshToken;
         }
 
+        public async Task ToggleTokensAsync()
+        {
+            var token = await GetTokenAsync();
+            var refreshToken = await GetRefreshTokenAsync();
+            var timeout = await _storageService.GetAsync<int?>(Constants.VaultTimeoutKey);
+            var action = await _storageService.GetAsync<string>(Constants.VaultTimeoutActionKey);
```
kspearrin, line 86: I dont see these being used anywhere?

src/Core/Services/VaultTimeoutService.cs
```diff
@@ -89,10 +96,19 @@ public async Task CheckLockAsync()
...
+                var action = await _storageService.GetAsync<string>(Constants.VaultTimeoutActionKey);
+                switch(action)
+                {
+                    case "lock":
+                        await LockAsync(true);
+                        break;
+                    case "logOut":
+                        await LogOutAsync();
+                        break;
+                }
```
kspearrin, line 111: If/else probably more simple here.

src/Core/Services/TokenService.cs
vincentsalucci: Correct - updated.

src/Core/Services/VaultTimeoutService.cs
vincentsalucci: Updated.

PR #842, opened 2020-04-17
Tracing bundles

mportune-bw:
- Enable AOT for performance boost (results in larger APK)
- Enable bundle support for Google Play (brings APK size down below where it was pre-AOT)
- Updated AndroidX Migration and Google Publisher deps


src/Android/Android.csproj
```diff
@@ -61,7 +61,7 @@
     <JavaMaximumHeapSize>1G</JavaMaximumHeapSize>
     <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
     <EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>
-    <AndroidLinkMode>Full</AndroidLinkMode>
```
kspearrin, line 64: So linker is no longer necessary?

mportune-bw: Full linking isn't supported with AOT, only "assemblies only" (which is what happens when this line is removed outright)

PR #844, opened 2020-04-20
[Auto-Logout] Settings Refactor - Application Layer Part 2

vincentsalucci:
## Code Changes
- **SettingsPage.xaml.cs**: Added action for clicking on the new action row. Refactored name of references.
- **SettingsPageViewModel.cs**: Refactored naming schemes to use `VaultTimeout`. Added `VaultTimeoutActions` list. Get/set stored values for timeout/action. Applied saving functions for both timeout and action. Added UI for new `VaultTimeoutAction` row (fixed but with biometrics ordinal). Added helper methods for retrieving values from list keys.
- **AppHelper.cs**: Fixed issue with retrieving potentially saved timeout action during install process.
- **String resources**: Updated existing strings (that will not need new translations) to use `VaultTimeout` scheme. Added new strings

## iOS
0-timeout-options
1-timeout-logout-warning
## Android
0-android-vault-timeout
1-android-log-out-warning


kspearrin:
Can you revert the changes to the other language files? We only need to make these changes to the english resources.

vincentsalucci:
> Can you revert the changes to the other language files? We only need to make these changes to the english resources.

Even for string resources with different names but the same value? (this is what I was asking about earlier)

On the surface it makes sense that if I change how a string is referenced, those should also be carried through to the already existing language references. However, when adding a NEW string with no current translation, only touch the english resource file (which is what is currently happening)

Just want to clarify the above before reverting.

kspearrin:
> 
> 
> > Can you revert the changes to the other language files? We only need to make these changes to the english resources.
> 
> Even for string resources with different names but the same value? (this is what I was asking about earlier)
> 
> On the surface it makes sense that if I change how a string is referenced, those should also be carried through to the already existing language references. However, when adding a NEW string with no current translation, only touch the english resource file (which is what is currently happening)
> 
> Just want to clarify the above before reverting.

Hmm, I am not sure what will happen here. I don't think I have done a language key change like this in the past. I usually don't touch existing string keys for this reason.

vincentsalucci:
> > > Can you revert the changes to the other language files? We only need to make these changes to the english resources.
> > 
> > 
> > Even for string resources with different names but the same value? (this is what I was asking about earlier)
> > On the surface it makes sense that if I change how a string is referenced, those should also be carried through to the already existing language references. However, when adding a NEW string with no current translation, only touch the english resource file (which is what is currently happening)
> > Just want to clarify the above before reverting.
> 
> Hmm, I am not sure what will happen here. I don't think I have done a language key change like this in the past. I usually don't touch existing string keys for this reason.

I can for sure just revert the name changes and then ADD new string refs while DELETING the old ones just in the english res file if that's how we want to proceed. I'm fine either way!

cscharf:
> > > > Can you revert the changes to the other language files? We only need to make these changes to the english resources.
> > > 
> > > 
> > > Even for string resources with different names but the same value? (this is what I was asking about earlier)
> > > On the surface it makes sense that if I change how a string is referenced, those should also be carried through to the already existing language references. However, when adding a NEW string with no current translation, only touch the english resource file (which is what is currently happening)
> > > Just want to clarify the above before reverting.
> > 
> > 
> > Hmm, I am not sure what will happen here. I don't think I have done a language key change like this in the past. I usually don't touch existing string keys for this reason.
> 
> I can for sure just revert the name changes and then ADD new string refs while DELETING the old ones just in the english res file if that's how we want to proceed. I'm fine either way!

That's how I would roll it @vincentsalucci... @kspearrin , even with this approach we could end up with orphaned keys in the translation files, if we're okay with that, then what Vince is proposing above should be okay; but otherwise we should probably clean up the other language files as well.

vincentsalucci:
> > > > > Can you revert the changes to the other language files? We only need to make these changes to the english resources.
> > > > 
> > > > 
> > > > Even for string resources with different names but the same value? (this is what I was asking about earlier)
> > > > On the surface it makes sense that if I change how a string is referenced, those should also be carried through to the already existing language references. However, when adding a NEW string with no current translation, only touch the english resource file (which is what is currently happening)
> > > > Just want to clarify the above before reverting.
> > > 
> > > 
> > > Hmm, I am not sure what will happen here. I don't think I have done a language key change like this in the past. I usually don't touch existing string keys for this reason.
> > 
> > 
> > I can for sure just revert the name changes and then ADD new string refs while DELETING the old ones just in the english res file if that's how we want to proceed. I'm fine either way!
> 
> That's how I would roll it @vincentsalucci... @kspearrin , even with this approach we could end up with orphaned keys in the translation files, if we're okay with that, then what Vince is proposing above should be okay; but otherwise we should probably clean up the other language files as well.

Updated!

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -371,5 +407,15 @@ public void BuildList()
...
+
+        private string getVaultTimeoutActionFromKey(string key)
+        {
+            return _vaultTimeoutActions.FirstOrDefault(o => o.Key == key).Value;
+        }
+
+        private int? getVaultTimeoutFromKey(string key)
+        {
+            return _vaultTimeouts.FirstOrDefault(o => o.Key == key).Value;
+        }
```
kspearrin, line 419: Should be PascalCased

src/App/Resources/AppResources.resx
```diff
@@ -518,23 +518,29 @@
...
     <value>1 minute</value>
   </data>
-  <data name="LockOption4Hours" xml:space="preserve">
+  <data name="VaultTimeout4Hours" xml:space="preserve">
     <value>4 hours</value>
   </data>
-  <data name="LockOptionImmediately" xml:space="preserve">
+  <data name="VaultTimeoutImmediately" xml:space="preserve">
     <value>Immediately</value>
   </data>
```
kspearrin, line 535: Should we just call these 15Minutes, 1Hour, etc and drop the prefix altogether?

vincentsalucci: I don't really have a problem with that since they don't needed have any real dependency with the VaultTimeout prefix.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
vincentsalucci: ....sorry..... pure muscle memory. Updating.

PR #847, opened 2020-04-22
Allow a BiometricPrompt to succeed after an initial failure

### LINKED ISSUE #791: [Android] Tresor won't unlock if fingerprint mismatches at the first try
## Describe the Bug
I setup Bitwarden to unlock with my fingerprint. If I want to use Bitwarden the following happens:
- I open the app
- fingerprint prompt appears
- first fingerprint mismatches
- second (or a following try) succeeds
- fingerprint prompt disappears
- tresor still locked

After that nothing happens. I have to tap on "use biometric data to unlock" or enter the password. That happens again and again until I succeed at the first try with my fingerprint.

## Steps To Reproduce

1. Setup Bitwarden to unlock with fingerprint
2. Lock Bitwarden if needed
3. Open Bitwarden / use biometric data to unlock
4. Fail on the first try at unlocking with your fingerprint
5. Succeed on a following try
6. Fingerprint prompt disappears and Bitwarden won't unlock.

## Expected Result

If the fingerprint matches at any try (even if the first try fails) your tresor should unlock.

## Actual Result

Nothing happens / tresor won't unlock. You're still on the main screen and have to unlock Bitwarden via password or biometric data.

## Screenshots or Videos

Can't take a screenshot or video from Bitwarden. The app blocks both of it.

## Environment

- Device: Huawei P30
- Operating system: Android 10 / EMUI 10.0.0
- Build Version (go to "Settings" → "About" in the app): 10.0.0.186 (C432E22R2P5)
- Is this a Beta release? No
- Bitwarden Version: 2.3.1

watk:
Previously a call to BiometricPrompt.AuthenticationCallback#OnAuthenticationFailed()
was treated as though an unrecoverable failure had occurred. However this is called on
each failed fingerprint match, for example, and is not a terminal failure. Now these
intermittent failures are ignored and a call to #OnAuthenticationError() is recognised
as an unrecoverable failure instead.

Fixes #791

### DIFF
--- a/src/Android/Services/DeviceActionService.cs
+++ b/src/Android/Services/DeviceActionService.cs
@@ -446,7 +446,8 @@ namespace Bit.Droid.Services
                     new BiometricAuthenticationCallback
                     {
                         Success = authResult => result.TrySetResult(true),
-                        Failed = () => result.TrySetResult(false),
+                        Error = () => result.TrySetResult(false),
+                        Failed = () => { },
                         Help = (helpCode, helpString) => { }
                     });
                 return result.Task;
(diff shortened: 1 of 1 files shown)


watk:
Great, thanks for your work!

PR #859, opened 2020-04-29
[Soft Delete] - Fix for iOS autofill index behavior

cscharf:
From prior PR comments, tested and working as expected with suggested changes.


kspearrin:
I am kind of confused. Didn't we just review most of this?

cscharf:
Yeah, it pulled in the other commits anyway even though they were already merged, PR closed.

PR #860, opened 2020-04-29
Fixes to uri parsing

mportune-bw:
Changed URI detection to utilize framework.  This fixes parsing failures on some (possibly a lot of) urls.

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -147,7 +147,7 @@ namespace Bit.Droid.Accessibility
             return uri;
         }
 
-        public static string ExtractUri(string uri, AccessibilityNodeInfo addressNode, Browser browser)
+        private static string ExtractUri(string uri, AccessibilityNodeInfo addressNode, Browser browser)
         {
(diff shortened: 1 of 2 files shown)


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -160,21 +160,13 @@ public static string ExtractUri(string uri, AccessibilityNodeInfo addressNode, B
...
                 {
-                    var parts = uri.Split(new string[] { ". " }, StringSplitOptions.None);
-                    if (parts.Length > 1)
-                    {
-                        var urlPart = parts.FirstOrDefault(p => p.StartsWith("http"));
-                        if (urlPart != null)
-                        {
-                            uri = urlPart.Trim();
-                        }
-                    }
```
kspearrin, line 177: This seems like something that was more custom. Are we sure it's not needed? I cannot recall.

mportune-bw: It's behind a check for `Build.VERSION.SdkInt <= BuildVersionCodes.KitkatWatch` which we no longer support.

PR #865, opened 2020-04-30
Additional URI parsing fix

mportune-bw:
Ran across an issue introduced by recent modifications where an expected failure did in fact, not fail, resulting in a 'no matches found' condition (calling `Uri.TryCreate` on a URL without a scheme worked, and passed along a scheme-less URI until it was rejected by subsequent code).

The solution (which lines up better with previous code) is to perform the http(s) scheme check / prefix creation _first_, and if not necessary _then_ fall back to attempting to create the URI with the string as-is.  This results in the scheme-less URI being properly prefixed with `http://`.  Also tested with the URLs fixed by our recent changes, and they still work properly.

I'm unsure why we didn't see this while testing the recent changes.  For reference, the uriString that exposed this was `rqpc45.synology.me:5000`

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -161,10 +161,6 @@ namespace Bit.Droid.Accessibility
             uri = browser.GetUriFunction(addressNode.Text)?.Trim();
             if (uri != null && uri.Contains("."))
             {
-                if (Uri.TryCreate(uri, UriKind.Absolute, out var uri2))
-                {
-                    return uri;
-                }
                 var hasHttpProtocol = uri.StartsWith("http://") || uri.StartsWith("https://");
                 if (!hasHttpProtocol && uri.Contains("."))
                 {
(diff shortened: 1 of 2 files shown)


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -161,18 +161,18 @@ private static string ExtractUri(string uri, AccessibilityNodeInfo addressNode,
             uri = browser.GetUriFunction(addressNode.Text)?.Trim();
             if (uri != null && uri.Contains("."))
             {
-                if (Uri.TryCreate(uri, UriKind.Absolute, out var uri2))
```
cscharf, line 164: Why not just call `return CoreHelpers.GetUri(uri)?.ToString() ?? uri;`

mportune-bw: Since the `Uri` class is a lot smarter than I am, I'm paranoid about the result of the string-to-uri-to-string not being identical to the original string, which is what we want to return here.

PR #867, opened 2020-05-01
[Auto Logout] iOS Autofill flow

vincentsalucci:
## Code Changes
- **CredentialProviderViewController.cs**: Added explicit `IsAuthed` conditionals that will launch a segue into the `Login` flow. Removed unnecessary method. Added setter for Credential Provider context into new view.
- **iOS.Autofill/LoginPasswordViewController.cs**: Created class for the autofill framework that will inherit from the shared code within the `Controllers` sub-module.
- **MainInterface.storyboard**: Added new `LoginPassword` segue that handles authentication for autofill service use.
- **iOS.Core/Controllers/LoginPasswordViewController.cs**: This controller was added to handle the shared logic between authentication to the server and autofill/extension usage.

## Known issues (must fix - any ideas/leads are greatly appreciated)
- ~~Hitting GO on the keyboard to launch the `LogInAsync` method bricks the application. Hitting `Log In` from the nav works fine.~~
-  After successfully authenticating and syncing (<- presumably) the logins list is empty for a known item (ex: github.com exists within the application, but after using the login segue and landing on the logins list, no results are found)
   -  Can anyone else test this with me: I'm actually seeing this succeed now.

## Upgrades (input needed on whether or not these should be added)
- Show Password option: **future release**
- ~~Get Master Password hint segue: not in scope~~

## What's next?
- 2FA authentication view
   - TODO: Duo Web View integration, YubiKey messages
   - TwoFactorSegue will be launched via the `LoginPasswordViewController`

## Login Credentials - AutoFill Service
0-login-creds-request

## Login Credentials Errors
1-login-creds-errors

## Login Credentials Items Found
2-login-found


src/iOS.Autofill/LoginPasswordViewController.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+using Foundation;
+using UIKit;
+
+namespace Bit.iOS.Autofill
+{
+	public partial class LoginPasswordViewController : Core.Controllers.LoginPasswordViewController
```
kspearrin, line 8: indenting seems off in this file

src/iOS.Autofill/MainInterface.storyboard
```diff
@@ -1,7 +1,9 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
kspearrin, line 1: I see a lot of changes in this file unrelated to your new View Controller. Might be fine, but I don't know.

src/iOS.Core/Controllers/LoginPasswordViewController.cs
```diff
@@ -0,0 +1,218 @@
...
+            BaseSubmitButton.Title = AppResources.LogIn;
+
+            var descriptor = UIFontDescriptor.PreferredBody;
+            
+            MasterPasswordCell.TextField.SecureTextEntry = true;
+            MasterPasswordCell.TextField.ReturnKeyType = UIReturnKeyType.Go;
+            MasterPasswordCell.TextField.ShouldReturn += (UITextField tf) =>
+            {
+                // TODO This isn't working - bricks application
+                LogInAsync().GetAwaiter().GetResult();
```
cscharf, line 59: This pattern will commonly cause a deadlock issue with thread blocking in .net. To properly call an async method syncronously, the method body being called (and any downstream methods) has to play along as well using `.ConfigureAwait(false)` on the return `await Task` result. You'll run into Exception unwrapping problems as well in this context (so basically it could be failing/throwing exceptions but you'll never know it and it'll never be logged because it's in an invalid thread-state).

If absolutely necessary you can also use the [nested message loop hack](https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#the-nested-message-loop-hack).

*ORrrrr*, it looks like there's a `.RunSynchronously()` method for `Task` and `Task<T>` in dotnet core 3.1 (from the docs): https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.runsynchronously?view=netcore-3.1

src/iOS.Core/Controllers/LoginPasswordViewController.cs
```diff
@@ -0,0 +1,218 @@
...
+                return true;
+            };
+
+            TableView.RowHeight = UITableView.AutomaticDimension;
+            TableView.EstimatedRowHeight = 70; // Should this be adjusted for two cells?
+            TableView.Source = new TableSource(this);
+            TableView.AllowsSelection = true;
+            
+            // Attempt to programatically set the user's email
+            EmailCell.TextField.Text = _storageService.GetAsync<string>("rememberedEmail").GetAwaiter().GetResult();
```
cscharf, line 69: This is another example of the pattern above, however `IStorageService` implementation here may only be going off of in-memory, but another implementation that actually waits on I/O operations or engages in true async behavior may cause a similar deadlock scenario.

src/iOS.Core/Controllers/LoginPasswordViewController.cs
```diff
@@ -0,0 +1,218 @@
...
+                await _deviceActionService.ShowLoadingAsync(AppResources.LoggingIn);
+                var response = await _authService.LogInAsync(email, password);
+                await _deviceActionService.HideLoadingAsync();
+                if (response.TwoFactor)
+                {
+                    // TODO Segue to Two Factor page
+                }
+                else
+                {
+                    var task = Task.Run(async () => await _syncService.FullSyncAsync(true));
```
cscharf, line 133: This method is already async, should just be reduced to:
`await _syncService.FullSyncAsync(true);`

vincentsalucci: Updaed!

src/iOS.Core/Controllers/LoginPasswordViewController.cs
vincentsalucci: Updated - Xamarin provides asnyc overrides for lifecycle methods.

src/iOS.Core/Controllers/LoginPasswordViewController.cs
vincentsalucci: Thanks! Since I don't really need to wait for the result of this method, I've updated it to remove the Awaiter/Result pattern above. It seems to be working as intended now.

src/iOS.Autofill/LoginPasswordViewController.cs
vincentsalucci: Updated.

vincentsalucci:
@kspearrin @cscharf @mportune-bw - What are your opinions on the `Upgrades` section? Should we add those missing pieces of functionality?

src/iOS.Autofill/MainInterface.storyboard
vincentsalucci: Not sure about this note as of now. I wonder if it could be related to using VS generation/editing versus using Rider/Xcode generation/editing.

src/iOS.Autofill/LoginPasswordViewController.cs
kspearrin: Still looks off. Looks like it is using tabs or something.

kspearrin:
Re upgrades: Hint: no, show password: maybe later. Let's do 2fa for now.

PR #880, opened 2020-05-05
Support for establishing a username field without a password field

mportune-bw:
Accessibility:  Added support for establishing a username field without the presence of a password field, or for overriding a username field with a different one (think github homepage user/email/pass form).  As there doesn't appear to be a reliable way to make this work automagically, this takes the manual route where we simply maintain a list of sites and their corresponding username field IDs.

I opted not to use the entire URI as some of the variations were endless depending on locale, continuation path args, etc.  So we're keying off the authority and the end of the path to establish a site & page match, then on the view ID for a username field match.

Additional data was needed earlier in the process to establish the username field, so some operations were shuffled around to prevent them from running twice.

device-2020-05-05-175044

device-2020-05-05-181429

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -123,6 +123,7 @@ namespace Bit.Droid.Accessibility
             new KnownUsernameField("amazon.com","signin", "ap_email_login"),
             new KnownUsernameField("github.com", "", "user[login]-footer"),
             new KnownUsernameField("paypal.com","signin", "email"),
+            new KnownUsernameField("signin.aws.amazon.com","signin", "resolving_input"),
             new KnownUsernameField("signin.ebay.com","eBayISAPI.dll", "userid"),
           
         }.ToDictionary(n => n.UriAuthority);
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -115,6 +115,17 @@ namespace Bit.Droid.Accessibility
             "com.ss.squarehome2",
             "com.treydev.pns"
         };
+        
+        // Be sure to keep these entries sorted alphabetically
+        public static Dictionary<string, KnownUsernameField> KnownUsernameFields => new List<KnownUsernameField>
+        {
+            new KnownUsernameField("accounts.google.com","ServiceLogin", "Email"),
+            new KnownUsernameField("amazon.com","signin", "ap_email_login"),
+            new KnownUsernameField("github.com", "", "user[login]-footer"),
+            new KnownUsernameField("paypal.com","signin", "email"),
+            new KnownUsernameField("signin.ebay.com","eBayISAPI.dll", "userid"),
+          
+        }.ToDictionary(n => n.UriAuthority);
 
         public static void PrintTestData(AccessibilityNodeInfo root, AccessibilityEvent e)
         {
@@ -280,17 +291,50 @@ namespace Bit.Droid.Accessibility
             return nodes;
         }
 
-        public static void GetNodesAndFill(AccessibilityNodeInfo root, AccessibilityEvent e,
-            IEnumerable<AccessibilityNodeInfo> passwordNodes)
+        public static AccessibilityNodeInfo GetUsernameEditText(string uriString, 
+            IEnumerable<AccessibilityNodeInfo> allEditTexts)
         {
-            var allEditTexts = GetWindowNodes(root, e, n => EditText(n), false);
-            var usernameEditText = GetUsernameEditTextIfPasswordExists(allEditTexts);
-            FillCredentials(usernameEditText, passwordNodes);
-            allEditTexts.Dispose();
-            usernameEditText = null;
-        }
+            string uriKey = null;
+            string uriLocalPath = null;
+            if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
+            {
+                uriKey = uri.Authority;
+                uriLocalPath = uri.LocalPath;
+            }
 
-        public static AccessibilityNodeInfo GetUsernameEditTextIfPasswordExists(
+            if (!string.IsNullOrEmpty(uriKey))
+            {
+                // Uncomment this to log values necessary for username field discovery
+                // foreach (var editText in allEditTexts)
+                // {
+                //     System.Diagnostics.Debug.WriteLine(">>> uriKey: {0}, uriLocalPath: {1}, viewId: {2}", uriKey,
+                //         uriLocalPath, editText.ViewIdResourceName);
+                // }
+
+                if (KnownUsernameFields.ContainsKey(uriKey))
+                {
+                    var usernameField = KnownUsernameFields[uriKey];
+                    if (uriLocalPath.EndsWith(usernameField.UriPathEnd))
+                    {
+                        foreach (var editText in allEditTexts)
+                        {
+                            foreach (var usernameViewId in usernameField.UsernameViewId.Split(","))
+                            {
+                                if (usernameViewId == editText.ViewIdResourceName)
+                                {
+                                    return editText;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            // no match found, attempt to establish username field based on password field
+            return GetUsernameEditTextIfPasswordExists(allEditTexts);
+        }
+        
+        private static AccessibilityNodeInfo GetUsernameEditTextIfPasswordExists(
             IEnumerable<AccessibilityNodeInfo> allEditTexts)
         {
             AccessibilityNodeInfo previousEditText = null;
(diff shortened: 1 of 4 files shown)


src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -115,6 +115,17 @@ public static class AccessibilityHelpers
             "com.ss.squarehome2",
             "com.treydev.pns"
         };
+        
+        // Be sure to keep these entries sorted alphabetically
+        public static Dictionary<string, KnownUsernameField> KnownUsernameFields => new List<KnownUsernameField>
+        {
+            new KnownUsernameField("accounts.google.com","ServiceLogin", "Email"),
+            new KnownUsernameField("amazon.com","signin", "ap_email_login"),
```
cscharf, line 123: Given just the authority, how would this work with sub-domains that behave differently, such as `aws.amazon.com` which has a username field prompted first (both iAM and Root accounts), `path="signin", field id="resolving_input"`

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -280,17 +291,50 @@ public static void FillEditText(AccessibilityNodeInfo editTextNode, string value
...
+                if (KnownUsernameFields.ContainsKey(uriKey))
+                {
+                    var usernameField = KnownUsernameFields[uriKey];
+                    if (uriLocalPath.EndsWith(usernameField.UriPathEnd))
+                    {
+                        foreach (var editText in allEditTexts)
+                        {
+                            foreach (var usernameViewId in usernameField.UsernameViewId.Split(","))
+                            {
+                                if (usernameViewId == editText.ViewIdResourceName)
```
cscharf, line 323: Should this be a case-insensitive comparison?

src/Android/Accessibility/AccessibilityHelpers.cs
mportune-bw: It has to be added to the list (same approach as `accounts.google.com`, unless I'm misunderstanding the question?):

`new KnownUsernameField("signin.aws.amazon.com","signin", "resolving_input")`

device-2020-05-06-092205

src/Android/Accessibility/AccessibilityHelpers.cs
mportune-bw: Hm, not sure.  My intent was for an exact match, but I'm curious what your thoughts are.

src/Android/Accessibility/AccessibilityHelpers.cs
mportune-bw: Additional clarification - this PR is only for showing the overlay where it wouldn't show before.  The matching algorithms used when the overlay is clicked have not changed.

src/Android/Accessibility/AccessibilityHelpers.cs
cscharf: Actually, I think exact match makes sense, HTML is case-sensitive for Ids so we probably should be as well, thanks.

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -115,6 +115,18 @@ public static class AccessibilityHelpers
             "com.ss.squarehome2",
             "com.treydev.pns"
         };
+        
+        // Be sure to keep these entries sorted alphabetically
+        public static Dictionary<string, KnownUsernameField> KnownUsernameFields => new List<KnownUsernameField>
+        {
+            new KnownUsernameField("accounts.google.com","ServiceLogin", "Email"),
+            new KnownUsernameField("amazon.com","signin", "ap_email_login"),
```
kspearrin, line 123: Some formatting issues here with spacing.

mportune-bw: Sorry about that - thanks @cscharf for fixing

contribucious:
Hello @mportune-bw,
&nbsp;

After uncommenting the 5 lines below:

https://github.com/bitwarden/mobile/blob/ff994629de7cbe4d961257ff21b64654b5d13ffd/src/Android/Accessibility/AccessibilityHelpers.cs#L321-L326

... I did a series of tests and here are the results obtained.

:information_source: To be completely precise, I used System.Console.WriteLine (because `Release` generated instead of `Debug` in this case).
&nbsp;

# Results
&nbsp;

## Google :x:

```csharp
new KnownUsernameField("accounts.google.com", "ServiceLogin", "Email"),
```

<details>
<summary>View me ...</summary>
&nbsp;

Fails to get prompt.
Which seems logical because, _from Belgium at least_ (despite arriving on _accounts.google.com_ also), I get this whatever the display language:
```
>>> uriKey: accounts.google.com, uriLocalPath: /signin/v2/identifier, viewId: identifierId
```
:arrow_right: **Context:** Clicked on the `Sign in` button located at the top right of the google.be or google.com home page, leading to `https://accounts.google.com/signin/v2/identifier?[many_parameters_here]&flowEntry=ServiceLogin`.

:information_source: **Just for info:** In my case `https://accounts.google.com/` redirects me to `https://accounts.google.com/ServiceLogin?[many_parameters_here]` which redirects me to `https://accounts.google.com/signin/v2/identifier?[many_parameters_here]&flowEntry=ServiceLogin`.

:bulb: **Proposal:** The line below is functional **in my case** (need to check on your side now):
```csharp
new KnownUsernameField("accounts.google.com", "identifier", "identifierId"),
```
:arrow_right_hook: Mobile version test: OK  :heavy_check_mark: **/** Desktop version test: OK  :heavy_check_mark: **/** International test: OK :heavy_check_mark:
</details>
&nbsp;

## Amazon :information_source:

```csharp
new KnownUsernameField("amazon.com", "signin", "ap_email_login"),
```

<details>
<summary>View me ...</summary>
&nbsp;

### 1. Mobile version test: OK  :heavy_check_mark:

### 2. Desktop version test: NOK  :x:
:bulb: **Proposal:**
```csharp
new KnownUsernameField("amazon.com", "signin", "ap_email_login,ap_email"), // 1st = Mobile, 2nd = Desktop
```
:arrow_right: It is particularly essential here to put the mobile value (`ap_email_login`) before the desktop value (`ap_email`) because it turns out that `ap_email` is also used on the mobile version page but for another use (in an account creation subsection located in _first_ position(*), displayed instantly when a checkbox is checked).

(*) Note that a useless prompt will always be available in this subsection btw, probably as the first `input[type="email"]` field found.

:computer: _Desktop mode_ is used in particular on tablets, sometimes automatically forced via an option available on certain browsers. So, useful to have.

### 3. International test (if available): NOK  :x:
```csharp
>>> uriKey: www.amazon.fr, uriLocalPath: /ap/signin, viewId: ap_email_login
```
```csharp
>>> uriKey: www.amazon.co.uk, uriLocalPath: /ap/signin, viewId: ap_email_login
```
:arrow_right_hook: As you can see, requires the _same_ line but for: amazon.fr, amazon.co.uk, etc. _(For security reasons, a wildcard system for the TLD part is probably not ideal.)_
</details>
&nbsp;

## GitHub / PayPal / Amazon Web Services (AWS) :heavy_check_mark:

```csharp
new KnownUsernameField("github.com", "", "user[login]-footer"),
new KnownUsernameField("paypal.com", "signin", "email"),
new KnownUsernameField("signin.aws.amazon.com", "signin", "resolving_input"),
```

<details>
<summary>View me ...</summary>
&nbsp;

### 1. Mobile version test: OK  :heavy_check_mark:
### 2. Desktop version test: OK  :heavy_check_mark:
### 3. International test (if available): OK  :heavy_check_mark:
</details>
&nbsp;

## eBay :information_source:

```csharp
new KnownUsernameField("signin.ebay.com", "eBayISAPI.dll", "userid"),
```

<details>
<summary>View me ...</summary>
&nbsp;

### 1. Mobile version test: OK  :heavy_check_mark:
### 2. Desktop version test: OK  :heavy_check_mark:
### 3. International test (if available): NOK  :x:
```csharp
>>> uriKey: signin.befr.ebay.be, uriLocalPath: /ws/eBayISAPI.dll, viewId: userid
```
```csharp
>>> uriKey: signin.benl.ebay.be, uriLocalPath: /ws/eBayISAPI.dll, viewId: userid
```
```csharp
>>> uriKey: signin.ebay.fr, uriLocalPath: /ws/eBayISAPI.dll, viewId: userid
```
:arrow_right_hook: As you can see, requires the _same_ line but for: signin.befr.ebay.be, signin.benl.ebay.be, signin.ebay.fr, etc. _(A wildcard system for the subdomain part could perhaps be envisaged — for security reasons, the TLDs being listed, as for them, manually.)_
</details>
&nbsp;

contribucious:
Finally, it might be good to add support for inputs that do not have an ID, just having a name (common in forms). :thumbsup:

#### Currently no way, it seems, to add [Backblaze.com](https://secure.backblaze.com/user_signin.htm) ([info](https://en.wikipedia.org/wiki/Backblaze)):
```csharp
>>> uriKey: secure.backblaze.com, uriLocalPath: /user_signin.htm, viewId:
```
```html
<input class="bz-input email-field" name="email-field" type="text" placeholder="Email" value="" data-focused="true">
```

## Thank you again for all your work!

contribucious:
:new: :arrow_double_up: **Google** part of the long post updated! :relaxed:

PR #881, opened 2020-05-06
Fixed spacing on known username field list

cscharf:
Fixed spacing on known username field list


mportune-bw:
Changes look clean but there are conflicts - I think you need to rebase against master, which already contains the first two commits

PR #890, opened 2020-05-08
Soft delete feature

cscharf:
Final pull request for the soft-delete / trash feature for the mobile client, includes merge from `master` to catch-up the `soft-delete` branch and to resolve conflicts.


cscharf:
Prod server / API has been updated now, this would no longer cause issues with beta testing.

PR #913, opened 2020-05-20
Workaround for pasting into editor within scrollview

### LINKED ISSUE #747: Android: Unable to paste data into Notes section of an item
Tested on Samsung Galaxy S10 running Android 10

**Steps to Reproduce:**

- Copy text from an application other than Bitwarden
- Open the Bitwarden App
- Log into your vault
- Add a new Item
- Scroll to the "Notes" section
- Long Press in the field

**Expected Result:** The option to paste appears

**Actual Result:** Unable to paste.

**Notes:** I have validated the ability to paste works in iOS as well as in the desktop app and browser extensions.

mportune-bw:
Workaround provided by Xamarin team for allowing long-press-in-editor-within-scrollview system prompts to appear (copy/paste/etc).  Fixes #747

### DIFF
--- a/src/Android/Renderers/CustomEditorRenderer.cs
+++ b/src/Android/Renderers/CustomEditorRenderer.cs
@@ -12,6 +12,15 @@ namespace Bit.Droid.Renderers
         public CustomEditorRenderer(Context context)
             : base(context)
         { }
+        
+        // Workaround for issue described here:
+        // https://github.com/xamarin/Xamarin.Forms/issues/8291#issuecomment-617456651
+        protected override void OnAttachedToWindow()
+        {
+            base.OnAttachedToWindow();
+            EditText.Enabled = false;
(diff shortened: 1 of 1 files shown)


src/Android/Renderers/CustomEditorRenderer.cs
```diff
@@ -12,6 +12,15 @@ public class CustomEditorRenderer : EditorRenderer
         public CustomEditorRenderer(Context context)
             : base(context)
         { }
+        
+        // Workaround for issue described here:
+        // https://github.com/xamarin/Xamarin.Forms/issues/8291#issuecomment-617456651
+        protected override void OnAttachedToWindow()
+        {
+            base.OnAttachedToWindow();
+            EditText.Enabled = false;
```
cscharf, line 21: For the love of all that's created, please add a code comment as to where this hack is coming from :-)

PR #920, opened 2020-05-23
Add support for hidden passwords

Hinton:
Adds support for the hidden password functionality, companion PR to https://github.com/bitwarden/server/pull/743.

This PR has only been tested using the Android emulator.

### DIFF
--- a/src/App/Pages/Vault/AddEditPageViewModel.cs
+++ b/src/App/Pages/Vault/AddEditPageViewModel.cs
@@ -275,7 +275,7 @@ namespace Bit.App.Pages
         public string ShowPasswordIcon => ShowPassword ? "" : "";
         public string ShowCardCodeIcon => ShowCardCode ? "" : "";
         public int PasswordFieldColSpan => Cipher.ViewPassword ? 1 : 4;
-        public int TotpColumnSpan => _cipher.ViewPassword ? 1 : 2;
+        public int TotpColumnSpan => Cipher.ViewPassword ? 1 : 2;
 
         public void Init()
         {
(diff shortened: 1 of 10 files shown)


src/App/Pages/Vault/AddEditPageViewModel.cs
```diff
@@ -274,6 +274,8 @@ public bool HasCollections
         public bool ShowAttachments => Cipher.HasAttachments;
         public string ShowPasswordIcon => ShowPassword ? "" : "";
         public string ShowCardCodeIcon => ShowCardCode ? "" : "";
+        public int PasswordFieldColSpan => Cipher.ViewPassword ? 1 : 4;
+        public int TotpColumnSpan => _cipher.ViewPassword ? 1 : 2;
```
kspearrin, line 278: Why Cipher for one and _cipher for the other?

Hinton:
Would be great if someone could verify the behavior on an iOS device or emulator.

kspearrin:
I am not sure. It would probably be the same issue on desktop devices.

Hinton:
@mportune-bw https://support.google.com/accessibility/android/answer/6006977?hl=en Seems that the user can configure the behavior. Not sure if there are any accessibility properties that can disable it.

Hinton:
Setting https://docs.microsoft.com/sv-se/xamarin/xamarin-forms/app-fundamentals/accessibility/automation-properties#isinaccessibletree to false would probably solve it.

Hinton:
I did some testing with Talkback, seems that all passwords in the Bitwarden apps are spoken as "bullet" unless "Show password" is toggled.

cscharf:
Will require some testing via Test Flight for iOS. Can get merged (reverse compatible with older API) and handle any updates/changes in another PR/patch if necessary.

PR #929, opened 2020-05-29
[Auto Logout] Update lock option to be default value

vincentsalucci:
## Code Changes
- **App.xaml.cs**: Changed conditional logic to make lock option the default action
- **SettingsPageViewModel.cs**: Added default for `lock` if saved action was null
- **VaultTimeoutService.cs**: Changed conditional logic to make lock option the default action

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -339,13 +339,13 @@ namespace Bit.App
             else if (vaultTimeout == 0)
             {
                 var action = await _storageService.GetAsync<string>(Constants.VaultTimeoutActionKey);
-                if (action == "lock")
+                if (action == "logOut")
                 {
-                    await _vaultTimeoutService.LockAsync(true);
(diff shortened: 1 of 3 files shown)


src/App/App.xaml.cs
```diff
@@ -339,13 +339,13 @@ private async Task HandleVaultTimeoutAsync()
...
-                if (action == "lock")
+                if (action == "logOut")
                 {
-                    await _vaultTimeoutService.LockAsync(true);
+                    await _vaultTimeoutService.LogOutAsync();
                 }
                 else
                 {
-                    await _vaultTimeoutService.LogOutAsync();
+                    await _vaultTimeoutService.LockAsync(true);
```
cscharf, line 348: Looks like there's extra whitespace at the end of the line?

vincentsalucci: Updated.

PR #937, opened 2020-05-31
Feature sync on refresh

aaxdev:
Added a way to sync vault with the pull down gesture (optional) as suggested here  https://github.com/bitwarden/mobile/issues/457#issuecomment-521321130. 
The option is into the menu Settings/Sync.
Tested on real android device and emulator.


src/App/Pages/Settings/SyncPageViewModel.cs
```diff
@@ -11,25 +12,53 @@ public class SyncPageViewModel : BaseViewModel
...
+        public async Task InitAsync()
+        {
+            await SetLastSyncAsync();
+            EnableSyncOnRefresh = await _syncService.IsSyncOnRefreshEnabledAsync();
+            _inited = true;
+        }
+
+        public async Task UpdateSyncOnRefreshAsync()
+        {
+            if(_inited) await _storageService.SaveAsync(Constants.SyncOnRefreshKey, _syncOnRefresh);
```
kspearrin, line 59: We do not format `if` statements this way. Please use syntax with brackets.

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
```diff
@@ -92,12 +92,12 @@ protected async override void OnAppearing()
                 else if (message.Command == "syncCompleted")
                 {
                     await Task.Delay(500);
-                    Device.BeginInvokeOnMainThread(() =>
+                    Device.BeginInvokeOnMainThread(async() =>
```
kspearrin, line 95: formatting. needs whitespace

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
```diff
@@ -92,12 +92,12 @@ protected async override void OnAppearing()
...
                     {
                         IsBusy = false;
                         if (_vm.LoadedOnce)
                         {
-                            var task = _vm.LoadAsync();
+                            await _vm.LoadAsync();
```
kspearrin, line 100: Did this need to be changed for a reason?

aaxdev: This was to look about await inside BeginInvokeOnMainThread but since it makes no difference i pushed now a commit to apply changes.

src/Core/Abstractions/ISyncService.cs
```diff
@@ -7,7 +7,7 @@ namespace Bit.Core.Abstractions
     public interface ISyncService
     {
         bool SyncInProgress { get; set; }
-
+        Task<bool> IsSyncOnRefreshEnabledAsync();
```
kspearrin, line 10: We have some longer-term goals to keep the `Core` project very non-platform specific and use it for more clients. Since a pull to refresh feature is very mobile-specific, I'd like it if this were not in a Core service. Can we just call the storage service each time this is needed instead of having a service function for it? Looks like a 1 liner anyways.

aaxdev: I did not know about the non-platform specific in core, i thought it was better to avoid duplicate code. I made the changes.

kspearrin:
We'll hold off on this for now and may merge it in a future release. Thanks for the PR.

aaxdev:
Any info about when this will be merged?

cscharf:
How about now? 😉

PR #943, opened 2020-06-01
Don't build the keyboard index for autofill if using logout action

kspearrin:
If users are using the vault timeout feature to log out of the app, the autofill index may not be cleared when they are logged out (ex, the app just terminates). This is because the traditional loggedout message event may never occur. I would consider this a security issue since it would lead to a logged out user still receiving hints in their keyboard about items that were in their vault. If a user is using the vault timeout == "logOut" feature, we cannot maintain the autofill index for keyboard hints.

### DIFF
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -112,7 +112,7 @@ namespace Bit.App.Pages
             {
                 fingerprint = await _cryptoService.GetFingerprintAsync(await _userService.GetUserIdAsync());
             }
-            catch (Exception e) when(e.Message == "No public key available.")
+            catch (Exception e) when (e.Message == "No public key available.")
             {
                 return;
             }
@@ -193,8 +193,10 @@ namespace Bit.App.Pages
 
         public async Task VaultTimeoutAsync()
         {
-            var options = _vaultTimeouts.Select(o => o.Key == _vaultTimeoutDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
-            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeout, AppResources.Cancel, null, options);
+            var options = _vaultTimeouts.Select(
+                o => o.Key == _vaultTimeoutDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
+            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeout,
+                AppResources.Cancel, null, options);
             if (selection == null || selection == AppResources.Cancel)
             {
                 return;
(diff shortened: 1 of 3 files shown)


src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -349,7 +357,11 @@ public void BuildList()
             var securityItems = new List<SettingsPageListItem>
             {
                 new SettingsPageListItem { Name = AppResources.VaultTimeout, SubLabel = _vaultTimeoutDisplayValue },
-                new SettingsPageListItem { Name = AppResources.VaultTimeoutAction, SubLabel = _vaultTimeoutActionDisplayValue },
+                new SettingsPageListItem 
+                {
+                    Name = AppResources.VaultTimeoutAction,
+                    SubLabel = _vaultTimeoutActionDisplayValue
+                },
```
kspearrin, line 364: Just a formatting fix. Exceeded max line length allowed.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -206,11 +208,13 @@ public async Task VaultTimeoutAsync()
...
-        
+
         public async Task VaultTimeoutActionAsync()
         {
-            var options = _vaultTimeoutActions.Select(o => o.Key == _vaultTimeoutActionDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
-            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeoutAction, AppResources.Cancel, null, options);
+            var options = _vaultTimeoutActions.Select(o =>
+                o.Key == _vaultTimeoutActionDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
+            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeoutAction,
+                AppResources.Cancel, null, options);
```
kspearrin, line 217: Just a formatting fix. Exceeded max line length allowed.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -193,8 +193,10 @@ public async Task LockAsync()
 
         public async Task VaultTimeoutAsync()
         {
-            var options = _vaultTimeouts.Select(o => o.Key == _vaultTimeoutDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
-            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeout, AppResources.Cancel, null, options);
+            var options = _vaultTimeouts.Select(
+                o => o.Key == _vaultTimeoutDisplayValue ? $"✓ {o.Key}" : o.Key).ToArray();
+            var selection = await Page.DisplayActionSheet(AppResources.VaultTimeout,
+                AppResources.Cancel, null, options);
```
kspearrin, line 199: Just a formatting fix. Exceeded max line length allowed.

src/iOS/AppDelegate.cs
```diff
@@ -120,7 +120,8 @@ public override bool FinishedLaunching(UIApplication app, NSDictionary options)
                         }
                     }
                 }
-                else if (message.Command == "addedCipher" || message.Command == "editedCipher" || message.Command == "restoredCipher")
+                else if (message.Command == "addedCipher" || message.Command == "editedCipher" ||
+                    message.Command == "restoredCipher")
```
kspearrin, line 124: Just a formatting fix. Exceeded max line length allowed.

PR #960, opened 2020-06-07
Fix for deadlock in iOS autofill & share extensions

mportune-bw:
Replaced use of `GetAwaiter().GetResult()` within autofill and sharing extension view controllers, fixes frozen and/or partially drawn autofill windows when setting up and using autofill.

Before:

before

After:

after

Also fixed the dark-text-on-dark-background issue for the autofill setup completion header (now matches "success" color used in share extension)

Autofill:

autofill_success

Share:

share_success

### DIFF
--- a/src/iOS.Autofill/CredentialProviderViewController.cs
+++ b/src/iOS.Autofill/CredentialProviderViewController.cs
@@ -41,7 +41,7 @@ namespace Bit.iOS.Autofill
             };
         }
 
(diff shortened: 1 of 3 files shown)


mportune-bw:
@kspearrin is the iOS ci still in development or is that failure legit?

kspearrin:
Ignore the GitHub ci for now.

PR #982, opened 2020-06-15
Resolve passwords in new ciphers being disabled

Hinton:
Resolve new ciphers defaulting to ViewPassword = false, causing passwords to be non editable.


Hinton:
Noticed another bug, when creating a new cipher in the mobile app the fields are disabled.

PR #983, opened 2020-06-15
Workaround for disabling predictive text in visible password fields

mportune-bw:
A workaround for disabling predictive text while editing visible password fields on Android (see code comments for details/links on what and why).  This should suffice until the Xamarin.Forms bug has been fixed:

https://github.com/xamarin/Xamarin.Forms/issues/10857


mportune-bw:
cherry-picked iOS build fix to get that glorious green

kspearrin:
Seems ok to me

PR #985, opened 2020-06-16
Disabled iOS PR builds for now

mportune-bw:
- Disabled iOS PR builds to prevent non-master build errors
- Restored prevention of secret decryption in non-master builds


kspearrin:
Seems that PR builds are not triggering our build currently, so we can just close this.

Issue #975, opened 2020-06-12
TOTP QR code scanner causes crash


lschober-bw:
## Describe the Bug

If you attempt to use the QR code scanner for TOTP, and Bitwarden is not already given Camera permissions, the system will ask you to allow the application access to the camera. This action causes the system and application to bind up and become unresponsive.

## Steps To Reproduce

1. If enabled, go to the system settings and revoke Bitwarden's camera permission. 
2. Open Bitwarden
3. Edit any Login type item
4. Select the camera icon in the Authenticator Key (TOTP) field
5. Grant Bitwarden camera access.

## Expected Result

The camera is enabled and you can scan your QR code.

## Actual Result

The camera viewfinder is stuck as a white screen. The application and system become unresponsive to input. The application crashes.

## Environment

- Device: Pixel 2XL
- Operating system: Android 10 (QQ2A.200501.001.B3)
- Build Version: 2.4.3 (2523)
- Is this a Beta release? Y

## Additional Context

Once the application restarts with the camera permission enabled, scanning a QR code is successful.

bcronce:
I'm stuck on Android 8 and was in darkmode, when instead of a white screen, I had a black one. I assume this is because dark mode.

Also mention that I just kept trying and eventually got it to take. Seemed like every other time crashed then worked. But it's late and I don't want to setup any more tonight to see if that pattern persists.

cscharf:
Resolved by #1005

OLLI-S:
@cscharf 
I switched to Bitwarden some days ago and still had this issue the last days (the Android app crashed while scanning the QR code).
So this issue is not fixed!

julian-alarcon:
I have the same error mentioned by @OLLI-S . But I think that this is a different issue than the reported bug here as the App already have the Camera permission and when I try to set the code scanning the QR the App crashes.

gaticssa:
I have (had?) the same problem on a Galaxy S10e. 

The first time I tried to scan a QR code (for an Amazon account) the Bitwarden app asked for the Camera permission.
I allowed that, but still the app crashed on every scanning attempt.

Then I noticed under App Settings that the Bitwarden app did not have the Storage permission. I don't recall whether it had asked for that on install, but it definitely didn't ask when scanning the QR code.

I allowed that, and I was able to successfully scan at least that one code. I haven't tried with any other sites, so I'm not sure if it was just a lucky coincidence that it worked.

Bitwarden app: Version 2.6.1 (3178)
[...]

PR #1005, opened 2020-07-08
Modified Permissions Flow

MattTheDev:
## Objective 

Permissions flow for Android was causing the white screen on initial permission grant on occasion. Moved permission grant to pre-TOTP page load. This resvoles #975 

## Code Changes

**src/Android/MainActivity.cs** - ZXing.Net.Mobile.Forms.Android.PermissionsHandler.OnRequestPermissionsResult was deprecated, and with new flow, was hanging. I've modified this to point to the new method.

**src/App/Pages/Vault/AddEditPage.xaml.cs** - Added permissions check on click to confirm access to camera. If access is denied, return to add page.

**src/App/Utilities/PermissionManager.cs** - Added static method for checking/requesting permissions using Xamarin Essentials, for this use case and potential future consumption. 

## Known Issues (Must Fix)

* None known. Requesting a test on a physical device before merging with master to confirm. Emulator runs without issue. Physical device confirmation would be preferred.

### DIFF
--- a/src/App/Pages/Vault/AddEditPage.xaml.cs
+++ b/src/App/Pages/Vault/AddEditPage.xaml.cs
@@ -1,12 +1,14 @@
 ﻿using Bit.App.Abstractions;
 using Bit.App.Models;
 using Bit.App.Resources;
+using Bit.App.Utilities;
 using Bit.Core;
 using Bit.Core.Abstractions;
 using Bit.Core.Enums;
 using Bit.Core.Utilities;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Xamarin.Essentials;
 using Xamarin.Forms;
 using Xamarin.Forms.PlatformConfiguration;
 using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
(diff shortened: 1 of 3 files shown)


MattTheDev:
If @mportune-bw or @vincentsalucci could validate functionality on a physical device - I'd be super grateful.

src/App/Pages/Vault/AddEditPage.xaml.cs
```diff
@@ -243,6 +245,12 @@ private async void ScanTotp_Clicked(object sender, System.EventArgs e)
         {
             if (DoOnce())
             {
+                var cameraPermission = await PermissionManager.CheckAndRequestPermissionAsync(new Permissions.Camera());
```
mportune-bw, line 248: Traditionally you'd want the page/activity that makes use of the permission in question to handle the permission check so you don't have to worry about multiple points of entry.  If we're certain this is the only way into ScanPage this will work for now, but we should consider fixing ScanPage at some point in the future (potentially a larger task as we'll need to safely block camera initialization while waiting for permission, etc.)

MattTheDev: @mportune-bw - This is the only place where I see the explicit instantiation of / navigation to ScanPage. Willing to implement it the "correct way" though if you reach out, I just need a little guidance.

mportune-bw: Ok cool, let's just roll with it for now then.  Better to stop the crashing sooner 👍

PR #1008, opened 2020-07-11
Don't break when CipherType data is null

Lockszmith:
as explained in issue https://github.com/bitwarden/mobile/issues/1006


Lockszmith:
This is my first PR to a project that isn't mine, please let me know if I missed anything.

PR #1010, opened 2020-07-13
Forcing autofill refresh on sync/modification of stored cred

### LINKED ISSUE #980: iOS Autocomplete filles old/wrong password
## Describe the Bug
I adjusted my Hotmail Password. The password got synced to my iPhone and iPad correctly, but when I autocomplete the password in any password field it fills the wrong one. When I copy and paste it manually it works.

## Steps To Reproduce

This was how I did it. Not sure how consistent reproduceable that is
1. Change Hotmail password
2. Start bitwarden and ensure that it is synced
3. Open app who uses Hotmail login and autofill password
4. Notice that password was incorrect (in my case I could see it by that it was much shorter - or the "login failed message)
5. Open bitwarden and copy the password
6. switch back to the application and paste password
7. Login successful

## Expected Result

The correct password is filled with Autofill Password

## Actual Result
It seems the old password is used in Autofill.

## Environment

iPhone 11 Pro (iOS 13.5.1)
iPad Pro 2020 (iOS 13.5.1)
- Operating system: [e.g. iOS 8.1]
- Build Version (go to "Settings" → "About" in the app): [e.g. 2.3.0 (2221)]: 2.4.3(69)
- Is this a Beta release? [Y/N] N

MattTheDev:
## Objective

Issue #980 appears to be happening in 2 places:

1. Changing credentials on the device, name or password, and going to the browser and attempting to auto fill. 
2. Syncing on device after credentials are changed on another device.

For some reason, the ASHelpers. ReplaceAllIdentities function is called within the stack ... but it wouldn't function 100% of the time within the existing process. Explicitly calling it after the credential change OR the sync, seems to resolve this inconsistency.

This fix will close #980.

## Code Changes

**src/App/Abstractions/IDeviceActionService.cs**
**src/Android/Services/DeviceActionService.cs**
**src/iOS.Core/Services/DeviceActionService.cs**

Added new RefreshAutoFillAsync function. This gives the App access to the individual devices. Android is just a return statement. This is only applicable to iOS, Android seems to be unaffected.

**src/App/Pages/Settings/SyncPageViewModel.cs**
**src/App/Pages/Vault/AddEditPageViewModel.cs**

Implemented call to new function. Sync is when the sync is explicitly called. Add/Edit is on credential change.

### DIFF
--- a/src/App/Pages/Vault/AddEditPageViewModel.cs
+++ b/src/App/Pages/Vault/AddEditPageViewModel.cs
@@ -455,7 +455,7 @@ namespace Bit.App.Pages
                 _platformUtilsService.ShowToast("success", null,
                     EditMode && !CloneMode ? AppResources.ItemUpdated : AppResources.NewItemCreated);
                 _messagingService.Send(EditMode && !CloneMode ? "editedCipher" : "addedCipher", Cipher.Id);
-               await _deviceActionService.RefreshAutoFillAsync();
+                await _deviceActionService.RefreshAutoFillAsync();
 
                 if (Page is AddEditPage page && page.FromAutofillFramework)
                 {
--- a/src/App/Pages/Vault/AddEditPageViewModel.cs
+++ b/src/App/Pages/Vault/AddEditPageViewModel.cs
@@ -455,6 +455,7 @@ namespace Bit.App.Pages
                 _platformUtilsService.ShowToast("success", null,
                     EditMode && !CloneMode ? AppResources.ItemUpdated : AppResources.NewItemCreated);
                 _messagingService.Send(EditMode && !CloneMode ? "editedCipher" : "addedCipher", Cipher.Id);
+               await _deviceActionService.RefreshAutoFillAsync();
 
                 if (Page is AddEditPage page && page.FromAutofillFramework)
                 {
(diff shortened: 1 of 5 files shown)


src/App/Pages/Vault/AddEditPageViewModel.cs
```diff
@@ -455,6 +455,7 @@ public async Task<bool> SubmitAsync()
                 _platformUtilsService.ShowToast("success", null,
                     EditMode && !CloneMode ? AppResources.ItemUpdated : AppResources.NewItemCreated);
                 _messagingService.Send(EditMode && !CloneMode ? "editedCipher" : "addedCipher", Cipher.Id);
+               await _deviceActionService.RefreshAutoFillAsync();
```
mportune-bw, line 458: formatting (need one more space before `await`, unless github is just rendering funny)

MattTheDev: Fixed.

kspearrin:
This should already be happening with the `syncCompleted` message in iOS AppDelegate.cs. We need to investigate why that isn't working as expected.

MattTheDev:
@kspearrin Will investigate that now.

MattTheDev:
Closing PR. We're doing this in the AppDelegate already. These changes are not required.

PR #1011, opened 2020-07-14
[KnownUsernameField] Engine update (make the system more flexible)

contribucious:
**CONTEXT:** This is an update of [this new system](https://github.com/bitwarden/mobile/pull/880) _(system allowing "user ID" field detection — i.e. email/username/phone/whatever — without "password" field using the accessibility service)_.

**UPDATED:** Engine.
:arrow_right_hook: _Prerequisite for my next PR which concerns the **entries**._
___
&nbsp;
**This:**
- allows to have multiple pairs "path / username view ID" per domain;

- adds more advanced path support:

  - in its case-sensitive version: `startswith:string`, `contains:string`, `endswith:string` and
  - in its case-insensitive version: `istartswith:string`, `icontains:string`, `iendswith:string`.
&nbsp;

:bulb: About using `StringComparison.Ordinal` and `StringComparison.OrdinalIgnoreCase`, this is recommended for a variety of reasons including better performance. See the page [Best Practices for Using Strings in .NET](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage) from Microsoft documentation.
&nbsp;
&nbsp;

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -364,7 +364,7 @@ namespace Bit.Droid.Accessibility
                         else if (curUriPathWanted.StartsWith("icontains:", StringComparison.Ordinal))
                         {
                             curUriPathWanted = curUriPathWanted.Substring(10);
-                            uriLocalPathMatches = uriLocalPath.IndexOf(curUriPathWanted, StringComparison.OrdinalIgnoreCase) >= 0;
+                            uriLocalPathMatches = uriLocalPath.Contains(curUriPathWanted, StringComparison.OrdinalIgnoreCase);
                         }
                         else if (curUriPathWanted.StartsWith("iendswith:", StringComparison.Ordinal))
                         {
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -133,12 +133,12 @@ namespace Bit.Droid.Accessibility
         // Be sure to keep these entries sorted alphabetically
         public static Dictionary<string, KnownUsernameField> KnownUsernameFields => new List<KnownUsernameField>
         {
-            new KnownUsernameField("accounts.google.com", "ServiceLogin", "Email"),
-            new KnownUsernameField("amazon.com", "signin", "ap_email_login"),
-            new KnownUsernameField("github.com", "", "user[login]-footer"),
-            new KnownUsernameField("paypal.com", "signin", "email"),
-            new KnownUsernameField("signin.aws.amazon.com", "signin", "resolving_input"),
-            new KnownUsernameField("signin.ebay.com", "eBayISAPI.dll", "userid"),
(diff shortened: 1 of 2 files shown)


contribucious:
When you have some time @mportune-bw, could you take a look at this slight update of your system?

My next PR is already ready @ local side (big update of **entries**).
This one being only a required preamble. :thumbsup:

Thanks!

cscharf:
@contribucious , is there a chance you can refactor a little bit using Tuples or actual types vs. multidimensional arrays (`string[,]`) for Access Options in the `KnownUsernameField` class? That way it can be explicit on how those dimensions are to be defined and used (code as documentation). It seems as if the only the first value of each dimension of the array is being used, explicitly, which feels more like a Tuple, Dictionary, etc.

src/Android/Accessibility/KnownUsernameField.cs
```diff
@@ -2,15 +2,13 @@
...
             UriAuthority = uriAuthority;
-            UriPathEnd = uriPathEnd;
-            UsernameViewId = usernameViewId;
+            AccessOptions = accessOptions;
         }
 
         public string UriAuthority { get; set; }
-        public string UriPathEnd { get; set; }
-        public string UsernameViewId { get; set; }
+        public string[,] AccessOptions { get; set; }
```
cscharf, line 12: Please see note regarding use of multidimensional arrays.

contribucious:
Hello @cscharf,

Sorry for the delay.

You are right. Is it better now?

src/Android/Accessibility/KnownUsernameField.cs
contribucious: Done. :thumbsup:

contribucious:
@cscharf Is last week's change OK for you?

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -330,15 +330,64 @@ public static void FillEditText(AccessibilityNodeInfo editTextNode, string value
...
+                        // Case-sensitive comparison
+                        if (curUriPathWanted.StartsWith("startswith:", StringComparison.Ordinal))
+                        {
+                            curUriPathWanted = curUriPathWanted.Substring(11);
+                            uriLocalPathMatches = uriLocalPath.StartsWith(curUriPathWanted, StringComparison.Ordinal);
+                        }
+                        else if (curUriPathWanted.StartsWith("contains:", StringComparison.Ordinal))
+                        {
+                            curUriPathWanted = curUriPathWanted.Substring(9);
+                            uriLocalPathMatches = uriLocalPath.Contains(curUriPathWanted, StringComparison.Ordinal);
```
cscharf, line 350: Why use `.Contains()` here but below use `.IndexOf() >= 0` for the same check (given explicit passing of `StringComparison` mode to each method?

contribucious: Fixed. Thanks. :thumbsup:

contribucious:
You're welcome. :thumbsup:

PR #1026, opened 2020-07-28
Invalidate biometric on change

Hinton:
This PR implements additional verification to ensure the biometrics on devices has not been modified. This is to prevent potential attack vectors where adding a fingerprint provides access to the vault.

### Android Implementation details

It works by creating a private key which requires authenticated to use. By default these keys also sets `setInvalidatedByBiometricEnrollment`[0] to true, which invalidates all keys whenever the biometrics has changed.

<details>
<summary>Screenshot</summary>

image

</details>

### iOS Implementation details

On iOS we use the `evaluatedPolicyDomainState`[1] to store the state of the biometrics during setup, this state is invalidated whenever the biometrics are changed, allowing us to simply compare if the state is the same as before.

[ 0 ] https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setInvalidatedByBiometricEnrollment(boolean)
[ 1 ] https://developer.apple.com/documentation/localauthentication/lacontext/1514150-evaluatedpolicydomainstate


src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            KeyGenerator keyGen = KeyGenerator.GetInstance(KEY_ALGORITHM, KEYSTORE_NAME);
+            KeyGenParameterSpec keyGenSpec =
+                new KeyGenParameterSpec.Builder(KEY_NAME, KeyStorePurpose.Encrypt | KeyStorePurpose.Decrypt)
+                    .SetBlockModes(BLOCK_MODE)
+                    .SetEncryptionPaddings(ENCRYPTION_PADDING)
+                    .SetUserAuthenticationRequired(true)
+                    .Build();
+            keyGen.Init(keyGenSpec);
+            keyGen.GenerateKey();
```
kspearrin, line 80: Do we know if this has compatibility back to android 5? I rememebr dealing with a lot of issues with KeyStore when we used to roll our own for secure storage, but that may have been Android 4.4 which we no longer support.

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        private void CreateKey()
+        {
+            KeyGenerator keyGen = KeyGenerator.GetInstance(KEY_ALGORITHM, KEYSTORE_NAME);
+            KeyGenParameterSpec keyGenSpec =
+                new KeyGenParameterSpec.Builder(KEY_NAME, KeyStorePurpose.Encrypt | KeyStorePurpose.Decrypt)
+                    .SetBlockModes(BLOCK_MODE)
+                    .SetEncryptionPaddings(ENCRYPTION_PADDING)
+                    .SetUserAuthenticationRequired(true)
```
kspearrin, line 77: Is this the part that makes it invalidate?

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+        static readonly string KEY_NAME = "com.8bit.bitwarden.biometric_integrity";
+
+        static readonly string KEYSTORE_NAME = "AndroidKeyStore";
+
+        static readonly string KEY_ALGORITHM = KeyProperties.KeyAlgorithmAes;
+        static readonly string BLOCK_MODE = KeyProperties.BlockModeCbc;
+        static readonly string ENCRYPTION_PADDING = KeyProperties.EncryptionPaddingPkcs7;
+        static readonly string TRANSFORMATION = KEY_ALGORITHM + "/" + BLOCK_MODE + "/" + ENCRYPTION_PADDING;
+
+        readonly KeyStore _keystore;
```
kspearrin, line 24: Please use `private`

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+using Java.Security;
+using Javax.Crypto;
+
+namespace Bit.Droid.Services
+{
+    class BiometricService : IBiometricService
+    {
+        static readonly string KEY_NAME = "com.8bit.bitwarden.biometric_integrity";
+
+        static readonly string KEYSTORE_NAME = "AndroidKeyStore";
```
kspearrin, line 17: These could be const

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        static readonly string KEY_ALGORITHM = KeyProperties.KeyAlgorithmAes;
+        static readonly string BLOCK_MODE = KeyProperties.BlockModeCbc;
+        static readonly string ENCRYPTION_PADDING = KeyProperties.EncryptionPaddingPkcs7;
+        static readonly string TRANSFORMATION = KEY_ALGORITHM + "/" + BLOCK_MODE + "/" + ENCRYPTION_PADDING;
```
kspearrin, line 22: Use PascalCasing for constants and camelCase for private members

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        readonly KeyStore _keystore;
+
+        Context context = Android.App.Application.Context;
```
kspearrin, line 26: Is this used anywhere? Not sure if this will have any issues or not, but we use this `CrossCurrentActivity.Current.Activity` plugin for getting context in other services. `CrossCurrentActivity.Current.Activity.Application.Context` should work.

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -238,6 +248,12 @@ public async Task SubmitAsync()
                     }
                     MasterPassword = string.Empty;
                     await SetKeyAndContinueAsync(key);
+
+                    // Re-enable biometrie
```
kspearrin, line 252: typo

src/Android/Services/BiometricService.cs
Hinton: Correct, from https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationRequired(boolean).

> The key will become irreversibly invalidated once the secure lock screen is disabled (reconfigured to None, Swipe or other mode which does not authenticate the user) or when the secure lock screen is forcibly reset (e.g., by a Device Administrator). Additionally, if the key requires that user authentication takes place for every use of the key, it is also irreversibly invalidated once a new biometric is enrolled or once\ no more biometrics are enrolled, unless setInvalidatedByBiometricEnrollment(boolean) is used to allow validity after enrollment. Attempts to initialize cryptographic operations using such keys will throw KeyPermanentlyInvalidatedException.

src/Android/Services/BiometricService.cs
Hinton: The Context was previously used when I experimented with things, will remove it.

src/Android/Services/BiometricService.cs
kspearrin: Looks like we need to limit this functionality to Android 6+

src/Android/Services/BiometricService.cs
kspearrin: Looks like we need to limit this functionality to Android 6+

kspearrin:
Looks good to me

src/iOS.Core/Services/BiometricService.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public bool ValidateIntegrity()
+        {
+            var oldState = _storageService.GetAsync<NSData>("biometricState").Result;
+            if (oldState == null)
+            {
+                SetupBiometric();
+
+                return true;
+            } else {
```
kspearrin, line 30: formatting should be on new line

src/iOS.Core/Services/BiometricService.cs
```diff
@@ -0,0 +1,45 @@
...
+                var state = getState();
+
+                return oldState == state;
+            }
+        }
+
+        private NSData getState()
```
kspearrin, line 37: methods should be pascal cased

src/iOS.Core/Services/BiometricService.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public BiometricService(IStorageService storageService)
+        {
+            _storageService = storageService;
+        }
+
+        public bool SetupBiometric()
+        {
+            var state = getState();
+            return _storageService.SaveAsync("biometricState", state).IsCompletedSuccessfully;
```
kspearrin, line 19: shouldnt we just make this all async so that we don't have to deal with `IsCompletedSuccessfully`

src/iOS.Core/Services/BiometricService.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public bool ValidateIntegrity()
+        {
+            var oldState = _storageService.GetAsync<NSData>("biometricState").Result;
```
kspearrin, line 24: shouldnt we just make this all async so that we don't have to deal with `Result`

src/iOS.Core/Services/BiometricService.cs
kspearrin, line 19: Also, I dont think our storage service supports storing NSData. Need to base64 it to a string.

Hinton: Hmm, yes. Checking the documentation it doesn't implement Serializeble.

Hinton: True, was a bit lazy and wanted to avoid modifying the interface.

Hinton:
This should be feature complete, we should do some more testing for Android on physical devices, worked last time I tried. iOS code is untested and someone with a setup dev environment for iOS should verify things are working as expected. Would be interesting to see if it also works for FaceID if we remove and re-enroll the face again.

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Threading.Tasks;
+using Android.OS;
+using Android.Security.Keystore;
+using Bit.Core.Abstractions;
+using Java.Security;
+using Javax.Crypto;
+
+namespace Bit.Droid.Services
+{
+    class BiometricService : IBiometricService
```
kspearrin, line 14: `public class` like the others

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private readonly KeyStore _keystore;
+
+        public BiometricService()
+        {
+            _keystore = KeyStore.GetInstance(KeyStoreName);
+            _keystore.Load(null);
+        }
+
+        public Task<bool> SetupBiometric()
```
kspearrin, line 33: Use async suffix. VS refactor rename should make an easy change.

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            if (Build.VERSION.SdkInt >= BuildVersionCodes.M)
+            {
+                CreateKey();
+            }
+
+            return Task.FromResult(true);
+        }
+
+        public Task<bool> ValidateIntegrity()
```
kspearrin, line 43: same

src/Android/Services/BiometricService.cs
Hinton: I was hesitant to make this change since there's nothing in the function that is Async which causes a warning to appear. But if that's our code style I'll change it.

src/Android/Services/BiometricService.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private readonly KeyStore _keystore;
+
+        public BiometricService()
+        {
+            _keystore = KeyStore.GetInstance(KeyStoreName);
+            _keystore.Load(null);
+        }
+
+        public async Task<bool> SetupBiometric()
```
kspearrin, line 33: I think you misunderstood what I meant. I meant to add the Async suffix to the method name.

Hinton: Ah yes, I re-read the comment and noticed you wrote "suffix". I just read async and figured you meant the keyword.

GurpreetKang:
> This should be feature complete, we should do some more testing for Android on physical devices, worked last time I tried. iOS code is untested and someone with a setup dev environment for iOS should verify things are working as expected. Would be interesting to see if it also works for FaceID if we remove and re-enroll the face again.

Not working on Essential PH-1 running Android 10.  Biometric change is always detected when Bitwarden locks (no new biometrics added).    Mobile build current to commit ae28de4.

Hinton:
Correct @GurpreetKang, I accidentally reversed the integrity check criteria on Android during testing, I've opened a PR which should solve this. Thanks for reporting it 👍 .

GurpreetKang:
Thanks @Hinton, new build working now.  :thumbsup:

PR #1047, opened 2020-08-22
Added SSO flows and functionality

mportune-bw:
SSO login support with 2FA & password creation/validation.  Makes use of `WebAuthenticator` from the Xamarin Essentials kit to manage external browser auth and callbacks.

I'll expand upon this further on Monday, but wanted to get the code in.


src/Android/WebAuthCallbackActivity.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Android.App;
+using Android.Content.PM;
+
+namespace Bit.Droid
+{
+    [Activity(NoHistory = true, LaunchMode = LaunchMode.SingleTop)]
+    [IntentFilter(new[] { Android.Content.Intent.ActionView },
+        Categories = new[] { Android.Content.Intent.CategoryDefault, Android.Content.Intent.CategoryBrowsable },
+        DataScheme = "bitwarden")]
+    public class WebAuthCallbackActivity : Xamarin.Essentials.WebAuthenticatorCallbackActivity { }
```
kspearrin, line 10: Should we register this with an explicit name? See other activities for example:


    [Register("com.x8bit.bitwarden.WebAuthCallbackActivity")]

If not, I think it creates some kind of random activity name in the manifest, which might change over time and break things.

We had this problem with the autofill service changing names and getting disabled from update to update, so I have always tried to be explicit since then.

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -224,18 +227,33 @@ public async Task SubmitAsync()
...
+                    if (storedKeyHash != null)
                     {
-                        await _secureStorageService.RemoveAsync("oldKey");
-                        await _cryptoService.SetKeyHashAsync(keyHash);
-                        storedKeyHash = keyHash;
+                        passwordValid = storedKeyHash == keyHash;
+                    }
+                    else
+                    {
+                        await _deviceActionService.ShowLoadingAsync(AppResources.LoggingIn);
```
kspearrin, line 240: Should probably show "Loading" here instead of "Logging in".

src/App/Pages/Accounts/LoginSsoPage.xaml.cs
```diff
@@ -0,0 +1,105 @@
...
+                    Application.Current.MainPage = new NavigationPage(new AutofillCiphersPage(_appOptions));
+                    return;
+                }
+            }
+            var previousPage = await _storageService.GetAsync<PreviousPageInfo>(Constants.PreviousPageKey);
+            if (previousPage != null)
+            {
+                await _storageService.RemoveAsync(Constants.PreviousPageKey);
+            }
+            await _vaultTimeoutService.LockAsync();
```
kspearrin, line 102: Why is this called on logged in?

src/App/Pages/Accounts/LoginSsoPage.xaml.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    Application.Current.MainPage = new NavigationPage(new AddEditPage(appOptions: _appOptions));
+                    return;
+                }
+                if (_appOptions.Uri != null)
+                {
+                    Application.Current.MainPage = new NavigationPage(new AutofillCiphersPage(_appOptions));
+                    return;
+                }
+            }
```
kspearrin, line 96: Can we actually navigate to these on SSO login?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+using Bit.Core.Models.Domain;
+using Xamarin.Essentials;
+using Xamarin.Forms;
+
+namespace Bit.App.Pages
+{
+    public class LoginSsoPageViewModel : BaseViewModel
+    {
+        private const string Keys_RememberedOrgIdentifier = "rememberedOrgIdentifier";
+        private const string Keys_RememberOrgIdentifier = "rememberOrgIdentifier";
```
kspearrin, line 18: I didn't make a checkbox for this on web. Maybe we should

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+            var codeChallenge = CoreHelpers.Base64UrlEncode(codeVerifierHash);
+            await _storageService.SaveAsync(Constants.SsoCodeVerifierKey, codeVerifier);
+
+            var state = await _passwordGenerationService.GeneratePasswordAsync(passwordOptions);
+            await _storageService.SaveAsync(Constants.SsoStateKey, state);
+
+            var redirectUri = "bitwarden://sso-callback";
+
+            var url = _apiService.IdentityBaseUrl + "/connect/authorize?" +
+                      "client_id=" + _platformUtilsService.IdentityClientId + "&" +
```
kspearrin, line 105: I forgot about _platformUtilsService.IdentityClientId. I dont think we use that on other clients.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+                    new Uri(redirectUri));
+            }
+            catch (TaskCanceledException e)
+            {
+                await _deviceActionService.HideLoadingAsync();
+                cancelled = true;
+            }
+            if (!cancelled)
+            {
+                if (authResult != null && authResult.Properties.TryGetValue("code", out var code))
```
kspearrin, line 126: According to spec, you are also suppose to verify that the state returned is the same as what you originally passed.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            await _deviceActionService.ShowLoadingAsync(AppResources.LoggingIn);
+
+            var passwordOptions = new PasswordGenerationOptions(true);
+            passwordOptions.Length = 64;
+
+            var codeVerifier = await _passwordGenerationService.GeneratePasswordAsync(passwordOptions);
+            var codeVerifierHash = await _cryptoFunctionService.HashAsync(codeVerifier, CryptoHashAlgorithm.Sha256);
+            var codeChallenge = CoreHelpers.Base64UrlEncode(codeVerifierHash);
+            await _storageService.SaveAsync(Constants.SsoCodeVerifierKey, codeVerifier);
```
kspearrin, line 97: Not sure you even need to save these to storage with this implementation.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            var state = await _passwordGenerationService.GeneratePasswordAsync(passwordOptions);
+            await _storageService.SaveAsync(Constants.SsoStateKey, state);
```
kspearrin, line 100: Not sure you even need to save these to storage with this implementation.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+                else
+                {
+                    await _deviceActionService.HideLoadingAsync();
+                    await _platformUtilsService.ShowDialogAsync(AppResources.LoginSsoError,
+                        AppResources.AnErrorHasOccurred);
+                }
+            }
+        }
+
+        private async Task Login(string code, string codeVerifier, string redirectUri)
```
kspearrin, line 139: Should probably be called `LogIn` since this is more of a verb action method name?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+                }
+                else if (response.ResetMasterPassword)
+                {
+                    StartSetPasswordAction?.Invoke();
+                }
+                else
+                {
+                    var disableFavicon = await _storageService.GetAsync<bool?>(Constants.DisableFaviconKey);
+                    await _stateService.SaveAsync(Constants.DisableFaviconKey, disableFavicon.GetValueOrDefault());
+                    var task = Task.Run(async () => await _syncService.FullSyncAsync(true));
```
kspearrin, line 165: I think we are missing this on other clients

src/App/Pages/Accounts/SetPasswordPage.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+                    Application.Current.MainPage = new NavigationPage(new AutofillCiphersPage(_appOptions));
+                    return;
+                }
+            }
+            var previousPage = await _storageService.GetAsync<PreviousPageInfo>(Constants.PreviousPageKey);
+            if (previousPage != null)
+            {
+                await _storageService.RemoveAsync(Constants.PreviousPageKey);
+            }
+            Application.Current.MainPage = new TabsPage(_appOptions, previousPage);
```
kspearrin, line 94: I feel like maybe we should try to move this into a helper or something. Seems to be used a few times in different places now.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
mportune-bw: Do you want to present the option on mobile?  (there's currently no toggle for saving email on the login page so I followed that pattern)

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
mportune-bw: Should I be using something else?

kspearrin: No, this is fine.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
kspearrin: You show a toggle for rememebr org identifier, or no?

mportune-bw: Currently no.

kspearrin: Ok, leave it as-is then.

src/Android/WebAuthCallbackActivity.cs
mportune-bw: I don't believe so, since the WebAuthenticator intent originates from within the app which can include the (known) activity name as an extra.  With autofill the intent originates from the system, which has no knowledge of the receiver's activity name.

kspearrin: Understood

mportune-bw:
> Does this work with the logging in for the autofill extension flows on iOS?

Not yet, doing that next.

src/App/Pages/Accounts/LoginSsoPage.xaml.cs
mportune-bw: It does now :)

mportune-bw:
@kspearrin Github won't let me respond to this comment inline for some reason, so doing it here:

> According to spec, you are also suppose to verify that the state returned is the same as what you originally passed.

Can you expand on this?  I'm unsure which values from `AuthResult` fall under the 'state' umbrella.

src/App/Pages/Accounts/LoginSsoPage.xaml.cs
```diff
@@ -0,0 +1,114 @@
...
+                RequestFocus(_orgIdentifier);
+            }
+        }
+
+        private void CopyAppOptions()
+        {
+            if (_appOptions != null)
+            {
+                // create an object copy of _appOptions to persist values when app is exited during web auth flow
+                _appOptionsCopy = new AppOptions();
```
mportune-bw, line 61: Further clarification on this:  Turns out the app is clearing the `Uri` from `Options` when it gets sent to the background during the web auth flow (see `App.xaml.cs > SetTabsPageFromAutofill(...)`).  To avoid messing with that method, I'm creating a copy of `AppOptions` then restoring the values back to the original upon web auth completion. I'm using a helper method instead of a copy constructor so the original object can be repopulated, and I added the method to the `AppOptions` class so it won't be overlooked if/when future options are added.

cscharf:
> @kspearrin Github won't let me respond to this comment inline for some reason, so doing it here:
> 
> > According to spec, you are also suppose to verify that the state returned is the same as what you originally passed.
> 
> Can you expand on this? I'm unsure which values from `AuthResult` fall under the `state` umbrella.

@mportune-bw , so you pass `state` in the initial request to login with SSO, from there what you get back in the final `AuthenticateResult` should contain an item with the literal key in the dictionary of `state`. That's the value you want to compare. see: https://github.com/bitwarden/jslib/blob/master/src/angular/components/sso.component.ts (where you have `submit()` set the state values and pass those in the URL for authentication. Those values should come back in the `AuthenticateResult` dictionary from `state` and can be compared from there (I believe that's how it goes).

mportune-bw:
> @mportune-bw , so you pass `state` in the initial request to login with SSO, from there what you get back in the final `AuthenticateResult` should contain an item with the literal key in the dictionary of `state`. That's the value you want to compare. see: https://github.com/bitwarden/jslib/blob/master/src/angular/components/sso.component.ts (where you have `submit()` set the state values and pass those in the URL for authentication. Those values should come back in the `AuthenticateResult` dictionary from `state` and can be compared from there (I believe that's how it goes).

Ohhh _that_ state.  ;)  Understood, thanks for clarification!

src/Core/Services/ApiService.cs
```diff
@@ -377,6 +377,34 @@ public async Task<string> GetActiveBearerTokenAsync()
...
         }
 
+        public async Task<object> PreValidateSso(string identifier)
+        {
+            var path = "/account/prevalidate?domainHint=" + WebUtility.UrlEncode(identifier);
+            using (var requestMessage = new HttpRequestMessage())
+            {
+                requestMessage.Version = new Version(1, 0);
+                requestMessage.Method = HttpMethod.Get;
+                requestMessage.RequestUri = new Uri(string.Concat(ApiBaseUrl, path));
```
cscharf, line 387: Should use the `Identity` service base Api, is that the same as `ApiBaseUrl` value?

mportune-bw: You are correct, should be `Identity` - fixed

mportune-bw:
I believe this is functionally complete.  There's a cosmetic issue I'm still fighting with (see below) but everything is ready.

Issue: When launching the web auth flow from an iOS extension, it's showing `(null)` instead of `Bitwarden` in the confirmation prompt.  I found a single (unanswered) SO post, and have tried setting numerous values on the page, controller, etc. to no avail.

https://stackoverflow.com/questions/59493957/null-wants-to-use-google-com-to-sign-in

From extension:
Screen Shot 2020-08-28 at 5 21 05 PM

From app:
Screen Shot 2020-08-28 at 5 25 02 PM

src/Core/Services/ApiService.cs
```diff
@@ -528,13 +528,20 @@ private async Task<ErrorResponse> HandleErrorAsync(HttpResponseMessage response,
                 await _logoutCallbackAsync(true);
                 return null;
             }
-            JObject responseJObject = null;
```
mportune-bw, line 491: Discovered by accident when I thought I restarted my DB container but actually stopped it, the `prevalidate` response error comes back as `application/json` so it passes the `IsJsonResponse` test, but the content is not json formatted (it's a SQL backend not found exception) so newton throws.  By returning null it falls back to a generic error instead of crashing the app.

src/App/Utilities/AppHelpers.cs
```diff
@@ -192,5 +193,34 @@ public static async Task SetPreconfiguredSettingsAsync(IDictionary<string, strin
                 }
             }
         }
+
+        public static bool HasAppOptions(AppOptions appOptions)
```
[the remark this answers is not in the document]
mportune-bw: Agreed - would you settle for `SetAlternateMainPage` since we're not actually modifying the AppOptions?

src/Core/Services/PolicyService.cs
```diff
@@ -66,5 +67,156 @@ public async Task Clear(string userId)
...
+        }
+
+        private int? GetPolicyInt(Policy policy, string key)
+        {
+            if (policy.Data.ContainsKey(key))
+            {
+                var value = policy.Data[key];
+                if (value != null)
+                {
+                    return (int)(long)value;
```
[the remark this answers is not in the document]
mportune-bw: This pattern was actually established during the initial implementation of policies where the use of `Convert` was removed (see #741 for history).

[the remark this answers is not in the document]
cscharf: Ah, okay, thanks. Better to stick with the pattern then.

src/App/Utilities/AppHelpers.cs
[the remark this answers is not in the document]
cscharf: that sounds good to me.

cscharf (dismissed):

cscharf:
@kspearrin , this is just waiting your final approval

kspearrin (changes requested):

kspearrin (approved):

src/App/Pages/Accounts/HomePage.xaml
```diff
@@ -11,12 +11,16 @@
     <ContentPage.BindingContext>
         <pages:HomeViewModel />
     </ContentPage.BindingContext>
+    
+    <ContentPage.ToolbarItems>
+        <ToolbarItem Text="{u:I18n Close}" Clicked="Close_Clicked" Order="Primary" Priority="-1" />
+    </ContentPage.ToolbarItems>
```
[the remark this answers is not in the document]
mportune-bw: The close option (only visible when home is shown from the extension) is to prevent the user from becoming trapped in the popup should they not want to login for [reasons]...

Screen Shot 2020-09-03 at 12 15 23 PM
Screen Shot 2020-09-03 at 12 14 40 PM

PR #1048, opened 2020-08-22
Fix auto biometric prompt on vault timeout

aaxdev:
A condition force to not show the auto biometric prompt (https://github.com/bitwarden/mobile/issues/1045).

This condition don't change the case about when you lock manually the vault from settings page or app bar, because the value of the function parameter is false in those cases, and the prompt is not shown, as expected.

However, in the case that the vault is not locked and you hit the home button or when you lock the phone screen, the value of the function parameter is true, and the condition set it on false, then the prompt will not be shown when you'll access back to the app. Whatever vault timeout selected, it's happen when the vault timed out. When it sets to "Immediatly" this behavior is annoying when you want to switch to another app and go back to it.

<details>
  <summary>GIF issue (expand)</summary>

gif
</details>

I don't get this issue on iOS version, it's maybe due to the sub condition vaultTimeout, i have always his value on null when i try it on Android but since i don't have any Mac environment, i can't debug it on iOS.

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -415,10 +415,6 @@ namespace Bit.App
                     autoPromptBiometric = false;
                 }
             }
-            else if (autoPromptBiometric && Device.RuntimePlatform == Device.Android)
-            {
-                autoPromptBiometric = false;
-            }
             PreviousPageInfo lastPageBeforeLock = null;
             if (Current.MainPage is TabbedPage tabbedPage && tabbedPage.Navigation.ModalStack.Count > 0)
             {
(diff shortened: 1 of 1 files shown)


aaxdev:
any update?

cscharf:
@Hinton ☝🏻

src/App/App.xaml.cs
```diff
@@ -407,18 +407,6 @@ private async Task TryClearCiphersCacheAsync()
         private async Task LockedAsync(bool autoPromptBiometric)
         {
             await _stateService.PurgeAsync();
-            if (autoPromptBiometric && Device.RuntimePlatform == Device.iOS)
```
mportune-bw, line 410: For the moment let's leave this iOS if block in place.  I was able to trigger a race condition resulting in a double-prompt and everything went sideways (though I've only been able to do it once).  Go ahead and remove the Android else block though, as that seems to achieve the desired result.

mportune-bw (changes requested):
See inline comment

cscharf (approved):

PR #1061, opened 2020-09-06
publish to google play internal test track

mportune-bw:
Publish directly to internal test track to skip play store approval process for internal testers (artifact can then be manually promoted to alpha > beta > prod like before).  This should bring the testing experience up to speed with TestFlight for iOS (i.e. push code > ci build > immediate availability).

Edit: For clarification the signup process requires maintaining a list of the testers' email addresses, combined with each tester clicking a "join the program" from a page that only renders for them if they're on the list.  Nothing goes public until we push it.


cscharf (approved):
Awesome!

PR #1096, opened 2020-09-28
Remap pt to pt-BR instead of pt-PT and handle valid mapping of pt to pt-BR initially

### LINKED ISSUE #979: iOS version ignoring Preferred Language Order setting
The iOS version apparently ignores the Preferred Language Order setting.

## Steps To Reproduce

1. On an iOS device, go to Settings, General, Language & Region. Under Preferred Language Order, tap on Add Language... and add both "Portuguese (Brasil)" and "Portuguese (Portugal)".

2. Edit the Preferred Language Order to put "Português" on the top and "Português (Portugal)" on the second row. That will change the iPhone Language to Portuguese (Brazil).

3. Open Bitwarden.

## Expected Result

The app should appear with the European Portuguese translation (since -- I suppose -- it doesn't have a Brazilian Portuguese translation yet).

## Actual Result

The app appears in English. Apparently it is ignoring the Preferred Language Order setting. If "Portuguese (Portugal)" is set above English, the app should use "Portuguese (Portugal)".

## Screenshots or Videos

iphone-language-order

## Environment

- Device: iPhone SE, iPhone 6S (tested on both)
- Operating system: iOS 13.4.1
- Build Version: 2.4.3 (69)
- Is this a Beta release? No

cscharf:
_Should_ fix and close #979. 

## Overview
In iOS, the mapping for Portuguese (Brazil) (`pt-BR`) is actually mapped to `pt` directly and therefore can't find `pt-BR` from our localization files. Since iOS considers this the default for Portuguese we need to fix the localization mapping to match the base Portuguese language 2-letter code to the locale version for Brazil. This _should_ in theory also keep those in Portugal or who have the language set to Portuguese (Portugal) (`pt-PT`) seeing the correct localization of the language.


kspearrin:
Have you looked at the `ILocalizeService` implementation for iOS? Perhaps this could just be corrected there?

cscharf:
So I've tried just toggling and mapping in the LocalizeService in the iOS Core project to map `pt` to `pt-BR` which didn't work. I've also tried going between `pt` and `pt-BR` without success.

Using
image

The only way I could get it to work
image

Was to rename the `AppResources.pt-BR.resx` file to `AppResources.pt.resx`. Once I did that I got the above as expected (and verified it is not pt-PT).

Any other thoughts @kspearrin , am I missing something?

cscharf:
I also pushed the commit where I was testing updates to the Localization service.

cscharf:
🤷🏻 , apparently I had to uninstall, close, uninstall and re-deploy when debugging for the update to get deployed.... this works w/o the file rename or Crowdin updates. Just from what I have now should resolve it.

kspearrin (approved):

idiosyn:
@cscharf Has this already been included in the app available in the App Store? I’m using 2.7.0 and the app still appears in English when iOS is in Portuguese (Brazil). Thanks.

cscharf:
@idiosyn , so, sorry, the app is still under review I believe; you should be looking out for v2.7.1.

athersus:
The translate still not working on v.2.8.0. Any news about that? iOS 14.4

PR #1097, opened 2020-09-30
Allow alphanumeric card numbers

TheDuffman85:
Removed Keyboard="Numeric" property from CardNumberEntry. European debit card numbers are alphanumeric. See also https://en.wikipedia.org/wiki/International_Bank_Account_Number.

### DIFF
--- a/src/App/Pages/Vault/AddEditPage.xaml
+++ b/src/App/Pages/Vault/AddEditPage.xaml
@@ -203,8 +203,7 @@
                         <Entry
                             x:Name="_cardNumberEntry"
                             Text="{Binding Cipher.Card.Number}"
-                            StyleClass="box-value"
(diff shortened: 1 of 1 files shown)


kspearrin:
Do we need this in other client apps?

TheDuffman85:
I don't think so. This is a mobile only issue, because the onscreen keyboard is switched to numeric only mode. I checked the Windows Client, Firefox Extension and the Web Vault. In all thoose clients you can enter alphanumeric characters just fine.

kspearrin (approved):

mmattel:
@kspearrin thanks for linking. When to excpect a rollout with this fix?

vachan-maker:
It is currently available in the beta version.

PR #1100, opened 2020-10-01
re-select support for android tabbed page buttons

mportune-bw:
Added support for tab button re-selection in Android to match behavior from iOS, enabling the user to return to the vault root `GroupingsPage` from within a Type page (Login, Card, etc.) simply by tapping `My Vault` again.  Since we're using bottom-nav a little extra song and dance was required to get ahold of the view and set the listener.

### DIFF
--- a/src/Android/Renderers/CustomTabbedRenderer.cs
+++ b/src/Android/Renderers/CustomTabbedRenderer.cs
@@ -7,7 +7,6 @@ using Xamarin.Forms.Platform.Android;
 using Xamarin.Forms.Platform.Android.AppCompat;
 
 [assembly: ExportRenderer(typeof(TabbedPage), typeof(CustomTabbedRenderer))]
-
 namespace Bit.Droid.Renderers
 {
     public class CustomTabbedRenderer : TabbedPageRenderer, BottomNavigationView.IOnNavigationItemReselectedListener
--- a/src/Android/Renderers/CustomTabbedRenderer.cs
+++ b/src/Android/Renderers/CustomTabbedRenderer.cs
@@ -1,0 +1,61 @@
+﻿using Android.Content;
+using Android.Views;
+using Bit.Droid.Renderers;
+using Google.Android.Material.BottomNavigation;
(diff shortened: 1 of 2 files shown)


src/Android/Renderers/CustomTabbedRenderer.cs
```diff
@@ -0,0 +1,61 @@
...
+                        {
+                            return bottomNavigationView;
+                        }
+                    }
+                }
+            }
+            return null;
+        }
+
+        async void BottomNavigationView.IOnNavigationItemReselectedListener.OnNavigationItemReselected(IMenuItem item)
```
cscharf, line 53: Any reason this interface is explicitly implemented here vs. generally implemented?

i.e.
```C#
async void BottomNavigationView.IOnNavigationItemReselectedListener.OnNavigationItemReselected(IMenuItem item)
// vs.
async void OnNavigationItemReselected(IMenuItem item)
```

src/Android/Renderers/CustomTabbedRenderer.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Android.Content;
+using Android.Views;
+using Bit.Droid.Renderers;
+using Google.Android.Material.BottomNavigation;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+using Xamarin.Forms.Platform.Android.AppCompat;
+
+[assembly: ExportRenderer(typeof(TabbedPage), typeof(CustomTabbedRenderer))]
+
```
cscharf, line 10: Remove extra line to follow convention of other custom render code-files.

src/Android/Renderers/CustomTabbedRenderer.cs
mportune-bw: it's actually generally implemented from `Google.Android.Material.BottomNavigation` - the view itself (`BottomNavigationView`) isn't part of the namespace.

cscharf (approved):

PR #1117, opened 2020-10-14
fix for stuck extension sheets when dismissed by swipe

mportune-bw:
On iOS 13+ the extension modals are displayed as sheets which enables the user to swipe them away as a method of cancelation.  We have varying levels of logic attached to our cancel buttons that was not firing when the user swiped the sheet(s) away, resulting in an un-dismiss-able sheet that requires the app that triggered it to be killed.  This PR attaches a delegate to each view controller to capture a dismiss-by-swipe event and fire off the same code used by the cancel buttons.  Viola, no more frozen treats!  According to Apple this callback only fires on dismiss-by-swipe, so using the cancel button won't result in code being executed twice.

### DIFF
--- a/src/iOS.Autofill/LockPasswordViewController.cs
+++ b/src/iOS.Autofill/LockPasswordViewController.cs
@@ -9,6 +9,7 @@ namespace Bit.iOS.Autofill
             : base(handle)
         {
             BiometricIntegrityKey = "autofillBiometricState";
+            DismissModalAction = () => Cancel();
         }
 
         public CredentialProviderViewController CPViewController { get; set; }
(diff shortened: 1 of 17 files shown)


src/iOS.Autofill/LockPasswordViewController.cs
```diff
@@ -9,6 +9,7 @@ public LockPasswordViewController(IntPtr handle)
             : base(handle)
         {
             BiometricIntegrityKey = "autofillBiometricState";
+            DismissModalAction = () => Cancel();
```
cscharf, line 12: You can assign the `Cancel` method directly for this delegate vs. wrapping it in an anonymous function.
i.e.
```c#
DismissModalAction = Cancel;
```

src/iOS.Autofill/LoginAddViewController.cs
```diff
@@ -8,7 +9,9 @@ public partial class LoginAddViewController : Core.Controllers.LoginAddViewContr
     {
         public LoginAddViewController(IntPtr handle)
             : base(handle)
-        { }
+        {
+            DismissModalAction = () => Cancel();
```
cscharf, line 13: Same, these assignments can be unwrapped and simplified.

kspearrin (approved):

PR #1121, opened 2020-10-19
Update support for several browsers and HTML viewer

stevenlele:
This patch adds the Chinese version of Via browser ([`mark.via`](https://www.coolapk.com/apk/mark.via)) and updates the uri view id for both Chinese and Google Play version.

I also added HTML viewer (`com.android.htmlviewer`) since my system used it to display the WiFi log-in page. **It does NOT have a url bar so I did NOT add it to `AccessibilityHelpers.cs`.**


stevenlele:
Added the following Chinese browsers:
- Browser [`com.mmbox.browser`](https://www.coolapk.com/apk/com.mmbox.browser)
- X Browser [`com.mmbox.xbrowser`](https://www.coolapk.com/apk/com.mmbox.xbrowser)
- Magic Browser [`com.z28j.feel`](https://www.coolapk.com/apk/com.z28j.feel)

stevenlele:
@mportune-bw Could you please have a look at this PR? Thanks a lot!

src/Android/Autofill/AutofillHelpers.cs
```diff
@@ -41,6 +41,7 @@ public static class AutofillHelpers
             "com.amazon.cloud9",
             "com.android.browser",
             "com.android.chrome",
+            "com.android.htmlviewer",
```
mportune-bw, line 44: Does this need to be added to `SupportedBrowsers` above as well?

stevenlele: idk, as I've said, it does not have a url bar, so I doubt if the accessibility service will work on this.

mportune-bw (approved):

src/Android/Autofill/AutofillHelpers.cs
contribucious: **[2020/12]** Precision now added in `SupportedBrowsers` about the non-presence of this entry. See PR https://github.com/bitwarden/mobile/pull/1187.

PR #1124, opened 2020-10-20
Additional support for system theme setting

mportune-bw:
- Splash screen now respects system dark/light setting (Android & iOS)
- Default app theme respects system dark light setting (Android only, already worked in iOS)
- On the fly switching between dark/light when using app's Default theme (Android only for now, this was a happy accident and will require a Forms update for iOS, which we're doing but not in this light-weight PR)
- Updated biometric lib to fix freeze on login screen when theme-switching while bio-prompt is showing (Android only)

ios_launch_dark
ios_launch
android_launch_dark
android_launch


cscharf (approved):

vachan-maker:
I have the beta version installed on Android P and this does not seem to work for me. If I change the app theme from light to dark, the splash screen still follows the white theme. Will this only work if I enable dark theme system wide?, because my Oppo phone doesn't have a system wide dark theme for some reason?

mportune-bw:
@vachan-maker System-wide light/dark was introduced in Android Q (10), which is needed for the splash screen to reflect the change.  All prior versions of Android will show light splash only.

PR #1127, opened 2020-10-20
Remove deprecated info.plist key

mportune-bw:
Apparently this key was deprecated and is no longer necessary (confirmed via simulator test on all versions back to iOS 10).  The presence of the key is preventing publishing, so away it goes.

### DIFF
--- a/src/iOS/Info.plist
+++ b/src/iOS/Info.plist
@@ -125,7 +125,5 @@
 	<string>Use Face ID to unlock your vault.</string>
 	<key>NFCReaderUsageDescription</key>
 	<string>Use Yubikeys for two-facor authentication.</string>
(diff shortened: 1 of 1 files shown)


src/iOS/Info.plist
```diff
@@ -125,7 +125,5 @@
 	<string>Use Face ID to unlock your vault.</string>
 	<key>NFCReaderUsageDescription</key>
 	<string>Use Yubikeys for two-facor authentication.</string>
-	<key>UILaunchImages</key>
```
cscharf, line 128: Ah, the good ol' `UILaunchImages` key, I'm going to miss you old friend...... not, lol

cscharf (approved):

PR #1135, opened 2020-11-02
Use plain shield icon for accessibility overlay for quick identification

mportune-bw:
Changed the icon for the accessibility overlay to the plain black shield to help quickly identify which service is being utilized.  Autofill proper will continue to use the white shield encapsulated in blue.

Screen Shot 2020-11-02 at 5 35 20 PM
Screen Shot 2020-11-02 at 5 42 44 PM


cscharf (approved):
That's a great idea!

PR #1150, opened 2020-11-14
Enhanced autofill settings

mportune-bw:
Reworked autofill settings screen to explain settings in greater detail depending on the Android version.  UI also matches paradigm defined throughout other settings screens.

A single `Autofill Services` section indicates `Enabled` if the combination of enabled settings are ready for use on a given version of Android:

settings

The phrasing for each setting is tweaked according to the capabilities & requirements of the given version of Android...

Android 11 (offers all options):

android_11

Android 8-10 (removal of inline autofill support):

android_8-10

Android 7 (only offers accessibility, support for quick-action autofill panel makes draw-over optional):

android_7

Android 6 (only offers accessibility, manual draw-over grant required):

android_6

Android 5 (only offers accessibility, draw-over on by default):

android_5


src/App/Resources/AppResources.resx
```diff
@@ -1744,4 +1744,43 @@
...
+    <value>Required to use the Autofill Quick-Action Tile, or to augment the Autofill Service by using Draw-Over (if enabled).</value>
+  </data>
+  <data name="DrawOver" xml:space="preserve">
+    <value>Use Draw-Over</value>
+  </data>
+  <data name="DrawOverDescription" xml:space="preserve">
+    <value>When enabled, allows the Bitwarden Accessibility Service to display a popup when login fields are selected.</value>
+  </data>
+  <data name="DrawOverDescription2" xml:space="preserve">
+    <value>If  enabled, the Bitwarden Accessibility Service will display a popup when login fields are selected to assist with auto-filling your logins.</value>
```
cscharf, line 1781: double space here after `If`

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -326,23 +326,11 @@ public void BuildList()
...
-                var accessibilityEnabled = _deviceActionService.AutofillAccessibilityServiceRunning() &&
-                    _deviceActionService.AutofillAccessibilityOverlayPermitted();
                 autofillItems.Add(new SettingsPageListItem
                 {
-                    Name = AppResources.AutofillAccessibilityService,
-                    SubLabel = accessibilityEnabled ?
-                        AppResources.Enabled : AppResources.Disabled
+                    Name = AppResources.AutofillServices,
+                    SubLabel = _deviceActionService.AutofillServicesEnabled() ?
+                            AppResources.Enabled : AppResources.Disabled
```
cscharf, line 333: Remove erroneous space at beginning of line here before `AppResources.Enabled`

src/App/Pages/Settings/AutofillServicesPage.xaml.cs
```diff
@@ -3,34 +3,33 @@
...
 
-        protected override void OnAppearing()
+        protected async override void OnAppearing()
         {
+            await _vm.InitAsync();
             _vm.UpdateEnabled();
-            _vm.UpdatePermitted();
             _timerStarted = DateTime.UtcNow;
-            Device.StartTimer(new TimeSpan(0, 0, 3), () =>
+            Device.StartTimer(new TimeSpan(0, 0, 0,0,500), () =>
```
cscharf, line 26: Please add some spaces in-between the numbers and commas in this constructor call.

src/Android/Services/DeviceActionService.cs
```diff
@@ -378,6 +378,34 @@ public void RateApp()
             }
         }
 
+        public void DisableAutofillService()
+        {
+            try
+            {
+                var activity = (MainActivity)CrossCurrentActivity.Current.Activity;
+                var type = Java.Lang.Class.FromType(typeof(AutofillManager));
+                var manager = activity.GetSystemService(type) as AutofillManager;
```
cscharf, line 387: Will the system service of the given `type` always be `AutofillManager` and alway exist? Generally using the `as` keyword for casting is a "safe" cast, so if the type doesn't match or the value returned is `null`, the return result will be `null`; in the line below there's no null check, so assuming it always succeeds, in this scenario it would be more appropriate to use an explicit cast, `var manager = (AutofillManager)activity.GetSystemService(type);` so that it will throw a proper exception if the cast is invalid or the type is missing vs. getting a null ref exception on the next line.

mportune-bw: It will _not_ always exist since Autofill wasn't introduced until Android 8.  I believe this is the reason the same approach was taken further down in the `SupportsAutofillService()` method, but correct me if I'm wrong @kspearrin .

kspearrin: Yes, this is probably throwing on those older devices, but you have a try/catch to avoid the null ref.

cscharf (approved):

vachan-maker:
I have autofill services turned ON but Bitwarden is saying that it is off for some reason.

mportune-bw:
@vachan-maker Can you provide more detail?  Is "Enabled" not showing, and/or is the switch turned off?  Is it just the settings screen, or is autofill not working at all?

vachan-maker:
> 
> 
> @vachan-maker Can you provide more detail? Is "Enabled" not showing, and/or is the switch turned off? Is it just the settings screen, or is autofill not working at all?

The switch is turned off and Disabled is being shown.

vachan-maker:
But in System Settings, Autofill for Bitwarden is enabled.

mportune-bw:
@vachan-maker Which device and Android version are you using?

vachan-maker:
> 
> 
> @vachan-maker Which device and Android version are you using?

I am using the latest version of Bitwarden on Beta.
Android 7.0 Nougat Honor EMUI

mportune-bw:
I'm unable to repro on a 7.0 emulator.  Also, the code for establishing if our accessibility service is running (which drives the "Enabled" label, "Use Accessibility" switch, and the error dialog you're seeing) is the same code used prior to this PR.  Did accessibility work for you before?

vachan-maker:
> 
> 
> I'm unable to repro on a 7.0 emulator. Also, the code for establishing if our accessibility service is running (which drives the "Enabled" label, "Use Accessibility" switch, and the error dialog you're seeing) is the same code used prior to this PR. Did accessibility work for you before?

Thanks for the reply.
1. Accessibility services for Bitwarden auto fill enabled in the system settings
2. The toggle in the **Auto-fill Services ** page for Use Accessibility greyed out. Tapping on it takes me to the settings.
Settings show ON
3. Auto-fill services shown as disabled in the settings page of Bitwarden.
4. Auto-fill tile cannot be used and an error is shown. (Screenshot attached above)

mportune-bw:
Can you install the previous release of our app and see if accessibility still works:

https://github.com/bitwarden/mobile/releases/tag/v2.6.1

vachan-maker:
> 
> 
> Can you install the previous release of our app and see if accessibility still works:
> 
> https://github.com/bitwarden/mobile/releases/tag/v2.6.1

I installed 2.6.1 and the accessibility is not turning on in the app for some reason. I used autofill many times on 2.6.1. I don't know why Bitwarden is showing disabled even though Accessibility service is turned on.

mportune-bw:
That is odd but it tells me something else on your device has changed.  Does EMUI take a more aggressive stance towards power management, or perhaps some other 3rd party task manager?

vachan-maker:
> 
> 
> That is odd but it tells me something else on your device has changed. Does EMUI take a more aggressive stance towards power management, or perhaps some other 3rd party task manager?

I installed 2.7.0 again and restarted my device. Problem fixed. But still I don't know what I happened. Thanks for your help.

mportune-bw:
Glad to hear it!

PR #1162, opened 2020-12-01
Fix for missing biometric integrity check in iOS extensions under certain conditions

mportune-bw:
A path was discovered where a user could modify biometric credentials on an iOS device then bypass the Bitwarden integrity check by using the autofill extension.

Reason: While the extensions do properly perform the integrity check per #1110 , the check would fail to catch the change **only if the extension had yet to be triggered once biometric unlock was enabled within the app**  (because the very first check results in a null value, then updates to the current integrity key value and returns true to indicate success - see [BiometricService line 39](https://github.com/bitwarden/mobile/blob/e27370cf32ece73ba44d7eadd6aebf519d82fdce/src/iOS.Core/Services/BiometricService.cs#L39)).

_Solution Part 1_: By performing the check regardless of the biometric unlock setting in the app, the null integrity value for the extension is replaced with a legitimate value when autofill is first enabled (by opening the extension for user verification).  By the time the scenario above plays out, a proper integrity value is in place and the extension catches the change and forces the user to enter their master password.

_Solution Part 2_:  While testing part 1 I witnessed the behavior occasionally revert to skipping the check.  As the issue with sharing the DB across processes is still a thorn in my side, I've moved the integrity values to pref storage (where I should have put them in the first place) and added migration to prevent any unnecessary panic brought on by an unplanned bio integrity mismatch.  After this change, I've been unable to repro the failure.

### DIFF
--- a/src/App/Resources/AppResources.Designer.cs
+++ b/src/App/Resources/AppResources.Designer.cs
@@ -2943,6 +2943,12 @@ namespace Bit.App.Resources {
             }
         }
         
+        public static string BiometricInvalidatedExtension {
+            get {
+                return ResourceManager.GetString("BiometricInvalidatedExtension", resourceCulture);
+            }
+        }
+        
         public static string EnableSyncOnRefresh {
             get {
                 return ResourceManager.GetString("EnableSyncOnRefresh", resourceCulture);
--- a/src/App/Resources/AppResources.resx
+++ b/src/App/Resources/AppResources.resx
@@ -1675,7 +1675,10 @@
     <comment>Confirmation alert message when soft-deleting a cipher.</comment>
   </data>
   <data name="BiometricInvalidated" xml:space="preserve">
-    <value>Biometric change detected, login using Master Password to enable again.</value>
+    <value>Biometric unlock disabled pending verification of master password.</value>
+  </data>
+  <data name="BiometricInvalidatedExtension" xml:space="preserve">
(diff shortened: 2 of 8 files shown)


mportune-bw:
After being away from this for a bit I realized that anyone who updates their app with this fix and has already enabled autofill _but has yet to use it_ could still experience this issue. Thinking...

mportune-bw:
Okay - after a lot of fiddling I don't believe we can reliably save the initial biometric validation state  without additional validation _and_ plug this hole.  I've removed the `null == assume update and generate new` logic and now require validation via the master password if a state value does not exist.  The only downside with the experience is that when a user enables biometrics for the first time, they'll have to verify their master password the first time they use autofill as well.  This applies to new users as well as existing users who enabled biometrics but haven't used autofill yet.  Existing users who actively use both are covered by the migration and won't notice anything.  And of course nobody will ever see the messages again until they add/remove a finger/face.

Regarding any potential panic with unplanned biometric changes, I've changed the messaging in the alert to simply indicate validation is required without specifying a reason.  I've also customized the message depending on app vs. extension to (hopefully) make it clearer that there's a reason for another validation request.  Maybe.

App:

app

Extension:

extension

cscharf (approved):

vachan-maker:
@mportune-bw 
Shouldn't Bitwarden ask for the Master Password right after the error has been displayed? Entering the PIN doesn't stop this error from appearing. 
Isn't it asking the master password more convenient?

Issue #1165, opened 2020-12-03
Android 11: Clipboard doesn't clear if password is never pasted


lynshi:
## Describe the Bug

On Android 11, I have set the Bitwarden option to clear the clipboard after 10 seconds. After I copied a password, the password is still available in my clipboard after 10 seconds if it is never pasted.

## Steps To Reproduce

1. Set the time for clearing the clipboard to 10 seconds.
2. Copy a password and leave the Bitwarden app open in the foreground for 10 seconds.
3. Put Bitwarden in the background and wait for 10 seconds.
4. Go to any text entry field and you are able to paste the password. You may also notice the copied text being suggested on the Google Keyboard.

## Expected Result

After 10 seconds, the clipboard is empty and nothing pastes.

## Actual Result

The copied text is still available to be pasted once after the time to clear the clipboard elapses. After pasting, the clipboard is cleared.

## Environment

- Device: Google Pixel 3 XL
- Operating system: Android 11
- Build Version (go to "Settings" → "About" in the app): 2.7.0 (3258)
- Is this a Beta release? N

lynshi:
This actually doesn't seem to be a Bitwarden problem. If you hold a text area and click "Paste" from the menu that shows up after the clipboard clear timeout, nothing pastes. However, the password is still suggested by the Google Keyboard (see [clipboard suggestions](https://www.androidpolice.com/2020/04/23/clipboard-suggestions-spotted-in-gboard/)), which is probably caching the latest non-empty clipboard entry. This might be fixed if the [`ClearClipboardAlarmReceiver`](https://github.com/bitwarden/mobile/blob/dcfdc7d0eaa31e935e63d3906041e1c0abffb059/src/Android/Receivers/ClearClipboardAlarmReceiver.cs#L11) sets the clipboard text to a string with a space in it like how the browser extension clears the clipboard.

PR #1172, opened 2020-12-10
Change text for clearing a clipboard to a space instead of an empty string on Android

lynshi:
Per #1165, on Android 11 the Google Keyboard will suggest passwords copied to the clipboard even after they have been cleared by Bitwarden.

This is because the latest suggestion is not updated after the primary text on the clipboard is set to an empty string. The suggestion is removed correctly if the clipboard clear text is a space instead. Using a space to clear the clipboard matches the behavior of the Chromium browser extension.

This is what the suggestion looks like after Bitwarden has set the text in the clipboard to a space.
image

### DIFF
--- a/src/Android/Receivers/ClearClipboardAlarmReceiver.cs
+++ b/src/Android/Receivers/ClearClipboardAlarmReceiver.cs
@@ -8,7 +8,7 @@ namespace Bit.Droid.Receivers
         public override void OnReceive(Context context, Intent intent)
         {
             var clipboardManager = context.GetSystemService(Context.ClipboardService) as ClipboardManager;
-            clipboardManager.PrimaryClip = ClipData.NewPlainText("bitwarden", string.Empty);
+            clipboardManager.PrimaryClip = ClipData.NewPlainText("bitwarden", " ");
         }
     }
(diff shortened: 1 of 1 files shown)


mportune-bw (commented):
I've been meaning to look at this so thanks for the nudge.  Have you tried [.clearPrimaryClip](https://developer.android.com/reference/android/content/ClipboardManager#clearPrimaryClip()) for API 28+?  I'm curious if it achieves the desired result since it seems less hacky.

lynshi:
Just tried it and it doesn't work for API 30. What happens is you lose the ability to hold down to paste (the paste option just won't show up), but Gboard will continue to suggest the previously copied password.

mportune-bw (approved):
> Just tried it and it doesn't work for API 30. What happens is you lose the ability to hold down to paste (the paste option just won't show up), but Gboard will continue to suggest the previously copied password.

I would expect paste to no longer be available if the clipboard is legitimately cleared.  The other part almost sounds like a caching issue with GBoard.  Let's go with your workaround for now, and revisit `clearPrimaryClip` at a later date.  Much appreciated!

Issue #798, opened 2020-03-27
[ios] Export vault not working


xtqqczze:
## Describe the Bug

Unable to export vault in iOS app: the master password is rejected even though it is valid.

## Steps To Reproduce

1. Open Bitwarden app
2. Go to settings
3. Select "Export vault"
4. Select either .json or .csv file format
5. Type valid master password
6. Select Export Vault button

## Expected Result

Vault is exported.

## Actual Result

Error message: "Invalid master password. Try again."

sdo9:
I also get the error about an invalid password, but then the exported file is actually created. So warning to OP: check if your exported unencrypted passwords might be lying around in your export location.

xtqqczze:
@sdo Where would the exported file be located? The mobile app does not give the opportunity to choose a location:

![Bitwarden export vault clipped](https://user-images.githubusercontent.com/45661989/80372397-61270400-888b-11ea-94e1-ed80c4e16223.png)

sdo9:
Never mind. So mine prompts for the master password, then there's a 
sharing dialog, I happened to pick Files (and in there picked Dropbox and 
Save). At this point I think it does the export for me, but doesn't 
confirm that it did, and then places me back at the screen with file format, master password, and 
Export button. I happened to just click Export, thinking to conclude the 
process, and it said the master password was invalid (because it had 
been blanked).

So, sorry, unrelated, kind of works for me except it's pretty confusing UX 
to be prompted again instead of getting confirmation that the export was done.

xtqqczze:
@kspearrin Any thoughts on this issue?

xtqqczze:
@kspearrin This feature is stil non-functional in version 2.5.0.

xtqqczze:
@kspearrin This feature is still non-functional in version 2.6.1

xtqqczze:
@kspearrin This feature is still non-functional in version 2.7.0 on iOS 14.2.

There is no doubt that the password my for account is valid as it is accepted at https://vault.bitwarden.com/

MGibson1:
@xtqqczze I'm not able to reproduce your issue. I downloaded your model iPhone (6s, right?) simulator on iOS 14.2 and json export worked for me on Bitwarden v2.7.0.

Is it possible you have a difficult-to-type master password and it is indeed wrong? Is there anything else unique about your setup?

xtqqczze:
[...]

PR #1174, opened 2020-12-11
Enable Encrypted json export of vaults

MGibson1:
# Overview

Allow for exporting vaults in an encrypted state.

# File changes

* **ExportVaultPage.xaml/ExportVaultPage.xaml.cs/ExportVaultPageViewModel.cs**: Add binding to update export warning message to appropriate message depending on export format.
* **AppResources.resx**: Add encrypted export warning message (matches jslib's)
* **Models/Export/\***: Update constructors to allow export of encrypted Domain types.

# Image

image


src/App/Pages/Settings/ExportVaultPage.xaml.cs
```diff
@@ -65,5 +65,13 @@ private async void ExportVault_Clicked(object sender, EventArgs e)
                 await _vm.ExportVaultAsync();
             }
         }
+
+        void _fileFormatPicker_SelectedIndexChanged(object sender, EventArgs e)
```
mportune-bw, line 69: Method naming convention is irregular here, maybe use something like `FileFormat_Changed`

MGibson1:
Note, jslib export nulls our Cipher CollectionIds prior to export: https://github.com/bitwarden/jslib/blob/master/src/services/export.service.ts#L121

while mobile does not (nor does this implementation)

mportune-bw:
> Note, jslib export nulls our Cipher CollectionIds prior to export: https://github.com/bitwarden/jslib/blob/master/src/services/export.service.ts#L121
> 
> while mobile does not (nor does this implementation)

Good catch, looks like an oversight on my part.  Go ahead and add that to mobile so the export results remain identical with other clients.

Looks legit otherwise 👍

src/App/Pages/Settings/ExportVaultPage.xaml.cs
```diff
@@ -65,5 +65,13 @@ private async void ExportVault_Clicked(object sender, EventArgs e)
                 await _vm.ExportVaultAsync();
             }
         }
+
+        void FileFormat_Changed(object sender, EventArgs e)
+        {
+            if (_vm != null)
+            {
+                _vm.UpdateWarning();
+            }
```
kspearrin, line 74:
```suggestion
            _vm?.UpdateWarning();
```

mportune-bw (approved):

kspearrin (approved):

xtqqczze:
Might this fix #798?

MGibson1:
@xtqqczze It is not specifically related, no. This is a new file format to export a vault in that is encrypted in such a way that Bitwarden can import it again.

PR #1175, opened 2020-12-12
Use monotonic clock for vault timeout

mportune-bw:
Protects the vault-timeout from clock changes by using a monotonic clock, validated on app resume. *

Android made this super easy.  iOS required a hack is required to get trustworthy values - instead of a timestamp in seconds we get a difference in seconds between now and the boot time, resulting in a duration that includes sleep time.  As both of those values are susceptible to clock changes in the same manner, the difference between them will be fixed, resulting in an anchored and forward-moving value that can be used as a monotonic clock.

* Unless I'm missing something @kspearrin this can replace the timers on both platforms.  I can remove them outright if you agree.

### DIFF
--- a/src/Android/Services/DeviceActionService.cs
+++ b/src/Android/Services/DeviceActionService.cs
@@ -752,6 +752,15 @@ namespace Bit.Droid.Services
             return false;
         }
 
+        public long GetActiveTime()
+        {
+            // Returns milliseconds since the system was booted, and includes deep sleep. This clock is guaranteed to
+            // be monotonic, and continues to tick even when the CPU is in power saving modes, so is the recommend
+            // basis for general purpose interval timing.
+            // ref: https://developer.android.com/reference/android/os/SystemClock#elapsedRealtime()
+            return SystemClock.ElapsedRealtime() / 1000;
+        }
+
         private bool DeleteDir(Java.IO.File dir)
         {
             if (dir != null && dir.IsDirectory)
--- a/src/Core/Services/VaultTimeoutService.cs
+++ b/src/Core/Services/VaultTimeoutService.cs
@@ -90,13 +90,13 @@ namespace Bit.Core.Services
             {
                 return;
             }
-            var lastActive = await _storageService.GetAsync<DateTime?>(Constants.LastActiveKey);
+            var lastActive = await _storageService.GetAsync<long?>(Constants.LastActiveKey);
             if (lastActive == null)
             {
                 return;
             }
-            var diff = DateTime.UtcNow - lastActive.Value;
-            if (diff.TotalSeconds >= vaultTimeout.Value)
+            var diff = _platformUtilsService.GetActiveTime() - lastActive;
+            if (diff >= vaultTimeout * 60)
             {
                 // Pivot based on saved action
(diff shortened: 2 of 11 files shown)


src/iOS.Core/Utilities/iOSHelpers.cs
```diff
@@ -7,7 +9,37 @@ namespace Bit.iOS.Core.Utilities
...
+                var pStr = Marshal.AllocHGlobal(length);
+                sysctlbyname(property, pStr, pLen, IntPtr.Zero, 0);
+                var timeVal = Marshal.PtrToStructure<TimeVal>(pStr);
+                var now = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+                if (timeVal.sec > 0 && now > 0)
+                {
+                    uptime = now - timeVal.sec;
+                }
+            }
+            catch { }
```
cscharf, line 38: Missing `finally { ... }` block to properly deallocate the memory being allocated using `Marshal.AllocHGlobal()`, otherwise this would cause a memory leak issue. Use `Marshal.FreeHGlobal(n)`

neat example can be found here: https://stackoverflow.com/a/17562399

src/Core/Services/VaultTimeoutService.cs
```diff
@@ -90,13 +90,13 @@ public async Task CheckVaultTimeoutAsync()
             {
                 return;
             }
-            var lastActive = await _storageService.GetAsync<DateTime?>(Constants.LastActiveKey);
+            var lastActive = await _storageService.GetAsync<long?>(Constants.LastActiveKey);
```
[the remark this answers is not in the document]
mportune-bw: On a cold start post-update when the app comes up locked, that code isn't executed.  Once the user unlocks, the pref is overwritten with the new long value so it's ready to go the first time vault timeout is checked.

cscharf (dismissed):

src/Android/Services/DeviceActionService.cs
```diff
@@ -752,6 +752,15 @@ public bool UsingDarkTheme()
             return false;
         }
 
+        public long GetActiveTime()
+        {
+            // Returns milliseconds since the system was booted, and includes deep sleep. This clock is guaranteed to
+            // be monotonic, and continues to tick even when the CPU is in power saving modes, so is the recommend
+            // basis for general purpose interval timing.
+            // ref: https://developer.android.com/reference/android/os/SystemClock#elapsedRealtime()
```
kspearrin, line 760: This comment says returns ms, but looks like you are returning seconds?

kspearrin (commented):
Yes, I don't think we need the timers anymore.

src/Android/Services/DeviceActionService.cs
mportune-bw: Correct, I standardized on seconds in the main app so platform-specific code converts accordingly.

kspearrin (approved):

steffenw94:
I have same issue on Google Pixel 3a.
Sometimes the vault is locking, sometimes not.
For me i wait for fix to use this app again because i think this is high security lack

Issue #1021, opened 2020-07-25
http URL was saved when adding a new login


yegle:
## Describe the Bug

When adding a new login from Android Chrome, a http:// URL was saved instead of https:// URL.

## Steps To Reproduce

1. In Chrome Android incognito tab, open a login form (e.g. https://github.com/login), confirm you are accessing using HTTPS.
2. When prompt "Autofill with Bitwarden", go to my vault.
3. Add a new login.

## Expected Result

The url `https://github.com/login` should be in the URL field.

## Actual Result

The url `http://github.com/login` is used in the URL field.

## Environment

- Device: Google Pixel 4 XL
- Operating system: Android 11 Beta 2
- Build Version (go to "Settings" → "About" in the app): 2.5.0 (3093)
- Is this a Beta release? N

RePacked:
I can't reproduce this. Shows correct URI for me.

Device: OnePlus 8
Operating system: Android 10 (OxygenOS 10.5.9.IN21BA)
Build Version: 2.5.0 (3093)
Is this a Beta release? No

yegle:
I can also reproduce it on an Android 10 Pixel 3A phone (QQ3A.200805.001).

MGibson1:
@yegle, What is the Autofill framework you have enabled? From your issue and description I think you're using Accessibility with Draw-Over enabled. Is this correct?

If so, this is _kind of_ an upstream issue with Chrome. The way accessibility grabs the URI is through the address bar in Chrome. Chrome has made the choice to remove the URI scheme from that bar, and so our accessibility code can't get it. We are currently attempting to prepend `http://` and see if that's a valid URI. If it is, we use that. That's why you get `http://github.com/login` instead of `https://...`. The reason for that choice is it's more likely to work. However, these days we really should assume `https://` first.

I'm submitting a PR to update this fallback case when we find no scheme to first attempt to prepend `https://`, then `http://`.

Again, the above assumes you're using Accessibility for your Autofill needs.

yegle:
Ah yes I use accessibility instead of the Autofill API.

Thanks for the fix! I agree the default should be https:// nowadays.

MGibson1:
I'm going to mark this closed since #1186 should have fixed this specific issue. We'll continue to have these problems with accessibility-based autofill, but hopefully the autofill framework fro android shores up in the future and it won't be a problem.

PR #1186, opened 2020-12-18
Attempt URI healing with https prior to http

MGibson1:
# Overview

Related to #1021. 

Browsers are moving away from displaying URI scheme in a way accessibility can easily grab. This causes this URI healing to be relied upon more frequently. It should attempt https prior to http due to prevelence of https and security concerns with passwords over http.

# Files changed

* **AccessibilityHelpers.cs**: This helper is responsible for grabbing the Uri out of  a browser's address node. If a URI scheme isn't present, it heals the URI with `https://`.  It seems unlikely that any URI that isn't healed by `https://` would be by prepending `http://`, but the attempt isn't very expensive.

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -390,12 +390,15 @@ namespace Bit.Droid.Accessibility
                 var hasHttpProtocol = uri.StartsWith("http://") || uri.StartsWith("https://");
                 if (!hasHttpProtocol && uri.Contains("."))
                 {
-                    if (Uri.TryCreate("http://" + uri, UriKind.Absolute, out var uri2))
+                    if (Uri.TryCreate("https://" + uri, UriKind.Absolute, out var _))
+                    {
+                        return string.Concat("https://", uri);
+                    } else if (Uri.TryCreate("http://" + uri, UriKind.Absolute, out var _))
                     {
                         return string.Concat("http://", uri);
                     }
                 }
-                if (Uri.TryCreate(uri, UriKind.Absolute, out var uri3))
+                if (Uri.TryCreate(uri, UriKind.Absolute, out var _))
                 {
                     return uri;
                 }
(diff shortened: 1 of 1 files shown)


mportune-bw (commented):
Not sure I see the need to keep `http` as well.  Would there ever be a situation where a URI would fail to create with `https` but not `http`?

cscharf:
> Not sure I see the need to keep `http` as well.  Would there ever be a situation where a URI would fail to create with `https` but not `http`?

Some legacy sites don't have https bindings so https may fail but http work

MGibson1:
I don't know the URI class well enough to say for sure.

I imaging it just does some semantic checks, but it sees _possible_ it actually makes a call to see if the server responds.

mportune-bw:
> Some legacy sites don't have https bindings so https may fail but http work

This is just URI creation though, not an actual network request.

MGibson1:
> Not sure I see the need to keep `http` as well. Would there ever be a situation where a URI would fail to create with `https` but not `http`?

I don't know the URI class well enough to say for sure.

I imagine it just does some semantic checks, but it sees possible it actually makes a call to see if the server responds.

src/Android/Accessibility/AccessibilityHelpers.cs
```diff
@@ -390,12 +390,15 @@ private static string ExtractUri(string uri, AccessibilityNodeInfo addressNode,
                 var hasHttpProtocol = uri.StartsWith("http://") || uri.StartsWith("https://");
                 if (!hasHttpProtocol && uri.Contains("."))
                 {
-                    if (Uri.TryCreate("http://" + uri, UriKind.Absolute, out var uri2))
+                    if (Uri.TryCreate("https://" + uri, UriKind.Absolute, out var _))
+                    {
+                        return string.Concat("https://", uri);
+                    } else if (Uri.TryCreate("http://" + uri, UriKind.Absolute, out var _))
```
cscharf, line 396: Please move the `else if` on the next line after the trailing `}`

cscharf: Also, this `else if` block is not necessary, so you should be able to remove it anyway. If `https://` doesn't work, `http://` won't either when using `Uri.TryCreate()`

cscharf (changes requested):

cscharf (approved):

PR #1189, opened 2020-12-22
Implemented Custom role and permissions

addisonbeck:
https://app.asana.com/0/1195018406457675/1153292148278664

### Narrative
As a user, I would like to be able to configure another user with access to manage user/group/collection assignment without being able to see organization passwords that are not directly assigned to them.

### Code Changes
1. Implemented the Permissions object and Custom role type as seen in the server PR
2. Converted any existing role checks in the app to Permission checks that verify a `ValidRole || ValidPermission`

### DIFF
--- a/src/App/Pages/Vault/AddEditPageViewModel.cs
+++ b/src/App/Pages/Vault/AddEditPageViewModel.cs
@@ -297,7 +297,7 @@ namespace Bit.App.Pages
                 if (org.Enabled && org.Status == OrganizationUserStatusType.Confirmed)
                 {
                     OwnershipOptions.Add(new KeyValuePair<string, string>(org.Name, org.Id));
-                    if (policies != null && org.UsePolicies && !org.IsAdmin && AllowPersonal)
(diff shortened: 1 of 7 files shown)


src/App/Pages/Vault/AddEditPageViewModel.cs
```diff
@@ -297,7 +297,7 @@ public async Task<bool> LoadAsync(AppOptions appOptions = null)
                 if (org.Enabled && org.Status == OrganizationUserStatusType.Confirmed)
                 {
                     OwnershipOptions.Add(new KeyValuePair<string, string>(org.Name, org.Id));
-                    if (policies != null && org.UsePolicies && !org.IsAdmin && AllowPersonal)
+                    if (policies != null && org.UsePolicies && !org.canManagePolicies && AllowPersonal)
```
cscharf, line 300: Assuming the idea here is that rather than `admin`, if the user has permissions to manage policies, they could just disable this policy anyway and do this (similar to an `admin` or `owner` before). Makes sense 👍🏻

cscharf (approved):

Issue #1197, opened 2020-12-28
[iOS Accessibility] VoiceOver always seeing fields for all entry types


pitermach:
## Describe the Bug
No matter what type of entry you're creating or viewing, VoiceOver is seeing the text fields for every other type of entry (IE, if creating a secure note you see text fields for username, password, credit card number and so on)

## Steps To Reproduce

1. Enable VoiceOver (Settings>Accessibility>VoiceOver, or ask siri tu turn VoiceOver on)
2. Note that now the screen will behave differently, touching something will first cause VoiceOver to highlight and speak it, you then need to tap the screen twice to actually press on a control.
3. In bitwarden, go to add an item
4. Even without changing the item type, touch on one of the first fields, like username, then continue swiping right to make VoiceOver move sequentially through all controls.
## Expected Result

Assuming the default item type of login is being added, you should only see the fields for name, username, password and URI
## Actual Result
You will also be able to focus the fields for credit card information, identities and so on. Also, the fields for URI for some reason appear at the bottom of these controls, so accessing them will take a long time to get past all the other fantom text fields. This issue also occurs in a similar fassion when viewing an entry

## Screenshots or Videos
[Screen Recording](https://cloud.tardis.pw/f/48f9adeb976845fdac1a/)
here I demonstrate the problem when trying to add a secure note with voice commentary for context.

## Environment

- Device: iPhone SE 2020
- Operating system: iOS 14.3
- Build Version: 2.7.2 (401)
- Is this a Beta release? No

## Additional Context
As I explained in the recording, this slows down navigation while viewing an entry considerably, there's also a security concern, because one of the things that's improperly hidden as a result while viewing an item is the plain text form of a password, no matter whether the button to toggle visibility is pressed or not so one could very easily broadcast a password to anyone in earshot of your phone.

eliykat:
Hi @pitermach, thanks for the detailed bug report - the video was especially helpful.

As discussed in the [pull request](https://github.com/bitwarden/mobile/pull/1213), this issue should be fixed with the next Xamarin.Forms update, which is being worked on at the moment. I'll leave this issue open until the update has been completed and we've tested it again to confirm the issue is resolved.

pitermach:
Excellent, thanks for looking into this. I'm looking forward to testing the fixed version once it releases.

eliykat:
Related to #1089

PR #1213, opened 2021-01-11
[iOS Accessibility] Bugfix: VoiceOver always seeing fields for all entry types

eliykat:
# Objective

Bugfix for #1197:

> No matter what type of entry you're creating or viewing, VoiceOver is seeing the text fields for every other type of entry (IE, if creating a secure note you see text fields for username, password, credit card number and so on)

The cause of the bug was that VoiceOver (the iOS screenreader) reads invisible elements (i.e. those with `IsVisible="False"`). However, `IsVisible` is the primary means for showing/hiding relevant/irrelevant fields and controls.

# Code Changes

Whether an element is “visible” to VoiceOver is determined by a separate property, `AutomationProperties.IsInAccessibleTree`. However, unlike `IsVisible`, this does not affect child items, so it cannot effectively be applied to a parent `StackLayout` to hide its children from VoiceOver.

I have therefore generally taken the following approach:
* Where `IsVisible` is set on an individual element, set `IsInAccessibleTree` on that element with the same value.
* Where `IsVisible` is set on a layout element, add `IsInAccessibleTree` to each child with the same value as the parent.
* However, I didn't do this for the main controls for each item type, because of the large number of elements. Instead, they are added/removed from the element tree as required using code behind.

# Issues

The **Add URI** button is still being picked up by VoiceOver on the AddEdit page, even when it's invisible and `IsInAccessibleTree` is set to `False`. I figure I'd looked at this long enough, maybe someone else can spot what's going on.

Other than that, I have tested this on my iPhone 6 and it all works as expected.

### DIFF
--- a/src/App/Pages/Vault/AddEditPage.xaml
+++ b/src/App/Pages/Vault/AddEditPage.xaml
@@ -95,7 +95,7 @@
                         Grid.Row="0"
                         Grid.Column="1"
                         Grid.RowSpan="2"
-                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.IsInAccessibleTree="{Binding Cipher.ViewPassword}"
                         AutomationProperties.Name="{u:I18n CheckPassword}"
                         IsVisible="{Binding Cipher.ViewPassword}" />
                     <controls:FaButton 
@@ -105,7 +105,7 @@
                         Grid.Row="0"
                         Grid.Column="2"
                         Grid.RowSpan="2"
-                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.IsInAccessibleTree="{Binding Cipher.ViewPassword}"
                         AutomationProperties.Name="{u:I18n ToggleVisibility}"
                         IsVisible="{Binding Cipher.ViewPassword}" />
                     <controls:FaButton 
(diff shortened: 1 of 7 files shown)


cscharf (commented):
Perhaps a custom render, or renderers would be appropriate or if there is another Xamarin.Forms extension that can be hooked into for this in a more generic way to handle screen reader updates across the board, especially for container/layout visibility changes.
@mportune-bw , any thoughts/ideas on this one?

eliykat:
After some investigation, I think this could be done much cleaner with a custom renderer to set the native iOS property `accessibilityElementsHidden` on StackLayout and Grid elements. This property appears to remove an element's children from the accessibility tree, but is not set by the Xamarin.Forms default renderer.

However, in the course of testing a solution to this, I upgraded the Xamarin.Forms package from 4.5.0.725 to 5.0.0.1874 (latest release), and this seems to have fixed the bug entirely.

To make sure, I cloned a fresh copy of the mobile repo, fetched nuget packages, upgraded Xamarin.Forms, and with no other changes the bug disappeared on my iPhone 6 (iOS 12.5).

Can someone else verify this - perhaps on a newer model iPhone?

@cscharf

mportune-bw:
Bumps to Forms require a fair amount of regression testing before we can ship.  I'm using 5.x in a work branch and it touches a _lot_ coming from 4.5.x.   If also fixes this bug entirely we should probably hold off on this PR.

cscharf:
> Bumps to Forms require a fair amount of regression testing before we can ship. I'm using 5.x in a work branch and it touches a _lot_ coming from 4.5.x. If also fixes this bug entirely we should probably hold off on this PR.

Agreed... @eliykat , let's hold off on this one since it appears the update Matt is already working on will resolve this issue. Good find on that!

eliykat:
Closing as this is the changes in this PR are unlikely to be relevant once the Xamarin Forms upgrade is complete.

PR #1219, opened 2021-01-13
Port send jslib to mobile

MGibson1:
# Overview

Port of Send PRs to jslib (bitwarden/jslib#190, bitwarden/jslib#192, bitwarden/jslib#205, bitwarden/jslib#242). This PR does not implement any actual use of the SendService.

#### Common

This project is a copy from the [bitwarden/server/Core.Test project](https://github.com/bitwarden/server/tree/master/test/Core.Test), with some additions. Given the similarity here, we should think about creating a new repo for common testing items and accessing them through a submodule.

#### Core.Test

This is a test bed for the Send feature I'm implementing. Crypto tests are copied from jslib tests. Send model and service tests are happy path tests to ensure proper function since I'm not implementing anything that uses them.

### Automated Testing

An Automated test runner is added to the `android` job. Unfortunately, `dotnet` is not capable of building the `android` and `iOS` projects, so it is run against only the `Core.Test` project. We'll have to add another line per test project we add.

# Files Changes/Added

* **build.yml**: Added running of tests in the `Core.Test` project to the android job pipeline.
* **bitwarden-mobile.sln**: Added test projects `test/Common.csproj` and `test/Core.Test`
* **Android.csproj**: This was automatically changed by VS Mac when adding the above projects. The new Guid is what was already in the solution file and seems like a good change to me, so I kept it. @mportune-bw, @kspearrin if you know why this Guid differed from the solution file, please let me know and I'll revert.
* **IApiService.cs/ApiService.cs**: Add Send methods. No Access methods on Mobile.
* **ICryptoFunctionService.cs/PclCryptoFunctionService.cs**: Add expanded hkdf implementation to function service. A basic implementation used to be private to CryptoFunctionService, but is not implemented in classes implementing this interface.
* **ICryptoService.cs/CryptoService.cs**: Add send MakeKey method
* **ISendService.cs/SendService.cs**: Service handling send functionality.
* **Enums/HkdfAlgorithm**: Hkdf is currently only valid for Sha256 and Sha512, so this represents a subset of valid CryptoHashAlgorithms. Conversion between these enums happens in the CryptoFunctionService when Hkdf is calculated
* **Enums/SendType.cs**: Types of valid sends
* **SendFileApi.cs**: Api interface model for File type send data.
* **SendTestApi.cs**: Api interface model for Text type send data.
* **SendData.cs/SendFileData.cs/SendTextData.cs**: Client storage model for Send / File Send info / and Text Send info, respectively
* **Domain/Send/SendFile/SendText** Encrypted data model for Send / File Send info / and Text Send info, respectively
* **SendView/SendFileView/SendTextView**: Decrypted model for Send / File Send info / and Text Send info, respectively
* **CipherString**: Allow optionally passing in decryption key 
* **Domain**: Optionally allow specifying decryption key
* **SendRequest**: Api request model
* **SendResponse**: Api response model
* **SyncResponse**: Added sends to sync response
* **SyncService**: Handle synced sends and replace current storage with received send data.
* **CoreHelpers**: Added this method for testing, but it exists in CoreHelpers for Server. Open to moving to TestHelpers
* **ServiceContainer.cs**: Make SendService available from ServiceContainer

#### Common
* **AutoSubDataAttribute/CustomAutoDataAttribute/SutAutoDataAttribute/InlineAutoSubDataAttribute/InlineCustomAutoDataAttribute/FixtureExtension**: Helper attributes to allow creation of autofixtures with customizations, substitutions, or sutProviders 
* **ISutProvider/SutProvider/SutProviderCustomization**: Autofixture customization to provide SUT objects with access to injected dependencies
* **common/TestHelper.cs**: Commonly used helper methods are placed here. 
* **Common.csproj**: Project to house testing code common across testing projects. We should think about splitting this to a new Repo to share between csharp projects.

#### Core.Test
* **Core.Test/\*\*/SymmetricCryptoKeyCustomization / SendCustomizations**: AutFixture customizations to produce valid fixture data for SymmetricCryptoKey and Send data structures, respectively.
* **Core.Test.csproj**: Project housing all Core tests and Core specific testing infrastructure (like above customizations)
* **SendDataTests.cs**: Tests for SendData, SendTextData, and SendFileData data models. These are model conversion tests
* **SendTests.cs**: Tests for Send, SendText, and SendFile domain models. These are model conversion tests
* **SendRequestTests.cs**: Tests for SendRequest, SendTextApi, and SendFileApi models. These are model conversion tests
* **CryptoFunctionServiceTests.cs**: Tests of Hkdf implementations. Tests are copied from [bitwarden/jslib](https://github.com/bitwarden/jslib/blob/master/spec/node/services/nodeCryptoFunction.service.spec.ts#L52-L89)
* **SendServiceTests.cs**: Happy path tests of send service functions


src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,270 @@
...
+using Bit.Core.Models.Request;
+using Bit.Core.Models.Response;
+using Bit.Core.Models.View;
+using Newtonsoft.Json;
+
+namespace Bit.Core.Services
+{
+    public class SendService : ISendService
+    {
+        public static string SEND_KEY(string userId) => string.Format("sends_{0}", userId);
```
kspearrin, line 20: Is this how we handle the key definitions in other services like cipher service or folder service?

src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,270 @@
...
+        private readonly IUserService _userService;
+        private readonly IApiService _apiService;
+        private readonly IStorageService _storageService;
+        private readonly II18nService _i18nService;
+        private readonly ICryptoFunctionService _cryptoFunctionService;
+        private Task<List<SendView>> _getAllDecryptedTask;
+
+
+        public SendService(ICryptoService cryptoService, IUserService userService, IApiService apiService,
+            IStorageService storageService, II18nService i18nService, ICryptoFunctionService cryptoFunctionService)
```
kspearrin, line 32: Should these be new lined?

kspearrin (commented):
looking great. just a few comments

src/Android/Android.csproj
```diff
@@ -206,7 +206,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\App\App.csproj">
-      <Project>{9F1742A7-7D03-4BB3-8FCD-41BC3002B00A}</Project>
+      <Project>{EE44C6A1-2A85-45FE-8D9B-BF1D5F88809C}</Project>
```
MGibson1, line 209: This was an autofix from VS Mac, but I imagine it should be done. It now matches the GUID in the solution.

src/Core/Services/SendService.cs
MGibson1: Not sure what you're referring to, but kind of. The keys look the same, the method of accessing it is slightly shorter.

This is how it's done in `FolderService`
```
private const string Keys_FoldersFormat = "folders_{0}";
```

accessed by
```
string.Format(Keys_FoldersFormat, userId)
```

I didn't like the `string.Format` all over the place so I just moved to a helper method

src/Core/Models/Domain/Send.cs
```diff
@@ -0,0 +1,98 @@
...
+            var cryptoService = ServiceContainer.Resolve<ICryptoService>("cryptoService");
+
+            try
+            {
+                view.Key = await cryptoService.DecryptToBytesAsync(Key, null);
+                view.CryptoKey = await cryptoService.MakeSendKeyAsync(view.Key);
+            }
+            catch (Exception e)
+            {
+                // TODO: error?
```
cscharf, line 79: @mportune-bw , any thoughts on whether this should be bubbled up? Not sure how logging or error handling is done in general for a scenario like this (coming out of a domain model); assuming we don't want an app crash, but at the same time behavior may be unexpected if the keys aren't set here.

src/Core/Models/Domain/SendFile.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class SendFile : Domain
+    {
+        public string Id { get; set; }
+        public string Url { get; set; }
+        public string Size { get; set; }
+        public string SizeName { get; set; }
+        public CipherString FileName { get; set; }
+
+        public SendFile() { }
```
cscharf, line 16: Would this also need to call the `base()` constructor? and perhaps the loaded constructor calls `this()`

src/Core/Models/Domain/SendFile.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public SendFile(SendFileData file, bool alreadyEncrypted = false) : base()
+        {
+            Size = file.Size;
+            BuildDomainModel(this, file, new HashSet<string> { "Id", "Url", "SizeName", "FileName" }, alreadyEncrypted, new HashSet<string> { "Id", "Url", "SizeName" });
+        }
+
+        public Task<SendFileView> DecryptAsync(SymmetricCryptoKey key)
+            => DecryptObjAsync(new SendFileView(this), this, new HashSet<string> { "FileName" }, null, key);
```
cscharf, line 25: move the `=>` to the top line, starting the anonymous function on the next line here

src/Core/Models/Domain/Send.cs
```diff
@@ -0,0 +1,98 @@
...
+        public CipherString Key { get; set; }
+        public int? MaxAccessCount { get; set; }
+        public int AccessCount { get; set; }
+        public DateTime RevisionDate { get; set; }
+        public DateTime? ExpirationDate { get; set; }
+        public DateTime DeletionDate { get; set; }
+        public string Password { get; set; }
+        public bool Disabled { get; set; }
+
+        public Send() { }
```
cscharf, line 31: Would this also need to call the `base()` constructor? and perhaps the loaded constructor calls `this()`

src/Core/Models/Domain/SendText.cs
```diff
@@ -0,0 +1,25 @@
...
+using Bit.Core.Models.View;
+
+namespace Bit.Core.Models.Domain
+{
+    public class SendText : Domain
+    {
+        public CipherString Text { get; set; }
+        public bool Hidden { get; set; }
+
+        public SendText() { }
```
cscharf, line 14: Same constructor thing, `base()`

src/Core/Models/Domain/SendText.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public SendText(SendTextData data, bool alreadyEncrypted = false) : base()
+        {
+            Hidden = data.Hidden;
+            BuildDomainModel(this, data, new HashSet<string> { "Text" }, alreadyEncrypted);
+        }
+
+        public Task<SendTextView> DecryptAsync(SymmetricCryptoKey key)
+            => DecryptObjAsync(new SendTextView(this), this, new HashSet<string> { "Text" }, null, key);
```
cscharf, line 23: Move `=>` up a line.

src/Core/Models/Request/SendRequest.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                case SendType.Text:
+                    Text = new SendTextApi
+                    {
+                        Text = send.Text?.Text?.EncryptedString,
+                        Hidden = send.Text.Hidden
+                    };
+                    break;
+                case SendType.File:
+                    File = new SendFileApi()
```
cscharf, line 44: Like the constructor call for `SendTextApi` above on line 37, remove the `()` when using a simple object creation construct.

src/Core/Models/View/SendFileView.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Bit.Core.Models.View
+{
+    public class SendFileView : View
+    {
+        public SendFileView() { }
+
+        public SendFileView(SendFile file)
+        {
+            Id = file.Id;
```
cscharf, line 12: would `file` parameter ever be `null`?

src/Core/Models/View/SendTextView.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public SendTextView() { }
+        public SendTextView(SendText text)
+        {
+            Hidden = text.Hidden;
+        }
+
+        public string Text { get; set; } = null;
+        public bool Hidden { get; set; }
+        public string MaskedText => Text != null ? "••••••••" : null;
```
cscharf, line 15: Maybe change `Text != null` check to `!string.IsNullOrEmpty(Text)` (whitespace is the norm, however here white-space could be a password I guess (funny password, all spaces, but you never know I guess)

src/Core/Models/View/SendView.cs
```diff
@@ -0,0 +1,45 @@
...
+        public int? MaxAccessCount { get; set; }
+        public int AccessCount { get; set; }
+        public DateTime RevisionDate { get; set; }
+        public DateTime DeletionDate { get; set; }
+        public DateTime? ExpirationDate { get; set; }
+        public string Password { get; set; }
+        public bool Disabled { get; set; }
+        public string UrlB64Key => Key == null ? null : CoreHelpers.Base64UrlEncode(Key);
+        public bool MaxAccessCountReached => MaxAccessCount.HasValue && AccessCount >= MaxAccessCount.Value;
+        public bool Expired => ExpirationDate.HasValue && ExpirationDate.Value <= DateTime.Now;
```
cscharf, line 42: On mobile do we use `Now` vs. `UtcNow` where it's using the device-localized time? Not saying this is wrong, just not sure and wanted to ask.

src/Core/Services/ApiService.cs
```diff
@@ -215,6 +215,28 @@ public Task DeleteFolderAsync(string id)
 
         #endregion
 
+        #region Send APIs
+
+        public Task<SendResponse> GetSendAsync(string id)
+            => SendAsync<object, SendResponse>(HttpMethod.Get, $"/sends/{id}", null, true, true);
```
cscharf, line 221: Move all `=>` up a line

src/Core/Services/ApiService.cs
```diff
@@ -215,6 +215,28 @@ public Task DeleteFolderAsync(string id)
...
+
+        public Task<SendResponse> PostSendAsync(SendRequest request)
+            => SendAsync<SendRequest, SendResponse>(HttpMethod.Post, $"/sends", request, true, true);
```
cscharf, line 224: Remove the `$` qualifier as there are no values to interpolate here.

src/Core/Services/PclCryptoFunctionService.cs
```diff
@@ -43,6 +44,61 @@ public Task<byte[]> Pbkdf2Async(byte[] password, byte[] salt, CryptoHashAlgorith
             return Task.FromResult(_cryptoPrimitiveService.Pbkdf2(password, salt, algorithm, iterations));
         }
 
+        public async Task<byte[]> HkdfAsync(byte[] ikm, string salt, string info, int outputByteSize, HkdfAlgorithm algorithm)
+            => await HkdfAsync(ikm, Encoding.UTF8.GetBytes(salt), Encoding.UTF8.GetBytes(info), outputByteSize, algorithm);
```
cscharf, line 48: Arrow function goes up 1 line `=>`

src/Core/Services/SendService.cs
cscharf: I like the helper method, however let's match casing (so it doesn't look like a `const` named value) and name it as a function, `GetSendKey(string userId)` then I think this is good.

src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            await _apiService.DeleteSendAsync(id);
+            await DeleteAsync(id);
+        }
+
+        public async Task<(Send send, CipherString encryptedFileData)> EncryptAsync(SendView model, byte[] fileData,
+            string password, SymmetricCryptoKey key = null)
+        {
+            var send = new Send();
+            CipherString encryptedFileData = null;
```
cscharf, line 84: encryptedFileData var + assignment can be moved down, and the `new Send { ... }` can be simplified into a simple object constructor format.

src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,275 @@
...
+            SendResponse response = null;
+            if (send.Id == null)
+            {
+                if (send.Type == SendType.Text)
+                {
+                    response = await _apiService.PostSendAsync(request);
+                }
+                else if (send.Type == SendType.File)
+                {
+                    var fd = new MultipartFormDataContent($"--BWMobileFormBoundary{DateTime.UtcNow.Ticks}")
```
cscharf, line 205: Is this intended to use `UtcNow` where other areas use `Now` because it's going to the server? Again, just not sure, sorry.

src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,275 @@
...
+                    {
+                        { new StringContent(JsonConvert.SerializeObject(request)), "model" },
+                        { new ByteArrayContent(encryptedFileData), "data", send.File.FileName.EncryptedString }
+                    };
+
+                    response = await _apiService.PostSendFileAsync(fd);
+                }
+                send.Id = response?.Id;
```
cscharf, line 213: This line handles a `null` `response`, however what happens in this scenario upstream perhaps (is it possible?)

src/Core/Services/SendService.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            var userId = await _userService.GetUserIdAsync();
+            await UpsertAsync(new SendData(response, userId));
+        }
+
+        public async Task UpsertAsync(params SendData[] sends)
+        {
+            var userId = await _userService.GetUserIdAsync();
+            var knownSends = await _storageService.GetAsync<Dictionary<string, SendData>>(SEND_KEY(userId))
+                ?? new Dictionary<string, SendData>();
```
cscharf, line 228: `??` operator goes to the line above this.

src/Core/Services/SyncService.cs
```diff
@@ -363,11 +367,14 @@ private async Task SyncSettingsAsync(string userId, DomainsResponse response)
...
-        private async Task SyncPolicies(List<PolicyResponse> response)
+        private async Task SyncPoliciesAsync(List<PolicyResponse> response)
         {
             var policies = response?.ToDictionary(p => p.Id, p => new PolicyData(p)) ??
                 new Dictionary<string, PolicyData>();
             await _policyService.Replace(policies);
         }
+
+        private Task SyncSendsAsync(string userId, List<SendResponse> sends)
+            => _sendService.ReplaceAsync(sends.ToDictionary(s => userId, s => new SendData(s, userId)));
```
cscharf, line 378: move `=> `up

test/Common/AutoFixture/SutProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+        public SutProvider()
+        {
+            _dependencies = new Dictionary<Type, Dictionary<string, object>>();
+            _fixture = new Fixture().WithAutoNSubstitutions();
+            _constructorParameterRelay = new ConstructorParameterRelay<TSut>(this, _fixture);
+            _fixture.Customizations.Add(_constructorParameterRelay);
+        }
+
+        public SutProvider<TSut> SetDependency<T>(T dependency, string parameterName = "")
+            => SetDependency(typeof(T), dependency, parameterName);
```
cscharf, line 28: Please move `=>` up to the prior line

test/Common/TestHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+            var relevantExcludedProperties = excludedPropertyStrings.Where(name => !name.Contains('.')).ToList();
+            if (expected == null)
+            {
+                Assert.Null(actual);
+                return;
+            }
+
+            if (actual == null)
+            {
+                throw new Exception($"Expected object is null but actual is not");
```
cscharf, line 23: No string interpolation is happening here, please remove the `$` string prefix.

test/Core.Test/Services/SendServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+using Bit.Core.Models.Request;
+using Bit.Core.Test.AutoFixture;
+using System.Linq.Expressions;
+using Bit.Core.Models.View;
+
+namespace Bit.Core.Test.Services
+{
+    public class SendServiceTests
+    {
+        private string SEND_KEY(string userId) => SendService.SEND_KEY(userId);
```
cscharf, line 29: Maybe rename this one to match the other.

test/Core.Test/Services/SendServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+        [InlineCustomAutoData(new[] { typeof(SutProviderCustomization), typeof(SymmetricCryptoKeyCustomization), typeof(TextSendCustomization) })]
+        [InlineCustomAutoData(new[] { typeof(SutProviderCustomization), typeof(SymmetricCryptoKeyCustomization), typeof(FileSendCustomization) })]
+        public async Task EncryptAsync_Success(SutProvider<SendService> sutProvider, SendView view, byte[] fileData, SymmetricCryptoKey privateKey)
+        {
+            var prefix = "encrypted_";
+            var prefixBytes = Encoding.UTF8.GetBytes(prefix);
+
+
+            byte[] getPbkdf(string password, byte[] key)
+                => prefixBytes.Concat(Encoding.UTF8.GetBytes(password)).Concat(key).ToArray();
```
cscharf, line 338: Please move (well, you know by now 😉 )... `=>`

test/Core.Test/Services/SendServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                {
+                    Assert.Equal(expectedKvp.Key, actualKvp.Key);
+                    TestHelper.AssertPropertyEqual(expectedKvp.Value, actualKvp.Value);
+                }
+                return true;
+            };
+            await sutProvider.GetDependency<IStorageService>().Received(1).SaveAsync(SEND_KEY(userId), Arg.Is<Dictionary<string, SendData>>(d => matchSendsPredicate(d)));
+        }
+
+        // TODO test encrypt
```
cscharf, line 327: It looks like you are testing encrypt, if I'm not mistaken, you can probably remove the `// TODO` here.

src/Core/Models/Domain/SendFile.cs
MGibson1: Do you mean _just_ the fat arrow (`=>`)?

src/Core/Models/View/SendTextView.cs
MGibson1: I'm hesitant include empty for the same reason. I copied the behavior from [`LoginView`'s MaskedPassword member](https://github.com/bitwarden/mobile/blob/master/src/Core/Models/View/LoginView.cs#L23).

src/Core/Models/View/SendView.cs
MGibson1: I'm glad you caught this. I wanted to discuss and forgot.

I'm not 100% sure it'll matter since I believe these are transmitted as universal, but If not, I'd imagine we want a Send to expire at the same instant everywhere, not an hour ealier/later depending on where the phone happens to be.

src/Core/Models/Domain/Send.cs
mportune-bw: I would track the exception in AppCenter (via `Crashes.TrackError(e);`) then return null from this method so the caller can present "there was a problem" in the UI.

src/Core/Models/View/SendView.cs
cscharf: @mportune-bw , @kspearrin , can you guys weigh in here, perhaps you'll have a quick answer?

src/Core/Services/SendService.cs
MGibson1: This parameter sets the boundary in the `Content-Type: multipart/form-data;` header. To my understanding it can be whatever you want and doesn't need to be unique, but I just [copied it from CipherService](https://github.com/bitwarden/mobile/blob/master/src/Core/Services/CipherService.cs#L560-L561)

src/Core/Models/View/SendView.cs
kspearrin: it might not matter here since dot net should account for timezone of now vs utc. better make it UtcNow, though, since that is what will be coming from the server with this data.

src/Core/Models/Domain/SendFile.cs
MGibson1: It probably should. I think the compiler was willing to slap a default constructor in there for me, but I like the consistency.

I can see the `this()` call going both ways in the future. It depends on if we think things added to the default constructor should be called before the loaded one. Personally I wouldn't be expecting that and might do something wrong because of it.

src/Core/Models/View/SendTextView.cs
cscharf: That's fair, let's leave it for now. Thanks!

src/Core/Models/View/SendFileView.cs
MGibson1: As intended to be used, no. All `View` objects are created exclusively by a `DecryptAsync` call on their respective `Domain` objects, which pass themselves in as the parameter here. For example, the only reference to this constructor is in SendFile.cs and reads `new SendFileView(this)`

This behavior matches all the other `View`s as well. I'm totally fine with a guard clause returning `null`, but I think I'd want to write one on all the other `View` objects too

src/Core/Models/Domain/Send.cs
MGibson1: Discussed with Matt. Since this is in `Core`, the right thing to do is just throw here and trust the `App` project to handle it.

In light of that, I'm just taking out the try/catch entirely.

src/Core/Services/SendService.cs
MGibson1: Nothing else ever accesses this `response`.

To answer your question though, apiService will never return null for any of these calls (because `hasResponse` is set for all of them)

I null checked here out of habit due to assigning response null earlier.

MGibson1: Actually looking again, this is a catch against unknown Send Types. I'm going to convert to a switch and make that explicit, and create an empty `SendResponse` if it happens

src/Core/Services/SendService.cs
```diff
@@ -185,32 +186,36 @@ public async Task<Send> GetAsync(string id)
...
+                        var fd = new MultipartFormDataContent($"--BWMobileFormBoundary{DateTime.UtcNow.Ticks}")
+                        {
+                            { new StringContent(JsonConvert.SerializeObject(request)), "model" },
+                            { new ByteArrayContent(encryptedFileData), "data", send.File.FileName.EncryptedString }
+                        };
+
+                        response = await _apiService.PostSendFileAsync(fd);
+                        break;
+                    default:
+                        response = new SendResponse();
```
MGibson1, line 215: This default behavior idea might just be a bad one. It's identifying a possible future bug and recovering from it in a hard-to-track-down manner.

I think throwing would be better. Thoughts?

cscharf: yes, please add `throw` here

MGibson1: I ended up adding the throw only here. 

I don't think we want to throw whenever we encounter one since that would do bad things like lock up Sync if clients get out of date relative to servers. 

Instead, we should just keep the client from ruining saved data by complaining last minute that it doesn't know what it's doing.

cscharf (approved):
Looks good, please hold merge until after the release [code-freeze]

PR #1220, opened 2021-01-15
Restore vault timeout timer for Android

mportune-bw:
This PR restores the vault timeout timer in Android in a way that coexists with the new locking mechanism, since we're unable to reproduce the failure of the new mechanism reported elsewhere.  It is my hope that we can remove the timer (again) once we're able to establish what is behind the failure on some devices.

Note that the restored timer is only utilized as a trigger (like before) but the actual time comparison is still using the new monotonic clock value.  If for some reason this still doesn't work, we can be reasonably sure that the problem lies with the returned value and not the way the lock trigger is fired.

Also changed:

- Normalized elapsed time to milliseconds instead of seconds so we can perform time checks with multiplication instead of division (Desperation Level 1)
- Changed name of last active pref key to make sure there's no possible overlap with old DateTime pref value (Desperation Level 2)
- Var name tweaks to make time comparison blocks more troubleshoot-able (Desperation Level 2.5)

Tested both mechanisms together and independently (by disabling the other) across several devices successfully.

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -38,6 +38,7 @@ namespace Bit.Droid
         private IAppIdService _appIdService;
         private IStorageService _storageService;
         private IEventService _eventService;
+        private PendingIntent _vaultTimeoutAlarmPendingIntent;
         private PendingIntent _clearClipboardPendingIntent;
         private PendingIntent _eventUploadPendingIntent;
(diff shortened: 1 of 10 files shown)


cscharf (approved):
Let's hope you don't get to desperation level 3 😉 , looks good! 🤞🏻

PR #1230, opened 2021-01-21
Add passphrase generator to iOS Extensions

### LINKED ISSUE #620: Passphrase option missing from password generator in iOS “Extension”
When using the Bitwarden "extension" in apps or safari:

Select Add Item > Generate Password > the password generator page is missing any options to generate a “passphrase” based password.

In the actual Bitwarden app, if you select the generator tab, you are able to select a “passphrase” type password to generate as expected.

I realize this is small, so hopefully easily to resolve. I just wanted to mention it.

iOS 13.1.2

eliykat:
## Objective

Fix #620: the iOS autofill and extension password generator does not give the user the options to generate a passphrase.

Edit: I also noticed that the "Avoid ambiguous characters" switch was missing when generating a password, so I added that control as well.

## Code changes

- add required controls and logic to `iOS.Core/Controllers/PasswordGeneratorViewController.cs`.
- add some event handlers to `FormEntryTableViewCell` and `PickerTableViewCell` so that we can regenerate the password when their value changes
- let developer specify a custom indentation for `FormEntryTableViewCell` objects - this is required to line it up with the other controls (not sure if there is a better way of doing this)

## Screenshots
New passphrase controls:

Added "Avoid ambiguous characters" control (note that this now pushes the "Option defaults are set..." text off-screen on an iPhone 12 - not sure if that's worth worrying about):

### DIFF
--- a/src/iOS.Core/Controllers/PasswordGeneratorViewController.cs
+++ b/src/iOS.Core/Controllers/PasswordGeneratorViewController.cs
@@ -327,7 +327,7 @@ namespace Bit.iOS.Core.Controllers
 
                 if (_controller._passType == "password")
                 {
-                    return 7;
+                    return 8;
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


mportune-bw (approved):
Looks and works great!  Hold off on merging this until @cscharf gives the all-clear.

GolfAMV:
Great work @eliykat, thank you!

PR #1231, opened 2021-01-21
CollectionView migration

mportune-bw:
Continuation of #934 (new PR due to age of 934 resulting in a lot of diff noise).  Screenshots below show Before (L) & After (R) for iOS & Android.  _Almost_ pixel perfect but struggling in some areas, possibly due to changes in dip calculations under the hood.

@cscharf I've been defeated by the same NPE you ran into regarding the menu button within the grid item.  Everything I've read about `BindingContext` tells me the Cipher should be accessible so _something_ (probably obvious) is still afoot.

```
System.NullReferenceException: Object reference not set to an instance of an object

  at Bit.App.Utilities.AppHelpers.CipherListOptions (Xamarin.Forms.ContentPage page, Bit.Core.Models.View.CipherView cipher) [0x000c6] in /Users/matt/Documents/Work/Companies/Bitwarden/Repos/mobile/src/App/Utilities/AppHelpers.cs:24 

  at Bit.App.Pages.GroupingsPageViewModel.CipherOptionsAsync (Bit.Core.Models.View.CipherView cipher) [0x0004b] in /Users/matt/Documents/Work/Companies/Bitwarden/Repos/mobile/src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs:517 
```

ios

android


src/App/Utilities/IconConverter.cs
```diff
@@ -0,0 +1,111 @@
...
+            string image = null;
+            switch (cipher.Type)
+            {
+                case CipherType.Login:
+                    var loginIconImage = GetLoginIconImage(cipher);
+                    icon = loginIconImage.Item1;
+                    image = loginIconImage.Item2;
+                    break;
+                case CipherType.SecureNote:
+                    icon = "";
```
cscharf, line 49: Are these encoded correctly? Just want to ensure the character is still valid/good. IMO would be more clear to use the unicode char code w/ escape character here explicitly defined, or hex representation of that char code, vs. the actual embedded unicode character (which depending on your editor is pretty hard to decipher/check).

cscharf (commented):
Looks good, will follow up on the issue you're seeing there, maybe with a fresh head now I'll be of some use there, 😉

cscharf:
@mportune-bw , I notice that in `RowSelected` for `GroupingsPage` it checks the following:

```c#
if (!(e.SelectedItem is GroupingsPageListItem item))
{
    return;
}
```

I wonder if that has anything to do with the `null` exception you're getting when referencing the `GroupingsPageListItem` in that `CipherOptionsAsync(CipherView cipher)` method that's being called in the button command for that menu. Feels like I'm stretching it, like perhaps the command is being fired the first time with `null` and then it blows up, but perhaps it would be called twice if the first time just ignored the call and `return;` early.

mportune-bw:
@cscharf No dice unfortunately, however this leads me to look at something else possibly connected, so thanks!

mportune-bw:
Replaced with #1374

Issue #1206, opened 2021-01-02
Android - Disconnecting a USB OTG Keyboard while logging in cancels the login flow and returns to home screen.


SergeantConfused:
## Describe the Bug
If you use an external keyboard connected via USB OTG to enter your email and master password in the Android client, and then unplug it to connect your YubiKey, the application would go back to the first screen showing 3 buttons to Log in, Create account, or Enterprise Single Sign-On.

## Steps To Reproduce
1. Connect a keyboard via USB OTG to your Android device.
2. Open your Bitwarden application.
3. Enter your email and master password using the keyboard.
4. Once you're asked to use your YubiKey, please disconnect your USB OTG keyboard.

## Expected Result
To remain on the same page.

## Actual Result
I'm returned to the first screen, and the login flow is cancelled.

## Screenshots or Videos
N/A.

## Environment

- Device: Samsung Note 10+.
- Operating system: Android 10.
- Build Version: 2.7.2 (3300).
- Is this a Beta release? N.

## Additional Context
N/A.

fpdragon:
I debugged into it a little bit and I noticed that the method MainActivity.OnPause() and OnCreate() is triggered.
This brought me to:
[Stack Overflow](https://stackoverflow.com/questions/41420025/android-app-fires-oncreate-whenever-bluetooth-keyboard-connects-disconnects)

"Alternatively, you can tell Android that your app will handle "keyboard" configuration changes and it will not restart your activities when the keyboard connects/disconnects. "

I am no expert in Xamarin but I guess we have to add something like:
    <activity android:configChanges="keyboard" />
somehow to the AndroidManifest.xml?! O.o
Here is the Android Java description:
[Click](https://developer.android.com/guide/topics/resources/runtime-changes.html#RetainingAnObject)

Maybe this could help.

Issue #350, opened 2018-07-31
Login resets to beginning or crashes on Yubikey/keyboard plug in/out at 2FA screen


bradfa:
When trying to log in on Android, when plugging in a Yubikey or a standard keyboard via USB as a 2FA device at the 2FA screen, the app will return to the beginning of the login process or crash entirely.  This makes using a Yubikey via USB impossible unless you insert it prior to opening the Bitwarden app to start the login process.

The app appears to crash if I wipe all the app's data from the device and then try to log in, plugging my Yubikey in at the 2FA screen.  The app appears to go back to the start page of the login process when plugging my Yubikey in at the 2FA screen if I've already once successfully logged in.  I'm unsure why there's a difference here, cached or saved data locally matters?

My testing has been on an LG G5 model RS988 running Android 7.0, a Yubikey NEO via USB, and Bitwarden version 1.17.1 from the Play store.

kspearrin:
Unfortunately I don't have an Android device with a USB port on it at this time, so I can't reproduce the issue. I'll see if there is anything in the crash logs.

kspearrin:
Could you give me an approx timestamp on a crash (make sure you open the app again after the crash so that the report can be uploaded)?

bradfa:
@kspearrin I reproduced it while writing up this issue, so some time just prior to 12:54pm EDT today was when I experienced the "crashes" (in quotes because the app exited but I'm not entirely sure why, it seemed like a crash to me).
In testing right now, I'm now unable to reproduce the crash but I still can reproduce the "return to the beginning of the login process" failure mode.

bradfa:
@kspearrin just had it crash again just now, 3pm EDT.

kspearrin:
Ok, thanks. I found the crash log:

```
Xamarin caused by: android.runtime.JavaProxyThrowable: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Java.Lang.IllegalStateException: Foreground dispatch can only be enabled when your activity is resumed
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <e44be0cdd50d455da1e7c0f6bc4979bf>:0 
[...]

PR #1246, opened 2021-01-29
Fix crash when using Yubikey via usb on Android

### LINKED ISSUE #769: Android + Yubikey not working
## Describe the Bug

When plugging in a Yubikey for 2FA while logging in the user gets sent back to the starting log in screen.

## Steps To Reproduce

1. Set up a Yubikey as a 2FA device for your account
2. Log in
3. Click on the box so that the Yubikey can type it's key
4. Plug in key
5. Get sent back to the log in screen

## Expected Result

Log into the Bitwarden app and get to the Vault

## Actual Result

Sent back to the start of the login process

## Environment

- Device: Pixel 3
- Operating system: Android 10
- Build Version: 2.2.8
- Is this a Beta release? No
- Yubikey Version: Yubikey 5C

## Additional Context

I just am getting started with Yubikey and Bitwarden so there's a chance I have misconfigured something, but I kind of doubt this is expected behavior. 

**Also it seems like Android is treating the Yubikey as a separate keyboard, my guess is that this is somehow tied to it**

eliykat:
## Objective
Fix #769 and #350. Inserting a Yubikey via USB on the 2FA page will close the 2FA page and revert the user back to the initial welcome page. See example here: https://youtu.be/bKYIPMlMJVU

This actually happens on any of the pre-login pages, but obviously the 2FA page is the biggest concern.

## Cause of issue

Android apps use "Activity" objects to manage views and entry points into the app. Xamarin.Forms implements this by creating a single `MainActivity` to manage the app.

When certain configuration changes occur, Android rapidly destroys and recreates Activities to help them gracefully handle the configuration change. It is the app developer's responsibility to make sure that their Activities handle this lifecycle without loss of data or interruption to the user.

The problem is that plugging in a Yubikey or USB keyboard triggers a configuration change, which in turn destroys and recreats `MainActivity`. We evidently have nothing in place to handle this part of the lifecycle, so the app loses its state and reverts back to the entry page.

## Code changes

* `MainActivity.cs` - add Keyboard and KeyboardHidden events to the list of configuration changes that the app will handle itself without destroying and recreating the Activity. 
  * Note: [the official Android documentation says this is a "last resort" only](https://developer.android.com/guide/topics/manifest/activity-element.html#config) because our app should handle its Activity lifecycle properly. However, we already use this option for configuration changes triggered by screen size and orientation, so I've done the same. It seems to "just work" but we may want to consider handling it properly.

* `TwoFactorPage.xaml.cs` - fixed so that the app will focus on the Yubikey input element when the 2FA page loads. The app will not receive the Yubikey code unless this element has focus.

## Testing considerations

Note: Yubikeys are recognised as USB keyboards.

Expected behaviour:
* given the user is on the Yubikey 2FA page, the user should be able to plug in the Yubikey via USB, press the Yubikey's button, and be logged in without any further action required.
* user should be able to insert a Yubikey or USB keyboard on any page within the app without changing the current page or causing any errors.

If we are adopting the solution proposed above, QA should try triggering various events related to the keyboard on different pages to make sure that they are still properly handled by the app. e.g. plugging/unplugging a physical keyboard, hiding/showing software keyboard, rotating to change the software keyboard dimensions.

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -28,7 +28,8 @@ namespace Bit.Droid
         Theme = "@style/LaunchTheme",
         MainLauncher = true,
         ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation |
-                               ConfigChanges.Keyboard | ConfigChanges.KeyboardHidden)]
+                               ConfigChanges.Keyboard | ConfigChanges.KeyboardHidden |
+                               ConfigChanges.Navigation)]
     [Register("com.x8bit.bitwarden.MainActivity")]
     public class MainActivity : Xamarin.Forms.Platform.Android.FormsAppCompatActivity
     {
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -27,7 +27,8 @@ namespace Bit.Droid
         Icon = "@mipmap/ic_launcher",
         Theme = "@style/LaunchTheme",
         MainLauncher = true,
-        ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation)]
+        ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation |
+                               ConfigChanges.Keyboard | ConfigChanges.KeyboardHidden)]
     [Register("com.x8bit.bitwarden.MainActivity")]
     public class MainActivity : Xamarin.Forms.Platform.Android.FormsAppCompatActivity
     {
--- a/src/App/Pages/Accounts/TwoFactorPage.xaml.cs
+++ b/src/App/Pages/Accounts/TwoFactorPage.xaml.cs
@@ -102,6 +102,9 @@ namespace Bit.App.Pages
                 if (_vm.TotpMethod)
                 {
                     RequestFocus(_totpEntry);
+                } else if (_vm.YubikeyMethod)
+                {
+                    RequestFocus(_yubikeyTokenEntry);
                 }
                 return Task.FromResult(0);
             });


mportune-bw (approved):
This will most likely fix #1206 as well.

eliykat:
Good pickup @mportune-bw. I did some more testing to confirm I was right about the USB keyboard issue, but found that I had to add `ConfigChanges.Navigation` as well to fix the keyboard.

Android documentation helpfully explains `ConfigChanges.Navigation` as:
> The navigation type (trackball/dpad) has changed. (This should never normally happen.)

mportune-bw:
> Good pickup @mportune-bw. I did some more testing to confirm I was right about the USB keyboard issue, but found that I had to add `ConfigChanges.Navigation` as well to fix the keyboard.
> 
> Android documentation helpfully explains `ConfigChanges.Navigation` as:
> 
> > The navigation type (trackball/dpad) has changed. **(This should never normally happen.)**
> 
> My guess is that it's triggered by keyboards with trackpads or usb receivers for keyboard + mouse combos.

Hmm I've never needed to add `.Navigation` in the past to handle HID add/removal.  What kind of keyboard are you using to test, and how are you hooking it up?

eliykat:
I've tested with a [Microsoft Wireless Desktop 3050](https://www.jbhifi.com.au/products/microsoft-wireless-desktop-3050-with-aes) and a cheaper Dell equivalent. Both are keyboard and mouse combos that use a USB transceiver. They're USB-A so I used an adapter to plug it into the phone's USB-C port.

mportune-bw (approved):
Let's go ahead and get this in so testers can hit it.

PR #1262, opened 2021-02-12
Send bugfixes & tweaks

mportune-bw:
- Fixed improper file field validation when editing a file Send
- Fixed "no file chosen" label after selecting file
- Fixed blank deletion date/time picker when current date is today or time is midnight
- Fixed all `Now` references to have minute-precision (strip seconds & milliseconds)
- Fixed `Name` field focus issue when switching between file and text types
- Moved `Name` field above Send type picker per discussion

Screen Shot 2021-02-12 at 12 27 15 PM
Screen Shot 2021-02-12 at 12 27 24 PM

### DIFF
--- a/src/App/Pages/Send/SendAddEditPage.xaml.cs
+++ b/src/App/Pages/Send/SendAddEditPage.xaml.cs
@@ -111,11 +111,17 @@ namespace Bit.App.Pages
         private void TextType_Clicked(object sender, EventArgs eventArgs)
         {
             _vm.TypeChanged(SendType.Text);
+            _nameEntry.ReturnType = ReturnType.Next;
+            _nameEntry.ReturnCommand = new Command(() => _textEditor.Focus());
+            RequestFocus(_nameEntry);
         }
         
         private void FileType_Clicked(object sender, EventArgs eventArgs)
         {
             _vm.TypeChanged(SendType.File);
+            _nameEntry.ReturnType = ReturnType.Done;
+            _nameEntry.ReturnCommand = null;
+            RequestFocus(_nameEntry);
         }
 
         private void OnMaxAccessCountTextChanged(object sender, TextChangedEventArgs e)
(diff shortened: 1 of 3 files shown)


src/App/Pages/Send/SendAddEditPage.xaml.cs
```diff
@@ -111,11 +111,17 @@ protected override void OnDisappearing()
         private void TextType_Clicked(object sender, EventArgs eventArgs)
         {
             _vm.TypeChanged(SendType.Text);
+            _nameEntry.ReturnType = ReturnType.Next;
+            _nameEntry.ReturnCommand = new Command(() => _textEditor.Focus());
+            RequestFocus(_nameEntry);
```
cscharf, line 116: Would we want this focused "above" the fold when the type is changed, but now since Name is above the Type, I would think changing the type shouldn't also change or re-focus the Name field, it should "in theory" already be populated and unchanging.

src/App/Pages/Send/SendAddEditPageViewModel.cs
```diff
@@ -184,6 +195,7 @@ public async Task<bool> LoadAsync()
             // TODO Policy Check
```
cscharf, line 195: Let's not forget about this ☝🏻

src/App/Pages/Send/SendAddEditPage.xaml.cs
mportune-bw: The physical touch of the type buttons grabs focus from the Name field, and this puts it back.  :)

mportune-bw: fixed per discussion, now checking for empty field before refocus

src/App/Pages/Send/SendAddEditPageViewModel.cs
mportune-bw: per discussion will be a new PR

cscharf (approved):

PR #1265, opened 2021-02-16
Expandable options for Add/Edit Send

mportune-bw:
Options section is collapsed by default when adding or editing Sends for simplification.

**Considerations:**

- `Share this Send upon save` switch is the last element before the Options panel (instead of the very end of the form) since the `Save` button on mobile is at the top.
- It's amazing how inconsistently apps us the "down" vs. "up" chevron to represent "expanded" vs. "collapsed".  I chose the "action to be performed when pressing this button" approach, so "down" is "collapsed".  I'm not super attached to it though if anybody has strong feelings otherwise.  (Some UIs show a right-facing chevron to indicate collapsed, but on mobile that can drive an expectation that we'll be taken to the "next screen", which isn't what we're doing here.  I'm probably overthinking it.)

**Screenshots:**

Screen Shot 2021-02-16 at 2 15 36 PM
Screen Shot 2021-02-16 at 2 15 41 PM
Screen Shot 2021-02-16 at 2 16 05 PM
Screen Shot 2021-02-16 at 2 16 08 PM

### DIFF
--- a/src/App/Pages/Send/SendAddEditPage.xaml
+++ b/src/App/Pages/Send/SendAddEditPage.xaml
@@ -89,22 +89,27 @@
                                 StyleClass="box-footer-label"
                                 Margin="0,5,0,0" />
                         </StackLayout>
-                        <StackLayout StyleClass="box-row"
-                                     IsVisible="{Binding EditMode, Converter={StaticResource inverseBool}}">
+                        <StackLayout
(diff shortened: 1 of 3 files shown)


cscharf (approved):
I think this looks good, and I think the chevron usage is consistent with expectations.

PR #1292, opened 2021-03-01
clear Send password if whitespace

addisonbeck:
Resolves https://app.asana.com/0/1199659118818122/1199987076477970/f

Sibling to https://github.com/bitwarden/jslib/pull/287

### DIFF
--- a/src/App/Pages/Send/SendAddEditPageViewModel.cs
+++ b/src/App/Pages/Send/SendAddEditPageViewModel.cs
@@ -337,7 +337,7 @@ namespace Bit.App.Pages
 
             UpdateSendData();
 
-            if (NewPassword != null && string.IsNullOrWhiteSpace(NewPassword)) 
+            if (string.IsNullOrWhiteSpace(NewPassword)) 
             {
                 NewPassword = null;
             }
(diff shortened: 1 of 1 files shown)


src/App/Pages/Send/SendAddEditPageViewModel.cs
```diff
@@ -337,6 +337,11 @@ public async Task<bool> SubmitAsync()
 
             UpdateSendData();
 
+            if (NewPassword != null && string.IsNullOrWhiteSpace(NewPassword))
```
cscharf, line 340: You can remove the bit of `NewPassword != null && `, `string.IsNullOrWhiteSpace()` already handles the `null` condition for you.

addisonbeck: In this case though, null is the value we are after. I'm just using IsNullOrWhiteSpace() for the empty/whitespace catch. If NewPassword is already null we don't need to null it out.

cscharf: setting `NewPassword` to `null` if it's already `null`... no harm, no foul. The `if` check just seems a little too complicated/confusing perhaps... maybe it's just me 🤷🏻

addisonbeck: I def see either way. Went ahead and removed it. Less code is usually better code.

cscharf (approved):

PR #1301, opened 2021-03-05
Added "About Send" icon & menu option

mportune-bw:
Added new "About Send" toolbar button (for iOS) and menu option (for Android) to send the user to our Send page.

Screen Shot 2021-03-05 at 6 29 55 PM


cscharf (commented):
Missing some file contents on the `.png`'s.

cscharf (approved):

PR #1313, opened 2021-03-12
Fix for incorrect deletion/expiration times during Send creation

mportune-bw:
When creating a new Send, the DateTimes created from the "simple" picker were being set to midnight instead of the current time.  Not really a showstopper until "one hour" is selected, resulting in a DateTime from the past ([today] at midnight).  Some investigation revealed that modifying the 'DeletionDate' property bound to `ExtendedDatePicker` was triggering property changes multiple times within the picker and resetting the time to midnight.  Those custom pickers are no fun, so for safety/sanity during this release I've given the "simple" pickers their own "simple" DateTimes to modify, isolating them from the values bound to the custom pickers.  I will revisit this later to come up with something more elegant.

Edit: Forgot to mention also fixed a missing "Send" capitalization


cscharf (approved):
Seems to work as I look at it, agree it's not that elegant but the logic seems to work in my head as I follow it through.

PR #1340, opened 2021-03-26
Hide email address in Sends

eliykat:
## Objective
Allow the creator of a Send to hide their email address from the recipient.

See the [web PR](https://github.com/bitwarden/web/pull/895) and [server PR](https://github.com/bitwarden/server/pull/1234) for more info and discussion.

## Code changes
* add "Hide my email address from recipients" toggle slider to the Send AddEdit page
* check and enforce new Send Options Policy > which can disable this new feature (`disableHideEmail`)
* add banner at the top of the Send AddEdit page if Send Options policy is in effect

## Screenshots

Screen Shot 2021-03-26 at 3 44 29 pm

Screen Shot 2021-03-26 at 3 44 42 pm

### DIFF
--- a/src/App/Pages/Send/SendAddEditPageViewModel.cs
+++ b/src/App/Pages/Send/SendAddEditPageViewModel.cs
@@ -94,6 +94,7 @@ namespace Bit.App.Pages
         public byte[] FileData { get; set; }
         public string NewPassword { get; set; }
         public bool ShareOnSave { get; set; }
+        public bool DisableHideEmailControl { get; set; }
         public List<KeyValuePair<string, SendType>> TypeOptions { get; }
         public List<KeyValuePair<string, string>> DeletionTypeOptions { get; }
         public List<KeyValuePair<string, string>> ExpirationTypeOptions { get; }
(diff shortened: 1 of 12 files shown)


src/App/Pages/Send/SendAddEditPageViewModel.cs
```diff
@@ -200,12 +203,25 @@ public bool ShowPassword
         public bool ShowDeletionCustomPickers => EditMode || DeletionDateTypeSelectedIndex == 6;
         public bool ShowExpirationCustomPickers => EditMode || ExpirationDateTypeSelectedIndex == 7;
         public string ShowPasswordIcon => ShowPassword ? "" : "";
+        public bool DisableHideEmail
```
mportune-bw, line 206: OCD alert: `DisableHideEmail` and `SendOptionsPolicyInEffect` should be above this section (probably right after `ShowPassword` and before `EditMode`), and `DisableHideEmailControl` should be up even higher (probably right after `ShareOnSave` and before `TypeOptions`)

src/App/Resources/AppResources.resx
```diff
@@ -1972,4 +1972,10 @@
     <value>About Send</value>
     <comment>'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated.</comment>
   </data>
+  <data name="HideEmail" xml:space="preserve">
+    <value>Hide my email address from recipients.</value>
+  </data>
+  <data name="SendOptionsPolicyInEffect" xml:space="preserve">
+    <value>One or more organization policies are affecting your Send options.</value>
```
mportune-bw, line 1979: Strings containing 'Send' should also include the `<comment>` about Send not being translated (see above)

mportune-bw (changes requested):
Looks good, just a few tweaks needed

mportune-bw (approved):

nonobio:
Hi,

As a user, is it already usable ?

Thanks

cscharf:
> As a user, is it already usable ?

It will be available in the clients as of next week when we plan to do our client deployments. Until then it is not yet usable (unless you're using a Beta client or direct build asset from one of our client `rc` (release candidate) branch builds.

nonobio:
Thanks. I installed Android Beta and the option is present :) .

PR #1343, opened 2021-03-29
Share-to-Send for Android

mportune-bw:
**Overview**

Added the ability to share from other apps directly to Bitwarden Send.

**Files Changed/Added**

Android:

- **DeviceActionService.cs**: Implemented `CloseMainApp()` method for returning to source app upon completing the share operation
- **MainActivity.cs**: Added `IntentFilter` to declare receivable mime types, and logic to parse the intent and place the results into `AppOptions`

App:

- **IDeviceActionService.cs**: Added `CloseMainApp()` method for use by Android
- **AppOptions.cs**: Added `CreateSend` property for maintaining received Send data through the auth/unlock process
- **SendGroupingsPage.xaml.cs**: Added `AppOptions` to constructor to load received Send data during `LoadOnAppearedAsync` process, and removed `ShowPreviousPageAsync` pasta
- **SendGroupingsPageViewModel.cs**: Added null appOptions arg when pushing modal `SendAddEditPage`
- **SendAddEditPage.xaml**: Adjusted component visibility based on smarter properties declared below
- **SendAddEditPage.xaml.cs**: Added `AppOptions` to constructor to load received Send data, and handled any existing create request by populating key Send values and adjusting UI accordingly
- **SendAddEditPageViewModel.cs**: Added `IsAddFromShare` property to drive UI and share completion behavior, and `ShowTypeButtons` property to include `IsAddFromShare` in the decision to display type buttons (they aren't displayed when sharing to Send as the type is declared by the mime type of the data to share)
- **SendsPageViewModel.cs**: Added null appOptions arg when pushing modal `SendAddEditPage`
- **TabsPage.cs**: Passed `AppOptions` reference to `SendGroupingsPage` and added programmatic support for jumping direction to Send tab
- **AppResources.resx**: Added new Send-specific 'premium required' string for when non-premium users attempt to share a file from a 3rd party app
- **AppHelpers.cs**: Added null appOptions arg when pushing modal `SendAddEditPage`, and included Send as an alternate MainPage if the appOptions data is populated
- **App.xaml.cs**: Added handler for `popAllAndGoToTabSend` message, and included Send as a MainPage if appOptions data is populated

iOS.Core:

- **DeviceActionService.cs**: Implemented `CloseMainApp()` method though only used for Android

**Screenshots**

_Sharing app from Play Store:_

_Sharing photo from Camera:_

_Sharing text from selection:_

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -30,6 +30,16 @@ namespace Bit.Droid
         ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation |
                                ConfigChanges.Keyboard | ConfigChanges.KeyboardHidden |
                                ConfigChanges.Navigation)]
+    [IntentFilter(
+        new[] { Intent.ActionSend },
+        Categories = new[] { Intent.CategoryDefault },
+        DataMimeTypes = new[]
+        {
+            @"application/*",
+            @"image/*",
+            @"video/*",
+            @"text/*"
+        })]
     [Register("com.x8bit.bitwarden.MainActivity")]
     public class MainActivity : Xamarin.Forms.Platform.Android.FormsAppCompatActivity
     {
@@ -169,13 +179,21 @@ namespace Bit.Droid
                     _appOptions.GeneratorTile = true;
                 }
             }
-            if (intent.GetBooleanExtra("myVaultTile", false))
+            else if (intent.GetBooleanExtra("myVaultTile", false))
             {
                 _messagingService.Send("popAllAndGoToTabMyVault");
                 if (_appOptions != null)
                 {
                     _appOptions.MyVaultTile = true;
                 }
+            }
+            else if (intent.Action == Intent.ActionSend && intent.Type != null)
+            {
+                if (_appOptions != null)
+                {
+                    _appOptions.CreateSend = GetCreateSendRequest(intent);
+                }
+                _messagingService.Send("popAllAndGoToTabSend");
             }
             else
             {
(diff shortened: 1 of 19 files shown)


src/Android/MainActivity.cs
```diff
@@ -30,6 +30,16 @@ namespace Bit.Droid
         ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation |
                                ConfigChanges.Keyboard | ConfigChanges.KeyboardHidden |
                                ConfigChanges.Navigation)]
+    [IntentFilter(
```
cscharf, line 33: Assuming the `IntentFilterAttribute` doesn't restrict by filter based on the intent actions exclusively (but that is how it's applying the filter when filtering instead; if types match then filter based on X)?

cscharf: Yeah, Microsoft's docs are useless here, figures: https://docs.microsoft.com/en-us/dotnet/api/android.app.intentfilterattribute?view=xamarin-android-sdk-9

cscharf: I'm assuming since the attribute target allows multiple, it would imply this acts as a filter for _when_ to apply the filter itself, vs. being part of the filter.

cscharf (approved):

src/Android/MainActivity.cs
mportune-bw: @cscharf I've read this 20 times and I'm still not sure what you're asking me.  :)  The MS docs are also confusing me since it keeps redirecting everything back to Android 9...

cscharf: Yeah, sorry Matt, it was just that I wasn't sure that the values being passed in the string[] array as the first argument of the `[IntentFilter]` would filter/limit the call to only those actions exclusively, making it "not fire" for any other actions than what's listed in the `[IntentFilter]`, so basically asking if this filter would cause any actions OTHER than `new[] { Intent.ActionSend }` to be filtered out/rejected. I don't think so however, so I believe you're good. I think you can ignore my concern.

mportune-bw: Ah ok, I think I follow now, and your assumption is correct - the filters only apply to this intent action.  Other actions can be declared with their own set of filters, even if they include the same mime types.

PR #1366, opened 2021-04-14
Use CipherByteArray to signify encrypted  byte[]

MGibson1:
# Overview

Similar PR to bitwarden/jslib#346. There was no bug in this repository, but the tiny type protection is nice to translate over to mobile as well.

All of these changes constitute changing `CipherService.EncrypteToBytes` return value to `CipherByteArray` rather than `byte[]` and spidering that through change to maintain the type until it is uploaded to the server.


cscharf (commented):
Looks good from a code perspective. I understanding wanting consistency between mobile and jslib, however not sure the abstraction is really necessary. Will leave to @mportune-bw  for final approval.

MGibson1:
I would say it's about more than just consistency.
The reason I went this route in jslib is because by changing type to distinguish between otherwise generic types we can catch these kind of issues before we even compile. It's the same argument given for the tiny type pattern: https://darrenhobbs.com/2007/04/11/tiny-types/

Because c# is also statically typed, we get the same benefit here.

mportune-bw:
My only nit is the use of `CipherByteArray` for Send, where the use of `byte[]` avoided that issue.  (If you recall parts of Send originally used `CipherString` which was confusing when I started building the UI).  A more generic name might be in order.  Thoughts?

MGibson1:
@mportune-bw, not quite sure what you mean. Send takes in `byte[]` cleartext documents and previously output `byte[]` ciphertext documents. The point is to make it clear that any `CipherByteArray` is encrypted.

Or are you saying that the `CipherString` class could handle this case as well?

mportune-bw:
I'm saying 'Cipher' in our naming convention has nothing to do with 'Send' (Our 'cipher' is a vault item, not a traditional cipher in the encryption sense.  I kinda wish this wasn't the case, but...)

MGibson1:
Ah I see what you're saying. Happy to change the name if you have something better. Perhaps ironically, I named this one after `CipherString`, which you don't like 😛

mportune-bw:
I imagine it's a bigger discussion since it affects more than mobile.  I recall the `Enc` prefix used elsewhere to represent encrypted data, so maybe `EncByteArray`?  Any thoughts @cscharf?

cscharf (dismissed):

mportune-bw (approved):

PR #1395, opened 2021-05-11
Explain how to verify email for file Sends

eliykat:
## Objective

Users are required to verify their email before creating a file Send, however the current message doesn't provide any explanation about how to do this. This is especially a problem because you can't verify your email in the mobile app, so it's not obvious that you have to go somewhere else to do it.

## Code changes

I've used the same UX flow as settings that can only be changed in the web vault:
* explain to the user that this can only be done on the "bitwarden.com web vault" (may be inaccurate for self hosted, but I've stuck to the current pattern)
* offer to take the user there now, which takes the user to the relevant help documentation

I'll replicate these changes for desktop and browser tomorrow.

Bonus: deleted unused private var I left there from the 'hide email' feature.

### DIFF
--- a/src/App/Pages/Send/SendAddEditPageViewModel.cs
+++ b/src/App/Pages/Send/SendAddEditPageViewModel.cs
@@ -45,6 +45,7 @@ namespace Bit.App.Pages
         };
         private bool _disableHideEmail;
         private bool _sendOptionsPolicyInEffect;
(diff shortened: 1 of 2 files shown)


MGibson1 (approved):
I don't see how you could possibly have caused that test to fail, but want feel free to fix if you want...

PR #1407, opened 2021-05-19
Updated MyVault string for spanish

sergioteula:
The string for MyVault in spanish was too long, so I've changed from "Mi caja fuerte" to "Caja fuerte".  This solves the issue where the title on the navbar appears cutted.


cscharf:
Hi @sergioteula , thank you for the contribution, however we manage all of our translations through Crowdin, so if we accepted this PR it would simply get overwritten the next time we sync from there. Please check out this key for the Spanish translation in Crowdin and feel free to contribute from there! https://crowdin.com/translate/bitwarden-mobile/45/en-es?filter=basic&value=0#q=MyVault

PR #1416, opened 2021-05-31
Fixes for password reprompt

Hinton:
## Objective
Resolves some scenarios where the master password prompt didn't work.

For autofill it has two scenarios:

- **Locked:** Since we don't have access to the decrypted item, we don't know if it needs MP prompt, therefore we first unlock using whatever method is available. If the MP was not used, we prompt for it using a dialog after unlocking.
- **Not locked:** If the item requires prompt, we display a customized lock screen prompting for the password.

### Code Changes
- **src/App/Pages/Vault/ViewPage.xaml.cs**: Add password prompt for "more" menu on iOS.
- **src/App/Services/MobilePlatformUtilsService.cs**: Remove "wrong password" dialog when clicking cancel.
- **src/App/Utilities/AppHelpers.cs**: Add password prompt for edit button.
- **src/iOS.Autofill/CredentialProviderViewController.cs**: Add password prompt during autofill.
- **src/iOS.Core/Controllers/LockPasswordViewController.cs**: Display a customized lock screen when items is protected, and set a flag when authenticated using password.


kspearrin (dismissed):

src/iOS.Core/Controllers/LockPasswordViewController.cs
```diff
@@ -28,6 +28,10 @@ public abstract class LockPasswordViewController : ExtendedUITableViewController
         private bool _pinLock;
         private bool _biometricLock;
         private bool _biometricIntegrityValid = true;
+        private int _invalidPinAttempts;
```
mportune-bw, line 31: `_invalidPinAttempts` should no longer exist

mportune-bw (commented):
I'm concerned about the re-introduction of that var.  Can you sanity check this against https://github.com/bitwarden/mobile/pull/1421/files from yesterday?

src/iOS.Core/Controllers/LockPasswordViewController.cs
Hinton: Yea it shouldn't be there.

mportune-bw (approved):

PR #1427, opened 2021-06-13
2FA screen: move continue to a button

### LINKED ISSUE #1419: Inconsistent placement of the "use another method" button in the 2FA authentication screen
## Describe the Bug
When logging in Bitwarden, the "log in" button is located at the bottom, but in the next screen, the "use another method" button is in the same place of the "log in" button, instead of the "continue" button. If you accidentally press the "use another method" button and press back, the whole process will be canceled. This inconsistency creates a non-intuitive design.

## Steps To Reproduce
1. Log in with your credentials (2FA must be enabled).

## Expected Result
The continue button should be at the bottom.

## Actual Result
N/A

## Screenshots or Videos
N/A

## Environment
- Device: A50
- Operating system: Android 11
- Build Version (go to "Settings" → "About" in the app): 2.10.0 (3516)
- Is this a Beta release? no

## Additional Context
N/A

hantatsang:
Close #1419

### Changes

- Move *"Continue"* to a button at the bottom
- Button is disabled until *Verification Code* field is not empty

### Demo

https://user-images.githubusercontent.com/11912225/121795363-511ae100-cc53-11eb-85d2-1a1c228f0a0b.mp4


src/App/Pages/Accounts/TwoFactorPage.xaml
```diff
@@ -123,6 +122,12 @@
                     Margin="10, 20, 10, 10"
                     HorizontalTextAlignment="Center" />
             </StackLayout>
+            <Button Text="{u:I18n Continue}"
+                    IsEnabled="False"
+                    IsVisible="{Binding EmailMethod}"
```
eliykat, line 127: I think this should be bound to `ShowContinue`, right? Currently I'm missing the button when using the authenticator method.

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
```diff
@@ -195,5 +183,17 @@ private async Task TwoFactorAuthSuccessAsync()
                 Application.Current.MainPage = new TabsPage(_appOptions, previousPage);
             }
         }
+
+        private void Token_TextChanged(object sender, TextChangedEventArgs e)
+        {
+            if (string.IsNullOrWhiteSpace(e.NewTextValue))
+            {
+                ContinueButton.IsEnabled = false;
```
eliykat, line 191: Please use a property like `EnableContinue`, which is bound to the `IsEnabled` property of the button in the XAML. This is more consistent with how the other buttons here are handled, and how you've handled `IsVisible` for this button.

eliykat (changes requested):
Thanks for volunteering to handle this, @hantatsang. I've left a couple of comments. 

Can you also move the "Use Another Two-Step Login Method" to a "three dots" menu in the top-right hand corner? (e.g. where the "Get your master password hint" is on the login screen.) I think this is a better place for secondary actions.

eliykat:
Also adding a `hold` label to remind myself not to merge before the next release.

src/App/Pages/Accounts/TwoFactorPage.xaml
hantatsang: Oops, damn typo 😬

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -302,5 +311,17 @@ public async Task<bool> SendEmailAsync(bool showLoading, bool doToast)
...
             }
         }
+
+        public void SetEnableContinue(bool isTokenEmpty)
+        {
+            if (isTokenEmpty)
+            {
+                EnableContinue = false;
+            }
+            else if (!EnableContinue)
```
eliykat, line 321: Checking `!EnableContinue` before setting it to `true` is unnecessary. The compiler would probably optimize it out anyway. I think you could remove this method entirely and set `_vm.EnableContinue` directly from `Token_TextChanged`.

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
```diff
@@ -22,6 +22,8 @@ public partial class TwoFactorPage : BaseContentPage
         private bool _authingWithSso;
         private string _orgIdentifier;
 
+        public Button ContinueButton { get; set; }
```
eliykat, line 25: You don't need this anymore after the last round of changes. You can get rid of the corresponding private property as well (`_continueButton`).

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
```diff
@@ -45,25 +47,11 @@ public TwoFactorPage(bool? authingWithSso = false, AppOptions appOptions = null,
...
-
-        public HybridWebView DuoWebView { get; set; }
 
-        public void AddContinueButton()
-        {
-            if (!ToolbarItems.Contains(_continueItem))
-            {
-                ToolbarItems.Add(_continueItem);
-            }
+            ContinueButton = _continue;
```
eliykat, line 51: You don't need this anymore after the last round of changes.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -171,11 +180,11 @@ public void Load()
             }
             if (SelectedProviderType == null || DuoMethod)
             {
-                page.RemoveContinueButton();
+                ShowContinue = false;
```
eliykat, line 183: You can simplify this `if` block to:
`ShowContinue = !(SelectedProviderType == null || DuoMethod)`

eliykat (changes requested):
Looking good! I've added a few comments, which are mostly to clean up after your first set of changes or to simplify style.

Probably the main task is to get the "..." menu working on iOS. Here's how it looks currently:

Screen Shot 2021-06-22 at 3 21 05 pm

It's been a while since I made similar changes to the login screen, but from memory you have to manually add the "..." icon in iOS, it doesn't just do it for you like with Android. Check the following parts of the login page for reference:
* `ResourceDictionary` in the xaml: https://github.com/bitwarden/mobile/blob/master/src/App/Pages/Accounts/LoginPage.xaml#L16
* check the current platform and add the required item in the code behind: https://github.com/bitwarden/mobile/blob/master/src/App/Pages/Accounts/LoginPage.xaml.cs#L45
* add event handler for the menu in iOS: https://github.com/bitwarden/mobile/blob/master/src/App/Pages/Accounts/LoginPage.xaml.cs#L95

I remember iOS being tricky to set up locally without our team credentials. If you can't get the iOS simulator running, then just do your best with the code and I'll test it at my end.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
hantatsang: Thanks @eliykat . I'll update the code over the weekend

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
hantatsang: ✅

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
hantatsang: ✅

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
hantatsang: ✅

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
hantatsang: ✅

hantatsang:
@eliykat I don't have iPhone, please help me test the implementation

src/App/Pages/Accounts/TwoFactorPage.xaml
```diff
@@ -16,14 +16,18 @@
...
     <ContentPage.Resources>
         <ResourceDictionary>
             <u:InverseBoolConverter x:Key="inverseBool" />
             <u:IsNullConverter x:Key="isNull" />
+            <ToolbarItem Icon="more_vert.png" Clicked="More_Clicked" Order="Primary"
+                x:Name="_moreItem" x:Key="moreItem"
+                 AutomationProperties.IsInAccessibleTree="True"
+                 AutomationProperties.Name="{u:I18n Options}" />
+            <ToolbarItem Text="{u:I18n UseAnotherTwoStepMethod}" Clicked="Methods_Clicked" Order="Secondary"
+                x:Name="_useAnotherTwoStepMethod" x:Key="useAnotherTwoStepMethod" />
```
eliykat, line 30: Formatting nit: please make this consistent with how other elements are layed out, with each attribute on a new line:
```suggestion
            <ToolbarItem Text="{u:I18n UseAnotherTwoStepMethod}" 
                Clicked="Methods_Clicked" 
                Order="Secondary"
                x:Name="_useAnotherTwoStepMethod" 
                x:Key="useAnotherTwoStepMethod" />
```

src/App/Pages/Accounts/TwoFactorPage.xaml.cs
```diff
@@ -145,6 +137,21 @@ private async void Methods_Clicked(object sender, EventArgs e)
             }
         }
 
+        private async void More_Clicked(object sender, EventArgs e)
+        {
+            if (DoOnce())
```
eliykat, line 142: This condition should be inverted. It's intended to prevent double-clicks, but at the moment it _requires_ a double-click.
```suggestion
            if (!DoOnce())
```

eliykat (changes requested):
Just a couple of minor changes, if you agree then we should be able to merge. Thanks!

eliykat (approved):
Great work, thanks!

PR #1428, opened 2021-06-13
Default usernames

michalchecinski:


fedemkr:
Hi @michalchecinski , is this a work in progress? I see that the `DefaultUsernames` is being filled but as far as I see it's not being used anywhere. Could you fill the PR template with the details for this?
Btw I think this is not compiling: `DefaultUsernames = _cipherService.GetTopUsernamesAsync();` it should break because you can't assign a `Task<IEnumerable<string>>` to a `IEnumerable<string>`

vvolkgang:
After talking with @michalchecinski, we can go ahead and close this for now.

PR #1433, opened 2021-06-17
Redefine cipher "share" to "move to organization"

MGibson1:
# Overview

The term "Share" is confusing to users because it implies maintaining ownership and control of a cipher while allowing others to access it as well. It also implies the ability to revoke access or to stop sharing -- neither of these things is true.

In addition to the share term being misleading, the [`fa-share-alt`](https://fontawesome.com/v4.7/icon/share-alt) icon implies a forking of the cipher item rather than transferring ownership.

When a cipher is "shared" it is irrevocably taken over by an organization. This occurs for good cryptographic and security reasons. The better solution to this is to revise the communication around "sharing." To that end, this work changes the term "share" to "move to organization" and the icon is changed to [`fa-cube`](https://fontawesome.com/v4.7/icon/cube), which is used to indicate a collection. The act of moving an item to an organization is denoted by [`fa-arrow-circle-o-right`](https://fontawesome.com/v4.7/icon/arrow-circle-o-right) rather than `fa-share-alt`.

# Screen shots

#### Item View Options
image
image

#### Item Edit Options
image
image

#### Move To Organization View
image
image

#### Move To Organization Toast
image
image

#### Item hint part of Organization
image
image


src/App/Resources/AppResources.resx
```diff
@@ -1390,11 +1397,14 @@
   <data name="ShareItem" xml:space="preserve">
     <value>Share Item</value>
   </data>
+  <data name="MoveToOrganization" xml:space="preserve">
+    <value>Move to Organization</value>
```
mportune-bw, line 1401: This title is really snug in a few of those iOS screenshots, and will probably exceed the available width in some translated versions.  That being said, I can't come up with a shorter way to convey 'move to organization' unless we start using 'Org' everywhere which isn't great.  Any thoughts?

mportune-bw: image

MGibson1: The other verb I pitched was `Transfer`. This might be a good reason to use it.

That being said. These images were taken using an ipod touch. Even on the 8, there's quite a bit more room:
image

mportune-bw (approved):
Cool, let's roll with it

PR #1439, opened 2021-06-24
Add help link to password reprompt

Hinton:
## Objective
Add a help link to the password reprompt option.

## Screenshot
image

### DIFF
--- a/src/App/Pages/Vault/AddEditPage.xaml
+++ b/src/App/Pages/Vault/AddEditPage.xaml
@@ -558,7 +558,14 @@
                 <StackLayout StyleClass="box-row, box-row-switch">
                     <Label
                         Text="{u:I18n PasswordPrompt}"
-                        StyleClass="box-label-regular"
+                        StyleClass="box-label-regular" />
+                    <controls:FaButton 
+                        StyleClass="box-row-button, box-row-button-platform"
+                        Text=""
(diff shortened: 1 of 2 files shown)


src/App/Pages/Vault/AddEditPage.xaml
```diff
@@ -558,7 +558,14 @@
                 <StackLayout StyleClass="box-row, box-row-switch">
                     <Label
                         Text="{u:I18n PasswordPrompt}"
-                        StyleClass="box-label-regular"
+                        StyleClass="box-label-regular" />
+                    <controls:FaButton 
+                        StyleClass="box-row-button, box-row-button-platform"
+                        Text=""
```
mportune-bw, line 564: Please use the unicode value here instead of the actual char (for example: `&#xf24d`)

Hinton: Didn't know that was possible, but definitely preferable over the unicode character.

mportune-bw: Is **[this](https://fontawesome.com/v4.7/icon/question-circle-o)** the correct char?  If so we should use `&#xf29c`

Hinton: Changed to hex, not sure where I got the decimal encoded unciode character from.

mportune-bw (approved):

PR #1441, opened 2021-06-25
MaximumAccessCountInfo locale string update

choutkamartin:
Hey, I noticed that the string doesn't have a capitalized "Send" word.

It a bit messes up with the Crowdin comparing system, as both browser and desktop versions have "Send" capitalized correctly - while mobile and web don't.

I also made a PR on the web version.


eliykat:
Thanks for this, we'll merge it after the next release.

cscharf (approved):

PR #1442, opened 2021-06-28
Fix for vault timeout locking issue on android

mportune-bw:
This PR closes a gap where the Android app would not lock after the vault timeout time elapsed.  It has some history, starting with a HackerOne report in December, where we introduced use of the system monotonic clock for tracking time and performing the time check when the app resumed from background.  This replaced a timer that started when the app first entered the background, but occasionally never fired if the system terminated the process. (PR #1175).

Shortly after that went live we started receiving reports of Android failing to lock at all (but just Android - iOS was fine).  At the time I was unable to figure out why, so we restored the lock timers for Android only (PR #1220).

This brings us to today, where a recent bug full of red herrings led me to discover why Android was (seemingly) randomly failing to lock after the first PR.  The culprit is the Android back button/gesture and how it relates to Xamarin Forms app lifecycles.  I incorrectly assumed Xamarin's `OnResume()` would always fire when the app was brought back to the foreground, regardless of how it got to the background (native Android does this).  Turns out this only works when the app is backgrounded via the home button/gesture.  When the back button/gesture is used, the next time the app is brought to the foreground it will call Xamarin's `OnStart()` method, which until this PR was _not_ checking the vault lock timer.  That hole is now plugged, fixing the reported bug as well as the original issue from December and (finally) negating the need for the timer (which has been removed again).

**TL;DR:**  Vault timeout clock check additionally needs to be performed in Xamarin `OnStart()` for Android to catch multiple backgrounding/foregrounding flows.


cscharf (commented):
Wow, nice find! Glad you figured this out. The only concern I have is this arithmatic change that was reverted in [DeviceActionService.cs](https://github.com/bitwarden/mobile/pull/1220/files#diff-baae9aac90018239aa3d5e2d0404727c3c15613ef10fb5f917bb01281bc5fa5bL761) that was not re-done here, not sure if it's necessary or the context however.

mportune-bw:
That's correct, in #1220 we normalized on returning milliseconds from `GetActiveTime()` so the time check could be performed using multiplication instead of division.  That change was left in place in this PR (calculation takes place **[here](https://github.com/bitwarden/mobile/blob/7c1cb27d5b39adeb8eae42ae67925da1cbac0982/src/Core/Services/VaultTimeoutService.cs#L98)**).

cscharf (approved):

PR #1443, opened 2021-06-29
Temporarily disable send tests

mportune-bw:
Temporarily disable Send test until the race condition can be found, or Github Actions is capable of re-running jobs on a per-platform basis.  This is becoming problematic when the test fails on a single platform when merging to master and re-publishing the successful build only results in more errors.


test/Core.Test/Models/Domain/SendTests.cs
```diff
@@ -37,6 +37,9 @@ public void Send_FromSendData_Success(SendData data)
         [InlineCustomAutoData(new[] { typeof(AutoNSubstituteCustomization), typeof(FileSendCustomization) })]
         public async void DecryptAsync_Success(ICryptoService cryptoService, Send send)
         {
+            // TODO restore this once race condition is fixed or GHA can re-run jobs on individual platforms
```
cscharf, line 40: Isn't there an attribute we can use vs. the `return;` statement? @MGibson1 ?

cscharf (approved):

PR #1447, opened 2021-06-30
Workaround for off-screen draw bug in XF4.5+

mportune-bw:
An issue where a BindableLayout that is modified off-screen isn't re-drawn to reflect the modification (Xamarin Forms 4.5+) broke the way we render organization & collection selection when adding new vault items.  This PR works around that in an _interesting_ way.

- `HasCollections` is set to `true` in order to make the layout visible before modifying the content
- After `Collections` is updated with the new content, it is set to a new instance of `ExtendedObservableCollection` containing the values already established in `ResetWithRange` (which must still occur or this has no effect)
- Upon scrolling the screen, the collections `RepeaterView.ItemsSource` is repeatedly set to `Collections` (limited by a timer to prevent scroll events from max'ing out the thread pool)

Things to consider:
- All of these seemingly useless steps are required to force collections to draw like they did before we updated to Forms 5.x. A lot of whittling took place to narrow this down.
- Android exhibits this bug far more consistently than iOS, but I didn't restrict the workaround since I witnessed it happen a few times in iOS.
- When the owner is changed to an organization and you begin to scroll, you may experience an initial lag in the scroll action as the workaround does its "magic".

Note to my future self: Please for the love of ham, revert this when that bug is fixed

### DIFF
--- a/src/App/Pages/Vault/AddEditPageViewModel.cs
+++ b/src/App/Pages/Vault/AddEditPageViewModel.cs
@@ -1,4 +1,5 @@
-﻿using Bit.App.Abstractions;
+﻿using System;
+using Bit.App.Abstractions;
 using Bit.App.Models;
 using Bit.App.Resources;
 using Bit.Core.Abstractions;
@@ -9,6 +10,7 @@ using Bit.Core.Utilities;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Bit.App.Controls;
 using Xamarin.Forms;
 using View = Xamarin.Forms.View;
 
@@ -39,6 +41,7 @@ namespace Bit.App.Pages
         private int _ownershipSelectedIndex;
         private bool _hasCollections;
         private string _previousCipherId;
+        private DateTime _lastHandledScrollTime;
         private List<Core.Models.View.CollectionView> _writeableCollections;
         private string[] _additionalCipherProperties = new string[]
         {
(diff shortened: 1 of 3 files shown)


cscharf:
> Note to my future self: Please for the love of ham, revert this when that bug is fixed

@mportune-bw , please create a tracking item in Asana for this assigned to yourself with a due date for say, 15 days, hopefully someone on the Xamarin team fixes this soon, but perhaps a reminder on the calendar every 15 days to follow up and a task in Asana to keep track of it would be worthwhile.

src/App/Pages/Vault/AddEditPageViewModel.cs
```diff
@@ -798,15 +802,32 @@ private void OrganizationChanged()
             }
             if (Cipher.OrganizationId != null)
             {
+                HasCollections = true;
```
cscharf, line 805: So this is forced `true` instead of using `cols.Any()` due to the drawing issue?

cscharf (commented):
I _think_ this is okay, just one quick question

src/App/Pages/Vault/AddEditPageViewModel.cs
mportune-bw: Yes, though you're correct I should be using `cols.Any()` in case the org has no collections.  I just need to do it here (before `cols` is applied to the `RepeaterView` to make sure the view is visible before modifying data).  Good catch, stand by for update...

cscharf (approved):

PR #1476, opened 2021-07-23
Feature/use hcaptcha if bot

MGibson1:
# Overview

Adding captcha responses to Login and Registration failures due to suspected bot requests. The basic flow is to attempt the request, fail with a captcha siteKey in the response, use that site key to show a web auth and retrieve a captcha token, and finally reattempt the request with the acquired token.

I couldn't find a good example for DI to a parent ViewModel in the solution already. I chose abstract properties to pass required services to the parent `CaptchaProtectedViewModel` because I still want to have the dependencies gathered at the highest level and passed down. It didn't make sense to me to grab the dependencies multiple times, but this would be my second favorite approach.

# Files Changed
* **loginPage.xaml/registerPage.xaml**: On my simulator I was getting long load times for the second request attempt post-captcha that was displaying the login or register page again. I'm disabling the submit buttons until the second attempt is completed.
* **LoginPageViewModel**: Inherits from CaptchaProtectedViewModel to get all the captcha handling goodies. Add loading binding to disable submit and handle captcha required by reattempting once the base class set `_captchaToken`.
* **RegisterPageViewModel**: Same as **`LoginPageViewModel`**, except the captcha requirement comes in as an ApiException. This is due to the differences in failure modes server-side. Indentity failures are just failed auth requests, whereas API failures are returned as Bad Requests and thrown.
* **CaptchaProtectedViewModel**: Initialize a `WebAuthenticator` when captcha is required and retrieve a token response on success. If we don't get a token, signal failure. The `EncodeB64` method matches the... odd... way we're encoding iframe data in the web vault (https://github.com/bitwarden/jslib/blob/master/common/src/misc/iframe_component.ts#L30-L34 && https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings)
* **AppResources**: Localize Captcha strings
* **ApiService**: Parse out IdentityCaptchaResponse as well as Token and Two Factor. This is primarily handled in the new return class.
* **AuthService**: Add Captcha to login. A two factor captcha bypass token is stored for subsequent calls

# Screenshots
## Login PreCaptcha
image

## Login/register captcha 
image

## Login/register captcha with images (pixel 3a)
image

## Login back out of captcha
image

## Register precaptcha
image

## Register back out of captcha
image


mportune-bw (approved):
This looks awesome!

mportune-bw (approved):

PR #1520, opened 2021-09-01
[SSO] Auto enroll during set password

vincentsalucci:
## Code Changes
- **SetPasswordPage.xaml**: Add new banner explaining auto enrollment
- **SetPasswordPageViewModel.cs**: Add new properties/methods/conditionals for handling auto enrollment once a successful password is set
- **AppResources.resx**: Added banner string
- **ApiService**: Added two new API methods to be used during the auto enrollment flow
- **PolicyService**: Added new getter method retrieving reset password policy options
- **UserService**: Added new method to retrieve an organization by `identifier`
- **PolicyType**: Added `ResetPassword` enumeration
- **OrganizationData**: Added `Identifier` property
- **Organization**: Added `Identifier` property
- **ResetPasswordPolicyOptions**: Created object
- **OrganizationUserResetPasswordEnrollmentRequest**: Created object
- **OrganizationKeysResponse**: Created object
- **ProfileOrganizationResponse**: Created object

### DIFF
--- a/src/App/Pages/Accounts/SetPasswordPageViewModel.cs
+++ b/src/App/Pages/Accounts/SetPasswordPageViewModel.cs
@@ -30,6 +30,7 @@ namespace Bit.App.Pages
 
         private bool _showPassword;
         private bool _isPolicyInEffect;
+        private bool _resetPasswordAutoEnroll;
         private string _policySummary;
         private MasterPasswordPolicyOptions _policy;
 
@@ -50,7 +51,7 @@ namespace Bit.App.Pages
             ToggleConfirmPasswordCommand = new Command(ToggleConfirmPassword);
             SubmitCommand = new Command(async () => await SubmitAsync());
         }
-
+        
         public bool ShowPassword
         {
             get => _showPassword;
@@ -62,6 +63,12 @@ namespace Bit.App.Pages
         {
             get => _isPolicyInEffect;
             set => SetProperty(ref _isPolicyInEffect, value);
+        }
+        
+        public bool ResetPasswordAutoEnroll
+        {
+            get => _resetPasswordAutoEnroll;
+            set => SetProperty(ref _resetPasswordAutoEnroll, value);
         }
 
         public string PolicySummary
@@ -86,10 +93,17 @@ namespace Bit.App.Pages
         public Action SetPasswordSuccessAction { get; set; }
         public Action CloseAction { get; set; }
         public string OrgIdentifier { get; set; }
+        public string OrgId { get; set; }
 
         public async Task InitAsync()
         {
             await CheckPasswordPolicy();
+            
+            var org = await _userService.GetOrganizationByIdentifierAsync(OrgIdentifier);
+            OrgId = org?.Id;
+            var policyList = await _policyService.GetAll(PolicyType.ResetPassword);
+            var policyResult = _policyService.GetResetPasswordPolicyOptions(policyList, OrgId);
+            ResetPasswordAutoEnroll = policyResult.Item2 && policyResult.Item1.AutoEnrollEnabled;
         }
 
         public async Task SubmitAsync()
@@ -171,6 +185,7 @@ namespace Bit.App.Pages
             try
             {
                 await _deviceActionService.ShowLoadingAsync(AppResources.CreatingAccount);
+                // Set Password and relevant information
                 await _apiService.SetPasswordAsync(request);
                 await _userService.SetInformationAsync(await _userService.GetUserIdAsync(),
                     await _userService.GetEmailAsync(), kdf, kdfIterations);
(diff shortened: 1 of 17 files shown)


cscharf:
@vincentsalucci , do you have screen shots to submit with the review?

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
```diff
@@ -50,7 +51,7 @@ public SetPasswordPageViewModel()
             ToggleConfirmPasswordCommand = new Command(ToggleConfirmPassword);
             SubmitCommand = new Command(async () => await SubmitAsync());
         }
-
+
```
cscharf, line 54: Please remove extra whitespace.

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
```diff
@@ -86,10 +93,17 @@ public MasterPasswordPolicyOptions Policy
...
+        public string OrgId { get; set; }
 
         public async Task InitAsync()
         {
             await CheckPasswordPolicy();
+            
+            var org = await _userService.GetOrganizationByIdentifierAsync(OrgIdentifier);
+            OrgId = org?.Id;
+            var policyList = await _policyService.GetAll(PolicyType.ResetPassword);
+            var policyResult = _policyService.GetResetPasswordPolicyOptions(policyList, OrgId);
```
cscharf, line 105: What happens if `OrgId` is `null`?

src/Core/Services/UserService.cs
```diff
@@ -167,6 +167,19 @@ public async Task<Organization> GetOrganizationAsync(string id)
...
+        {
+            var userId = await GetUserIdAsync();
+            var organizations = await GetAllOrganizationAsync();
+
+            if (organizations == null || organizations.Count == 0)
+            {
+                return null;
+            }
+
+            return organizations.Find(o => o.Identifier == identifier);
```
cscharf, line 181: Would `.FirstOrDefault()` be more appropriate here?

src/Core/Services/PolicyService.cs
```diff
@@ -180,6 +181,23 @@ public async Task Clear(string userId)
...
+            if (policies == null || orgId == null)
+            {
+                return new Tuple<ResetPasswordPolicyOptions, bool>(resetPasswordPolicyOptions, false);
+            }
+
+            var policy = policies.FirstOrDefault(p =>
+                p.OrganizationId == orgId && p.Type == PolicyType.ResetPassword && p.Enabled);
+            resetPasswordPolicyOptions.AutoEnrollEnabled = GetPolicyBool(policy, "autoEnrollEnabled") ?? false;
+
+            return new Tuple<ResetPasswordPolicyOptions, bool>(resetPasswordPolicyOptions, policy?.Enabled ?? false);
```
cscharf, line 198:
```suggestion
            return new Tuple<ResetPasswordPolicyOptions, bool>(resetPasswordPolicyOptions, policy != null);
```

The policy is `null` if not `Enabled`, so this check is redundant or at the least a little confusing. I would recommend just checking if the `policy` is `null` since the filter on `Enabled` was done when fetching the policy originally.

src/Core/Services/ApiService.cs
```diff
@@ -177,7 +177,7 @@ public Task PostAccountVerifyPasswordAsync(PasswordVerificationRequest request)
             return SendAsync<PasswordVerificationRequest, object>(HttpMethod.Post, "/accounts/verify-password", request,
                 true, false);
         }
-
+
```
cscharf, line 180: Looks like perhaps your IDE is added these extra whitespace perhaps?

cscharf (commented):
Hi @vincentsalucci , just a few quick questions/feedback items.

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
vincentsalucci: The `policyResult` will return default values, which will resolve the `ResetPasswordAutoEnroll` property to `false`

src/Core/Services/UserService.cs
vincentsalucci: Yes, indeed. TS -> C# failed syntax translations :D

src/Core/Services/ApiService.cs
vincentsalucci: I don't believe so. Every other region has the same spacing except this one, so just added it to be uniform with the others.

vincentsalucci:
> @vincentsalucci , do you have screen shots to submit with the review?

Will gather and post!

src/Core/Services/PolicyService.cs
vincentsalucci: Slightly redundant, but very explicit! Hah. Regardless, good suggestion. Thanks!

cscharf (approved):

mportune-bw (approved):

PR #1533, opened 2021-09-08
Move policy checks inside PolicyService

eliykat:
## Objective

Mobile implementation of https://github.com/bitwarden/jslib/pull/466.

## Code changes

* `PolicyService` - add `PolicyAppliesToUser`. Note that this follows the same pattern as `jslib` after I tweaked it.
* `Organization` - add `isExemptFromPolicies` getter
* all other files - use `PolicyAppliesToUser` for all policy checks

Note that mobile does not have any concept of providers, so providers are not considered "owners" of orgs such that they are exempt from policies. If providers are not using the mobile app then this shouldn't matter.

## Code freeze

A reminder to my future self.


eliykat:
Code freeze has lifted and this is ready for review.

mportune-bw (approved):

PR #1534, opened 2021-09-08
friendly message for webauthn errors

mportune-bw:
_Edited with final screenshots_

Prepend WebAuthn error messages with a friendly/helpful message to hopefully cut down on support requests.

Key points conveyed:

1. User's browser must support WebAuthn, and 
2. It has to be set as the default browser (per WebAuthenticator requirements) 

New message:

device-2021-09-08-183325

Old message:

before2

### DIFF
--- a/src/App/Pages/Accounts/TwoFactorPageViewModel.cs
+++ b/src/App/Pages/Accounts/TwoFactorPageViewModel.cs
@@ -240,15 +240,8 @@ namespace Bit.App.Pages
             else
             {
                 await _deviceActionService.HideLoadingAsync();
-                if (authResult != null && authResult.Properties.TryGetValue("error", out var resultError))
-                {
-                    await _platformUtilsService.ShowDialogAsync(resultError, AppResources.AnErrorHasOccurred);
-                }
-                else
-                {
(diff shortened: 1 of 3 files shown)


Hinton:
I believe there's some value in keeping the exact error message. There are a quite a few possible error messages which could help the user debug the error. Like `SecurityError` for insecure contexts (http not https).

mportune-bw:
> I believe there's some value in keeping the exact error message. There are a quite a few possible error messages which could help the user debug the error. Like `SecurityError` for insecure contexts (http not https).

Perhaps a combination of the two?

device-2021-09-08-182228

Hinton:
Looks good. I would change it from `WebAuthn Error: ` to `Additional information`, or remove it all together.

mportune-bw:
> Looks good. I would change it from `WebAuthn Error: ` to `Additional information`, or remove it all together. It also feels a bit duplicate with `An error has occurred.` `Something Went Wrong`

I believe that's part of the `error` returned by the connector.

mportune-bw:
device-2021-09-08-183325

Hinton (approved):

Hinton:
We should look into removing the `WebAuthn error:` prefix from the connectors.

PR #1541, opened 2021-09-15
Icon Updates

Hinton:
## Objective
Updates the Mobile Application with the new icons. This was made fairly difficult due to us the unicode representative of the icons throughout the whole application. To avoid this issue in the future, and also to provide clarity in which icons are used where, I created a `BitwardenIcons` class which holds the unicode mappings.

https://app.asana.com/0/1200804338582616/1200487492949147


Hinton:
@mportune-bw Should I rename `FaLabel` and `FaButton` now as well? Perhaps to `IconLabel` and `IconButton`.

src/Core/BitwardenIcons.cs
```diff
@@ -0,0 +1,118 @@
...
+        // TODO: Missing icons
+        public const string File = FilePdf;
+        public const string FolderOpenO = Frown;
+        public const string ChevronUp = Frown;
+        public const string Paste = Frown;
+        public const string Ban = Frown;
+        public const string Camera = Frown;
+        // Eye-slash not aligned with eye
+
+        public const string User = "";
```
mportune-bw, line 14: We've been transitioning towards using the unicode values for these instead of the actual chars since they aren't readable everywhere (like here for example).

mportune-bw:
> @mportune-bw Should I rename `FaLabel` and `FaButton` now as well? Perhaps to `IconLabel` and `IconButton`.

Yeah I think that makes sense.

mportune-bw:
Some of the white pngs for iOS were replaced with black versions (`info.png` for example).  Is there a plan to handle this programmatically?

Hinton:
I completely missed that some icons were white, curiously they still work both in normal and dark mode. Do we want to change them back to white?

mportune-bw:
> I completely missed that some icons were white, curiously they still work both in normal and dark mode. Do we want to change them back to white?

What does 'still work' mean here?  Are they showing up as white?

Hinton:
They show up as white.

src/iOS/iOS.csproj
```diff
@@ -153,6 +152,19 @@
     <ImageAsset Include="Resources\Assets.xcassets\LaunchScreen.imageset\logo_white%403x.png" />
     <ImageAsset Include="Resources\Assets.xcassets\ic_warning.imageset\Contents.json" />
     <ImageAsset Include="Resources\Assets.xcassets\ic_warning.imageset\ic_warning-1.pdf" />
+    <BundleResource Include="Resources\bwi-font.ttf" />
```
mportune-bw, line 155: Does this change in organization matter? (vs. the previous individual `ItemGroup`)

@vvolkgang @jlf0dev @fedemkr

jlf0dev: I'm pretty sure we're good. I also found this in the [documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/itemgroup-element-msbuild?view=vs-2022):

> In a simple project file, you normally use a single ItemGroup element, but you can also use multiple ItemGroup elements. When multiple ItemGroup elements are used, items are combined into a single ItemGroup. For example, some items might be included by a separate ItemGroup element that's defined in an imported file.

mportune-bw (approved):

src/iOS/iOS.csproj
vvolkgang: MSBuild wise the end result will be the same. 

@Hinton which IDE did you use? I would suggest opening the .csproj in VS4Mac and Rider just to if this changes between those two. VS4Mac formats .csproj's differently than VS4win for instance, which leads to always having that pending change. We should follow the format used by our most common IDE.

fedemkr:
We should check `LockPasswordViewController -> ViewDidLoad` because it's expecting `FontAwesome` as the font family. I think we should change it, right? @mportune-bw @Hinton 
image

mportune-bw:
Good catch @fedemkr , @Hinton can you fix?

Hinton:
@mportune-bw I'll look into it.

src/iOS/iOS.csproj
Hinton: Visual Studio.

PR #1544, opened 2021-09-22
Update XF to remove AndroidX dependency hack

mportune-bw:
The SR5 release of Xamarin Forms (5.0.0.2125) fixed the broken AndroidX & Material dependencies of the previous release, allowing us to remove the extra dependencies added in #1471 and update others to the latest without breaking the build.  This will close [this Asana task](https://app.asana.com/0/1169444489336079/1200632790235501/f).

(Note this does not yet fix [this Asana task](https://app.asana.com/0/1169444489336079/1200531933783245/f) which we worked around in #1447 and #1454)

Also updated LiteDB, AppCenter, and .NET Test SDK to the latest.

### DIFF
--- a/src/Android/Renderers/CustomTabbedRenderer.cs
+++ b/src/Android/Renderers/CustomTabbedRenderer.cs
@@ -49,7 +49,7 @@ namespace Bit.Droid.Renderers
             return null;
         }
 
-        async void BottomNavigationView.IOnNavigationItemReselectedListener.OnNavigationItemReselected(IMenuItem item)
+        public async void OnNavigationItemReselected(IMenuItem item)
         {
             if (_page?.CurrentPage?.Navigation != null && _page.CurrentPage.Navigation.NavigationStack.Count > 0)
             {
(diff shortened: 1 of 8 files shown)


src/Android/Renderers/CustomTabbedRenderer.cs
```diff
@@ -49,7 +49,7 @@ private BottomNavigationView GetBottomNavigationView()
             return null;
         }
 
-        async void BottomNavigationView.IOnNavigationItemReselectedListener.OnNavigationItemReselected(IMenuItem item)
+        public async void OnNavigationItemReselected(IMenuItem item)
```
cscharf, line 52: Should this be `async void` or `async Task`? _Feels_ like we need `async Task` to avoid latent exceptions or app crashes that aren't handled through this async call.

cscharf (commented):
Just one small item.

src/Android/Renderers/CustomTabbedRenderer.cs
mportune-bw: Looks like we don't have a choice...

image

cscharf: Bummer, I didn't realize it was an interface method implementation. With `async void` you'll need to ensure that there's proper error handling/try-catch blocks to ensure exceptions are bubbled up outside of latent execution where we'd lose the stack-trace, etc.

cscharf (approved):

PR #1558, opened 2021-10-04
Split collections permissions

MGibson1:
# Overview

Asana: https://app.asana.com/0/0/1201116867140797/f

Splits collections permissions. These permissions are currently unused, but keeps the models in line with bitwarden/jslib#504

### DIFF
--- a/src/Core/Models/Data/Permissions.cs
+++ b/src/Core/Models/Data/Permissions.cs
@@ -1,3 +1,5 @@
+using System;
+
 namespace Bit.Core.Models.Data 
 {
     public class Permissions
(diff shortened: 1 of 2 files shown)


src/Core/Models/Data/Permissions.cs
```diff
@@ -1,3 +1,5 @@
+using System;
```
mportune-bw, line 1: Necessary import?

MGibson1: No, it's vestigial from when I was obsoleting permissions. thanks I'll remove

mportune-bw (approved):

PR #1563, opened 2021-10-06
[Linked fields] Add Linked Field as a custom field type

eliykat:
## Objective

Mobile implementation of https://github.com/bitwarden/browser/pull/1963.

Short version:
* this adds a new custom field type called "Linked". It "links" the value of a custom field to a default field, so that (in browser) you can effectively autofill a default field using a custom selector.
* mobile does not autofill custom fields, so the changes here are just to allow the configuration of custom linked fields.

## Code changes

These changes follow the same patterns as Angular clients unless noted otherwise.

* **src/Core/Models/View/ItemView.cs**: new base class for all item types.
* **src/Core/Enums/LinkedIdType.cs**: enum containing the saved values for linked fields (in browser, these are used to associate it with the property it's linked to). In Typescript I could use a separate enum for each cipher type to avoid clashing enum keys, but I couldn't figure out how to make that play nice with C#'s type system. Instead it's just one big enum with prefixes on each key.
* **LoginView, CardView, IdentityView, SecureNoteView**: add `LinkedFieldOptions`, which is just a getter containing the LinkedIds and i18n keys for that cipher type. Secure Notes don't support linked fields so that just returns null, which we test for where required.
  * This is a simpler approach than the Angular clients because we don't have to care about autofilling custom fields
* **Field models (various)**: add `LinkedId` property
* **AddEditPage**: add the new type when adding a new custom field. It uses a wheel selector to allow you to select from the `LinkedFieldOptions`.
* **ViewPage**: if it's a linked field, display a little "link" symbol next to its value to indicate that it's a pointer and not a literal value

## Screenshots

Screen Shot 2021-10-06 at 4 35 32 pm

Screen Shot 2021-10-06 at 4 35 46 pm

Screen Shot 2021-10-06 at 4 35 50 pm


src/App/Pages/Vault/ViewPageViewModel.cs
```diff
@@ -741,16 +743,38 @@ public bool ShowHiddenValue
...
+            get
+            {
+                if (IsBooleanType)
+                {
+                    return _field.Value == "true" ? "" : "";
+                }
+                else if (IsLinkedType)
+                {
+                    var i18nKey = _cipher.LinkedFieldI18nKey(Field.LinkedId.GetValueOrDefault());
+                    return " " + _i18nService.T(i18nKey);
```
jlf0dev, line 757: Special chars should be formatted as unicode values (ex: "\uf24a")

mportune-bw: We can probably skip this for now (I forgot @Hinton already has a full conversion (to unicode values in a constants class) in motion with the icon updates.

mportune-bw (approved):
I'm good with this unless @jlf0dev finds anything else

src/App/Pages/Vault/ViewPageViewModel.cs
eliykat: I'll follow @mportune-bw 's advice and leave it for now

PR #1572, opened 2021-10-11
Website favicon only shows if hostname is first element

### LINKED ISSUE #1120: Website favicon only shows if hostname is first element
## Describe the Bug

Website Icon (favicon) only shows if website url (hostname) is in the first url element.
If the first url is an app link (eg. androidapp:// ) and the website url as second element, the favicon will not show.

## Steps To Reproduce

1. Go to 'My Vault'
2. Click on 'Add Item'
3. Select type 'Login'
4. Insert random name, username, password
5. In URIs section add first element as 'androidapp://example.com'
6. In URIs section add New URI second element with website hotname 'http://github.com'
7. Save

## Expected Result

Credentials icon from website favicon.

## Actual Result

Default item icon without customization.

## Environment

- Device: Xiaomi
- Operating system: Android 9

jlf0dev:
# Overview
Website Icon (favicon) only shows if website url (hostname) is in the first url element.
If the first url is an app link (eg. androidapp:// ) and the website url as second element, the favicon will not show.

This fix iterates through the list of urls and looks for the first website url for the icon. If no website is found, it defaults to the android, apple, or globe icon.

Closes #1120

# Changes

- **IconImageConverter.cs** - Create static class for uri logic (keeping converter for future)
- **CipherViewCellViewModel.cs** - Create new property in viewmodel for icon source, making sure to cache result
    - Check if icons are enabled and source for icon before showing default glyph
- **CipherViewCell.xaml** - Bind source to new property in viewmodel

### DIFF
--- a/src/App/Controls/CipherViewCell/CipherViewCell.xaml
+++ b/src/App/Controls/CipherViewCell/CipherViewCell.xaml
@@ -46,7 +46,7 @@
         WidthRequest="22"
         HeightRequest="22"
         IsVisible="{Binding ShowIconImage}"
-        Source="{Binding Cipher, Converter={StaticResource iconImageConverter}}"
+        Source="{Binding IconImageSource, Mode=OneTime}"
         AutomationProperties.IsInAccessibleTree="False" />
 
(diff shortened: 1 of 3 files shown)


adripo:
Thank you!

PR #1588, opened 2021-10-15
fix gap in background color application resulting in flashing during transitions

mpbw2:
There was a gap in the application of background color (Android only) when activities are first created, resulting in unpleasant transitions between screens when selecting non-item things (types, folders, collections).  In addition due to the way our splash screen is implemented you could also see our logo during the gap, making the experience even less pleasant.  This was previously covered by the app restarting on theme change, which is no longer the case.


mportune-bw:
Disregard draft state - thought there was an issue with Android 11 but it was just my emulator image

vachan-maker:
@mportune-bw  I got the update. But the startup logo still appears when I press the search icon.

PR #1602, opened 2021-10-22
Fix accessibility tile bugs

jlf0dev:
- AccessibilityActivity.cs: Tile not working when Bitwarden is open in background
- AccessibilityService.cs: Tile not working when Autofill is enabled and suggestions are showing

### DIFF
--- a/src/Android/Accessibility/AccessibilityActivity.cs
+++ b/src/Android/Accessibility/AccessibilityActivity.cs
@@ -18,6 +18,7 @@ namespace Bit.Droid.Accessibility
         protected override void OnCreate(Bundle bundle)
         {
             base.OnCreate(bundle);
+            MoveTaskToBack(true);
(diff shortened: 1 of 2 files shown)


mportune-bw (approved):

fedemkr:
@jlf0dev  @mp-bw should this be still on hold?

jlf0dev:
These fixes were intended for inconsistencies I encountered a while ago using the Accessibility Tile. I haven't been able to reproduce these in a while, and haven't heard of any reports, so I'm hesitant to check in these changes.

fedemkr:
oh ok, @vvolkgang shall we ask QA if they can reproduce this on other devices? So that we can either merge/close it

vvolkgang:
@fedemkr Let's close for now and re-evaluate if issues come up in the future. Thanks both for following up!

PR #1625, opened 2021-11-02
[SupportedBrowsers] Add Captive Login and drop Alook

stevenlele:
## Captive Login

Some android systems show captive login activity (`com.google.android.captiveportallogin`) when connecting to wireless network. This activity shows URL in its ActionBar subtitle, so there is no view ID.

See: https://cs.android.com/android/platform/superproject/+/master:packages/modules/CaptivePortalLogin/src/com/android/captiveportallogin/CaptivePortalLoginActivity.java;l=600

## Alook

This browser was added in https://github.com/bitwarden/mobile/pull/1281, but the developer seems to be unwilling to add auto-fill compatibility. Auto-fill was partially functioning before adding the support, but afterwards it was completely broken (or it might be subject to browser version). Dropping the support might be a good choice.

### DIFF
--- a/src/Android/Accessibility/AccessibilityHelpers.cs
+++ b/src/Android/Accessibility/AccessibilityHelpers.cs
@@ -31,7 +31,6 @@ namespace Bit.Droid.Accessibility
             // So keep them in sync with:
             //   - AutofillHelpers.{TrustedBrowsers,CompatBrowsers}
             //   - Resources/xml/autofillservice.xml
-            new Browser("alook.browser", "search_fragment_input_view"),
             new Browser("com.amazon.cloud9", "url"),
             new Browser("com.android.browser", "url"),
             new Browser("com.android.chrome", "url_bar"),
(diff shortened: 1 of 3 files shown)


mportune-bw (approved):

LiuDeng:
@stevenlele Hello, we are Alook browser developers, we will adapt bitwarden's autofill in the latest 5.6 version, can you help add us to the list? 
We used to use the old Chromium kernel that cannot adapt to autofill. Recently, we upgraded the browser kernel to Chromium 96, and will complete the adaptation of autofill in the new version.
Now the jumping past shows the url of androidapp://, which cannot be filled by matching the URL. Our package names are **alook.browser** and **alook.browser.google**
[http://static.alookweb.com/WechatIMG126.jpeg](url)
Thank you very much.
Best Regards~

stevenlele:
@LiuDeng Sure. I will revert this commit.

stevenlele:
@LiuDeng it is already added in #1814.

Issue #385, opened 2018-09-29
ios app won’t add attachments other than photos 


suyrulola:
I’m trying to add a pdf attachment to a secure bitwarden note. The pdf file resides in the apple files app. 

1. create a secure note.
2. select “attachments@
3. select file
4. search
5. pick the pdf file (or any other)
6. now the filename should show but doesn’t. Nothing is attached.

Adding fotos from camera/ foto roll works however.

App is v1.19.0 (37), iphone with ios 12.

kspearrin:
Do you have a premium membership?

suyrulola:
Yes, premium is active.

cfis:
I also have a premium membership and have the same issue.
iOS 12, iPhone X.  I am trying to upload a PDF file saved to the built in ios File App.

ShirokaiLon:
Having a similar problem on the Android app. On Android 6.0.1.

Can add images through the Camera or Gallery apps just fine. But if I try to add a file (even an image) through the My Files/Documents view that first opens up, I'm taken back to the front page of the app and no file is attached.

vvolkgang:
Issue migrated to https://github.com/bitwarden/mobile/issues/385

PR #1626, opened 2021-11-02
Supress lock and logout when showing fileswitcher on Android

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If vault timeout was set to "Immediate", the file picker on Android was locking and logging out Bitwarden. This meant you could never choose a file.

This fix adds a `DelayLockAndLogoutMs` long to the VaultTimeoutService and sets it to 60 seconds while choosing an attachment on Android.

I also removed the extraneous logic for lock/logout from App.xaml.cs. I consolidated this logic in `CheckVaultTimeoutAsync()` inside the VaultTimeoutService.

## Code changes

* **IVaultTimeoutService.cs:** Add DelayLockAndLogoutMs long to TimeoutService
* **VaultTimeoutService.cs:** Add logic to check for DelayLockAndLogoutMs when checking VaultTimeout settings
* **AttachmentsPageViewModel.cs:** If Android, set DelayLockAndLogoutMs to 60s before showing filepicker
* **App.xaml.cs:** Reset DelayLockAndLogoutMs and move lock/logout logic to VaultTimeoutService

## Testing requirements

- User can pick an attachment while the VaultTimeout option is set to "Immediate"
- Functionality of VaultTimeout still works

## Before you submit


djsmith85:
Possibly also related to #385

eliykat (commented):
I don't know enough about mobile to do the code review, but my question about this approach is whether it could accidentally leave `SuppressLockLogout` set to `true`. e.g.
* what if the user locks the screen from the file picker?
* what if the user changes app from the file picker?
* any other code flows / user interactions that could evade setting it back to `false`?

jlf0dev:
These are good questions @eliykat and I'm not completely happy with how I fixed this. Let me take a second look and see if we can accomplish this any other way.

mportune-bw:
Good points brought up by @eliykat .  I don't think there's a clean way to handle points 1 & 2 without over-engineering something truly awful, but handling points 3 & 4 should probably cover it.  To expand on point 4, how about making `SuppressLockLogout` a timestamp (nullable `long`) instead of a bool, then compare to a hardcoded threshold upon returning to the app if it's not null?  You can use `_deviceActionService.GetActiveTime()` to get a monotonic timestamp in milliseconds, and give the user, let's say 45-60 seconds to select an attachment.

jlf0dev:
Changed my approach here to follow @mportune-bw's suggestion. I like this strategy better, should be a little more bulletproof.

mportune-bw (approved):

PR #1647, opened 2021-11-16
Add Share app Extension on iOS for Send

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

This PR introduces Share app extension on iOS, i.e. the ability to make a Send from other apps to Bitwarden, e.g. from photos select a picture and share it to Send.

## Code changes

Mainly a new project was added: **"iOS.ShareExtension"** .
Also there were some code improvements and try catches added on places where the app may crash.
* **AppOptions**, **SendAddEditPage.xaml**, **SendAddEditPageViewModel.cs**:
The "Share this Send upon save" becomes "Copy Send Link on save" on the iOS share extension due to some OS limitations to display the native share dialog on the extension after saving.
* **iOSCoreHelpers:** Fixed bug where if we send `details.ConfirmText` `null` then the behavior was not the intended one leading the app to crash on some places.

## Screenshots

You can see the Bitwarden icon to share on other apps now working as expected.
image

## Testing requirements

The main focus for this is to share text and files from other apps to Bitwarden.
So we should test sharing from the Photos app, from Files app (other than photos, like a movie or a PDF), text selection from any app like Safari.
Also, just in case test that only on the iOS sharing app extension the "Share this Send upon save" toggle changes to "Copy Send Link on save" and that the behavior works as expected.

## Before you submit


src/Core/Services/SendService.cs
```diff
@@ -219,12 +219,12 @@ public async Task<string> SaveWithServerAsync(Send send, EncByteArray encryptedF
                         {
                             response = await LegacyServerSendFileUpload(request, send, encryptedFileData);
                         }
-                        catch (Exception e) 
+                        catch (Exception genE)
```
cscharf, line 222: Would it make more sense to just `catch` with a general `throw`? Since you're not checking the exception, or exception type, or using the exception message, etc. That way the stack-trace is left unadulterated by being re-thrown explicitly.

catch (Exception genE) 
 

 
Would it make more sense to just `catch` with a general `throw`? Since you're not checking the exception, or exception type, or using the exception message, etc. That way the stack-trace is left unadulterated by being re-thrown explicitly.

```c#
                        catch
                        {
                            if (response != default){
                                await _apiService.DeleteSendAsync(response.Id);
                            }
                            throw;
                        }
```

src/iOS.ShareExtension/iOS.ShareExtension.csproj
```diff
@@ -0,0 +1,220 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\iOS.Core\iOS.Core.csproj">
+      <Name>iOS.Core</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\Core\Core.csproj">
+      <Name>Core</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <BundleResource Include="Resources\logo.png" />
```
cscharf, line 209: Shouldn't these bundle resources, which likely already exists elsewhere in the project, simply be linked vs. copies?

src/Core/Services/SendService.cs
fedemkr: yes, sure that's a good improvement; I'll make it happen

src/iOS.ShareExtension/iOS.ShareExtension.csproj
fedemkr: Oh, didn't think about that, I'll give it a try by referencing to the other extension project that holds all the same icons

cscharf:
@vgrassia , can you please review the changes to `.github/workflows/build.yml`?

cscharf (commented):
Looks good to me, will approve after Vince has a look at the build workflow.

vgrassia:
> @vgrassia , can you please review the changes to `.github/workflows/build.yml`?

The `build` workflow looks good, but the file `./.github/secrets/dist_share_extension.mobileprovision.gpg` is missing in the PR.

cscharf:
> > @vgrassia , can you please review the changes to `.github/workflows/build.yml`?
> 
> The `build` workflow looks good, but the file `./.github/secrets/dist_share_extension.mobileprovision.gpg` is missing in the PR.

@fedemkr , can you please resolve this?

joseph-flinn:
@cscharf I've added the encrypted version of the profile as I don't think Fede has access to those secrets. 

@fedemkr It looks like something is breaking in the CI pipeline. I'm not familiar enough with iOS building to be able to pinpoint the issue. Could you take a look?

fedemkr:
I'll push the fix in a minute

vgrassia:
@fedemkr Can you resolve the merge conflicts on `src/App/Resources/AppResources.resx`?

vgrassia (approved):

PR #1648, opened 2021-11-17
Show SSO login if using key-connector without bio or pin

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Use SSO if using key connector without bio or pin on autofill
Asana Task: https://app.asana.com/0/1201363553288890/1201363818962345/f
https://app.asana.com/0/1201363553288890/1201363818962343/f

## Code changes

* **src/iOS.Core/Controllers/LockPasswordViewController.cs:** Add call to SSO login page if using key connector

## Testing requirements

- Autofill authentication with key connector
- Extension authentication with key connector

## Before you submit

### DIFF
--- a/src/iOS.Core/Controllers/LockPasswordViewController.cs
+++ b/src/iOS.Core/Controllers/LockPasswordViewController.cs
@@ -11,6 +11,9 @@ using System.Threading.Tasks;
 using Bit.App.Utilities;
 using Bit.Core.Models.Domain;
 using Bit.Core.Enums;
+using Bit.App.Pages;
+using Bit.App.Models;
+using Xamarin.Forms;
 
 namespace Bit.iOS.Core.Controllers
 {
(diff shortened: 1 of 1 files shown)


src/iOS.Core/Controllers/LockPasswordViewController.cs
```diff
@@ -258,6 +265,7 @@ private async void DoContinue(bool masterPassword = false)
             {
                 await _storageService.SaveAsync(Bit.Core.Constants.PasswordVerifiedAutofillKey, true);
             }
+            await EnableBiometrics();
```
mportune-bw, line 268: Can you elaborate on this change?

jlf0dev: Good callout, I created a method to check if biometrics have been validated yet so we can use SSO to enable them instead of MP if they are using key connector.

Regarding 2ecbc4b2c155fbd927af5558fb677b1d34f37821, I accidentally included an extra call in `CheckPasswordAsync()` that I removed in this commit (and did some housekeeping).

I admit this method will now also be called whenever biometrics are successful (unnecessarily), but I thought it was a little cleaner.

mportune-bw (approved):

PR #1656, opened 2021-11-22
[Key Connector] Hide MP input in iOS extensions

eliykat:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Implements https://github.com/bitwarden/mobile/pull/1654 for the iOS extension & autofill projects:

> If a key connector user has biometrics set up on the lock screen, we want to hide the master password prompt. [...]
>
> If the user is using a PIN with biometrics, the screen should remain the same and prompt for PIN input.

## Code changes

* **src/iOS.Core/Controllers/LockPasswordViewController.cs**:
  * hide the Submit button if using biometric unlock only (this is done by setting text to `null` and disabling it)
  * change the title text to "Unlock Vault" if using biometric unlock only
  * skip setup for the `MasterPasswordCell` if using biometric unlock only
  * if using biometric unlock only, move the "biometric cell" up to the first row, and adjust the length of the grid accordingly. To assist this, I've extracted the 'builder' logic for the biometric cell into a getter.

Please pay particular attention to the conditional logic involved - there are quite a few different combinations, and I'd like simpler logic, but I think this works.

## Screenshots

Screen Shot 2021-11-22 at 2 04 39 pm

## Testing requirements

* confirm that the iOS unlock page conforms to the Figma and behaves as expected
* full regression on iOS extension ("share") and autofill unlock page

## Before you submit

### DIFF
--- a/src/iOS.Core/Controllers/LockPasswordViewController.cs
+++ b/src/iOS.Core/Controllers/LockPasswordViewController.cs
@@ -107,14 +107,14 @@ namespace Bit.iOS.Core.Controllers
                 _biometricIntegrityValid = _biometricService.ValidateIntegrityAsync(BiometricIntegrityKey).GetAwaiter()
                     .GetResult();
                 _usesKeyConnector = await _keyConnectorService.GetUsesKeyConnector();
-                _biometricUnlockOnly = _usesKeyConnector && !_pinLock && _biometricLock;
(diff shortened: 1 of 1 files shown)


jlf0dev (commented):
Nice work! Approved but @mportune-bw might want to take a look as well

mportune-bw (approved):

PR #1659, opened 2021-11-23
Support installing development and prod concurrently

Hinton:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Other

## Objective

Configures the debug build to use a different app bundle id than release, allowing you to install the development build alongside a regular instance of the application.

Only the `Debug` build is affected and all other builds will use the production app bundle id, which should ensure that the AppStore and Testflight builds behave identical as before.

## Code changes

I duplicated the `Info.plist` files to `Info-dev.plist` and modified the bundle id. And changed the projects to conditionally load the `Info.plist` files depending on which configuration is run.

## Before you submit


vvolkgang (approved):
LGTM but, supporting multiple plists can get tricky down the line. Has anyone raised using [CAKE](https://cakebuild.net)?

mportune-bw:
I'm hesitant to merge this as it means we'll be releasing a different binary than what QA is testing. While it _shouldn't_ make a difference, strange things happen.  I'd like to think the upcoming account switching feature will help with simultaneous personal use and testing.

Hinton:
Cake is outside my area of expertise @vvolkgang. I mostly wanted a quick way to run Bitwarden on my devices without having to replace my regular app.

fedemkr:
I think this would be super helpful when trying to compare things on dev against prod without [un]installing prod every time this is needed. Also for people that also uses the prod app in the normal life of their device and have to test something they are currently working on, they have to keep switching which it can be a waste of time.
If we're going to use CAKE, then here is a really useful extension for editing the plists [Cake.Plist](https://cakebuild.net/extensions/cake-plist/)
I can do it on a BEEEP if needed.
What I'd add is some sort of automatic verification check in case someone needs to modify some plist or config file so that they update cake as well if needed

Hinton:
Superseded by https://github.com/bitwarden/mobile/pull/2161

PR #1660, opened 2021-11-24
Add Share app Extension on iOS for Send (re-PR)

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

This PR introduces Share app extension on iOS, i.e. the ability to make a Send from other apps to Bitwarden, e.g. from photos select a picture and share it to Send.
This is a new PR because the previous one (#1647 ) I merged it before all reviews were done.

## Code changes

Mainly a new project was added: **"iOS.ShareExtension"** .
Also there were some code improvements and try catches added on places where the app may crash.
* **AppOptions**, **SendAddEditPage.xaml**, **SendAddEditPageViewModel.cs**:
The "Share this Send upon save" becomes "Copy Send Link on save" on the iOS share extension due to some OS limitations to display the native share dialog on the extension after saving.
* **iOSCoreHelpers:** Fixed bug where if we send `details.ConfirmText` `null` then the behavior was not the intended one leading the app to crash on some places.

## Screenshots

You can see the Bitwarden icon to share on other apps now working as expected.
image

## Testing requirements

The main focus for this is to share text and files from other apps to Bitwarden.
So we should test sharing from the Photos app, from Files app (other than photos, like a movie or a PDF), text selection from any app like Safari.
Also, just in case test that only on the iOS sharing app extension the "Share this Send upon save" toggle changes to "Copy Send Link on save" and that the behavior works as expected.

## Before you submit
- [x] This change requires a **documentation update** (notify the documentation team)


src/App/Pages/Send/SendAddEditPageViewModel.cs
```diff
@@ -96,6 +100,16 @@ public SendAddEditPageViewModel()
         public bool ShareOnSave { get; set; }
         public bool DisableHideEmailControl { get; set; }
         public bool IsAddFromShare { get; set; }
+        public bool CopyInsteadOfShareAfterSaving
```
mportune-bw, line 103: Move this down so it's grouped with the other properties with custom get/set logic like `SendEnabled`, etc.)

src/Core/Utilities/TaskExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        public static async void FireAndForget(this Task task, Action<Exception> onException = null)
+        {
+            try
+            {
+                await task.ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                onException?.Invoke(ex);
+                // TODO: Add app center exception handling in here, but we have to check whether we can add the AppCenter package to Core
```
mportune-bw, line 23: I think it's ok to add AppCenter to Core, just remember to do the usual exclusions for F-Droid.

mportune-bw (commented):
Looks good, though I wouldn't mind @jlf0dev taking a pass on the iOS-specific stuff since my extension experience is full of holes.

mportune-bw (dismissed):

jlf0dev (approved):
Sorry for the wait, my Github notifications were a mess last year.

I like that `FireAndForget` utility, but are you using it here? Maybe I'm missing it.

Other than that, looks good to me.

jlf0dev (approved):

vvolkgang (approved):

PR #1662, opened 2021-11-26
[EC-528] Refactor Custom Fields into separate components

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Refactor custom fields to separate the different types on components to reuse them on the different views.

## Code changes

* **ViewPage.xaml, AddEditPage.xaml:** Replaced `RepeaterView` with `StackLayout` with `BindableLayout` in order to use the feature provided by Xamarin.Forms.
* **RepeaterView.cs:** Marked as Obsolete because we should use a layout `BindableLayout` properties so that we take advantage of the power of Xamarin.Forms.
* **AppSetup.cs:** Added this setup to configure things specifics to the `App` project, like registering services (this is called from both platforms Android and iOS).
* **ViewPageViewModel, AddEditPageViewModel:** Replaced logic of the fields creation to use the new factory and the custom fields interface for the items. So that we have better control and scalability in case new field types are added.
* **IconGlyphExtensions.cs:** Added this file in order to not rely only on the Converter to have the proper glyph, also added some enum to use to get the correct glyph when there are more than one for a given type, e.g. boolean we have for the `eye` and the `checkbox`
* **Lists:** Added this folder on the `App` project in order to have a place to have list related things reused across the app. 

## Testing requirements

Regression test custom fields on edition and also when viewing them. The behavior should be the same as before

## Before you submit


mp-bw (approved):

i3130002:
@fedemkr 

Should someone update this? 
https://github.com/bitwarden/mobile/blob/90e0b5dcf0783603c825159fb7a74612a8bd7722/src/Android/MainApplication.cs#L53

fedemkr:
HI there, thanks for letting us now; there is already an internal issue to refactor this.

PR #1663, opened 2021-11-26
BEEEP: Abstract and Centralize Logging

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Added a Logger in order to centralize logging exceptions

## Code changes

* **ILogger, Logger.cs:** Added this file to centralize and abstract the logging logic
* **StubLogger.cs:** Added this file to use it on FDROID so that it doesn't use AppCenter
* **INativeLogService:** Renamed `ILogService` to `INativeLogService` to be more specific and differentiate it from the new one.
* **Other:** Implemented new `ILogger` where `Crashes.TrackError` was being used except on some specific cases where for now it's more convenient to keep using the old way just in case the `ILogger` couldn't be resolved because of some scenario

## Testing requirements

Maybe a quick test on delete account and export vault to check that the action is done correctly. Very little was changed there. And check that overall the app behaves and doesn't crash when opening some screen.

## Before you submit


fedemkr:
For now I created it really simple, but we could add extra data to be passed when reporting an exception, so that we fill the properties. Or make use of the crash file to send more details about the context on which the exception happened (like the last navigations of the user)

mportune-bw (approved):

PR #1667, opened 2021-11-29
only prompt for sso if using key connector

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Key Connector changes for autofill on iOS were showing the SSO login flow even when a user wasn't a SSO user. This was a logic error, we weren't checking for key connector use while seeing if the user was using biometric without being validated. 

Needs to be cherry picked to `rc`
Asana Task: https://app.asana.com/0/1201363553288890/1201432735007902/f

## Code changes

* **src/iOS.Core/Controllers/LockPasswordViewController.cs:** Add check for key connector before sso login flow to autofill

## Testing requirements

- iOS autofill with/without key connector & biometric authentication

## Before you submit

### DIFF
--- a/src/iOS.Core/Controllers/LockPasswordViewController.cs
+++ b/src/iOS.Core/Controllers/LockPasswordViewController.cs
@@ -185,9 +185,13 @@ namespace Bit.iOS.Core.Controllers
             base.ViewDidAppear(animated);
 
             // Users with key connector and without biometric or pin has no MP to unlock with
-            if (_usesKeyConnector && (!(_pinLock || _biometricLock)) || ( _biometricLock && !_biometricIntegrityValid))
+            if (_usesKeyConnector)
             {
-                PromptSSO();
+                if (!(_pinLock || _biometricLock) ||
+                    (_biometricLock && !_biometricIntegrityValid))
+                {
+                    PromptSSO();
+                }
             }
             else if (!_biometricLock || !_biometricIntegrityValid)
             {


mportune-bw (approved):

src/iOS.Core/Controllers/LockPasswordViewController.cs
```diff
@@ -185,9 +185,13 @@ public override async void ViewDidAppear(bool animated)
             base.ViewDidAppear(animated);
 
             // Users with key connector and without biometric or pin has no MP to unlock with
-            if (_usesKeyConnector && (!(_pinLock || _biometricLock)) || ( _biometricLock && !_biometricIntegrityValid))
+            if (_usesKeyConnector)
             {
-                PromptSSO();
```
fedemkr, line 190: I'm not sure which is the standard here, but maybe these bools combinations could be properties in order to improve readability:
`bool IsLocked => _pinLock || _biometricLock;`
`bool IsBiometricLockInvalid => _biometricLock && !_biometricIntegrityValid`

```c#
if(!IsLocked || IsBiometricLockInvalid)
{
     PromptSSO();
}

```
Aside from this, I think it's ok

jlf0dev: I'm hesitant to add extra bool properties because they might be a little too specific. They would have to be
`bool NotUsingPinOrBio` and `bool IsBiometricLockInvalid` which seem to be describing the conditional in almost the same amount of characters. If we go this route I would also want to refactor the entire class to use them and that is a little scary since we're picking this to `rc`

fedemkr: Yes, makes sense. Just in case I wasn't aming to reduce the amount of chars, only readability since reading a bunch of "||" and "&" can be hard to follow sometimes (not this particular case), but here I believe you have a point and can be left as is 😄

PR #1673, opened 2021-12-06
remove "singleTask" launch mode on Android 11+

### LINKED ISSUE #1645: Android 12 change in lifecycle autofill bug with Draw Over/Accessibility Quick Tile
### Steps To Reproduce

1. Open Bitwarden from launcher
2. Press back
3. Go to website/app with autofill
4. Click on either Auto-fill with Bitwarden pop over or Autofill Quick Tile in notification bar.

### Expected Result

Bitwarden should open with "Items for ..." page allowing you to select the credential to auto-fill.

### Actual Result

Bitwarden loads "My Vault" page not allowing you to select an item to auto-fill

### Screenshots or Videos

_No response_

### Additional Context

Autofill from IME is unaffected

If you use the task manager to swipe the previously launched bitwarden app close. Then the quick tile works. I suspect its an interaction with the lifecycle changes in [Android 12](https://developer.android.com/about/versions/12/behavior-changes-all#activity-lifecycle)

### Operating System

Android

### Operating System Version

12

### Device

Pixel 6 Pro and Pixel 4 XL

### Build Version

2.14.2 (4005)

### Beta

- [X] Using a pre-release version of the application.

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This bug fix has multiple parts:
1) Android 11 has introduced new lifecycle changes and fixes that break our Accessibility Service. When our app is open in the background, the Accessibility Service uses that instance and runs through `OnNewIntent`. Although `OnNewIntent` is not new, we haven't seen this behavior, thus we had no code yet in `OnNewIntent` to handle accessibility interactions. 

2) The second problem is that even though we now handle accessibility calls in `OnNewIntent`, the autofill page does not return our login information to our Accessibility Service. Our Accessibility Service uses a `StartActivityForResult` call that opens the main app and loads the autofill page. `StartActivityForResult` is incompatible with `LaunchMode.SingleTask` because it pulls the existing app to the front, that app is in a different Task. Tasks cannot communicate, meaning no data was being passed back to the Accessibility Service. We require `LaunchMode.SingleTask` to mitigate Strandhogg vulnerabilities in earlier versions of Android, but it can be removed in Android 11+.

**tl;dr**
The Android Accessibility Service on Android 11+ should now correctly show the autofill page and autofill credentials when selected.

Asana Task: https://app.asana.com/0/1169444489336079/1201373745150061/f
Closes #1645

## Code changes

* **src/Android/MainActivity.cs:** Remove `[Activity]` and `[IntentFilter]` attributes and declare manually in AndroidManifest.xml so we can use a resource string for LaunchMode. Handle Accessibility Service calls in `OnNewIntent`.
* **src/App/App.xaml.cs:** Show Autofill Page if Accessibility Service comes through `OnNewIntent`.
* **src/Android/Properties/AndroidManifest.xml:** Create `<activity>` and `<intent-filter>` manually so we can use a resource string for LaunchMode.
* **src/Android/Resources/values-v31/manifest.xml:** Set LaunchMode to `LaunchMode.Multiple` if Android 11+.
* **src/Android/Resources/values/manifest.xml:** Set LaunchMode to `LaunchMode.SingleTask` if Android 10-.
* **src/Android/Android.csproj:** New files

## Testing requirements

- Android 10- Accessibility Service should still work correctly. 
    - With/without app open in background.
- Android 11+ Accessibility Service should now work correctly.
    - With/without app open in background. 
- Testing around the Android lifecycle to make sure the new LaunchMode in Android 11+ didn't make any unintentional changes. 
    - Two Factor Authentication workflow.
    - Returning from application in background.
    - Autofill Service.
    - Main app launcher still works correctly.

## Before you submit


src/Android/MainActivity.cs
```diff
@@ -22,25 +22,8 @@
...
-        Categories = new[] { Intent.CategoryDefault },
-        DataMimeTypes = new[]
-        {
-            @"application/*",
-            @"image/*",
-            @"video/*",
-            @"text/*"
-        })]
+    // Activity and IntentFilter declarations have been moved to Properties/AndroidManifest.xml
+    // They have been hardcoded so we can use the default LaunchMode on Android 12+
```
mportune-bw, line 26: Can you add a note about making sure any future changes are made in both manifest files?  (I can see forgetting about the `values-v31` version)

src/Android/Android.csproj
```diff
@@ -206,6 +206,14 @@
     <AndroidResource Include="Resources\values-night\styles.xml" />
     <AndroidResource Include="Resources\values\styles.xml" />
     <AndroidResource Include="Resources\values\colors.xml" />
+    <AndroidResource Include="Resources\values\manifest.xml">
+      <SubType></SubType>
```
mportune-bw, line 210: Remove the empty `SubType` and `Generator` tags for both manifests

src/Android/Properties/AndroidManifest.xml
```diff
@@ -49,6 +49,20 @@
 
     <!-- Support for LG "Dual Window" mode (for Android < 7.0 users) -->
     <meta-data android:name="com.lge.support.SPLIT_WINDOW" android:value="true" />
+		<activity android:name="com.x8bit.bitwarden.MainActivity" android:configChanges="keyboard|keyboardHidden|navigation|orientation|screenSize|uiMode" android:exported="true" android:icon="@mipmap/ic_launcher" android:label="Bitwarden" android:launchMode="@integer/launchModeAPIlevel" android:theme="@style/LaunchTheme">
```
mportune-bw, line 52: Formatting - can you break this line up

src/Android/MainActivity.cs
```diff
@@ -159,7 +142,16 @@ protected override void OnNewIntent(Intent intent)
             base.OnNewIntent(intent);
             try
             {
-                if (intent.GetBooleanExtra("generatorTile", false))
+                var uri = intent?.GetStringExtra("uri");
+                if (uri != null)
```
fedemkr, line 146: I think we could make it one line (not sure if there is an standard that is against this):
```c#
if(intent?.GetStringExtra("uri") is string uri)
{
    ....
}
```

mportune-bw (approved):

PR #1679, opened 2021-12-09
Fix Clipboard clear after time on iOS

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This PR fixes the Clipboard clear after time set by the user on Options on iOS

## Code changes

* **ClipboardService.cs:** Added this service to centralize the clipboard logic
* **MainActivity.cs/AppDelegate.cs:** Removed clipboard and messenger related logic and moved it to the service
* **OtherFiles:** Changed the old calls to point to the new `ClipboardService`

## Testing requirements

Given that the copy feature has been refactored a bit in order to fix this issue, we should check that values are being copied correctly in the next places of the app (both Android and iOS):

- Cyphers all values that can be copied (Username, Password, TOTP, Card Number, Security Code, Notes, Send URL)
- Password generator password
- Password history password
- Settings → About

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/ViewPageViewModel.cs
+++ b/src/App/Pages/Vault/ViewPageViewModel.cs
@@ -9,6 +9,7 @@ using Bit.Core.Utilities;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Xamarin.Forms;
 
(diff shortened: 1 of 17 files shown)


src/App/Pages/Vault/ViewPageViewModel.cs
```diff
@@ -9,6 +9,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
mportune-bw, line 12: Is this used?

src/App/Services/MobilePlatformUtilsService.cs
```diff
@@ -6,21 +6,23 @@
 using Plugin.Fingerprint.Abstractions;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.SymbolStore;
```
mportune-bw, line 9: Also this

mportune-bw (commented):
This is good, just some import cleanup unless I'm missing something

mportune-bw (approved):

PR #1680, opened 2021-12-09
remove re-throws of exceptions, hiding stack trace

jlf0dev:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Other

## Objective

remove `throw e` from project as it hides stack trace
https://app.asana.com/0/1153292148278596/1201459506316842/f

## Before you submit

### DIFF
--- a/src/Core/Services/AuthService.cs
+++ b/src/Core/Services/AuthService.cs
@@ -268,7 +268,7 @@ namespace Bit.Core.Services
                     kdfIterations = preloginResponse.KdfIterations;
                 }
             }
-            catch (ApiException e) 
+            catch (ApiException e)
             {
                 if (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
                 {
--- a/src/Core/Services/AuthService.cs
+++ b/src/Core/Services/AuthService.cs
@@ -268,12 +268,9 @@ namespace Bit.Core.Services
                     kdfIterations = preloginResponse.KdfIterations;
                 }
             }
-            catch (ApiException e)
+            catch (ApiException e) when (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
             {
-                if (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
-                {
-                    throw e;
-                }
+                throw;
             }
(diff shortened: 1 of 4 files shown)


src/Core/Services/AuthService.cs
```diff
@@ -268,12 +268,9 @@ private async Task<SymmetricCryptoKey> MakePreloginKeyAsync(string masterPasswor
...
                 }
             }
-            catch (ApiException e)
+            catch (ApiException e) when (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
             {
-                if (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
-                {
-                    throw e;
-                }
+                throw;
```
jlf0dev, line 273: I'm not sure why we catch this error just to rethrow it. I left it in there (removing the rethrow of course), but i'll remove it if anyone can confirm that its unnecessary.

fedemkr, line 273: If I'm not mistaken what is written there is not the same as before:
```c#
catch (ApiException e)
{
      if (e.Error == null || e.Error.StatusCode != System.Net.HttpStatusCode.NotFound)
      {
             throw e;
      }
}
```
Because on this code (before), all `ApiException` are being catched but only a few rethrow the exception (the others are ignored). On yours, it'll only catch the `ApiException` that comply with the condition and others will be rethrown as well (so all `ApiException` will thrown). E.g., `ApiException` with `StatusCode` equal to `NotFound` on the original code it'll be ignored grabbed by the catch, and on your code it won't be catched so the exception will be bubbled up to the caller

jlf0dev: I was a little confused by this as well. In another file we used the condition:
```c#
if (e.Error != null && e.Error.StatusCode == System.Net.HttpStatusCode.NotFound)
{
    return new List<BreachAccountResponse>();
}
throw e;
```
which makes more sense to me. 

Either way I've reverted this to catch all `ApiException`. I'm not quite sure what the expected behavior is either, but for now we are following the same rules as before.

vvolkgang (approved):

PR #1682, opened 2021-12-10
Fix Progress dialog crash on tombstoning

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Android: Fix crash produced by the Progress dialog on tombstoning and also to change it because it was deprecated for the new way.

## Code changes

* **DeviceActionService.cs:** Changed `ProgressDialog` for `AlertDialog` with a custom layout for the progress and improved the `HideLoadingAsync` to take care of more situations so that the dismissal is done correctly and doesn't crash the app
* **AddEditPageViewModel.cs:** Added a check before popping the modal because it could lead to a crash on tombstoning

## Testing requirements

Check that the UI of the loading hasn't been affected dramatically, check that the loading is done correctly in different screens, e.g. logging in or saving a cypher.
Also test that while loading (without letting the loading finish) tombstone the app (a simpler way, but not the same, is to set `Don't keep activities` in Developer options of the settings of the phone before opening the app) and then go back to the app and check that it doesn't crash.

## Before you submit

### DIFF
--- a/src/Android/Services/DeviceActionService.cs
+++ b/src/Android/Services/DeviceActionService.cs
@@ -13,6 +13,7 @@ using Android.OS;
 using Android.Provider;
 using Android.Text;
 using Android.Text.Method;
+using Android.Views;
 using Android.Views.Autofill;
 using Android.Views.InputMethods;
 using Android.Webkit;
@@ -27,6 +28,7 @@ using Bit.Core.Enums;
 using Bit.Core.Models.View;
 using Bit.Core.Utilities;
 using Bit.Droid.Autofill;
+using Bit.Droid.Utilities;
 using Plugin.CurrentActivity;
 
 namespace Bit.Droid.Services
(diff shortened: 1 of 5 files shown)


src/Android/Services/DeviceActionService.cs
```diff
@@ -108,22 +112,102 @@ public async Task ShowLoadingAsync(string text)
...
+
+            if (@object is Activity activity)
             {
-                _progressDialog.Dismiss();
-                _progressDialog.Dispose();
-                _progressDialog = null;
+                if (activity.IsFinishing)
+                    return false;
+
+                if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBeanMr1 &&
```
mportune-bw, line 205: Fwiw we only support 5.x +

mportune-bw (dismissed):

src/Android/Services/DeviceActionService.cs
fedemkr: I forgot to remove that sorry, making the change asap

vvolkgang (approved):

mportune-bw (approved):

Issue #1689, opened 2021-12-14
Crash on password generator history when changing theme


fedemkr:
### Steps To Reproduce

1. Go to a Password Generator
2. Go to Password Generator history (three dots at the top right corner)
3. Put the app on background
3. Change light/dark mode 🌕 🌑 
4. Go back to the app
6. The app crashes

### Expected Result

The app should open again without crashing

### Actual Result

The app crashes when opening

### Operating System

iOS

### Operating System Version

15.0

### Device

iPhone 11, iPhone 11 Pro

### Build Version

2.15.0 (1225)

### Beta

- [ ] Using a pre-release version of the application.

fedemkr:
Possible related to https://github.com/xamarin/Xamarin.Forms/issues/12045 as well

PR #1690, opened 2021-12-14
Fix by workaround crash on LabelRenderer and when changing themes #1689

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix crash on iOS LabelRenderer when changing themes on password generator screen and password generator history. These are **workarounds** only

I think this may need to be further tested and check what happens on certain scenarios before going to PRD. I saw that when the app is supposed to crash, but now it doesn't because of the try...catch, there are certain elements that don't update the style, so maybe I could spend more time in it trying to make everything work correctly

## Code changes

* **CustomLabelRenderer.cs:** Added try...catch when calling base so that we can ignore the exception if it happens, it's the only workardound I could find so far (already submitted PR on Xamarin Forms to fix this https://github.com/xamarin/Xamarin.Forms/pull/14982)
* **ThemeManager.cs:** Sometimes the app crashes when closing a view when coming from background having changed the theme, so I've added a try...catch as a workaround here as well.

## Testing requirements

Scenario 1:
- Go to a cypher -> Edit -> tap on the icon for regenerate password
- Put the app on background
- Change light/dark mode
- Go back to the app
- Without waiting the password label to be updated tap on Select
- The app crashes, and now it shouldn't

Scenario 2:
- Go to Password generator -> Password history (on the toolbar menu three dots)
- Put the app on background
- Change light/dark mode
- Go back to the app
- The app crashes, and now it shouldn't (sometimes it doesn't crash, it needs a few attempts)

## Before you submit

### DIFF
--- a/src/App/Utilities/ThemeManager.cs
+++ b/src/App/Utilities/ThemeManager.cs
@@ -4,6 +4,9 @@ using Bit.App.Services;
 using Bit.App.Styles;
 using Bit.Core;
 using Xamarin.Forms;
+#if !FDROID
+using Microsoft.AppCenter.Crashes;
+#endif
 
 namespace Bit.App.Utilities
 {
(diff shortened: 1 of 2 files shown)


mportune-bw (approved):

src/App/Utilities/ThemeManager.cs
```diff
@@ -14,61 +17,79 @@ public static class ThemeManager
 
         public static void SetThemeStyle(string name, ResourceDictionary resources)
         {
-            Resources = () => resources;
+            try
```
vvolkgang, line 20: NP, code style suggestion:

```csharp
        private static (ResourceDictionary Resources, bool UseLightTheme) GetConfig(string name) => name switch
        {
            "dark" => (new Dark(), false),
            "black" => (new Black(), false),
            "nord" => (new Nord(), false),
            "light" => (new Light(), true),
            _ => OsDarkModeEnabled() ? (new Dark(), false) : (new Light(), true)
        };

//..

            var config = GetConfig(name);
            resources.MergedDictionaries.Add(config.Resources);
            UsingLightTheme = config.UseLightTheme;

```

vvolkgang (approved):
LGTM, small suggestion, thought we could improve this code while at it with newer C# features.

src/App/Utilities/ThemeManager.cs
fedemkr: @vvolkgang  The other day I tested the new c# features but weren't compiling because the project is set with target framework Net Standard 2.0 thus it accepts C# 7.3 by default (and it's marked as using Default c# lang) [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults).
We might need to change that config in order to use new stuff and also check whether the CI configuration is also able to compile on latest C#

vvolkgang: Fair point, we'll merge this for now and deal with that later! Thanks!

PR #1700, opened 2021-12-27
Possible fix for iOS crash after long time idling

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix crash produced on iOS after long time in the background

## Code changes

* **AppDelegate.cs:** Added null checks on some services that were being called on the app lifecycle and may have been disposed before.

I'm not sure yet what is the root cause of the problem, so this'd be a baindaid for the problem, if it keeps appearing we'd need to dig in more to check which is the flow or root cause that leads to the blink crash.

## Testing requirements

Steps to reproduce:

- Open the Bitwarden app on iOS
- Stop using the phone for a while leaving it on background (no idea yet how long, to make sure of my testing I left it on the background until the next day)
- Open the Bitwarden app again. On the first try, the app will just blink (opening then closed suddenly).
- It will open successfully on the second try.

The app shouldn't crash when opened again.

## Before you submit

### DIFF
--- a/src/iOS/AppDelegate.cs
+++ b/src/iOS/AppDelegate.cs
@@ -201,8 +201,8 @@ namespace Bit.iOS
 
         public override void DidEnterBackground(UIApplication uiApplication)
         {
-            _storageService.SaveAsync(Constants.LastActiveTimeKey, _deviceActionService.GetActiveTime());
-            _messagingService.Send("slept");
+            _storageService?.SaveAsync(Constants.LastActiveTimeKey, _deviceActionService.GetActiveTime());
+            _messagingService?.Send("slept");
             base.DidEnterBackground(uiApplication);
         }
 
@@ -220,7 +220,7 @@ namespace Bit.iOS
 
         public override void WillEnterForeground(UIApplication uiApplication)
         {
-            _messagingService.Send("resumed");
+            _messagingService?.Send("resumed");
             base.WillEnterForeground(uiApplication);
         }
 


src/iOS/AppDelegate.cs
```diff
@@ -220,7 +220,7 @@ public override void OnActivated(UIApplication uiApplication)
 
         public override void WillEnterForeground(UIApplication uiApplication)
         {
-            _messagingService.Send("resumed");
+            _messagingService?.Send("resumed");
```
vvolkgang, line 223: This will run when the app is transitioning from Background to Active state, if this is null, the app memory was garbage collected so we need to re-init everything otherwise, even if this doesn't crash, the rest of the app might be unusable. 

I suggest merging it for now but revisiting for the IoC / ServiceLocator refactor.

src/iOS/AppDelegate.cs
```diff
@@ -201,8 +201,8 @@ public override void OnResignActivation(UIApplication uiApplication)
 
         public override void DidEnterBackground(UIApplication uiApplication)
         {
-            _storageService.SaveAsync(Constants.LastActiveTimeKey, _deviceActionService.GetActiveTime());
-            _messagingService.Send("slept");
+            _storageService?.SaveAsync(Constants.LastActiveTimeKey, _deviceActionService.GetActiveTime());
+            _messagingService?.Send("slept");
```
vvolkgang, line 205: `DidEnterBackground` only runs for roughly ~5s, we might have to run this in a background task.

https://docs.microsoft.com/en-us/xamarin/ios/app-fundamentals/backgrounding/ios-backgrounding-techniques/ios-backgrounding-with-tasks

src/iOS/AppDelegate.cs
fedemkr: yes, totally agree

src/iOS/AppDelegate.cs
fedemkr: Those methods don't take much time AFAIK, but yes it would be a great idea to run them in a background task

vvolkgang (approved):

PR #1707, opened 2022-01-03
Improve Theming

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Improve theming logic and performance, also fix issues on theme applying on password generator and password generator history. Also fix exception raise on applying theme because of collection modification on enumeration because of the merged dictionaries.

It should fix [https://appcenter.ms/users/kspearrin/apps/bitwarden/crashes/errors/848906559u/overview](https://appcenter.ms/users/kspearrin/apps/bitwarden/crashes/errors/848906559u/overview
)

## Code changes

* **ThemeManager.cs:** Changed several things here to improve performance like checking whether the current theme is already the one we want to set. Also only change the theme ResourceDictionary instead of clearing everything and adding all resources again, this was crashing the app on previous versions and raising exceptions leading to inconsistent theming state on some screens
* **MainActivity.cs and AppDelegate.cs:** Added messenger send to report when the app appeared so that we can update the password generated screens on theme change
* **Generator...:** Added logic to update the theming correctly on these screens
* **App.xaml.cs:** Changed UpdateTheme so that it's done in proper order and applied without any issue on race conditions

## Testing requirements

Steps

- Unlock the app
- Put the app to background
- Change OS theme
- Wait the vault timeout (you can set it previously on the app to 1 minute so you don't have to wait that much)
- Return to the app
- The colors and theme are weird on some places of the lock screen

This also affects Password Generator and Password Generator History screens, i.e.:

- Go to Password Generator screen and then to its history (three dots)
- Change the OS theme without waiting the vault timeout
- Go back to the app without waiting the vault timeout
- You'll see that not all passwords themes are changed
- Also if you go back, the generated password of the Password Generator screen will be with previous theme 

## Before you submit


jlf0dev (approved):
Cool idea with the `IThemeResourceDictionary`, looks good to me unless @mportune-bw or @vvolkgang have any issues.

mportune-bw (approved):

src/App/Utilities/ThemeManager.cs
```diff
@@ -21,47 +22,31 @@ public static void SetThemeStyle(string name, ResourceDictionary resources)
...
+                var newTheme = NeedsThemeUpdate(name, resources);
+                if (newTheme is null)
+                {
+                    return;
+                }
+
+                var currentTheme = resources.MergedDictionaries.FirstOrDefault(md => md is IThemeResourceDictionary);
+                if (currentTheme != null)
+                {
+                    resources.MergedDictionaries.Remove(currentTheme);
```
vvolkgang, line 34: Worst case scenario we're might be going through the list 3 times (`.Any` + `FirstOrDefault` + `.Remove`), wish we could come up with something faster given how often this will be used.

vvolkgang (commented):
Looks good overall, `ThemeManager` is definitely better, just wish we could stop using the messaging service.

src/App/Pages/Generator/GeneratorPage.xaml.cs
```diff
@@ -61,25 +61,31 @@ public async Task InitAsync()
         protected async override void OnAppearing()
         {
             base.OnAppearing();
+
+            _vm?.RedrawPassword(); // this is needed when coming back from another view to this one and the theme has changed
```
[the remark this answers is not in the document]
fedemkr: it shouldn't be, at least not on all cases I've tested, but I don't want the app to crash because we didn't consider some scenario. At the top of my head, it should never be null

src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
```diff
@@ -57,5 +61,29 @@ private async void CopyAsync(GeneratedPasswordHistory ph)
             _platformUtilsService.ShowToast("info", null,
                 string.Format(AppResources.ValueHasBeenCopied, AppResources.Password));
         }
+
+        public void Subscribe()
+        {
+            _broadcasterService?.Subscribe(nameof(GeneratorHistoryPageViewModel), async (message) =>
```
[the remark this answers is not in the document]
fedemkr: The problem is that the `Page` has no idea when the app goes back to Foreground, this is only available on the `AppDelegate -> OnActivated` method; so we need to have the messenger or some sort of pub/sub or event to tell the page that the app is now on foreground again to update the theme of the passwords. That's why I used the `messenger`, I think this is a particular case where the messenger is handy.

For 2., we could subscribe to the event `App.Current.RequestedThemeChanged` on the pages that need to update the theme and apply it there; the problem with that is that this event doesn't actually work well when coming from background to foreground, because the things that needs to applied on foreground on the main thread may be done before the desired page is actually visible, thus the changes may not be applied entirely. So iOS `OnActivated` / Droid `OnResume`, are better places that actually tells that you can do some work on the UI that will be applied correctly to the View.

Maybe there's a better way, but this is the simplest and that works on all cases that I've found.

src/App/Utilities/ThemeManager.cs
fedemkr: It's not often at all, unless you change themes every little time hehe, before that code you have a check to see if the theme has changed, if not this code is not triggered at all.

src/Android/MainActivity.cs
```diff
@@ -133,6 +138,21 @@ protected override void OnResume()
                 .GetResult();
         }
 
+        // TODO: When #1660 gets merged then we can use the task extension FireAndForget instead of this method
```
vvolkgang, line 141: Another option (not necessarily better, just a different approach) for this is to expose a synchronous method that saves the task internally, tend to prefer this approach like this to hidden tasks, easier to debug and figure out when things aren't working. 

`ThemeManager.UpdateThemeOnPages();`

```
public void UpdateThemeOnPages()
{
    _updateThemeTask = UpdateThemeOnPagesAsync();
}
```

vvolkgang (approved):

vvolkgang (approved):

PR #1714, opened 2022-01-10
Expand Settings TimePicker tap recognition to entire row

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Every option in settings allows you to tap anywhere within the row to enable the action. Every option except "Custom" timeouts. This fix pulls the option into its own data template and adds a tap recognizer on the entire row for the TimePicker.

Asana: https://app.asana.com/0/1169444489336079/1201172188894451/f

## Code changes

* **src/App/Pages/Settings/SettingsPage/SettingsPage.xaml:** Create new DataTemplate for Custom option in Settings. Include DataTemplate in the DataTemplateSelector
* **src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs:** If TapGestureRecognizer is triggered, find the TimePicker from the StackLayout and trigger.
* **src/App/Pages/Settings/SettingsPage/SettingsPageListItemSelector.cs:** Return new DataTemplate for Custom option

## Screenshots

https://user-images.githubusercontent.com/24985544/148834516-153919be-d983-43e6-b247-2a8352507515.mov

## Testing requirements

- Test functionality of Settings options

## Before you submit

### DIFF
--- a/src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
@@ -1,4 +1,5 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.ComponentModel;
 using System.Linq;
 using System.Threading.Tasks;
 using Bit.App.Abstractions;
@@ -43,6 +44,17 @@ namespace Bit.App.Pages
                 return true;
             }
             return base.OnBackButtonPressed();
+        }
+
+        void ActivateTimePicker(object sender, EventArgs args)
+        {
+            var stackLayout = (ExtendedStackLayout)sender;
+            SettingsPageListItem item = (SettingsPageListItem)stackLayout.BindingContext;
+            if (item.ShowTimeInput)
+            {
+                var timePicker = stackLayout.Children.Where(x => x is TimePicker).FirstOrDefault();
+                ((TimePicker)timePicker).Focus();
+            }
         }
 
         async void OnTimePickerPropertyChanged(object sender, PropertyChangedEventArgs args)
--- a/src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
@@ -53,7 +53,7 @@ namespace Bit.App.Pages
             if (item.ShowTimeInput)
             {
                 var timePicker = stackLayout.Children.Where(x => x is TimePicker).FirstOrDefault();
-                ((TimePicker)timePicker).Focus();
+                ((TimePicker)timePicker)?.Focus();
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
```diff
@@ -45,6 +46,17 @@ protected override bool OnBackButtonPressed()
             return base.OnBackButtonPressed();
         }
 
+        void ActivateTimePicker(object sender, EventArgs args)
+        {
+            var stackLayout = (ExtendedStackLayout)sender;
+            SettingsPageListItem item = (SettingsPageListItem)stackLayout.BindingContext;
+            if (item.ShowTimeInput)
+            {
+                var timePicker = stackLayout.Children.Where(x => x is TimePicker).FirstOrDefault();
```
vvolkgang, line 55: What if we add more time pickers down the line?

src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
```diff
@@ -45,6 +46,17 @@ protected override bool OnBackButtonPressed()
...
+                ((TimePicker)timePicker).Focus();
```
vvolkgang, line 56: Given that `.FirstOrDefault()`, we should probably check for null before executing this just to be safe.

vvolkgang (changes requested):

src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
Hinton: I don't think that will happen. The setting page is a bit messy as it is and we should think of ways to re-design it in the long run.

vvolkgang: Looking forward to that as well! Until then, don't we have an ID or something we could use to pinpoint the exact item here?

jlf0dev: There might be some confusion here. The `sender` will actually be the `ExtendedStackLayout` inside the `DataTemplate` where the `TimePicker` lives. Its not the `CollectionView` with the list of cells. This means that while there may be additional `TimePicker` cells added, the sender will always refer back to the one tapped. 

We could hardcode the `View` # inside the `ExtendedStackLayout`, but that seems like a bad idea if we ever want to reformat the cell.

vvolkgang (approved):

src/App/Pages/Settings/SettingsPage/SettingsPage.xaml.cs
vvolkgang: Misunderstood, thanks for clearing it up!

PR #1721, opened 2022-01-18
Account Deletion on SSO with CME

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Added possibility to delete an account that was logging on SSO with CME. Also, added a `VerificationCodePage` that can be reused when we need action code verification.

## Code changes

* **VerificationActionsFlowHelper.cs:** Added this helper that interacts with the different verifications (master password/verification code) and the actions to perform (like delete account)
* **DeleteAccountViewModel.cs:** Modified to adapt it to the `VerificationActionsFlowHelper`
* **VerificationCodePage.xaml/ViewModel.cs:** Added this view to reuse it when we need verification code with actions
* **ApiService:** Added possibility of avoid logging out when `HttpStatusCode.Unauthorized`

## Testing requirements

* Perform same tests as in #1621 
* Test that an account logged in on SSO with CME can also be deleted with the Verification Code

## Before you submit

### DIFF
--- a/src/App/Utilities/VerificationActionsFlowHelper.cs
+++ b/src/App/Utilities/VerificationActionsFlowHelper.cs
@@ -1,0 +1,226 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading.Tasks;
+using Bit.App.Abstractions;
+using Bit.App.Pages;
+using Bit.App.Pages.Accounts;
+using Bit.App.Resources;
+using Bit.Core.Abstractions;
+using Bit.Core.Enums;
+using Bit.Core.Utilities;
+using Newtonsoft.Json;
+using Xamarin.Forms;
+
+namespace Bit.App.Utilities
+{
+    //public interface IVerificationFlowHandler
+    //{
+    //    Task Handle();
+    //}
+
+    //public abstract class VerificationFlowHandler : IVerificationFlowHandler
+    //{
+    //    IVerificationFlowHandler _nextHandler;
+
+    //    public VerificationFlowHandler(IVerificationFlowHandler nextHandler)
+    //    {
+    //        _nextHandler = nextHandler;
+    //    }
+
+    //    public virtual Task Handle()
+    //    {
+    //        return _nextHandler.Handle();
+    //    }
+    //}
+
+    //public class MasterPasswordVerificationFlowHandler : VerificationFlowHandler
+    //{
+    //    readonly IUserVerificationService _userVerificationService;
+
+    //    public MasterPasswordVerificationFlowHandler(IVerificationFlowHandler nextHandler)
+    //        : base(nextHandler)
+    //    {
+    //    }
+
+    //    public override async Task Handle()
+    //    {
+    //        if (!await _userVerificationService.VerifyUser(Secret, Core.Enums.VerificationType.MasterPassword))
+    //        {
+    //            return;
+    //        }
+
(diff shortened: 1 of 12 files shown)


src/App/Utilities/VerificationActionsFlowHelper.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private readonly Dictionary<VerificationFlowAction, IActionFlowExecutioner> _actionExecutionerDictionary = new Dictionary<VerificationFlowAction, IActionFlowExecutioner>();
+
+        public VerificationActionsFlowHelper(IKeyConnectorService keyConnectorService,
+            IPasswordRepromptService passwordRepromptService)
+        {
+            _keyConnectorService = keyConnectorService;
+            _passwordRepromptService = passwordRepromptService;
+
+            _actionExecutionerDictionary.Add(VerificationFlowAction.DeleteAccount, ServiceContainer.Resolve<IDeleteAccountActionFlowExecutioner>("deleteAccountActionFlowExecutioner"));
```
jlf0dev, line 75: I like what you're doing with the `VerificationActionsFlowHelper` and the `IActionFlowExecutioner`, trying to keep the two steps decoupled. I guess I'm wondering what you see are some other use cases for this workflow? This might just help me understand it better.

If the goal _is_ to keep it decoupled from the ViewModel and allow us to use it during export as well, I'm not sure if registering the `IDeleteAccountActionFlowExecutioner` as a service and keeping a reference to it here is the best way to tackle it. Do you think we need access to the Executioner outside of the Helper? Could we just create a new object inside the helper?

fedemkr: Yes, the other current use case would be "Export Vault" and any other future action that need verification ✅ .
I thought it as this will be the helper needed for any interaction with the master password/verification code/any future verification process to perform an action.

About the decoupling, I think it's better to have it registered as a service in order to be more testable (I forgot to mention that I'd need to add unit tests on the PR), so we can just inject the `Executioner` from the testing container. Maybe we could improve that and just pass it on the constructor so we avoid resolving it there. Or put a public method to setup the executioners from outside.
Also I wasn't sure on having the `Executioner` on the same `ViewModel` file or in another file in some specific folder or `Utilities`.

jlf0dev: Ah I see, that sounds good to me. After thinking about it I think including the `Executioner` in the `VerificationActionsFlowHelper` as a service is a better idea than exposing it and leaving it in the constructor or something. 

I'm also ok with leaving the `Executioner` in the `ViewModel`. Makes more sense to me than throwing it in a random folder somewhere.

src/App/Utilities/VerificationActionsFlowHelper.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        public IVerificationActionsFlowHelper Configure(VerificationFlowAction action,
+            IActionFlowParmeters parameters = null,
+            string verificatioCodeMainActionText = null,
+            bool isVerificationCodeMainActionStyleDanger = false)
+        {
+            _action = action;
+            _parameters = parameters;
+            _verificatioCodeMainActionText = verificatioCodeMainActionText;
```
jlf0dev, line 85: Tiny spelling mistake here

jlf0dev (approved):

PR #1722, opened 2022-01-18
Check for disable save prompt option before sending fill request in Android Autofill

### LINKED ISSUE #1017: Disable Save Prompt not working
On Android Pie 9.0 with May 2019 security patch for LG G6 Thin Q, unrooted, with autofill and accessibility services enabled, I get spurious and annoying save prompts, that also carry the danger of overwriting a valid password with bad data. 
I enabled accessibility service as the autofill is missing nearly all login fields. Since then, I've been getting a lot of BW's popups to either save what it thinks is a new password (not the case) or to update an existing one that hasn't changed. Also, it seems to ignore the URL blacklist for autofill.

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

closes #1017 
Asana: https://app.asana.com/0/1169444489336079/1185097609703843/f

We're having reports of some applications and browsers repeatedly asking if the user would like to save login information. While this is most likely an issue with the Autofill service or the application itself, our option to "Disable Save Prompts" isn't working correctly to stop all prompts.

This change moves the check for "Disable Save Prompts" to before sending the fill package. That way, we can determine if the we need to include Save Information in the `FillResponse` before the prompt shows.

## Code changes

* **src/Android/Autofill/AutofillHelpers.cs:** Change `BuildFillResponse` to `CreateFillResponse` as we are only returning the `FillResponse.Builder` now in order to add save info later on.
* **src/Android/Autofill/AutofillService.cs:** Move `AddSaveInfo` call to the `AutofillService` as we now check the DisableSavePrompts before calling it.

## Testing requirements

- Testing around Android Autofill Service and the "Disable Save Prompts" option
    - When enabled, there shouldn't be a prompt to save the login

## Before you submit

### DIFF
--- a/src/Android/Autofill/AutofillHelpers.cs
+++ b/src/Android/Autofill/AutofillHelpers.cs
@@ -160,7 +160,7 @@ namespace Bit.Droid.Autofill
             return new List<FilledItem>();
         }
 
-        public static FillResponse BuildFillResponse(Parser parser, List<FilledItem> items, bool locked,
+        public static FillResponse.Builder CreateFillResponse(Parser parser, List<FilledItem> items, bool locked,
             bool inlineAutofillEnabled, FillRequest fillRequest = null)
         {
             // Acquire inline presentation specs on Android 11+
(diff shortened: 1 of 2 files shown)


fedemkr (commented):
I think the `AutofillService` method should have a try..catch with logging because if some exception is raised there the service will probably crash and we should log that if possible.

fedemkr:
Yes, exactly. I'm not sure if we can provide any feedback to the user saying that something went wrong there but at least logging it should be fine 😄

fedemkr (approved):

PR #1730, opened 2022-01-24
Fix for initial login token storage paradox

mportune-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix NPE on login with account that has never logged in before (recent change in the way skipping token storage is handled combined with a fresh install)

## Code changes

* **TokenService.cs:** Added `forDecodeOnly` bool to `SetAccessTokenAsync` method to allow token decoding without saving to the account state (which doesn't exist until the next step in the auth process, which requires values from the decoded token)
* **AuthService.cs:** Utilize the new bool

## Testing requirements

Login with a fresh install and boom.  Apparently it's been awhile since I cleared app data.

## Before you submit

### DIFF
--- a/src/Core/Services/AuthService.cs
+++ b/src/Core/Services/AuthService.cs
@@ -352,7 +352,7 @@ namespace Bit.Core.Services
             {
                 await _tokenService.SetTwoFactorTokenAsync(tokenResponse.TwoFactorToken, email);
             }
-            await _tokenService.SetTokensAsync(tokenResponse.AccessToken, tokenResponse.RefreshToken);
+            await _tokenService.SetAccessTokenAsync(tokenResponse.AccessToken, true);
             await _stateService.AddAccountAsync(
                 new Account(
                     new Account.AccountProfile()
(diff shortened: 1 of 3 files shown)


src/Core/Services/AuthService.cs
```diff
@@ -352,7 +352,7 @@ private async Task<SymmetricCryptoKey> MakePreloginKeyAsync(string masterPasswor
             {
                 await _tokenService.SetTwoFactorTokenAsync(tokenResponse.TwoFactorToken, email);
             }
-            await _tokenService.SetTokensAsync(tokenResponse.AccessToken, tokenResponse.RefreshToken);
+            await _tokenService.SetAccessTokenAsync(tokenResponse.AccessToken, true);
```
fedemkr, line 355: Won't there be a problem by not setting the `RefreshToken` there? Or is that planned to be moved to another place to be saved? Because other than in `ApiService -> DoRefreshTokenAsync` I don't see any other place where that's being saved with this change

vvolkgang (approved):

src/Core/Services/AuthService.cs
mportune-bw: Both tokens are properly "set" on the next line (new `Account` constructor):

```
new Account.AccountTokens()
{
    AccessToken = tokenResponse.AccessToken,
    RefreshToken = tokenResponse.RefreshToken,
}
```

...and are then immediately available via `StateService`, which is used to perform addition get/set operations on the RefreshToken within `TokenService`.  The only reason I couldn't remove this particular line outright is because `TokenService` requires an in-memory copy of the AccessToken to decode the values necessary for the other parts of the Account constructor (userId, email).

PR #1731, opened 2022-01-25
Add logs to help solve iOS PN problem

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Add logs and information to help figure out what's wrong on iOS push notifications. **This PR should be reverted after we fix the iOS push notifications on TestFlight** But we need this PR to trigger a new TestFlight version with additional logs

## Code changes

* Added logs and info on Settings about regarding PN registration

## Testing requirements

This is merely for Dev logs.

## Before you submit

### DIFF
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using Bit.Core.Enums;
 using Bit.Core.Models.Domain;
 using Xamarin.Forms;
+using ZXing.Client.Result;
 
 namespace Bit.App.Pages
 {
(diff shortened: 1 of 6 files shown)


jlf0dev:
Could we run a manual Build on your branch instead? Or is this a master specific issue?

@joseph-flinn is a manual Build on a certain branch possible? Just to send it to App Store Connect

mportune-bw:
@jlf0dev Builds happen on all pushed branches (see the `Actions` tab to grab artifacts).  Publishing to the test tracks (both google and testflight) isn't something we can cleanly one-off however, which is why it only happens in a go-forward manner (meaning this has to go to master for that to happen)

fedemkr:
Also, we can't grab the `ipa` generated by the Actions build and install it directly on device because it's being built with `AppStore` configuration, so we need to go to `master` to generate the `TestFlight` build and so I can check what's going on with the iOS push notifications.

For the future maybe we could set up an `AdHoc` build that we can trigger to reproduce or test stuff in production-like mode without going to TestFlight, or add some kind of environment on `AppCenter` where we can trigger builds from specific branches to test this kind of stuff

mportune-bw:
Just so everyone knows we can commit this to master safely.  The automatic publishing only goes to internal users on both platforms, whereas public test tracks require action by a human.  As long as it's reverted before we cut `rc` (which is when I typically start a public beta) we'll be good.

jlf0dev:
Ah gotcha, ok sounds good to me then. I was just worried since we have a short timeframe before we cut for `rc`. Maybe we can figure out an App Center or `Ad Hoc` build in the future :)

jlf0dev (approved):

PR #1746, opened 2022-02-02
Fix iOS top space on Account switching list overlay after modal

fedemkr:
## Type of change
- [X] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Fix top space shown on iOS on the Account Switching overlay after opening and closing a modal on the Vault and then opening the Account Switching list.

## Code changes

* **GroupingsPage.xaml:** Added effect to change the behavior of content inset adjustment on the list view of account switching overlay so that it never takes into account the Safe Area which was causing the issue.

## Testing requirements

In the Vault, after opening and closing a modal like adding a cypher, and opening the account switching overlay shouldn't present any weird spaces at the top.

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
+++ b/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
@@ -189,13 +189,17 @@
                         ItemSelected="AccountRow_Selected"
                         BackgroundColor="{DynamicResource BackgroundColor}"
                         VerticalOptions="Start"
-                        RowHeight="{OnPlatform 70, iOS=70, Android=74}">
+                        RowHeight="{OnPlatform 70, iOS=70, Android=74}"
+                        effects:ScrollViewContentInsetAdjustmentBehaviorEffect.ContentInsetAdjustmentBehavior="Never">
                         <ListView.ItemTemplate>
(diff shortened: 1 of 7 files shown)


src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
```diff
@@ -189,13 +189,17 @@
                         ItemSelected="AccountRow_Selected"
                         BackgroundColor="{DynamicResource BackgroundColor}"
                         VerticalOptions="Start"
-                        RowHeight="{OnPlatform 70, iOS=70, Android=74}">
+                        RowHeight="{OnPlatform 70, iOS=70, Android=74}"
```
mportune-bw, line 192: We should probably add this to Home, Login, and Lock as well (Home for sure has the same problem, the others could potentially at some point)

mportune-bw:
Awesome!  Just one inline comment

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
fedemkr: oh ok, didn't think there were any modals there, but yes I'll add this everywhere

mportune-bw (approved):

PR #1753, opened 2022-02-04
Create and use Account Switching overlay control

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add an account switching overlay control to be reused on all the views. Also add its own ViewModel to handle the logic and configuration.

## Code changes

* **AccountSwitchingOverlayView:** This is the new control for the overlay
* **AccountSwitchingOverlayViewModel:** This is the overlay's viewmodel to handle logic and configuration
* **...Page.xaml:** Changed to use the new overlay control
* **...Page.xaml.cs:** Changed to use the new overlay control and removed old code
* **...PageViewModel.cs:** Added and created new `AccountSwitchingOverlayViewModel` on each to offer the binding context for the new overlay control and configuration 
* **BaseContentPage.cs:** Moved Account switching overlay logic to the new control
* **StateService:** Change the `AccountViews` to be a `List` instead of an `ExtendedObservableCollection` given that we don't change items while the view is open; it only changes when adding/removing accounts and that happens on pages transition so it doesn't make sense to use all the notifying events of the `ExtendedObservableCollection`. And also the List is replaced completely when refreshing the accounts so a simple `List` makes more sense for this case given that there are just a few items.

## Testing requirements

Everything should work as before.

## Before you submit

### DIFF
--- a/src/App/App.csproj
+++ b/src/App/App.csproj
@@ -128,6 +128,7 @@
   <ItemGroup>
     <Folder Include="Resources\" />
     <Folder Include="Behaviors\" />
+    <Folder Include="Controls\AccountSwitchingOverlay\" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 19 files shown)


mportune-bw (commented):
oops, meant to approve but merged instead.  Guess the coffee wore off.

PR #1775, opened 2022-02-13
Fix for vault timeout not firing on resume

### LINKED ISSUE #1772: Android app no longer locks on vault timeout in version 2.16.1
### Steps To Reproduce

1. In Settings, set Vault Timeout to any time increment (e.g., 1 minute)
2. In Settings, set Vault Timeout Action to Lock
3. On your device, navigate away from Bitwarden so that it is no longer in the foreground
4. Wait until the time interval selected above has passed
5. Re-open Bitwarden and there is no prompt to unlock the vault

### Expected Result

If the select Vault Timeout period has elapsed, the vault should automatically lock and the user should be presented with a challenge to unlock the vault, which is how it has worked in past versions (i.e., up to and including app version 2.15).

### Actual Result

In release 2.16.1, the vault does not lock and the user is not presented with a challenge to unlock the vault, even though the vault timeout period has elapsed.

### Screenshots or Videos

_No response_

### Additional Context

Issue reported on the Community Forums at: 
https://community.bitwarden.com/t/android-app-vault-lock-not-working/38010

### Operating System

Android

### Operating System Version

Android 12, kernel version 4.19.113-23195093

### Device

Samsung Galaxy S20 5G

### Build Version

2.16.1

### Beta

- [ ] Using a pre-release version of the application.

mportune-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix for vault timeout not firing on app resume in most recent release.  Closes #1772

## Code changes

* **App.xaml.cs:** The [recent change](https://github.com/bitwarden/mobile/pull/1707) to make `UpdateTheme` async resulted in a race condition which allowed `BaseContentPage.SaveActivity()` to execute prior to `VaultTimeoutService.CheckVaultTimeoutAsync()`.  This resulted in the `LastActiveTime` value being reset before it could be compared to the current time, making the app think only a few milliseconds had passed regardless of how long it was in the background.  By checking vault timeout before executing the improved theme update process, we're able to establish the correct elapsed time again.  @fedemkr Let me know if you see any issues with this interfering with the theme updater.

## Testing requirements

Vault timeout should function as expected again.

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -97,7 +97,7 @@ namespace Bit.App
                 {
                     if (Device.RuntimePlatform == Device.iOS)
                     {
-                        ResumedAsync();
+                        ResumedAsync().FireAndForget();
                     }
                 }
                 else if (message.Command == "slept")
@@ -205,7 +205,7 @@ namespace Bit.App
             _isResumed = true;
             if (Device.RuntimePlatform == Device.Android)
             {
-                ResumedAsync();
+                ResumedAsync().FireAndForget();
             }
         }
 
@@ -215,7 +215,7 @@ namespace Bit.App
             _messagingService.Send("stopEventTimer");
         }
 
-        private async void ResumedAsync()
+        private async Task ResumedAsync()
         {
             await _vaultTimeoutService.CheckVaultTimeoutAsync();
             _messagingService.Send("startEventTimer");
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -217,10 +217,9 @@ namespace Bit.App
 
         private async void ResumedAsync()
         {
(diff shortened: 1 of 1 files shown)


src/App/App.xaml.cs
```diff
@@ -217,10 +217,9 @@ private async Task SleptAsync()
 
         private async void ResumedAsync()
```
vvolkgang, line 218: Let's take the opportunity to change this to async Task and use the FireAndForget method @fedemkr added in a previous PR, logging exceptions.

vvolkgang (changes requested):

fedemkr (approved):
It shouldn't interfere with the theme update, but just in case let QA know

vvolkgang (approved):

PR #1781, opened 2022-02-15
Usability improvements to handling logout on vault timeout

mportune-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Improvements to the handling and presentation of logging out on vault timeout.  Some high-level details:
- Being "logged out" on timeout is presented to the user the same way as being "locked" (i.e. the account is still displayed in the account switcher and can be selected as the active account)
- User settings are maintained so when the user logs back in the app behaves like they're picking up where they left off.  Vault data however (pretty much everything encrypted for that matter) is removed, otherwise it would defeat the purpose of logging out on timeout.  Logging back in will sync and reconstruct that data.
- When the user logs out via the settings menu (known internally as "user initiated"), then we clean house of all settings and data, making subsequent logins by the same user a from-scratch affair.
- This flow feels the most natural in the mobile app since we don't run timers to manage account auth states

## Code changes

* **AppHelpers.cs:** Flow updates and added automatic account switch toast on user initiated logout
* **StateService.cs:** Added support for `userInitiated` bool to logout and account removal flow to keep or remove data accordingly, removed userId support for non user-specific data, removed unused methods
* **StateMigrationService.cs/SyncService.cs:** Removed support for ForcePasswordReset value since that was moved to the auth flow awhile back
* **VaultTimeoutService.cs:** Check for "lock" action set before establishing `isLocked` to prevent result from having multiple meanings
* **Constants.cs/MobileStorageService.cs:** Removed unused storage/pref keys
* **AppResources.resx:** Added string for automatic account switch toast
* **LoginPageViewModel.cs:** Removed `RememberEmail` property since it was never set and is always true
* **LoginSsoPageViewModel.cs:** Removed `RememberOrgIdentifier` property since it was never set and is always true
* **ExtensionPageViewModel.cs:** Removed use of `StateService` since the values were never set and defaulted to false

## Testing requirements

Use "Logout" and "Lock" vault timeout actions with multiple accounts simultaneously

## Before you submit

### DIFF
--- a/src/App/Utilities/AppHelpers.cs
+++ b/src/App/Utilities/AppHelpers.cs
@@ -449,7 +449,6 @@ namespace Bit.App.Utilities
 
         public static async Task LogOutAsync(string userId, bool userInitiated = false)
         {
-            var syncService = ServiceContainer.Resolve<ISyncService>("syncService");
             var tokenService = ServiceContainer.Resolve<ITokenService>("tokenService");
             var cryptoService = ServiceContainer.Resolve<ICryptoService>("cryptoService");
             var settingsService = ServiceContainer.Resolve<ISettingsService>("settingsService");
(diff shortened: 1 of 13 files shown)


jlf0dev (approved):

src/App/Utilities/AppHelpers.cs
```diff
@@ -473,21 +472,21 @@ public static async Task LogOutAsync(string userId, bool userInitiated = false)
...
+                stateService.LogoutAccountAsync(userId, userInitiated));
+
+            // check if we switched accounts automatically
+            if (userInitiated && await stateService.GetActiveUserIdAsync() != null)
+            {
+                var messagingService = ServiceContainer.Resolve<IMessagingService>("messagingService");
+                messagingService.Send("switchedAccount");
+                
+                var platformUtilsService = ServiceContainer.Resolve<IPlatformUtilsService>("platformUtilsService");
+                platformUtilsService.ShowToast("info", null, AppResources.AccountSwitchedAutomatically);
```
fedemkr, line 489: If possible I think it'd be better to tell the user which account is the current one on the message as well, like "Switched to next available account: myaccount@domain.com" or directly "Switched to myaccount@domain.com".
And is it ok for it to be here instead of inside the logic of the subscription of "switchedAccount" `messagingService`? Because the logic makes async stuff and that is not awaited on `messagingService.Send("switchedAccount");` so the message could be shown before the account is actually switched. Furthermore, if any other place calls "switchedAccount" would need to add the toast thus IMO it makes more sense to put it in the `SwitchedAccountAsync` of the `App.xaml.cs`

src/Core/Services/VaultTimeoutService.cs
```diff
@@ -88,7 +88,8 @@ public async Task<bool> ShouldTimeoutAsync(string userId = null)
             {
                 return false;
             }
-            if (await IsLockedAsync(userId))
+            var vaultTimeoutAction = await _stateService.GetVaultTimeoutActionAsync(userId);
+            if (vaultTimeoutAction == "lock" && await IsLockedAsync(userId))
```
fedemkr, line 92: Maybe not for this PR, but I'd very like to change the `VaultTimeoutAction` to be an enum instead of a `string`, it's easier to compare, less error prone, allocates less memory and you can see all of the values in one place.

src/App/Utilities/AppHelpers.cs
mportune-bw: My only concern with including the email address is the potential for really long addresses being crammed into a toast.  I'll run it by design though.

I see what you're saying on the second point - taking another look now

PR #1792, opened 2022-02-20
Fix for invalid PIN lock state when switching accounts

mportune-bw:
## Type of change
- [X] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Fixes to data handling for PIN lock flow caused by my misinterpretation of a specific "key" used for crypto purposes vs a "key" used for storage purposes.

## Code changes

* **Account.cs:** Moved `PinProtected` from `AccountSettings` to `PinProtectedKey` in `AccountKeys` as it represents a value held in memory only (`AccountKeys` is never written to state storage)
* **Everything else:** Updated references to changes made in `Account.cs`, and added `await` to a calls to `_stateService.GetPinProtectedKeyAsync(..)` so null check actually works as expected

## Testing requirements

https://app.asana.com/0/1166330464508357/1201846609152550

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LockPageViewModel.cs
+++ b/src/App/Pages/Accounts/LockPageViewModel.cs
@@ -40,7 +40,8 @@ namespace Bit.App.Pages
         private string _biometricButtonText;
         private string _loggedInAsText;
         private string _lockedVerifyText;
-        private Tuple<bool, bool> _pinSet;
+        private bool _isPinProtected;
+        private bool _isPinProtectedWithKey;
 
         public LockPageViewModel()
         {
@@ -134,8 +135,9 @@ namespace Bit.App.Pages
 
         public async Task InitAsync()
         {
-            _pinSet = await _vaultTimeoutService.IsPinLockSetAsync();
-            PinLock = (_pinSet.Item1 && await _stateService.GetPinProtectedKeyAsync() != null) || _pinSet.Item2;
+            (_isPinProtected, _isPinProtectedWithKey) = await _vaultTimeoutService.IsPinLockSetAsync();
+            PinLock = (_isPinProtected && await _stateService.GetPinProtectedKeyAsync() != null) ||
+                      _isPinProtectedWithKey;
             BiometricLock = await _vaultTimeoutService.IsBiometricLockSetAsync() && await _cryptoService.HasKeyAsync();
 
             // Users with key connector and without biometric or pin has no MP to unlock with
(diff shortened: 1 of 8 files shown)


src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -135,7 +135,7 @@ public string LockedVerifyText
         public async Task InitAsync()
         {
             _pinSet = await _vaultTimeoutService.IsPinLockSetAsync();
-            PinLock = (_pinSet.Item1 && _stateService.GetPinProtectedAsync() != null) || _pinSet.Item2;
+            PinLock = (_pinSet.Item1 && await _stateService.GetPinProtectedKeyAsync() != null) || _pinSet.Item2;
```
fedemkr, line 138: Could we use explicit names here instead of `Item1` and `Item2`?
instead of 
`_pinSet = await _vaultTimeoutService.IsPinLockSetAsync();` 
you could have
`(isPinProtected, isPinProtectedWithKey) = await _vaultTimeoutService.IsPinLockSetAsync();`

src/Core/Services/VaultTimeoutService.cs
```diff
@@ -146,7 +146,7 @@ public async Task LockAsync(bool allowSoftLock = false, bool userInitiated = fal
 
             if (await _keyConnectorService.GetUsesKeyConnector()) {
                 var pinSet = await IsPinLockSetAsync(userId);
-                var pinLock = (pinSet.Item1 && _stateService.GetPinProtectedAsync(userId) != null) || pinSet.Item2;
+                var pinLock = (pinSet.Item1 && await _stateService.GetPinProtectedKeyAsync(userId) != null) || pinSet.Item2;
```
fedemkr, line 149: Could we use explicit names here instead of `Item1` and `Item2`?
instead of 
`pinSet = await IsPinLockSetAsync();` 
you could have
`(isPinProtected, isPinProtectedWithKey) = await IsPinLockSetAsync();`

src/iOS.Core/Controllers/LockPasswordViewController.cs
```diff
@@ -101,7 +101,7 @@ public override async void ViewDidLoad()
             else
             {
                 _pinSet = _vaultTimeoutService.IsPinLockSetAsync().GetAwaiter().GetResult();
-                _pinLock = (_pinSet.Item1 && _stateService.GetPinProtectedAsync() != null) || _pinSet.Item2;
+                _pinLock = (_pinSet.Item1 && await _stateService.GetPinProtectedKeyAsync() != null) || _pinSet.Item2;
```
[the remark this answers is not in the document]
mportune-bw: I think `ViewDidLoad()` was changed to async after I added `GetAwaiter().GetResult()` and I just missed it.  Updated those with proper `await`

fedemkr (approved):

PR #1793, opened 2022-02-21
Global theme and account removal

mportune-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

- Apply theme selections globally (took this approach so we can restore per-account themes in the future by removing a contained block of code)
- Support for account removal from login page (via logout on vault timeout)
- Check for existing account on add-account and prompt to switch if available

## Code changes

* **LoginPage.*:** Added "Remove Account" menu option and supporting code, also added checking for already-added account when logging in a new account
* **OptionsPageViewModel.cs:** Apply theme globally when applied in Settings
* **StateService.cs:** Added `GetUserIdAsync` and `ApplyThemeGloballyAsync` methods

## Screenshots

Screen Shot 2022-02-20 at 8 45 54 PM

Screen Shot 2022-02-20 at 8 46 17 PM

Screen Shot 2022-02-20 at 8 48 03 PM

## Testing requirements

- Change theme then switch users - theme will remain consistent
- Switch to an active logged-out account and expand the overflow menu to use the "Remove Account" option
- Add an account you've already added

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPageViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPageViewModel.cs
@@ -101,7 +101,7 @@ namespace Bit.App.Pages
             }
         }
 
-        public async Task LogInAsync(bool showLoading = true)
+        public async Task LogInAsync(bool showLoading = true, bool checkForExistingAccount = false)
         {
             if (Xamarin.Essentials.Connectivity.NetworkAccess == Xamarin.Essentials.NetworkAccess.None)
             {
@@ -133,6 +133,23 @@ namespace Bit.App.Pages
             ShowPassword = false;
             try
             {
+                if (checkForExistingAccount)
+                {
+                    var userId = await _stateService.GetUserIdAsync(Email);
+                    if (!string.IsNullOrWhiteSpace(userId))
+                    {
+                        var switchToAccount = await _platformUtilsService.ShowDialogAsync(
+                            AppResources.SwitchToAlreadyAddedAccountConfirmation,
+                            AppResources.AccountAlreadyAdded, AppResources.Yes, AppResources.Cancel);
+                        if (switchToAccount)
+                        {
+                            await _stateService.SetActiveUserAsync(userId);
+                            _messagingService.Send("switchedAccount");
+                        }
+                        return;
+                    }
+                }
+
                 if (showLoading)
                 {
                     await _deviceActionService.ShowLoadingAsync(AppResources.LoggingIn);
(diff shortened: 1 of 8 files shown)


src/App/Pages/Accounts/LoginPageViewModel.cs
```diff
@@ -190,5 +207,15 @@ public void TogglePassword()
...
         }
+
+        public async Task RemoveAccountAsync()
+        {
+            var confirmed = await _platformUtilsService.ShowDialogAsync(AppResources.RemoveAccountConfirmation,
+                AppResources.RemoveAccount, AppResources.Yes, AppResources.Cancel);
+            if (confirmed)
+            {
+                _messagingService.Send("logout");
+            }
```
fedemkr, line 218: Add try...catch

src/Core/Services/StateService.cs
```diff
@@ -67,6 +67,28 @@ public async Task<bool> IsAuthenticatedAsync(string userId = null)
...
+                throw new Exception("email cannot be null");
+            }
+
+            await CheckStateAsync();
+            if (_state?.Accounts != null)
+            {
+                foreach (var account in _state.Accounts)
+                {
+                    var accountEmail = account.Value?.Profile?.Email;
+                    if (accountEmail != null && accountEmail == email)
```
fedemkr, line 83: Given that `email` can't be `null` because of aforementioned check, then this can be directly written as 
`if (accountEmail == email)`

src/Core/Services/StateService.cs
```diff
@@ -843,6 +865,26 @@ public async Task SetThemeAsync(string value, string userId = null)
...
+                {
+                    var uid = account.Value?.Profile?.UserId;
+                    if (uid == null || uid == activeUserId)
+                    {
+                        // active user theme already set, skip
+                        continue;
+                    }
+                    await SetThemeAsync(value, uid);
+                }
+            }
```
fedemkr, line 885: In order to avoid nesting and some lines, I'd rewrite this as:
```c#
if (_state?.Accounts == null)
{
    return;
}

var activeUserId = await GetActiveUserIdAsync();
foreach (var account in _state.Accounts)
{
    var uid = account.Value?.Profile?.UserId;
    if (uid != null && uid != activeUserId)
    {
        // active user theme not already set, so set it
        await SetThemeAsync(value, uid);
    }
}
```

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPageViewModel.cs
mportune-bw: Can you clarify?  This is triggered directly by the user with the app in the foreground.

fedemkr: @mportune-bw if the `ShowDialogAsync` or the `_messagingService` throw an exception the app will crash given that the `RemoveAccountAsync` is called from the event `RemoveAccount_Clicked`. So a `try...catch` is needed here or in `RemoveAccount_Clicked` to prevent the crash if any exception happen.

PR #1795, opened 2022-02-21
Fix vault item icon image size to be adaptive on Large Font Size Accessibility

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix icon image size to be adaptive on Large Font Size Accessibility which also fixes the row height being really high on large vault items.

## Code changes

* **CipherViewCell:** Change the Width/HeightrRequest of the CachedImage icon to be calculated with the font scale.

### Before:

image
### After:
image
image

## Testing requirements

Test that the icons have the correct size and check this on large vault as well.

## Before you submit

### DIFF
--- a/src/App/Controls/CipherViewCell/CipherViewCell.xaml.cs
+++ b/src/App/Controls/CipherViewCell/CipherViewCell.xaml.cs
@@ -21,8 +21,10 @@ namespace Bit.App.Controls
         {
             InitializeComponent();
 
-            var deviceActionService = ServiceContainer.Resolve<IDeviceActionService>("deviceActionService");
-            _iconColumn.Width = new GridLength(40 * deviceActionService.GetSystemFontSizeScale(), GridUnitType.Absolute);
+            var fontScale = ServiceContainer.Resolve<IDeviceActionService>("deviceActionService").GetSystemFontSizeScale();
+            _iconColumn.Width = new GridLength(40 * fontScale, GridUnitType.Absolute);
+            _iconImage.WidthRequest = 22 * fontScale;
+            _iconImage.HeightRequest = 22 * fontScale;
         }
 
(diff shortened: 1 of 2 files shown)


src/App/Controls/CipherViewCell/CipherViewCell.xaml.cs
```diff
@@ -21,8 +21,10 @@ public CipherViewCell()
         {
             InitializeComponent();
 
-            var deviceActionService = ServiceContainer.Resolve<IDeviceActionService>("deviceActionService");
-            _iconColumn.Width = new GridLength(40 * deviceActionService.GetSystemFontSizeScale(), GridUnitType.Absolute);
+            var fontScale = ServiceContainer.Resolve<IDeviceActionService>("deviceActionService").GetSystemFontSizeScale();
+            _iconColumn.Width = new GridLength(40 * fontScale, GridUnitType.Absolute);
+            _iconImage.WidthRequest = 22 * fontScale;
+            _iconImage.HeightRequest = 22 * fontScale;
```
jlf0dev, line 27: Does this `22` need to be the same value as the original `WidthRequest` and `HeightRequest`? If so, we should consider putting it in a `const` or grabbing the value from `_iconColumn` before using it.

fedemkr: I'll put it in a cost, we can't grab it from _iconColumn (unless we do the math proportion: 40/22=1.82 => _iconImage.WidthRequest = _iconColumn.Width.PropetyForTheActualValue / 1,82 but we'd need a const here as well and it's simpler with the 22 const) 😄 . I'll make the change

jlf0dev (approved):

PR #1811, opened 2022-02-24
Apply Disable Favicon setting globally to match desktop

mportune-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Make the "Disable Website Icons" option global to match desktop.  I'm continuing to follow the `Apply*Globally` pattern established with the Theme change so a single block of code can be removed to restore per-account functionality in the future.

## Code changes

* **StateService.cs:** Created `SetValueGloballyAsync` method to support key/value based settings (values not in the `Account` object)
* **OptionsPageViewModel.cs:** Removed previous `Apply*Global` calls since everything is contained within `StateService` now

## Testing requirements

https://app.asana.com/0/1201803072708593/1201880420113232

## Before you submit

### DIFF
--- a/src/Core/Services/StateService.cs
+++ b/src/Core/Services/StateService.cs
@@ -579,26 +579,9 @@ namespace Bit.Core.Services
                 await GetDefaultStorageOptionsAsync());
             var key = Constants.DisableFaviconKey(reconciledOptions.UserId);
             await SetValueAsync(key, value, reconciledOptions);
-        }
 
-        public async Task ApplyDisableFaviconGloballyAsync(bool? value)
-        {
-            // TODO remove this method (ApplyDisableFaviconGloballyAsync) to restore per-account icon support
-            await CheckStateAsync();
-            if (_state?.Accounts == null)
-            {
-                return;
-            }
-            var activeUserId = await GetActiveUserIdAsync();
-            foreach (var account in _state.Accounts)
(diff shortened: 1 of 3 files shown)


src/Core/Services/StateService.cs
```diff
@@ -581,6 +581,26 @@ public async Task SetDisableFaviconAsync(bool? value, string userId = null)
...
+            foreach (var account in _state.Accounts)
+            {
+                var uid = account.Value?.Profile?.UserId;
+                // skip active user (theme already set)
+                if (uid != null && uid != activeUserId)
+                {
+                    await SetDisableFaviconAsync(value, uid);
+                }
+            }
+        }
```
fedemkr, line 602: This is using almost the same code as `Task ApplyThemeGloballyAsync(string value)`. We could refactor a bit to share some more code, given that the only thing that changes is the call in the `if`; like a method that has a Task as a parameter that will be triggered in the i

mportune-bw: Took a slightly different approach to keep everything contained within `StateService`

fedemkr (approved):

src/Core/Services/StateService.cs
fedemkr: I really like the new approach 😄

PR #1813, opened 2022-02-25
BEEEP: Colorize hidden custom field when value visible

djsmith85:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
The standard password field displays number and special characters in a different color for better readability. A hidden custom-field currently does not colorize it's value when shown.

Asana task: https://app.asana.com/0/1200804338582616/1201884248598927/f

## Code changes
* **src/App/Pages/Vault/ViewPageViewModel.cs:** Add new public property ColoredHiddenValue which formats the hidden custom field value
* **src/App/Pages/Vault/ViewPage.xaml:** Adjust hidden fieldvalue binding to new property and add text-html to StyleClass

## Screenshots
**BEFORE:**
image

**AFTER:**
image

## Testing requirements
When the value of a hidden custom field is shown, it should be colorized as the standard password field

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/ViewPage.xaml
+++ b/src/App/Pages/Vault/ViewPage.xaml
@@ -579,8 +579,8 @@
                                                          Grid.Row="1"
                                                          Grid.Column="0">
                                                 <controls:MonoLabel
-                                                    Text="{Binding ValueText, Mode=OneWay}"
-                                                    StyleClass="box-value"
+                                                    Text="{Binding ColoredHiddenValue, Mode=OneWay}"
+                                                    StyleClass="box-value, text-html"
                                                     IsVisible="{Binding ShowHiddenValue}" />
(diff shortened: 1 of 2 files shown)


fedemkr (approved):

fedemkr:
Just one thing @djsmith85 , does it behave correctly when changing themes?

djsmith85:
@fedemkr Thanks for the review and seems to look good on the other themes.

Dark:
image
Black:
image
Nord:
image

djsmith85:
Good catch: Setting the theme to OS default and then switching the OS theme while having the password and custom field values exposed, doesn't work (doesn't re-draw). Closing the cipher and re-opening it re-draws the item and works fine.

A bit out of scope for a mobile newbie 😉

PR #1818, opened 2022-02-28
Fixed migration and account removal issues

mportune-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fixed issue where migration was skipped when using default environment, "Remove Account" was shown on iOS when it shouldn't be, and creating a new account was not carrying forward the current `DisableFavicon` setting

## Code changes

* **LoginPage.xaml.cs:** Only show "Remove Account" in the iOS action sheet when `_email` is disabled (same condition as Android)
* **StateMigrationService.cs:** Check for existence of standalone `userId` to signify v2 and trigger migration to v3 (was previously checking if `envUrls` were stored in Prefs (instead of LiteDb) but those particular prefs are only saved when a custom env is set on the Environment page (which I always did throughout the dev process, classic) so the migrator assumed it was a fresh install and never ran
* **StateService.cs:** Copy the current `DisableFavicon` setting when creating a new account (same approach as theme)

## Testing requirements

- Remove existing installation to clear data
- Install v2.16.2 (1434) via TestFlight and login + configure any settings
- Install latest TestFlight build over top and observe migration

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPage.xaml.cs
+++ b/src/App/Pages/Accounts/LoginPage.xaml.cs
@@ -147,8 +147,10 @@ namespace Bit.App.Pages
                 return;
             }
 
-            var selection = await DisplayActionSheet(AppResources.Options, 
-                AppResources.Cancel, null, AppResources.GetPasswordHint, AppResources.RemoveAccount);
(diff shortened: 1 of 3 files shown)


vvolkgang (approved):
LGTM!

PR #1833, opened 2022-03-08
Fix for short profile Name value crashing app

mpbw2:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

- Using a one or two character "Name" in your account profile would result in an invisible avatar, which then caused iOS to crash when our `CustomNavigationRenderer` tried to apply a rendering mode to the non-existent avatar.  The latter issue is now protected from crashing, and the former issue has been fixed so both platforms create the avatar properly if "Name" is 1 or 2 chars.  Special thanks to "AJ" for the help tracking this down. :)
- Discovered "Name" was never updated after account was initially added, so now it's updated on sync

## Code changes

* **AvatarImageSource.cs:** Make sure `chars` and `upperData` are both populated if the first two conditions fail in `Draw()`, and return the original value if both conditions fail in `GetFirstLetters(..)`
* **CustomNavigationRenderer.cs:** Null check `uiBarButtonItem.Image` before applying rendering mode
* **SyncService.cs & StateService.cs:** Added support for updating "Name" on sync (previously was only set when adding account)

## Before you submit

### DIFF
--- a/src/App/Controls/AvatarImageSource.cs
+++ b/src/App/Controls/AvatarImageSource.cs
@@ -62,6 +62,10 @@ namespace Bit.App.Controls
                 upperData = _data.ToUpper();
                 chars = GetFirstLetters(upperData, 2);
             }
+            else
+            {
+                chars = upperData = _data.ToUpper();
+            }
 
             var bgColor = StringToColor(upperData);
             var textColor = Color.White;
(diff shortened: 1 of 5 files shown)


jlf0dev (approved):
Nice!

PR #1850, opened 2022-03-17
Fix iOS 15.4 crash on adding first item on collection

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix crash produced on iOS 15.4 when adding an item on an empty list: Vault, Send mainly. Also applied the fix on Settings and Autofill Ciphers.
This has a drawback that it takes a bit longer to load the collection and there is an animation when loading the items, but it's something much better than the app crashing. It's because adding a delay between item adding, the crash is related to https://github.com/xamarin/Xamarin.Forms/issues/13323
Also fixed crash when calling base on `OpenUrl` on `AppDelegate`

## Code changes

* **AppDelegate:** Removed call to base on `OpenUrl` given that that throws an exception for optional iOS method
* **...xaml:** Added common item header template, removed grouping and `GroupHeaderTemplate`
* **...Selector:** Added the `HeaderTemplate` and logic to return it depending on the VM item
* **...ViewModel:** Changed logic to use `ObservableRangeCollection` and on iOS to have a bit of delay after clearing items and also after adding each header that is what is causing the measure crash on iOS 15.4
* **...ItemViewModel:** Added header item VM and also common interface to use on the lists.

## Testing requirements

Test that Vault items, Send items, Settings and Autofill ciphers list are loading correctly even if you remove all items so that the list gets empty and also when the list is empty add a new item and test that the app behaves as expected without crashing.

## Before you submit

### DIFF
--- a/src/App/Pages/Send/SendGroupingsPage/ISendGroupingsPageListItem.cs
+++ b/src/App/Pages/Send/SendGroupingsPage/ISendGroupingsPageListItem.cs
@@ -1,0 +1,6 @@
+﻿namespace Bit.App.Pages
+{
+    public interface ISendGroupingsPageListItem
+    {
+    }
+}
--- a/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
+++ b/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
@@ -73,7 +73,29 @@
                 </controls:ExtendedStackLayout>
             </DataTemplate>
 
+            <DataTemplate
+                x:Key="headerTemplate"
+                x:DataType="pages:SendGroupingsPageHeaderListItem">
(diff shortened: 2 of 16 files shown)


mportune-bw (commented):
Do we need a `ExtendedObservableRangeCollection` to implement any of the customizations made in `ExtendedObservableCollection`?

fedemkr:
They are pretty much the same @mportune-bw, IMO [ObservableRangeCollection](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/ObservableRangeCollection.shared.cs) has some better additions as well

mportune-bw (approved):

PR #1861, opened 2022-03-24
Fix for missing token when checking for key connector migration

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Crash data from both platforms show a common thread where `TokenService.DecodeToken` is throwing `Access token not found` when called from `KeyConnectorService.UserNeedsMigration` > `TokenService.GetIsExternal`.  Based on the 1:1 ratio of users to crashes, this seems like a race condition that occurs once before an active account token is written to storage.  An additional null check of `_accessTokenForDecoding` before calling `DecodeToken` from `GetIsExternal` (following the same logic of the `decoded?["amr"]` null check immediately below it) should prevent this.

## Code changes

* **TokenService.cs:** Return `false` if `_accessTokenForDecoding` if still null after `GetTokenAsync` and before `DecodeToken`

## Before you submit

### DIFF
--- a/src/Core/Services/TokenService.cs
+++ b/src/Core/Services/TokenService.cs
@@ -209,6 +209,10 @@ namespace Bit.Core.Services
             {
                 await GetTokenAsync();
             }
+            if (_accessTokenForDecoding == null)
+            {
+                return false;
+            }
             var decoded = DecodeToken();
             if (decoded?["amr"] == null)
             {


src/Core/Services/TokenService.cs
```diff
@@ -209,6 +209,10 @@ public async Task<bool> GetIsExternal()
             {
                 await GetTokenAsync();
             }
+            if (_accessTokenForDecoding == null)
```
cscharf, line 212: Would it make more sense to change this to a `while` instead of `if` with the call to `await GetTokenAsync()`, perhaps with a cancellation token or counter within that loop? What impacts does `GetIsExternal()` have on the UI/UX if it returns `false` (since this is a race condition vs. a valid `null` condition)?

mp-bw: From what I can see returning false prevents the `RemoveMasterPasswordPage` from launching.  As for looping vs returning false - I've gone through what feels like hundreds of flow possibilities and I keep disproving my own theories, so the only thing I'm sure of is that I don't know (we've never been able to repro this in-house). I'm counting on subsequent calls to `TabsPage.OnAppearing` to trigger the page launch once the token becomes available, but looping with a counter _could_ be a better approach.  Need input from @jlf0dev on this one.

jlf0dev: I'm not sure how we could improve this either, besides maybe loading the token a few lines sooner? @mp-bw we should be filling the token from the `LogInHelper` but is it not getting there in time?

That being said, I _think_ I like it more without a loop. The delay in relying on subsequent calls to `TabsPage.OnAppearing` to check for the token feels safer than the danger of hanging the UI while we wait.

cscharf (approved):

src/Core/Services/TokenService.cs
mp-bw: > we should be filling the token from the `LogInHelper` but is it not getting there in time?

It _should_ be there unless we intentionally don't store it based on configuration (combination of logout on timeout and no biometrics I believe prevent token storage).  ~~Perhaps in that state, the app is backgrounded on timeout, brought to the foreground and firing `OnAppearing` before the vaultTimeoutService re-routes to the login screen.  Though If that were the case I would expect to see a higher number of crashes per user.~~ Except a "resume" flow like that would still have the token in memory, so add that to the dead theory pile.

src/Core/Services/TokenService.cs
```diff
@@ -209,6 +209,10 @@ public async Task<bool> GetIsExternal()
...
+            {
+                return false;
+            }
```
fedemkr, line 215: IMO, even though it's something really small this should be done after the await inside the previous if, otherwise if `_accessTokenForDecoding` is not `null` at beginning it's being compared again where the result will be again not null.

```
if (_accessTokenForDecoding == null)
{
     await GetTokenAsync();
     if (_accessTokenForDecoding == null)
     {
            return false;
     }
}
```

PR #1865, opened 2022-03-25
[BEEEP] Improved the ServiceContainer Register/Resolve

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Improve the `ServiceContainer` so that we can register/resolve service a lot easier.
Also made the collection a `ConcurrentDictionary` so we don't have threading issues.

## Code changes

* **ServiceContainer.cs:** Changed `RegisteredServices` to be a `ConcurrentDictionary` so that we don't run in threading issues. Also added several methods that don't need the `serviceName` string to register/resolve the services.

## Testing requirements

Nothing, right now it's just an addition to the component so nothing on the app has actually changed.

## Before you submit


src/Core/Utilities/ServiceContainer.cs
```diff
@@ -129,6 +129,59 @@ public static T Resolve<T>(string serviceName, bool dontThrow = false)
...
+        {
+            Register(typeof(T), obj);
+        }
+
+        public static void Register(Type type, object obj)
+        {
+            var serviceName = GetServiceRegistrationName(type);
+            if (!RegisteredServices.TryAdd(serviceName, obj))
+            {
+                throw new Exception($"Service {serviceName} has already been registered.");
```
vvolkgang, line 143: Do we really need to throw? We could just register the new one and discard the previous reference.

src/Core/Utilities/ServiceContainer.cs
```diff
@@ -1,14 +1,15 @@
...
 using System.Threading.Tasks;
+using Bit.Core.Abstractions;
+using Bit.Core.Services;
 
 namespace Bit.Core.Utilities
 {
     public static class ServiceContainer
     {
-        public static Dictionary<string, object> RegisteredServices { get; set; } = new Dictionary<string, object>();
+        public static ConcurrentDictionary<string, object> RegisteredServices { get; set; } = new ConcurrentDictionary<string, object>();
```
vvolkgang, line 12: Can't this be private?

src/Core/Utilities/ServiceContainer.cs
fedemkr: I wanted not to change much of the previous logic, but yes definitely we can do that and it's the way most of the DI engines work. And today it's throwing an exception so if nothing is currently broken we can overwrite the registration without any possible drawbacks

src/Core/Utilities/ServiceContainer.cs
fedemkr: It could be private but currently it's being used from outside to check whether there are services registered. I could change the callers logic to just call a `HasAnyRegistrations()` method/property of the `ServiceContainer`. 
Not sure if we can use the `bool Inited` current property or if we might get into a wall because it's being set to `true` just at the beginning of the `Init` method instead of at the end, thus it's just working to prevent calling Init several times right now and not to check whether all services have been registered, so if we use this property we might get some timing issues on some places, so a different property/method should be better for this that checks directly against the collection or at the end of the registrations.

mp-bw (approved):

PR #1869, opened 2022-04-01
Crash fixes

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fixes for several crashes reported in AppCenter

## Code changes

* **BiometricService.cs:** Extended `try` scope around keystore calls (`_keyStore.GetKey(..)` was throwing `UnrecoverableKeyException` which we catch, but only when calling `cipher.Init(..)`)
* **DeviceActionService.cs:** Null check `cipher` in addition to `cipher.Login`
* **AutofillServicesPageViewModel.cs** Platform check Android before calling `deviceActionService.AutofillServiceEnabled()` which throws unsupported on iOS
* **AutofillCiphersPageViewModel.cs** null check `appOptions` before getting `appOptions.Uri`
* **TokenService.cs** move second `_accessTokenForDecoding` null check inside initial null check block as recommended in #1861 

## Testing requirements

N/A as these crashes haven't been reproduced in house

## Before you submit

### DIFF
--- a/src/App/Pages/Settings/AutofillServicesPageViewModel.cs
+++ b/src/App/Pages/Settings/AutofillServicesPageViewModel.cs
@@ -4,6 +4,7 @@ using Bit.App.Resources;
 using Bit.App.Services;
 using Bit.Core.Abstractions;
 using Bit.Core.Utilities;
+using Xamarin.Forms;
 
 namespace Bit.App.Pages
 {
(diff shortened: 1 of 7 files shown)


src/App/Pages/Settings/AutofillServicesPageViewModel.cs
```diff
@@ -192,7 +193,8 @@ public void ToggleDrawOver()
         
         public void UpdateEnabled()
         {
-            AutofillServiceToggled = _deviceActionService.AutofillServiceEnabled();
+            AutofillServiceToggled =
+                Device.RuntimePlatform == Device.Android && _deviceActionService.AutofillServiceEnabled();
```
fedemkr, line 197: Even though this is fine, I think the `_deviceActionService` should be the one indicating whether `AutofillServiceEnabled()` should be called given that following the **Information expert** pattern that service is the one with the information to check whether that method is implemented and IMO we should minimize the runtime platform checks on the VMs if possible.
So something like this would be better
`AutofillServiceToggled = _deviceActionService.HasAutofillService && _deviceActionService.AutofillServiceEnabled();`

mp-bw: Makes sense - I went ahead and made the change and will save the remainder for a tech debt task (there's a lot)

fedemkr (approved):

PR #1878, opened 2022-04-08
[BEEEP] add context to search titles

jlf0dev:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

Currently, search screens in our clients use generic titles such as “Search collection” or “Search folder”. In desktop and browser the folder name is shown but in mobile there is no clue as to which folder, collection, or item type you are searching for.

This PR adds the name of the collection/folder/type to the title of the page in mobile. For Sends, it also adds whether its a Text or File Send.

## Code changes

* **src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs:** Pass the page title to search screen
* **src/App/Pages/Vault/CiphersPage.xaml.cs:** Removes previous titles and uses page title instead
* **src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml.cs:** Passes the type of Send to the search screen
* **src/App/Pages/Send/SendsPage.xaml.cs:** Uses type of Send as search screen title

## Screenshots

https://user-images.githubusercontent.com/24985544/162494216-880e1b41-5281-4012-840f-87d16ab7d462.mov

https://user-images.githubusercontent.com/24985544/162494239-f4fe54bc-0712-4776-aa8e-580e8d8bd957.mp4

## Testing requirements

- Test search titles for correct names

## Before you submit

### DIFF
--- a/src/App/Pages/Send/SendsPage.xaml.cs
+++ b/src/App/Pages/Send/SendsPage.xaml.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using Bit.App.Controls;
 using Bit.App.Resources;
+using Bit.Core.Enums;
 using Bit.Core.Models.View;
 using Xamarin.Forms;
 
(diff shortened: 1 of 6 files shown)


src/App/Pages/Send/SendsPage.xaml.cs
```diff
@@ -12,15 +13,22 @@ public partial class SendsPage : BaseContentPage
         private SendsPageViewModel _vm;
         private bool _hasFocused;
 
-        public SendsPage(Func<SendView, bool> filter, bool type = false)
+        public SendsPage(Func<SendView, bool> filter, SendType? type = null)
```
vvolkgang, line 16: Would suggest moving away from nullable types as much as possible on new developments so we can eventually turn on [C# 8.0 (non) nullable references](https://endjin.com/blog/2020/04/dotnet-csharp-8-nullable-references-non-nullable-is-the-new-default). For enums in general, we should add a `Null` case ideally as the first or `= 0` field as that will be the default value used if a serialisation goes wrong. 

E.g:

```
    public enum SendType
    {
        Null,
        Text,
        File,
    }
```

vvolkgang (changes requested):

src/App/Pages/Send/SendsPage.xaml.cs
cscharf: Can we please wait and not add `null` enum types and stick with `Nullable<t>` for the time being, until C# 8.0 has null references in-place or we're using it? Easier to refactor as well.

vvolkgang (approved):

PR #1882, opened 2022-04-15
Account switching in autofill UI (Android)

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Added support for account switching within the autofill UI on Android

## Code changes

* **AutofillCiphersPage.xaml:** Added `AvatarImageSource` to toolbar, additional `ShowNoData` property (explained below), and `AccountSwitchingOverlayView` control to layout
* **AutofillCiphersPage.xaml.cs/AutofillCiphersPageViewModel.cs:** Added avatar & overlay support, plus separate loading & "NoData" states to smooth out UX when switching accounts (previously "NoData" was simply the negated "HasData" resulting in "No Matches" being shown momentarily before showing the matches)
* **App.xaml.cs:** Added additional condition for clearing `Options.Uri` upon exiting the autofill flow (required to prevent main app from falling back into "autofill matches" view when switching accounts, as that triggers the `SetMainPageAsync()` flow that shows autofill matches if `Options.Uri` has value)

## Testing requirements

Testing should consist of three parts:

1. **Obvious:** Trigger autofill, switch users, unlock if necessary, select matches
2. **Less Obvious:** Make sure main app continues to behave normally after using autofill flow (including saving new creds during autofill)
3. **Not Remotely Obvious:** File selection flow (via file attachments and file Sends)

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -208,7 +208,10 @@ namespace Bit.App
                 {
                     await _stateService.SetLastActiveTimeAsync(_deviceActionService.GetActiveTime());
                 }
-                SetTabsPageFromAutofill(isLocked);
+                if (!SetTabsPageFromAutofill(isLocked))
+                {
+                    ClearAutofillUri();
+                }
                 await SleptAsync();
             }
         }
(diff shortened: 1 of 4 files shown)


vvolkgang (approved):
LGTM!

PR #1905, opened 2022-05-12
PS-518 - Add setting to block AppCenter / Analytics - Mobile

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
Add a new entry in settings to allow user to disable sending information of crash reports to Appcenter.

## Code changes
* **SettingsPageViewModel.cs:** Added new method, `AppCenterReportingAsync`, to handle activation and deactivation of Crashes SDK. All changes are protected by compilation tags and shouldn't appear in **FDroid** build or **Debug** (Appcenter is only running in release compilations).
* **SettingsPage.xaml:** Add call to AppCenterReportingAsync 

## Testing Requirements
Note: This change will only not appear in **Debug** or **FDroid** builds.
1. Login into your account
2. Go to **Settings** tab
3. Scroll down to **Others** section
4. Click on **Report crash logs**

## Screenshots
image
image

## Before you submit


src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -286,6 +294,26 @@ public async Task VaultTimeoutAsync(bool promptOptions = true, int? newTimeout =
             }
         }
 
+        public async Task AppCenterReportingAsync()
+        {
+#if !FDROID && !DEBUG
```
mp-bw, line 299: Same here

mp-bw (changes requested):
See inline comments

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -286,6 +294,26 @@ public async Task VaultTimeoutAsync(bool promptOptions = true, int? newTimeout =
...
+            var options = _appcenterActions.Select(s => s == _appCenterReportsDisplayValue ? $"✓ {s}" : s).ToArray();
```
fedemkr, line 300: I don't see much value on having the allocation for _appcenterActions, I'd just construct the array in situ avoiding the Select and the conversion from List to the Array, even more with just two options. And also I'd extract a method to compose the string with the tick and reuse it on the other methods on the ViewModel:
```c#
var options = new [] 
{  
     CreateSelectableOption(AppResources.Yes, _appCenterReportsEnabled),
     CreateSelectableOption(AppResources.No, !_appCenterReportsEnabled),
};

string CreateSelectableOption(string option, bool selected) => selected ? $"✓ {option}" : option;
```

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -286,6 +294,26 @@ public async Task VaultTimeoutAsync(bool promptOptions = true, int? newTimeout =
...
+            var selection = await Page.DisplayActionSheet(AppResources.ReportAppLogsDescription, AppResources.Cancel, null, options);
+
+            if (selection == null || selection == AppResources.Cancel)
+            {
+                return;
+            }
+
+            var cleanSelection = selection.Replace("✓ ", string.Empty);
+            var enableAppcenter = cleanSelection == AppResources.Yes;
+            await Microsoft.AppCenter.Crashes.Crashes.SetEnabledAsync(enableAppcenter);
```
fedemkr, line 310: Even though I know it's used like this on other parts of the code, we can avoid the replace and just compare directly:
```c#
await Microsoft.AppCenter.Crashes.Crashes.SetEnabledAsync(CompareSelection(selection, AppResources.Yes));

// Directly
bool CompareSelection(string selection, string compareTo) => selection == $"✓ {compareTo}";

// Substring from 2nd index (avoiding the tick comparison)
bool CompareSelection(string selection, string compareTo) => selection?.Substring(2) == compareTo;
```

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -286,6 +294,26 @@ public async Task VaultTimeoutAsync(bool promptOptions = true, int? newTimeout =
...
+            _appCenterReportsEnabled = await Microsoft.AppCenter.Crashes.Crashes.IsEnabledAsync();
+            _appCenterReportsDisplayValue = _appCenterReportsEnabled ? AppResources.Enabled : AppResources.Disabled;
```
fedemkr, line 312: if the options array is created in situ using the `_appCenterReportsEnabled` flag, then we can delete this line and the `_appCenterReportsDisplayValue` field

fedemkr (changes requested):
Overall it seems fine, just a few code improvements

src/App/Resources/AppResources.resx
```diff
@@ -2181,4 +2181,10 @@
   <data name="EnterTheVerificationCodeThatWasSentToYourEmail" xml:space="preserve">
     <value>Enter the verification code that was sent to your email</value>
   </data>
+  <data name="ReportAppLogs" xml:space="preserve">
+    <value>Report crash logs</value>
+  </data>
```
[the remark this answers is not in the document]
andrebispo5: you're right, thanks!

src/App/Resources/AppResources.resx
```diff
@@ -2181,4 +2181,10 @@
...
+  <data name="ReportAppLogsDescription" xml:space="preserve">
```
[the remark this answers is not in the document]
andrebispo5: Agreed.

fedemkr (commented):
Some comments for improvements, also I'd like @mp-bw view on this just in case I've missed something.

src/Android/MainActivity.cs
```diff
@@ -85,6 +80,8 @@ protected override void OnCreate(Bundle savedInstanceState)
             _appOptions = GetOptions();
             LoadApplication(new App.App(_appOptions));
 
+            ServiceContainer.Resolve<ILogger>("logger").InitAsync();
```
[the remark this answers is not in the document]
andrebispo5: Ye you are right at one point i was using Xam Essentials and forgot to change it back, nice catch!

src/Core/Services/Logging/Logger.cs
```diff
@@ -29,6 +40,60 @@ protected Logger()
...
+                case Enums.DeviceType.Android:
+                    AppCenter.Start(DroidAppSecret, typeof(Crashes));
+                    break;
+                case Enums.DeviceType.iOS:
+                    AppCenter.Start(iOSAppSecret, typeof(Crashes));
+                    break;
+                default:
+                    return;
+
+            }
```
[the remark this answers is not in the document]
andrebispo5: Ye i think this is the case that needs an exception. If someone adds a new device type in the future, for any reason we cannot predict now, it would be secretly sending reports to the "else case" of the if and it would be pretty hard to notice it.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -286,6 +288,27 @@ public async Task VaultTimeoutAsync(bool promptOptions = true, int? newTimeout =
             }
         }
 
+        public async Task LoggerReportingAsync()
+        {
+            _reportLoggingEnabled = await _loggerService.IsEnabled();
```
fedemkr, line 293: Here you don't need to have this line `_reportLoggingEnabled = await _loggerService.IsEnabled();` you're already updating `_reportLoggingEnabled` in line 308 so you don't need to do this again here unless I'm mistaken and there is some flow in which it's outdated. If so, let me know because I can't think of any in which line 293 is necessary

andrebispo5: I let that call stay just in case. But i think we can remove it.

fedemkr (commented):
IMO I'd approve this but maybe we could add in here also the update of the `userId` on AppCenter when Account Switching occurs. What do you think @mp-bw ? Or maybe it could be part of a different PR to have a specific PR change for that issue

mp-bw:
> IMO I'd approve this but maybe we could add in here also the update of the `userId` on AppCenter when Account Switching occurs. What do you think @mp-bw ? Or maybe it could be part of a different PR to have a specific PR change for that issue

Let's save that for a different PR since it may require some additional planning (for example, if a push payload for a non-active user somehow throws, do we log the userId for the payload, or the current active userId? etc.)

mp-bw (approved):

Issue #1664, opened 2021-11-27
iOS - Avoid ambiguous characters is activated inside the main client, but is deactivated when creating a vault item from the autofill prompt.


SergeantConfused:
### Steps To Reproduce

1. Activate 'Avoid ambiguous characters' in the Password Generator settings in the iOS client.
2. Open Safari and navigate to the Web Vault (https://vault.bitwarden.com/#/).
3. Tap on the Master password field.
4. Tap on 'Passwords' above the keyboard.
5. Tap on the plus symbol (+) to create a new vault item.
6. Tap on Generate Password.

### Expected Result

To have 'Avoid ambiguous characters' be activated.

### Actual Result

'Avoid ambiguous characters' is deactivated.

### Screenshots or Videos

N/A.

### Operating System

iOS

### Operating System Version

15.1.

### Device

iPhone SE (2020).

### Build Version

2.14.2 (1105).

### Beta

- [ ] Using a pre-release version of the application.

PR #1906, opened 2022-05-12
PS-591 Fix avoid ambiguous characters #1664

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Fix **avoid ambiguous characters** option. On mobile the value was being stored locally inverted of what the user had selected. Wasn't noticed during the session due to caching.

## Code changes
Refactor the name of the property `Ambiguous` to `AvoidAmbiguous`, this naming was misleading.
Fixed bug where the boolean value for the `AvoidAmbiguous` property was being stored inverted.

* **GeneratorPageViewModel.cs:140:** value was being stored inverted of what UI was showing

## Before you submit

### DIFF
--- a/src/Core/Models/Domain/PasswordGenerationOptions.cs
+++ b/src/Core/Models/Domain/PasswordGenerationOptions.cs
@@ -9,7 +9,7 @@
             if (defaultOptions)
             {
                 Length = 14;
-                AvoidAmbiguous = false;
+                AllowAmbiguousChar = true;
                 Number = true;
                 MinNumber = 1;
                 Uppercase = true;
@@ -27,7 +27,7 @@
         }
 
         public int? Length { get; set; }
-        public bool? AvoidAmbiguous { get; set; }
+        public bool? AllowAmbiguousChar { get; set; }
         public bool? Number { get; set; }
         public int? MinNumber { get; set; }
         public bool? Uppercase { get; set; }
@@ -45,7 +45,7 @@
         public void Merge(PasswordGenerationOptions defaults)
         {
             Length = Length ?? defaults.Length;
-            AvoidAmbiguous = AvoidAmbiguous ?? defaults.AvoidAmbiguous;
+            AllowAmbiguousChar = AllowAmbiguousChar ?? defaults.AllowAmbiguousChar;
             Number = Number ?? defaults.Number;
             MinNumber = MinNumber ?? defaults.MinNumber;
             Uppercase = Uppercase ?? defaults.Uppercase;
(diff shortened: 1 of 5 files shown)


fedemkr (dismissed):

src/Core/Models/Domain/PasswordGenerationOptions.cs
```diff
@@ -9,7 +9,7 @@ public PasswordGenerationOptions(bool defaultOptions)
             if (defaultOptions)
             {
                 Length = 14;
-                Ambiguous = false;
+                AvoidAmbiguous = false;
```
mp-bw, line 12: I'm not a fan of this particular change since it's inconsistent with all the other vars in the model (i.e. bools are named so `true` means "use this", not "don't use this").  In addition by coupling the name with the expectation in the UI, it will forever require updating should we ever change the presentation (i.e. if we change the UI from "Avoid Ambiguous" to "Allow Ambiguous").  I think the original naming should remain.  Thoughts @fedemkr ?

andrebispo5: @mp-bw I usually follow that approach as well but in this particular case it is misleading because how we perceive the feature. In the UI the "true" is to avoid, and that was the cause of the bug in the first place.

mp-bw: @andrebispo5 My thought process is that the data model should not be forced to bend to the UI paradigm, which could change at some point.  Changing the name in this way forces our data model to have a single inconsistently named var for the sake of the UI, forcing the dev to do mental gymnastics if/when using the model elsewhere, when the inversion should just take place in the ViewModel, since the UI is the only place where inversion is required.

fedemkr: I hadn't thought about it but I agree with @mp-bw we should be consistent model wise. Maybe we could take the UX name and name it as `AllowAmbiguous` in the model so to avoid confusion, because just the word `Ambiguous` could be misleading.
And even though it could be a fraction of a thinking, having not operator with a negative variable is less readable than not operator with positive name: `!AvoidAmbiguous` vs `!AllowAmbiguous`; furthermore the worst would be `!NotAllowAmbiguous`

andrebispo5: @fedemkr ye maybe changing the variable name to AllowAmbiguous might help with readability, what do you think @mp-bw? i think is the best trade-off overall.

mp-bw: I'm good with `AllowAmbiguous`

mp-bw (approved):

Issue #1687, opened 2021-12-13
Name field is not prioritized in search results


jonschlu:
### Steps To Reproduce

Searching for items in any part of the iOS app

### Expected Result

When searching for an item, the value from the "name" field should be prioritized and always at the top of the results. If I have one login with the name "Yahoo" in my vault, this entry should be at the very top of the results search results when I search for "Yahoo". This is how it works in the browser extension.

### Actual Result

Results are sorted alphabetically based on both the field "username" and "name". 

If I would search for "Yahoo" in my vault and my email was aaron[at]yahoo.com, all my logins where the username is my e-mail address would be above my login named "Yahoo".

### Operating System

iOS

### Operating System Version

iOS 15.1

### Device

iPhone 12 Pro

### Build Version

2.15.0 (1225)

### Beta

- [ ] Using a pre-release version of the application.

novocaiin:
Same issue on Android.

sawft99:
Yeap. About to put in a bug report but found this. Still the same on Android 12.

PR #1907, opened 2022-05-12
PS-592 Name field is not prioritised in search results #1687

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
When search for ciphers or sends give priority to matches with the `Name` property, show them first in the results list.

## Code changes
* **SearchService.cs:** While running the search methods, created two list, one for matches and another for matches with lower priority. After the search, merged the two lists and retuned the results. This ensures that we only iterate through the list once and give it the order we desire.   

## Screenshots
image

## Before you submit

### DIFF
--- a/src/Core/Services/SearchService.cs
+++ b/src/Core/Services/SearchService.cs
@@ -74,28 +74,34 @@ namespace Bit.Core.Services
             CancellationToken ct = default, bool deleted = false)
         {
             ct.ThrowIfCancellationRequested();
+            var matchedCiphers = new List<CipherView>();
+            var lowPriorityMatchedCiphers = new List<CipherView>();
             query = query.Trim().ToLower();
-            return ciphers.Where(c =>
+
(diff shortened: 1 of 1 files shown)


mp-bw (approved):

kspearrin:
Should we apply this logic to JS clients too?

mp-bw:
> Should we apply this logic to JS clients too?

Hm for some reason I thought it already was, but I just re-read the github issue and I'm not sure how I jumped to that conclusion.  @andrebispo5 Can you apply this change to the other clients for consistency?

andrebispo5:
@mp-bw search is only giving correct `Name` priority in the vault, the Sends are missing this logic. Ye I'll do it.

andrebispo5:
Related to: https://github.com/bitwarden/jslib/pull/808

PR #1921, opened 2022-05-23
PS-90: App keeps asking for 2FA even though I've checked the "remember me" checkbox.

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
App keeps asking for 2FA, even if "remember me" was previously selected.  

## Code changes
* **AuthService.cs:** Previous 2FA code gets cleared if there is an auth error regarding 2FA.
* **StateService.cs:** Removed code that was deleting user 2FA token. Not having this token was causing 2FA login after user logout. 

## Before you submit

### DIFF
--- a/src/Core/Services/AuthService.cs
+++ b/src/Core/Services/AuthService.cs
@@ -346,6 +346,7 @@ namespace Bit.Core.Services
                 TwoFactorProvidersData = response.TwoFactorResponse.TwoFactorProviders2;
                 result.TwoFactorProviders = response.TwoFactorResponse.TwoFactorProviders2;
                 CaptchaToken = response.TwoFactorResponse.CaptchaToken;
+                await _tokenService.ClearTwoFactorTokenAsync(email);
                 return result;
             }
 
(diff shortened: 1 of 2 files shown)


src/Core/Services/AuthService.cs
```diff
@@ -346,6 +346,7 @@ private async Task<SymmetricCryptoKey> MakePreloginKeyAsync(string masterPasswor
                 TwoFactorProvidersData = response.TwoFactorResponse.TwoFactorProviders2;
                 result.TwoFactorProviders = response.TwoFactorResponse.TwoFactorProviders2;
                 CaptchaToken = response.TwoFactorResponse.CaptchaToken;
+                await _tokenService.ClearTwoFactorTokenAsync(email);
```
mp-bw, line 349: I _think_ this may be unneccesary once the other change is made in `StateService`.  Once that's done can you remove this and re-test?

mp-bw (approved):
Disregard my previous comments, I see what you did now.

PR #1923, opened 2022-05-24
[PS-672] Accessibility - File/Text controls unintuitive, "Text" accessible name incorrect

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Change the segmented button (File/Text) in Sends tab, to have a better voice feedback to the user using TalkBack or VoiceOver. Added a missing "Text" resource and dynamic voice over based on button state.

## Code changes

* **src/App/Pages/Send/SendAddEditPageViewModel.cs:** Label properties for button states
* **src/App/Pages/Send/SendAddEditPage.xaml.cs:** Accessibility help text change on button click and screen init.

## Before you submit

### DIFF
--- a/src/App/Pages/Send/SendAddEditPage.xaml.cs
+++ b/src/App/Pages/Send/SendAddEditPage.xaml.cs
@@ -126,6 +126,7 @@ namespace Bit.App.Pages
                     }
                     AdjustToolbar();
                 });
+                SetAccessibilityLabels();
             }
             catch (Exception ex)
             {
@@ -173,6 +174,7 @@ namespace Bit.App.Pages
             {
                 RequestFocus(_nameEntry);
             }
+            SetAccessibilityLabels();
         }
 
         private async void FileType_Clicked(object sender, EventArgs eventArgs)
@@ -184,6 +186,7 @@ namespace Bit.App.Pages
             {
                 RequestFocus(_nameEntry);
             }
+            SetAccessibilityLabels();
         }
 
         private void OnMaxAccessCountTextChanged(object sender, TextChangedEventArgs e)
@@ -355,6 +358,12 @@ namespace Bit.App.Pages
                 TextType_Clicked(null, null);
             }
             _appOptions.CreateSend = null;
+        }
+
+        private void SetAccessibilityLabels()
+        {
+            AutomationProperties.SetHelpText(_typeFileButton, _vm.FileTypeAccessibilityLabel);
+            AutomationProperties.SetHelpText(_typeTextButton, _vm.TextTypeAccessibilityLabel);
         }
     }
 }
--- a/src/App/Pages/Send/SendAddEditPageViewModel.cs
+++ b/src/App/Pages/Send/SendAddEditPageViewModel.cs
@@ -23,6 +23,7 @@ namespace Bit.App.Pages
         private readonly IStateService _stateService;
         private readonly ISendService _sendService;
         private readonly ILogger _logger;
+        private readonly II18nService _i18nService;
         private bool _sendEnabled;
         private bool _canAccessPremium;
         private bool _emailVerified;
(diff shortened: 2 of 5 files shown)


src/App/Pages/Send/SendAddEditPageViewModel.cs
```diff
@@ -231,6 +233,8 @@ public bool SendOptionsPolicyInEffect
         public bool ShowDeletionCustomPickers => EditMode || DeletionDateTypeSelectedIndex == 6;
         public bool ShowExpirationCustomPickers => EditMode || ExpirationDateTypeSelectedIndex == 7;
         public string ShowPasswordIcon => ShowPassword ? BitwardenIcons.EyeSlash : BitwardenIcons.Eye;
+        public string FileTypeAccessibilityLabel => $"{_i18nService.T("File")} {_i18nService.T(IsFile ? "TypeIsSelected" : "TypeIsNotSelected")}";
+        public string TextTypeAccessibilityLabel => $"{_i18nService.T("Text")} {_i18nService.T(IsText ? "TypeIsSelected" : "TypeIsNotSelected")}";
```
fedemkr, line 237: Composition should be avoided in strings translations as this because it's just the way it's written in english but in another language it could vary like in Spanish: "Archivo tipo es seleccionado" makes no sense.
So, in this case you should have "FileTypeIsSelected", "TextTypeIsSelected", "FileTypeIsNotSelected" and "TextTypeIsNotSelected" as keys so they can be well translated to any language.

src/App/Pages/Send/SendAddEditPage.xaml.cs
```diff
@@ -356,5 +359,11 @@ private async Task HandleCreateRequest()
             }
             _appOptions.CreateSend = null;
         }
+
+        private void SetAccessibilityLabels()
+        {
+            AutomationProperties.SetHelpText(_typeFileButton, _vm.FileTypeAccessibilityLabel);
+            AutomationProperties.SetHelpText(_typeTextButton, _vm.TextTypeAccessibilityLabel);
+        }
```
fedemkr, line 367: This should be done in xaml with bindings, if the binding doesn't exist we should create a new custom xaml binding for this as it will be used in other parts of the app. And if this is done, all related code-behind is not necessary.

fedemkr (changes requested):

src/App/Pages/Send/SendAddEditPageViewModel.cs
fedemkr: Also `_is18nService` is not necessary, you can directly call `AppResources.File`

andrebispo5: Totally agree.

src/App/Pages/Send/SendAddEditPage.xaml.cs
andrebispo5: Yes and it makes the code much cleaner, thanks!

src/App/Resources/AppResources.resx
```diff
@@ -1880,11 +1880,17 @@
...
-  <data name="TypeIsNotSelected" xml:space="preserve">
-    <value>type is, not selected, click to select.</value>
+  <data name="FileTypeIsNotSelected" xml:space="preserve">
+    <value>File type is, not selected, click to select.</value>
+  </data>
+  <data name="TextTypeIsSelected" xml:space="preserve">
+    <value>Text type is, selected.</value>
+  </data>
+  <data name="TextTypeIsNotSelected" xml:space="preserve">
+    <value>Text type is, not selected, click to select.</value>
```
fedemkr, line 1893: Remove comma after "...type is," and change "click" to "tap" as in mobile the user taps a control, e.g. "File type is, not selected, click to select" should be "File type is not selected, tap to select"

fedemkr (changes requested):

src/App/Resources/AppResources.resx
andrebispo5: I added the comma because it makes the Voice Over be easier to understand, without it gets confusing because of the automatic robotic voice. Tap makes more sense though.

mp-bw: The commas should be excluded.  TTS can vary wildly between engines & languages, and we can't rely on crowdsourcing to test & adjust every variation.  What sounds better in one situation could be disastrous in another. 
 It's the responsibility of the synthesizer to manage cadence & delivery.

PR #1928, opened 2022-05-27
[SG-79] Mobile Vault Filter

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Added the ability to filter vault ciphers/folders/collections by all/personal/organization when specific conditions are met (account is part of an organization(s) & the Personal Ownership Policy does _not_ apply)

## Code changes

* **GroupingsPage.xaml/xaml.cs/ViewModel.cs:** Added UI & logic for displaying & handling vault filter options & selection
* **TabsPage.cs:** Added `UpdateVaultButtonTitleAsync()` method to change vault button title to `Vaults` or `My Vault` based on filter availability.  Triggered on page appearing as well as sync completion via `BroadcasterService`
* **AppHelpers.cs:** Added `IsShowingVaultFilterAsync()` convenience method
* **FolderService.cs:** Added support for passing in existing `FolderView` list to `GetAllNestedAsync` method to support required flexibility during filter selection

## Screenshots

Example 1: Filter conditions apply and selection UI is displayed.  Page and tab vault icon title changed to "Vaults" to prevent confusion with "My Vault" selection.

Screen Shot 2022-05-27 at 12 28 29 PM

Example 2: Specific organization selected by user.  Page and tab vault icon title changed to "Vaults" to prevent confusion with "My Vault" selection.

Screen Shot 2022-05-27 at 12 28 56 PM

Example 3: Filter conditions do _not_ apply to this user so filter selector row is hidden.  Page and tab vault icon title remains "My Vault".

Screen Shot 2022-05-27 at 12 30 33 PM

## Before you submit

### DIFF
--- a/src/App/Pages/TabsPage.cs
+++ b/src/App/Pages/TabsPage.cs
@@ -12,6 +12,7 @@ namespace Bit.App.Pages
 {
     public class TabsPage : TabbedPage
     {
+        private readonly IBroadcasterService _broadcasterService;
         private readonly IMessagingService _messagingService;
         private readonly IKeyConnectorService _keyConnectorService;
 
@@ -21,6 +22,7 @@ namespace Bit.App.Pages
 
         public TabsPage(AppOptions appOptions = null, PreviousPageInfo previousPage = null)
         {
+            _broadcasterService = ServiceContainer.Resolve<IBroadcasterService>("broadcasterService");
             _messagingService = ServiceContainer.Resolve<IMessagingService>("messagingService");
             _keyConnectorService = ServiceContainer.Resolve<IKeyConnectorService>("keyConnectorService");
 
@@ -80,11 +82,24 @@ namespace Bit.App.Pages
         protected override async void OnAppearing()
         {
             base.OnAppearing();
+            _broadcasterService.Subscribe(nameof(TabsPage), async (message) =>
+            {
+                if (message.Command == "syncCompleted")
+                {
+                    Device.BeginInvokeOnMainThread(async () => await UpdateVaultButtonTitleAsync());
+                }
+            });
             await UpdateVaultButtonTitleAsync();
             if (await _keyConnectorService.UserNeedsMigration())
             {
                 _messagingService.Send("convertAccountToKeyConnector");
             }
+        }
+
+        protected override void OnDisappearing()
+        {
+            base.OnDisappearing();
+            _broadcasterService.Unsubscribe(nameof(TabsPage));
         }
 
         public void ResetToVaultPage()
--- a/src/App/Pages/TabsPage.cs
+++ b/src/App/Pages/TabsPage.cs
@@ -1,6 +1,8 @@
-﻿using Bit.App.Effects;
+﻿using System.Threading.Tasks;
+using Bit.App.Effects;
 using Bit.App.Models;
 using Bit.App.Resources;
+using Bit.App.Utilities;
 using Bit.Core.Abstractions;
 using Bit.Core.Models.Data;
 using Bit.Core.Utilities;
@@ -78,6 +80,7 @@ namespace Bit.App.Pages
         protected override async void OnAppearing()
         {
             base.OnAppearing();
+            await UpdateVaultButtonTitleAsync();
             if (await _keyConnectorService.UserNeedsMigration())
             {
                 _messagingService.Send("convertAccountToKeyConnector");
@@ -130,6 +133,12 @@ namespace Bit.App.Pages
             {
                 groupingsPage.HideAccountSwitchingOverlayAsync().FireAndForget();
             }
+        }
+
+        private async Task UpdateVaultButtonTitleAsync()
+        {
+            var isShowingVaultFilter = await AppHelpers.IsShowingVaultFilterAsync();
+            _groupingsPage.Title = isShowingVaultFilter ? AppResources.Vaults : AppResources.MyVault;
         }
     }
 }
--- a/src/App/Pages/TabsPage.cs
+++ b/src/App/Pages/TabsPage.cs
@@ -1,8 +1,8 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using Bit.App.Effects;
 using Bit.App.Models;
 using Bit.App.Resources;
-using Bit.App.Utilities;
 using Bit.Core.Abstractions;
 using Bit.Core.Models.Data;
 using Bit.Core.Utilities;
@@ -15,6 +15,7 @@ namespace Bit.App.Pages
         private readonly IBroadcasterService _broadcasterService;
         private readonly IMessagingService _messagingService;
         private readonly IKeyConnectorService _keyConnectorService;
+        private readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>("logger");
 
         private NavigationPage _groupingsPage;
         private NavigationPage _sendGroupingsPage;
@@ -152,8 +153,16 @@ namespace Bit.App.Pages
 
         private async Task UpdateVaultButtonTitleAsync()
         {
-            var isShowingVaultFilter = await AppHelpers.IsShowingVaultFilterAsync();
-            _groupingsPage.Title = isShowingVaultFilter ? AppResources.Vaults : AppResources.MyVault;
+            try
+            {
+                var policyService = ServiceContainer.Resolve<IPolicyService>("policyService");
+                var isShowingVaultFilter = await policyService.ShouldShowVaultFilterAsync();
+                _groupingsPage.Title = isShowingVaultFilter ? AppResources.Vaults : AppResources.MyVault;
+            }
+            catch (Exception ex)
+            {
+                _logger.Value.Exception(ex);
+            }
         }
     }
 }
--- a/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
+++ b/src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
@@ -107,6 +107,30 @@
 
             <StackLayout x:Key="mainLayout" x:Name="_mainLayout">
                 <StackLayout
+                    IsVisible="{Binding ShowVaultFilter}"
+                    Orientation="Horizontal"
+                    HorizontalOptions="FillAndExpand"
+                    Margin="0,5,0,0">
+                    <Label
+                        Text="{Binding VaultFilterDescription}"
+                        LineBreakMode="TailTruncation"
+                        Margin="10,0"
+                        StyleClass="text-md, text-muted"
+                        VerticalOptions="Center"
+                        HorizontalOptions="StartAndExpand"
+                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.Name="{u:I18n Filter}" />
+                    <controls:MiButton
+                        Text="&#xe5d3;"
+                        StyleClass="list-row-button, list-row-button-platform, btn-disabled"
+                        Clicked="VaultFilter_Clicked"
+                        VerticalOptions="Center"
+                        HorizontalOptions="End"
+                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.Name="{u:I18n Filter}" />
+                </StackLayout>
+
+                <StackLayout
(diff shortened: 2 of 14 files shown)


src/App/Pages/TabsPage.cs
```diff
@@ -78,12 +82,26 @@ public TabsPage(AppOptions appOptions = null, PreviousPageInfo previousPage = nu
         protected override async void OnAppearing()
         {
             base.OnAppearing();
+            _broadcasterService.Subscribe(nameof(TabsPage), async (message) =>
+            {
+                if (message.Command == "syncCompleted")
+                {
+                    Device.BeginInvokeOnMainThread(async () => await UpdateVaultButtonTitleAsync());
+                }
```
fedemkr, line 90: Add try...catch, if an exception occurs there it will crash the app given that this is similar to an async void method. Or at least try catch inside `UpdateVaultButtonTitleAsync`

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
```diff
@@ -106,6 +106,30 @@
...
+                        Text="{Binding VaultFilterDescription}"
+                        LineBreakMode="TailTruncation"
+                        Margin="10,0"
+                        StyleClass="text-md, text-muted"
+                        VerticalOptions="Center"
+                        HorizontalOptions="StartAndExpand"
+                        AutomationProperties.IsInAccessibleTree="True"
+                        AutomationProperties.Name="{u:I18n Filter}" />
+                    <controls:MiButton
+                        Text="&#xe5d3;"
```
fedemkr, line 124: I think this should be like `Text="{Binding Source={x:Static core:BitwardenIcons.MyCoolIcon}}"` so that we keep all icons centralized in `BitwardenIcons`

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
```diff
@@ -106,6 +106,30 @@
...
+                        StyleClass="list-row-button, list-row-button-platform, btn-disabled"
+                        Clicked="VaultFilter_Clicked"
```
fedemkr, line 126: Could we use the `Command` version of this?

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
```diff
@@ -230,6 +234,11 @@ private async void Search_Clicked(object sender, EventArgs e)
             }
         }
 
+        private async void VaultFilter_Clicked(object sender, EventArgs e)
+        {
+            await _vm.VaultFilterOptionsAsync();
+        }
```
fedemkr, line 240: Not needed if `Command` approach is used.

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -87,7 +92,7 @@ public GroupingsPageViewModel()
         public bool HasCiphers { get; set; }
         public bool HasFolders { get; set; }
         public bool HasCollections { get; set; }
-        public bool ShowNoFolderCiphers => (NoFolderCiphers?.Count ?? int.MaxValue) < NoFolderListSize &&
+        public bool ShowNoFolderCipherGroup => (NoFolderCiphers?.Count ?? int.MaxValue) < NoFolderListSize &&
             (!Collections?.Any() ?? true);
```
fedemkr, line 96: I think it's more readable like this:
```c#
public bool ShowNoFolderCipherGroup => NoFolderCiphers != null 
                                       && NoFolderCiphers.Count < NoFolderListSize 
                                       && Collections != null 
                                       && !Collections.Any();
```

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -564,6 +642,42 @@ private async Task LoadDataAsync()
             }
         }
 
+        private bool IsVaultFilterMyVault => 
+            _vaultFilterSelection != null && _vaultFilterSelection == AppResources.MyVault;
```
fedemkr, line 646: `_vaultFilterSelection != null` is not needed given that `null` is different than `AppResources.MyVault`, you can write it directly as `private bool IsVaultFilterMyVault => _vaultFilterSelection == AppResources.MyVault;`

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -564,6 +642,42 @@ private async Task LoadDataAsync()
...
+        private async Task<string> GetVaultFilterOrgIdAsync()
+        {
+            foreach (var org in _organizations)
+            {
+                if (org.Name == _vaultFilterSelection)
+                {
+                    return org.Id;
+                }
+            }
+            return null;
```
fedemkr, line 661: This is the same as `return _organizations.FirstOrDefault(o => o.Name == _vaultFilterSelection)?.Id;`

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -564,6 +642,42 @@ private async Task LoadDataAsync()
...
+                if (_allCiphers.Any(c => c.FolderId == folder.Id))
+                {
+                    if (folders == null)
+                    {
+                        folders = new List<FolderView>();
+                    }
+                    folders.Add(folder);
+                }
+            }
+            return folders;
```
fedemkr, line 678: This can be rewritten as:
```c#
var folders = decFolders.Where(f => _allCiphers.Any(c => c.FolderId == folder.Id)).ToList();
return folders.Any() ? folders : null;
```

src/App/Resources/AppResources.resx
```diff
@@ -2220,4 +2220,19 @@
...
+  </data>
+  <data name="Vaults" xml:space="preserve">
+    <value>Vaults</value>
+  </data>
+  <data name="VaultFilterDescription" xml:space="preserve">
+    <value>Vault: {0}</value>
+  </data>
+  <data name="VaultFilterDescriptionAll" xml:space="preserve">
+    <value>All</value>
+  </data>
```
fedemkr, line 2237: Shouldn't we just use `All` as the key here? I think it's more 1:1 reference and easier to know the content of the string.

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -342,6 +371,24 @@ public void DisableRefreshing()
...
+            var selection = await Page.DisplayActionSheet(AppResources.FilterByVault, AppResources.Cancel, null, options.ToArray());
+            if (selection == AppResources.Cancel ||
+                (_vaultFilterSelection == null && selection == AppResources.AllVaults) ||
+                (_vaultFilterSelection != null && _vaultFilterSelection == selection))
+            {
+                return;
+            }
+            VaultFilterDescription = selection;
+            await LoadAsync();
+        }
```
fedemkr, line 390: Add try...catch or some sort of exception handling

src/App/Utilities/AppHelpers.cs
```diff
@@ -382,6 +383,17 @@ public static async Task<bool> IsHideEmailDisabledByPolicyAsync()
...
+        public static async Task<bool> IsShowingVaultFilterAsync()
+        {
+            var organizationService = ServiceContainer.Resolve<OrganizationService>("organizationService");
+            var policyService = ServiceContainer.Resolve<IPolicyService>("policyService");
+            
+            var organizations = await organizationService.GetAllAsync();
+            var personalOwnershipPolicyApplies =
+                    await policyService.PolicyAppliesToUser(PolicyType.PersonalOwnership);
+            return (organizations?.Any() ?? false) && !personalOwnershipPolicyApplies;
+        }
```
fedemkr, line 395: I personally don't like much the static methods on AppHelpers that can be placed on a service or somewhere else.
In this case I'd just add this method to the `OrganizationService`. I think when we add unit testing it's better to have these in non-static places in order to be able to mock them easier and we don't strictly need to resolve each service in each call to the method (as a matter of fact placing it in the `OrganizationService` saves us one `Resolve` call from the service container) 😄

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -424,8 +472,23 @@ public async Task SyncAsync()
...
+            }
+            else if (IsVaultFilterOrgVault)
+            {
+                orgId = await GetVaultFilterOrgIdAsync();
+                _allCiphers = decCiphers.Where(c => c.OrganizationId == orgId).ToList();
+            }
+            else
+            {
+                _allCiphers = decCiphers;
+            }
```
fedemkr, line 489: Can we extract it to another method? `LoadDataAsync` is IMO already too large, it could be something like:
`async Task<string> FillAllCiphersAndGetOrgIdIfNeededAsync();`

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -439,12 +502,27 @@ private async Task LoadDataAsync()
...
+                    Folders = decFolders;
+                    Collections = decCollections;
+                }
+                NestedFolders = await _folderService.GetAllNestedAsync(Folders);
                 HasFolders = NestedFolders.Any(f => f.Node?.Id != null);
-                Collections = await _collectionService.GetAllDecryptedAsync();
-                NestedCollections = await _collectionService.GetAllNestedAsync(Collections);
-                HasCollections = NestedCollections.Any();
+                NestedCollections = Collections != null ? await _collectionService.GetAllNestedAsync(Collections) : null;
+                HasCollections = NestedCollections?.Any() ?? false;
```
fedemkr, line 525: Can we extract it to another method? Same reason as before: `LoadDataAsync` is IMO already too large, it could be something like:
`async Task FillFoldersAndCollectionsAsync();`

fedemkr (changes requested):

src/App/Utilities/AppHelpers.cs
mp-bw: I ended up using `PolicyService` to prevent a circular dependency when constructing `OrganizationService`

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
mp-bw: I agree this line is... _less_ readable (let's go ahead and say unreadable).  In fact I renamed the var after wasting a lot of time caused by a misinterpretation.  Unfortunately your modified version returns a different result, and I _think_ it's because of how the second condition is handled in the unmodified version (`?? true`).  In any case I left this one alone for fear of breaking it.  I can revisit when I'm thinking clearly again.

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -76,6 +77,7 @@ public GroupingsPageViewModel()
                 await LoadAsync();
             });
             CipherOptionsCommand = new Command<CipherView>(CipherOptionsAsync);
+            VaultFilterCommand = new Command(async () => await VaultFilterOptionsAsync());
```
fedemkr, line 80: We should use the `AsyncCommand` from [Xamarin.Community.Toolkit](https://docs.microsoft.com/en-us/xamarin/community-toolkit/objectmodel/asynccommand), it has `onException` as a parameter that's useful to trigger the logger 😄

fedemkr (changes requested):

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
mp-bw: I tried to update the other Commands as well but ended up rolling those back due to some unusual events in the debug log (apparently further adjustments will be needed to make those work cleanly with `AsyncCommand`, skipping that for now since it's not part of this work)

fedemkr (approved):

PR #1929, opened 2022-05-27
PS-689 Android: Accessibility - back buttons in search and vault > bin lack appropriate accessible name

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Android back buttons throughout the app don't have accessibility text. Add text to be read by the VoiceOver\TalkBack on the back button.

## Code changes
* **src/Android/Renderers/CustomPageRenderer.cs:** Toolbar back button was only accessible through a custom renderer.

## Before you submit

### DIFF
--- a/src/Android/Renderers/CustomPageRenderer.cs
+++ b/src/Android/Renderers/CustomPageRenderer.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using Android.App;
+using Android.Content;
+using AndroidX.AppCompat.Widget;
+using Bit.App.Resources;
+using Bit.Droid.Renderers;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+
+[assembly: ExportRenderer(typeof(ContentPage), typeof(AndroidNavigationPageRenderer))]
+namespace Bit.Droid.Renderers
+{
+    public class AndroidNavigationPageRenderer : PageRenderer
+    {
+        public AndroidNavigationPageRenderer(Context context) : base(context)
+        {
+        }
+
+        protected override void OnElementChanged(ElementChangedEventArgs<Page> e)
+        {
+            base.OnElementChanged(e);
+
+            Activity context = (Activity)this.Context;
+            var toolbar = context.FindViewById<Toolbar>(Resource.Id.toolbar);
+            if(toolbar != null)
+            {
+                toolbar.NavigationContentDescription = AppResources.GoBack;
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/App/Resources/AppResources.resx
```diff
@@ -2214,4 +2214,7 @@
   <data name="SpecialCharacters" xml:space="preserve">
     <value>Special Characters (!@#$%^&amp;*)</value>
   </data>
+  <data name="GoBack" xml:space="preserve">
```
fedemkr, line 2217: IMO the key and the value should be the same, except on cases where it's too large, like descriptions. So in this case it'd be `TapToGoBack`. Furthermore, maybe we need "Go back" as a string in other place and a developer may use this string because the key is misleading, and also if we'd need "Go back" we couldn't use the key given that it'd be already in use by this.

src/Android/Renderers/CustomPageRenderer.cs
```diff
@@ -0,0 +1,31 @@
...
+
+            Activity context = (Activity)this.Context;
+            var toolbar = context.FindViewById<Toolbar>(Resource.Id.toolbar);
+            if(toolbar != null)
+            {
+                toolbar.NavigationContentDescription = AppResources.GoBack;
+            }
+        }
+    }
+}
```
fedemkr, line 31: I have to check, but isn't there a way to set this with an `Effect` or a custom renderer for the `Toolbar` instead of the whole page?

fedemkr (changes requested):

src/Android/Renderers/CustomPageRenderer.cs
andrebispo5: i'll take a look if it works with an effect or a `Toolbar` renderer.

andrebispo5: Toolbar support seems pretty weak. Didn't find a better solution to add the accessibility text to the toolbar.

patrickhlauke:
Note that if this is exposed as a button already, there's no need to add instructions on how to operate it. i.e. shouldn't need to say "Tap to go back", as that will be announced by TalkBack as "Tap to go back, button" ... which is both redundant, and inaccurate (as you'd double-tap). Long story short: if the control already exposes a proper role like button, just give it a name like "Back". Users will know how to operate it.

andrebispo5:
> Note that if this is exposed as a button already, there's no need to add instructions on how to operate it. i.e. shouldn't need to say "Tap to go back", as that will be announced by TalkBack as "Tap to go back, button" ... which is both redundant, and inaccurate (as you'd double-tap). Long story short: if the control already exposes a proper role like button, just give it a name like "Back". Users will know how to operate it. (I'll review this once it's released and suggest wording tweaks, as I currently don't have the setup to compile and test this)

Thanks for the feedback @patrickhlauke, i'll check and make changes taking into account your insight.

PR #1941, opened 2022-06-06
PS-587 - Fix searchbar search textfield and icon colors for dark themes

LRNcardozoWDF:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
With these changes, autofill iOS extension searchbar's search text and icon should appear as white for non light themes.

## Code changes
* **LoginSearchViewController.cs**: Invoked new method extension ApplyDarkThemesColors from UISearchBarExtensions.cs
* **UISearchBarExtensions.cs**: Applied white color to search text and icon only for iOS versions greater or equal than 13 because this solution isn't supported for prior versions.

## Screenshots
Without the fix:
image

With the fix:
Dark theme

image

Black theme
image

Nord theme
image

## Before you submit

### DIFF
--- a/src/iOS.Autofill/LoginSearchViewController.cs
+++ b/src/iOS.Autofill/LoginSearchViewController.cs
@@ -36,6 +36,7 @@ namespace Bit.iOS.Autofill
             if (!ThemeHelpers.LightTheme)
             {
                 SearchBar.KeyboardAppearance = UIKeyboardAppearance.Dark;
+                SearchBar.ApplyDarkThemesColors();
             }
 
             TableView.RowHeight = UITableView.AutomaticDimension;
--- a/src/iOS.Autofill/LoginSearchViewController.cs
+++ b/src/iOS.Autofill/LoginSearchViewController.cs
@@ -33,11 +33,7 @@ namespace Bit.iOS.Autofill
             CancelBarButton.Title = AppResources.Cancel;
             SearchBar.Placeholder = AppResources.Search;
             SearchBar.BackgroundColor = SearchBar.BarTintColor = ThemeHelpers.ListHeaderBackgroundColor;
-            if (!ThemeHelpers.LightTheme)
-            {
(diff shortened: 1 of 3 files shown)


src/iOS.Core/Utilities/UISearchBarExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Bit.iOS.Core.Utilities
+{
+    public static class UISearchBarExtensions
+    {
+        public static void ApplyDarkThemesColors(this UISearchBar searchBar)
+        {
+            var versionParts = UIDevice.CurrentDevice.SystemVersion.Split('.');
+            if (versionParts.Length > 0 && int.TryParse(versionParts[0], out var version))
+            {
+                if (version >= 13)
```
fedemkr, line 12: This can be done easier by doing: `if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))`

src/iOS.Autofill/LoginSearchViewController.cs
```diff
@@ -36,6 +36,7 @@ public async override void ViewDidLoad()
             if (!ThemeHelpers.LightTheme)
             {
                 SearchBar.KeyboardAppearance = UIKeyboardAppearance.Dark;
+                SearchBar.ApplyDarkThemesColors();
```
fedemkr, line 39: I'd change the name to `ApplyDarkTheme()`, given that then we may need to update other things and there should be all in the same method. Now it's the same but IMO theming should be always called in a general way and in the extension or helper class call any particular things for the theme; so the caller doesn't need to call multiple methods to apply the whole theme (e.g. ApplyDarkThemeColors, ApplyDarkThemeIcons, ApplyDarkThemeSeparators, etc)

fedemkr (changes requested):
Great work just a few improvements 😄

PR #1947, opened 2022-06-09
Clipboard handling adjustments for Android 13

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Android 13 introduces some changes to clipboard behavior when copying text:
https://developer.android.com/about/versions/13/features/copy-paste

This PR does the following:
* Prevent toasts when copying text on Android 13+ so as to not conflict with the new indicator
* Pass the `IS_SENSITIVE` flag to the `ClipboardManager` to prevent showing copied text in the new popup in Android 13+

_**Note 1: Currently this can't be tested in the emulator as the latest image (at least for ARM64) is still internally returning `32` for the API version (should be `33`).  I installed the June 8 beta on a Pixel 4a for testing.**_

_**Note 2: As of the latest beta released yesterday (June 8) the OS behavior doesn't line up with the documented behavior (Instead of obfuscated dots in the popup it shows "Tap to view" which then shows the copied text in a separate clipboard viewer).  I'm unsure if the behavior is still in flux, or the documentation hasn't caught up, or my implementation isn't correct. Regardless these changes prevent copied content from appearing in the clear within the popup.  I'll continue to observe each new release in case that changes before Android 13 goes live.**_

## Code changes

Android:
* **ClipboardService.cs:** Added `isSensitive` arg to `CopyTextAsync` method, defaults to `true`.  Added `CopyNotificationHandledByOs` method to establish if the OS has built-in support for notifying the user when something is copied to the clipboard, used to prevent conflicting toasts
* **DeviceActionService.cs (Android):** Added `ClipboardService` for TOTP copying with the new handler (removed private method)
* **MainApplication.cs:** Adjustments to service initialization

iOS:
* **ClipboardService.cs:** Added `isSensitive` arg to `CopyTextAsync` method (unused w/ comment) and `CopyNotificationHandledByOs` method (with comment about future use)

General:
* **All other files:** Check if `CopyNotificationHandledByOs` before showing toasts when copying text to the clipboard

## Before you submit

### DIFF
--- a/src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
+++ b/src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
@@ -1,7 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Threading.Tasks;
 using Bit.App.Resources;
-using Bit.App.Utilities;
 using Bit.Core.Abstractions;
 using Bit.Core.Models.Domain;
 using Bit.Core.Utilities;
@@ -59,7 +58,7 @@ namespace Bit.App.Pages
         private async void CopyAsync(GeneratedPasswordHistory ph)
         {
             await _clipboardService.CopyTextAsync(ph.Password);
-            AppHelpers.ShowToastForCopiedValue(_platformUtilsService, _clipboardService, AppResources.Password);
+            _platformUtilsService.ShowToastForCopiedValue(AppResources.Password);
         }
 
         public async Task UpdateOnThemeChanged()
--- a/src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
+++ b/src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
@@ -58,8 +58,11 @@ namespace Bit.App.Pages
         private async void CopyAsync(GeneratedPasswordHistory ph)
         {
             await _clipboardService.CopyTextAsync(ph.Password);
-            _platformUtilsService.ShowToast("info", null,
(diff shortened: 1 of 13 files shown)


src/App/Pages/Generator/GeneratorHistoryPageViewModel.cs
```diff
@@ -58,8 +58,11 @@ public async Task ClearAsync()
         private async void CopyAsync(GeneratedPasswordHistory ph)
         {
             await _clipboardService.CopyTextAsync(ph.Password);
-            _platformUtilsService.ShowToast("info", null,
+            if (!_clipboardService.CopyNotificationHandledByOs())
+            {
+                _platformUtilsService.ShowToast("info", null,
                 string.Format(AppResources.ValueHasBeenCopied, AppResources.Password));
+            }
```
fedemkr, line 65: Given that this is becoming a boilerplate and being used quite often, we could make an extension method for this:
```c#
public static void ShowToastForCopiedValue(this IPlatformUtilsService platformUtilsService, IClipboardService clipboardService, string valueNameCopied)
{
    if (!clipboardService.CopyNotificationHandledByOs())
    {
        platformUtilsService.ShowToast("info", null, string.Format(AppResources.ValueHasBeenCopied, valueNameCopied));
    }
}
```
So we can just use:
```c#
_platformUtilsService.ShowToastForCopiedValue(_clipboardService, AppResources.Password);
```
Or the other way is just to add the method to `IPlatformUtilsService`

fedemkr (approved):

PR #1951, opened 2022-06-13
[SG-79] [SG-389] Additional logic around vault filter display

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

- Hide the "My Vault" option if the Personal Ownership Policy is enabled in an organization
- Hide the vault filters altogether if the Personal Ownership & Single Org policies are both enabled in an organization

## Code changes

* **GroupingsPageViewModel/CiphersPageViewModel.cs:** Added `_hideMyVaultFilterOption` bool based on `_policyService.PolicyAppliesToUser(PolicyType.PersonalOwnership)`
* **PolicyService.cs:** Added additional policy checks in `ShouldShowVaultFilterAsync()` method

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/CiphersPageViewModel.cs
+++ b/src/App/Pages/Vault/CiphersPageViewModel.cs
@@ -46,9 +46,6 @@ namespace Bit.App.Pages
 
             Ciphers = new ExtendedObservableCollection<CipherView>();
             CipherOptionsCommand = new Command<CipherView>(CipherOptionsAsync);
-            VaultFilterCommand = new AsyncCommand(VaultFilterOptionsAsync,
-                onException: ex => _logger.Exception(ex),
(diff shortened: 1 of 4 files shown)


fedemkr (changes requested):
Could you create a `VaultFilterViewModel` to centralize the logic regarding vault filtering?
So that we can reuse that component on both `GroupingsPageViewModel` and `CiphersPageViewModel`. Currently, the logic is repeated and if issues occur the work needs to be done twice.

fedemkr (approved):
Approving to speed up QA, but we should move the `VaultFilterViewModel` to a component approach instead of a base class for both of the ViewModels. Also, the services that are now abstract should be passed through the base call constructor

PR #1952, opened 2022-06-15
[SG-386] iOS Update user state when coming from background

fedemkr:
## Type of change
- [X] Bug fix
- [X] New feature development
- [ ] Other

## Objective

On iOS, when coming from background to the app we should check if any extension has switched users, and if so, change the user as well on the app.

## Code changes

* **MainApplication.cs:** Updated `StateService` to include the messenger
* **App.xaml.cs:** Calls check for extension user id change when resuming the app (background to foreground)
* **AccountSwitchingOverlayViewModel.cs:** Added `FromIOSExtension` and if that's the case save the user id if the account has been switched. Also, refactored a bit the if to have less branches and applied the constants.
* **AccountsManager.cs:** Changed invoking on main thread to be awaited, thus the exception can bubble up
* **Constants.cs:** Added `iOSExtensionActiveUserIdKey` to be used to store the active user id on the extension if it has changed so that it can be updated on the main app when coming to foreground again.
* **StateService.cs:** Added `CheckExtensionActiveUserAndSwitchIfNeededAsync()` to do the actual verification and switch the account if necessary and the method to get/save the extension active user id.
* **AccountsManagerMessageCommands.cs:** Moved to `Core` project to be used in `StateService`
* **AccountSwitchingOverlayHelper.cs:** Applied the `FromIOSExtension` to `true` to enable saving the extension user id
* **iOSCoreHelpers.cs:** Updated `StateService` to include the messenger
* **MobileStorageService.cs:** Added `iOSExtensionActiveUserIdKey` to the preference storage keys

## Before you submit

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -25,6 +25,7 @@
         public static string iOSExtensionBiometricIntegrityKey = "iOSExtensionBiometricIntegrityState";
         public static string iOSShareExtensionClearCiphersCacheKey = "iOSShareExtensionClearCiphersCache";
         public static string iOSShareExtensionBiometricIntegrityKey = "iOSShareExtensionBiometricIntegrityState";
+        public static string iOSExtensionActiveUserIdKey = "iOSExtensionActiveUserId";
         public static string EventCollectionKey = "eventCollection";
         public static string RememberedEmailKey = "rememberedEmail";
         public static string RememberedOrgIdentifierKey = "rememberedOrgIdentifier";
--- a/src/Core/Services/StateService.cs
+++ b/src/Core/Services/StateService.cs
@@ -15,16 +15,20 @@ namespace Bit.Core.Services
     {
         private readonly IStorageService _storageService;
         private readonly IStorageService _secureStorageService;
+        private readonly IMessagingService _messagingService;
 
         private State _state;
         private bool _migrationChecked;
 
         public List<AccountView> AccountViews { get; set; }
 
-        public StateService(IStorageService storageService, IStorageService secureStorageService)
+        public StateService(IStorageService storageService,
+                            IStorageService secureStorageService,
+                            IMessagingService messagingService)
         {
             _storageService = storageService;
             _secureStorageService = secureStorageService;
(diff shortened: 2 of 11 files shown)


src/Core/Services/StateService.cs
```diff
@@ -67,6 +71,28 @@ public async Task SetActiveUserAsync(string userId)
...
 
+        public async Task CheckExtensionActiveUserAndSwitchIfNeededAsync()
+        {
+            var extensionUserId = await GetExtensionActiveUserIdFromStorageAsync();
+            if (string.IsNullOrEmpty(extensionUserId))
+            {
+                return;
+            }
+
+            if (_state?.ActiveUserId == extensionUserId)
```
mp-bw, line 82: Any reason not to use `GetActiveUserIdAsync()` here instead of `_state?.ActiveUserId`?

src/Core/Constants.cs
```diff
@@ -25,6 +25,7 @@ public static class Constants
         public static string iOSExtensionBiometricIntegrityKey = "iOSExtensionBiometricIntegrityState";
         public static string iOSShareExtensionClearCiphersCacheKey = "iOSShareExtensionClearCiphersCache";
         public static string iOSShareExtensionBiometricIntegrityKey = "iOSShareExtensionBiometricIntegrityState";
+        public static string iOSExtensionActiveUserIdKey = "iOSExtensionActiveUserId";
```
mp-bw, line 28: I would also add `Constants.iOSExtensionActiveUserIdKey` to `MobileStorageService._preferenceStorageKeys`  to store this in the native prefs storage instead of the main app DB.

src/Core/Services/StateService.cs
fedemkr: I don't want to get the user id from storage. If the extension changed the user then it also changed the active user id on storage thus the validation wouldn't make any sense because the storage and the extension user id would be the same.
What I want is to check whether the in memory user id (that is the one in place when sending the app to background to open another app that opens the extension) is different from the one that the extension changed.

mp-bw (approved):

PR #1953, opened 2022-06-15
[SG-390] Fix for missing org items with single org & personal ownership enabled

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This fixes an issue where org items were not appearing when the vault filter control was hidden due to the single org and personal ownership policies both being enabled.

## Code changes

* **VaultFilterViewModel.cs:** Update logic to handle mixed policy variations/changes better on initial load and refresh.  A delay is required before setting `ShowVaultFilter` to properly draw the org name in the vault filter if the filter was previously hidden but now shown after a refresh.

## Before you submit

### DIFF
--- a/src/App/Pages/VaultFilterViewModel.cs
+++ b/src/App/Pages/VaultFilterViewModel.cs
@@ -19,7 +19,8 @@ namespace Bit.App.Pages
         protected abstract ILogger logger { get; }
 
         protected bool _showVaultFilter;
-        protected bool _hideMyVaultFilterOption;
+        protected bool _personalOwnershipPolicyApplies;
+        protected bool _singleOrgPolicyApplies;
         protected string _vaultFilterSelection;
         protected List<Organization> _organizations;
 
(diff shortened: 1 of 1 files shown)


src/App/Pages/VaultFilterViewModel.cs
```diff
@@ -64,15 +65,21 @@ public string GetVaultFilterOrgId()
...
         {
             _organizations = await organizationService.GetAllAsync();
-            ShowVaultFilter = await policyService.ShouldShowVaultFilterAsync();
-            if (ShowVaultFilter)
+            if (_organizations?.Any() ?? false)
             {
-                _hideMyVaultFilterOption = await policyService.PolicyAppliesToUser(PolicyType.PersonalOwnership);
-                if (_vaultFilterSelection == null)
+                _personalOwnershipPolicyApplies = await policyService.PolicyAppliesToUser(PolicyType.PersonalOwnership);
+                _singleOrgPolicyApplies = await policyService.PolicyAppliesToUser(PolicyType.OnlyOrg);
```
fedemkr, line 71: `_singleOrgPolicyApplies` doesn't need to be a field, it can be a var or directly put the await call inside the if

fedemkr (changes requested):

fedemkr (approved):

PR #1954, opened 2022-06-15
[SG-392] Fix for vault filter selection carryover to secondary page

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

`InitVaultFilterAsync` wasn't being called on the secondary view page, resulting in missing data.  In order to do that from `GroupingsPage` the logic for setting the `ShowVaultFilter` property was broken out into its own method, and only called from the MainPage.  Both methods are called back to back on `CiphersPage` to maintain previous behavior.

## Code changes

* **GroupingsPageViewModel.cs:** Call `InitVaultFilterAsync` regardless of page, and `ShowVaultFilterIfNeededAsync` on MainPage only
* **CiphersPageViewModel.cs:** Call `InitVaultFilterAsync` and `ShowVaultFilterIfNeededAsync` back to back to maintain previous behavior
* **VaultFilterViewModel.cs:** Separate setting of `ShowVaultFilter` property to its own method `ShowVaultFilterIfNeededAsync`

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
+++ b/src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
@@ -181,9 +181,10 @@ namespace Bit.App.Pages
                 return;
             }
 
+            await InitVaultFilterAsync();
             if (MainPage)
             {
-                await InitVaultFilterAsync();
+                await ShowVaultFilterIfNeededAsync();
                 PageTitle = ShowVaultFilter ? AppResources.Vaults : AppResources.MyVault;
             }
 
--- a/src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
+++ b/src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
@@ -181,10 +181,9 @@ namespace Bit.App.Pages
                 return;
             }
 
-            await InitVaultFilterAsync();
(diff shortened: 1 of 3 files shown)


src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -181,9 +181,10 @@ public async Task LoadAsync()
                 return;
             }
 
+            await InitVaultFilterAsync();
             if (MainPage)
             {
-                await InitVaultFilterAsync();
+                await ShowVaultFilterIfNeededAsync();
                 PageTitle = ShowVaultFilter ? AppResources.Vaults : AppResources.MyVault;
             }
```
fedemkr, line 189: Instead of splitting this into different methods I'd just change the method to 
```c#
public async Task InitVaultFilterAsync(bool shouldUpdateShowVaultFilter)
{
    ...
    if (shouldUpdateShowVaultFilter)
    {
         await Task.Delay(100);
         ShowVaultFilter = await policyService.ShouldShowVaultFilterAsync();
    }
    ...
}
```

Then here the call would just be: `await InitVaultFilterAsync(MainPage);` and on `CiphersPageViewModel` would be `await InitVaultFilterAsync(true);`

fedemkr (approved):

PR #1967, opened 2022-06-29
[PS 920] Fix selfhosted url validations

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Added feedback to user when saving a bad formed URL in Settings page or trying to perform login with a previously saved bad formed url.

## Code changes

- **src/App/Pages/Accounts/EnvironmentPage.xaml**: Refactor to use AsyncCommand; 
- **src/App/Pages/Accounts/EnvironmentPage.xaml.cs**: Since AsyncCommand has proper validation to ensure multiple clicks aren't allowed I removed this validation;
- **src/App/Pages/Accounts/EnvironmentPageViewModel.cs**: Added validation for URLs saved and show pop up with error message;
- **src/App/Resources/AppResources.resx**: Added string with error message for when user tries to save a bad formed URL;
- **src/Core/Services/ApiService.cs**: Added validation for bad formed URL previously saved in Settings and show pop up with error. (Note: Since Bit.Core doesn't have access to src/App/Resources/AppResources.resx the message is hard coded for now).

## Before you submit


src/App/Pages/Accounts/EnvironmentPageViewModel.cs
```diff
@@ -57,5 +72,16 @@ public async Task SubmitAsync()
...
+        public bool FormValidation()
+        {
+            BaseUrlValid = string.IsNullOrEmpty(BaseUrl) || Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative);
+            ApiUrlValid = string.IsNullOrEmpty(ApiUrl) || Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative);
+            IdentityUrlValid = string.IsNullOrEmpty(IdentityUrl) || Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative);
+            WebVaultUrlValid = string.IsNullOrEmpty(WebVaultUrl) || Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative);
+            IconsUrlValid = string.IsNullOrEmpty(IconsUrl) || Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative);
+
+            return BaseUrlValid && ApiUrlValid && IdentityUrlValid && WebVaultUrlValid && IconsUrlValid;
+        }
```
fedemkr, line 85: A couple of things:
- Unless you're doing or plan to do something with these public properties, e.g. `BaseUrlValid`, I'd just use local variables inside this method. Right now they are being "cached" with no use.
- Also, you're using `Uri.IsWellFormedUriString(BaseUrl, UriKind.Relative )` on all cases, it should change to each url case, i.e. `ApiUrlValid`, `IdentityUrlValid`, and so on
- Also, I'd change the  name of the method to something more readable => `public bool ValidateUrls()` or `public bool AreUrlsValid()`
- Furtheremore, you could have a local function and just return using the && concatenation of the call to it with each url, i.e.

```c#
public bool ValidateUrls()
{
    bool IsUrlValid(string url)
    {
        return string.IsNullOrEmpty(url) || Uri.IsWellFormedUriString(url, UriKind.Relative);
    }

    return IsUrlValid(BaseUrl)
           &&
           IsUrlValid(ApiUrl)
           &&
           IsUrlValid(IdentityUrl)
           &&
           IsUrlValid(WebVaultUrl)
           &&
           IsUrlValid(IconsUrl);
}
```

fedemkr (changes requested):
A couple of things to improve and on the PR:
- Change the title to `[PS-920] Fix environment url validations` so the format is alike the other PRs.
- On the description, I'd clarify that `Settings` page is the environment's one or it could be misleading to the Settings tab in the app's home.

src/App/Pages/Accounts/EnvironmentPageViewModel.cs
```diff
@@ -22,10 +23,10 @@ public EnvironmentPageViewModel()
             IdentityUrl = _environmentService.IdentityUrl;
             IconsUrl = _environmentService.IconsUrl;
             NotificationsUrls = _environmentService.NotificationsUrl;
-            SubmitCommand = new Command(async () => await SubmitAsync());
+            SubmitCommand = new AsyncCommand(SubmitAsync, allowsMultipleExecutions: false);
```
fedemkr, line 26: Here for the `SubmitAsync` we need to add some kind of exception handling, at least to log the exception to AppCenter in case that happens; you can use `ILogger` for that and the parameter of `AsyncCommand` -> `onException:` (check example [here](https://github.com/bitwarden/mobile/blob/3aef86bd3402f21d2f14249dfa04f5f5e3a8e04a/src/App/Controls/AccountSwitchingOverlay/AccountSwitchingOverlayViewModel.cs#L26))
Also you can display a generic error to the user after logging it.

fedemkr (changes requested):
Good work, just one thing I forgot to mention on the first review, and even though the method didn't have exception handling before, we should add it.

fedemkr (approved):

fedemkr (approved):

PR #1973, opened 2022-07-04
[EC-295][BEEEP] Add Swipe to Copy on vault login items

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add swipe to copy username/password as a shortcut on vault list login items.
Swipe left to right to copy username and right to left to copy password.

## Code changes

* **IconFontImageSource.cs:** `FontImageSource` with the correct font type applied
* **CipherHelper.cs:** Added this class to move the static `CipherListOptions` on `AppHelpers` to be here so the class can be injected, is better prepared for unit testing and also to reuse copy username/password; furthermore now the services are just resolved once instead of every time the method is called. Now is called `ShowCipherOptionsAsync(...)`
* **GroupingsPage.xaml:** Added `DataTemplate` for login cipher item to have the swipe behavior
* **GroupingsPage.xaml.cs:** Added a helper property to calculate the threshold of the swipe depending on the width of the page.
* **GroupingsPageListItemSelector.cs:** Added logic to choose the swipe data template for login ciphers
* **GroupingsPageViewModel.cs:** Added commands and behavior for copy username/password and updated code to use `ICipherHelper` methods instead of `AppHelpers`
* **AutofillCiphersPageViewModel.cs:** Changed call to `AppHelpers.CipherListOptions` for the new one on `CipherHelper`
* **CiphersPageViewModel.cs:** Changed call to `AppHelpers.CipherListOptions` for the new one on `CipherHelper`
* **AppHelpers.cs:** Removed `CipherListOptions(...)` given that now is on `ICipherHelper`
* **iOSCoreHelpers.cs:** Register new `CipherHelper`
* **MainApplication.cs:** Register new `CipherHelper`
* **ExtendedGrid.cs:** Added `ApplyRipple` flag to control whether the grid needs to apply ripple
* **Android/Renderers/ExtendedGridRenderer.cs:** Added check for `ApplyRipple` if not it overrides the behavior of the swipe background

## Screenshots

iOS Behavior:
https://user-images.githubusercontent.com/15682323/177174276-6ee098b4-a936-4555-8e8b-66163fc81392.mov

iOS Dark Mode:
Swipe to copy iOS Dark

## Before you submit


fedemkr:
Closing this now as it's currently blocked by an [issue](https://github.com/xamarin/Xamarin.Forms/issues/14640) in Xamarin.Forms regarding scrolling and lists, and trying to workaround it would take much effort vs the value provided by it.

PR #2000, opened 2022-07-20
[PS-1080] Added text alternative to Boolean custom field icon

aj-rosado:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Screen Reader should be able to read if the Boolean icon value on custom fields is Enabled or Disabled

## Code changes

* **ViewPage.xaml**: Added `AutomationProperties.Name` accessibility property to the CustomField boolean IconLabel.
* **ViewPageViewModel**: Added a new property named `ValueAccessibilityText` that has the value to use on the xaml accessibility property.

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/ViewPage.xaml
+++ b/src/App/Pages/Vault/ViewPage.xaml
@@ -571,6 +571,7 @@
                                                 IsVisible="{Binding IsLinkedType}" />
                                             <controls:IconLabel
                                                 Text="{Binding ValueText, Mode=OneWay}"
+                                                AutomationProperties.Name="{Binding ValueAccessibilityText, Mode=OneWay}"
                                                 StyleClass="box-value"
                                                 Grid.Row="1"
                                                 Grid.Column="0"
--- a/src/App/Pages/Vault/ViewPage.xaml
+++ b/src/App/Pages/Vault/ViewPage.xaml
@@ -571,6 +571,7 @@
                                                 IsVisible="{Binding IsLinkedType}" />
                                             <controls:IconLabel
                                                 Text="{Binding ValueText, Mode=OneWay}"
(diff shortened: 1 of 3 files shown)


src/App/Pages/Vault/ViewPage.xaml
```diff
@@ -571,6 +571,7 @@
                                                 IsVisible="{Binding IsLinkedType}" />
                                             <controls:IconLabel
                                                 Text="{Binding ValueText, Mode=OneWay}"
+                                                AutomationProperties.Name="{Binding ValueAccessibilityText, Mode=OneWay}"
```
fedemkr, line 574: We need to add here `AutomationProperties.IsInAccessibleTree="true"` [(docs)](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/accessibility/automation-properties#automationpropertiesisinaccessibletree) as well so we don't have any weird issues on some devices

src/App/Pages/Vault/ViewPageViewModel.cs
```diff
@@ -764,6 +765,19 @@ public string ValueText
             }
         }
 
+        public string ValueAccessibilityText
+        {
+            get
+            {
+                if (IsBooleanType)
+                {
+                    return _field.Value == "true" ? AppResources.Enabled : AppResources.Disabled;
```
fedemkr, line 774: instead of comparing to that hardcoded string I'd use parse:
```c#
return bool.TryParse(_field.Value, out var b) && b ? AppResources.Enabled : AppResources.Disabled;
```

fedemkr (changes requested):
Great work, just a few improvements 😄

fedemkr (approved):

PR #2001, opened 2022-07-21
[PS-686] Change disabled/enabled to on/off

coroiu:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [x] Other

## Objective
Change disabled/enabled to on/off

## Code changes
* **AppResources.resx** Translations
* **SettingsPageViewModel.cs** Dynamic translation labels
* **SettingsPageListItem.cs** Match new labels to correct color

## Screenshots
image

image

## Before you submit


coroiu:
`src/App/Resources/AppResources.Designer.cs` is completely rewritten by VS Code for Mac. Not sure how I should deal with that

fedemkr (approved):
About `AppResources.Designer.cs` I don't see a problem on that for now, we can leave it as is. If a merge is needed we can just take one and then generate it again before pushing. Not quite sure if we can do something about it, like explicit format, I'll dig in on that (one way would be to get them out of the version control...)

PR #2011, opened 2022-07-25
[EC-341] Fix show alternative 2FA on iOS extensions

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

On iOS Extensions when 2FA view is presented, it should allow to change between different 2FA if setup.

## Code changes

* **TwoFactorPage.xaml:** Changed Toolbar item More Clicked event to Command
* **TwoFactorPage.xaml.cs:** Removed More_Clicked event and moved logic to VM command
* **TwoFactorPageViewModel.cs:** Added MoreCommand with the logic of the More_Clicked and changed it to use `DeviceActionService` instead of directly `Page` to show the action sheet in order for it to work on the extensions.
* **iOS.Autofill.csproj:** Added reference to the "more"/"three dots" images to get in the bundle
* **DeviceActionService.cs:** Improved null check of the ViewController, if that one was null on some cases the Task was not finishing given that it was not setting the result.
* **iOS.Extension.csproj:** Added reference to the "more"/"three dots" images to get in the bundle
* **iOS.ShareExtension.csproj:** Added reference to the "more"/"three dots" images to get in the bundle

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/TwoFactorPage.xaml
+++ b/src/App/Pages/Accounts/TwoFactorPage.xaml
@@ -17,15 +17,19 @@
         <ToolbarItem Text="{u:I18n Cancel}" Clicked="Close_Clicked" Order="Primary" Priority="-1"
                      x:Name="_cancelItem" />
     </ContentPage.ToolbarItems>
-
+            
     <ContentPage.Resources>
         <ResourceDictionary>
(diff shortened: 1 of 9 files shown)


jlf0dev (approved):
LGTM

PR #2017, opened 2022-07-28
[SG-471] Passwordless device login screen

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
- Screen which the user will land after clicking remote notification for device login request
- Screen will populate with information coming from the notification
- Will need server endpoint to accept/reject the login request
- Navigation to vault after accept/reject action
- Show information toast after server response from action

## Code changes

This PR is being merged into its feature branch in order to have smaller pull requests until we have the full feature to go to master.

* **src/App/Pages/Accounts/LoginPasswordlessViewModel.cs:** `AcceptRequestAsync()` and `RejectRequestAsync()` methods have mock service calls, this may change in the future based on service requirements.
* **src/Core/Services/AuthService.cs:** passwordless methods will be implemented in the future in another PR, for now they are has just mocks.

## Screenshots

image
image
image
image

## Before you submit


src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                var res = await _authService.LogInPasswordlessAcceptAsync();
+                await Page.Navigation.PopModalAsync();
+                _platformUtilsService.ShowToast("info", null, AppResources.LogInAccepted);
+            }
+            catch (Exception ex)
+            {
+                _logger.Exception(ex);
+            }
+        }
```
fedemkr, line 175: You already have the try...catch in the `AsyncCommand` so it's not needed here, same on `RejectRequestAsync`

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+            _logger = ServiceContainer.Resolve<ILogger>("logger");
+
+            PageTitle = AppResources.LogInRequested;
+
+            AcceptRequestCommand = new AsyncCommand(AcceptRequestAsync,
+                onException: ex => _logger.Exception(ex),
+                allowsMultipleExecutions: false);
+            RejectRequestCommand = new AsyncCommand(RejectRequestAsync,
+                onException: ex => _logger.Exception(ex),
+                allowsMultipleExecutions: false);
```
fedemkr, line 42: for the `onException`, besides logging you should also inform the user that an unexpected error has occurred. Otherwise the user won't have any feedback on the UI.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                _logger.Exception(ex);
+            }
+        }
+
+        private async Task RejectRequestAsync()
+        {
+            try
+            {
+                var res = await _authService.LogInPasswordlessRejectAsync();
```
fedemkr, line 181: I believe this will have an API call, so you should display a loading here. Same in `AcceptRequestAsync`

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            get => _fingerprintPhraseFormatted;
+            set => SetProperty(ref _fingerprintPhraseFormatted, value);
+        }
+
+        public string LogInAttempByLabel
+        {
+            get => _logInAttempByLabel;
+            set => SetProperty(ref _logInAttempByLabel, value);
+        }
```
fedemkr, line 81: Instead of having a full property here and assign it on the `Email`, this should be a `get` only property which returns the formatted string.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        public string TimeOfRequestText
+        {
+            get => _timeOfRequest;
+            set
+            {
+                SetProperty(ref _timeOfRequest, value);
+            }
+        }
```
fedemkr, line 118: Idem `LogInAttempByLabel`

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        public DateTime RequestDate
+        {
+            get => _requestDate;
+            set
+            {
+                TimeOfRequestText = CreateRequestDate();
+                SetProperty(ref _requestDate, value);
+            }
+        }
```
fedemkr, line 109: These and Origin, Email and Fingerprint phase should be in a class/struct so that it can be directly assigned from the Page and therefore we avoid having these many full properties here and it's easier to maintain

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+
+                fs.Spans.Add(new Span
+                {
+                    Text = "-",
+                    TextColor = ThemeManager.GetResourceColor("DangerColor")
+                });
+            }
+
+            return fs;
+        }
```
fedemkr, line 146: could `fingerprintPhrase` be `null`?
This method should have a try...catch, if an exception is raised then it'll crash the app
Also I'd extract `'-'` to a const in some place so it's not hardcoded and we can reuse it. Also if at some moment the separator is changed or can be changed by the user it'd be easier to maintain or modify.

You can also write this simpler with Linq Aggregate:
```c#
private FormattedString CreateFingerprintPhrase(string fingerprintPhrase)
{
    try
    {
        return fingerprintPhrase?.Split(new [] { SomeConsts.FingerprintPhraseSeparator }, StringSplitOptions.RemoveEmptyEntries).Aggregate(new FormattedString(), (fs, fingerprint) => 
        {
            if(fs.Spans.Any())
            {
                fs.Spans.Add(new Span
                {
                    Text = SomeConsts.FingerprintPhraseSeparator,
                    TextColor = ThemeManager.GetResourceColor("DangerColor")
                });
            }

            fs.Spans.Add(new Span
            {
                Text = fingerprint
            });
        });
    }
    catch (Exception ex)
    {
        _logger.Exception(ex);
        return null;
    }
}
```

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,191 @@
...
+        private string CreateRequestDate()
+        {
+            var minutesSinceRequest = RequestDate.ToUniversalTime().Minute - DateTime.UtcNow.Minute;
+            if (minutesSinceRequest < 5)
+            {
+                return AppResources.JustNow;
+            }
+            if (minutesSinceRequest < 59)
+            {
+                return $"{minutesSinceRequest} {AppResources.MinutesAgo}";
```
fedemkr, line 157: Not all languages have the format "x minutes ago", so this should be a string format:
`string.Format(AppResources.XMinutesAgo, minutesSinceRequest)`
and in the AppResources `XMinutesAgo` would be: `{0} minutes ago`

src/App/Resources/AppResources.resx
```diff
@@ -2275,4 +2275,46 @@
...
   </data>
+  <data name="LogInRequested" xml:space="preserve">
+    <value>Log in requested</value>
+  </data>
+  <data name="AreYouTryingToLogIn" xml:space="preserve">
+    <value>Are you trying to log in?</value>
+  </data>
+  <data name="LogInAttemptByOn" xml:space="preserve">
+    <value>Log in attempt by {0} on {1}</value>
+  </data>
```
fedemkr, line 2286: In order be easier to guess, I try to use XYZ letters to indicate that it needs parameters, so I'd change it to:
`LogInAttemptByXOnY`

src/Core/Abstractions/IAuthService.cs
```diff
@@ -25,6 +25,11 @@ public interface IAuthService
         Task<AuthResult> LogInSsoAsync(string code, string codeVerifier, string redirectUrl, string orgId);
         Task<AuthResult> LogInCompleteAsync(string email, string masterPassword, TwoFactorProviderType twoFactorProvider, string twoFactorToken, bool? remember = null);
         Task<AuthResult> LogInTwoFactorAsync(TwoFactorProviderType twoFactorProvider, string twoFactorToken, string captchaToken, bool? remember = null);
+
+        Task<AuthResult> GetLogInPasswordlessRequestsAsync();
+        Task<AuthResult> LogInPasswordlessAcceptAsync();
+        Task<AuthResult> LogInPasswordlessRejectAsync();
```
fedemkr, line 31: I think it'd be better to have the verb as the prefix to be easier to read so:
`LogInPasswordlessAcceptAsync` => `AceptLogInPasswordlessAsync`
`LogInPasswordlessRejectAsync ` => `RejectLogInPasswordlessAsync`

src/App/Resources/AppResources.resx
```diff
@@ -2275,4 +2275,46 @@
...
+  <data name="DenyLogIn" xml:space="preserve">
+    <value>Deny Log In</value>
+  </data>
+  <data name="JustNow" xml:space="preserve">
+    <value>Just Now</value>
+  </data>
+  <data name="MinutesAgo" xml:space="preserve">
+    <value>minutes ago</value>
+  </data>
```
fedemkr, line 2313: Change this as to the format way, as explained in `LoginPasswordlessViewModel`

fedemkr (changes requested):
Great work, I've left some improvements you can make. Lmk what you think

src/App/Pages/Accounts/LoginPasswordlessPage.xaml
```diff
@@ -0,0 +1,86 @@
...
+        </ResourceDictionary>
+    </ContentPage.Resources>
+
+    <ScrollView Padding="7, 0, 7, 20">
+        <StackLayout>
+            <Label
+                Text="{u:I18n AreYouTryingToLogIn}"
+                FontSize="Title"
+                FontAttributes="Bold"
+                Margin="0,14,0,21"/>
```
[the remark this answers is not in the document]
andrebispo5: yes, for labels the default behaviour of screen readers is to ready the text itself. We have to be careful with controls that have state or are not implicit like toggle switches for example.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: you are right!

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: it was supposed to always come and have that structure but it's prettier with the linq aggregate 👍

src/Core/Abstractions/IAuthService.cs
andrebispo5: This will probably be merged into one method with different arguments, it will certainly change in the future. But to be coherent with the other methods in the service i'll leave it like this.

src/App/Pages/Accounts/LoginPasswordlessPage.xaml.cs
```diff
@@ -0,0 +1,42 @@
...
+            _vm.LoginRequest = new LoginPasswordlessDetails()
+            {
+                Origin = origin,
+                Email = email,
+                DeviceType = deviceType,
+                IpAddress = ipAddress,
+                NearLocation = location,
+                FingerprintPhrase = fingerprintPhrase,
+                RequestDate = requestDate,
+            };
```
fedemkr, line 26: What I meant on the previous request was to change the constructor so it's more direct and easier to maintain if one parameter is added/modified/removed, i.e.:
```c#
public LoginPasswordlessPage(LoginPasswordlessDetails loginRequest)
{
    InitializeComponent();
    _vm = BindingContext as LoginPasswordlessViewModel;
    _vm.Page = this;
    _vm.LoginRequest = loginRequest;
```
In this manner, if anything changes, you only have to change the `LoginPasswordlessDetails` and update the bindings and that's it. In the current way you'd have to change the constructor, the assignment here and the bindings and any place that is calling the constructor

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        private void HandleServerErrorException(Exception ex)
+        {
+            Xamarin.Essentials.MainThread.BeginInvokeOnMainThread(async () =>
+            {
+                await _deviceActionService.HideLoadingAsync();
+                await _platformUtilsService.ShowDialogAsync(AppResources.GenericErrorMessage);
+            });
+            _logger.Exception(ex);
+        }
```
fedemkr, line 136: I'd change the name, the exception could not be necessary from server, maybe the `ShowLoading` throws and that wouldn't be server error. And Server errors are handled by the `ApiException` if you wanted some custom behavior there.
Aside from that if an exception happens inside the invoke on main thread it will crash the application because of the context. If you need to do that, either add a try...catch internally or use the awaitable version of that which bubbles up the exception, i.e. `MainThread. InvokeOnMainThreadAsync(Func<Task>)`
Btw, is the invoke on main thread really needed here?

src/App/Resources/AppResources.resx
```diff
@@ -2275,4 +2275,46 @@
...
+  <data name="DenyLogIn" xml:space="preserve">
+    <value>Deny Log In</value>
+  </data>
+  <data name="JustNow" xml:space="preserve">
+    <value>Just Now</value>
+  </data>
+  <data name="MinutesAgo" xml:space="preserve">
+    <value>{0} minutes ago</value>
+  </data>
```
fedemkr, line 2313: To keep the format and better guess that it takes a parameter, please add X as a prefix in the name, i.e. `<data name="XMinutesAgo" xml:space="preserve">`

src/Core/Constants.cs
```diff
@@ -29,6 +29,7 @@ public static class Constants
         public static string EventCollectionKey = "eventCollection";
         public static string RememberedEmailKey = "rememberedEmail";
         public static string RememberedOrgIdentifierKey = "rememberedOrgIdentifier";
+        public static string FingerprintPhraseSeparator = "-";
```
fedemkr, line 32: I'd change that to a `const` instead of being `static`

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessPage.xaml.cs
andrebispo5: oh ye it makes total sense.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: Had to add the invoke on main thread because if you try to call the hide loading it will throw.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -0,0 +1,157 @@
...
+            _platformUtilsService.ShowToast("info", null, AppResources.LogInDenied);
+        }
+
+        private void HandleException(Exception ex)
+        {
+            Xamarin.Essentials.MainThread.InvokeOnMainThreadAsync(async () =>
+            {
+                await _deviceActionService.HideLoadingAsync();
+                await _platformUtilsService.ShowDialogAsync(AppResources.GenericErrorMessage);
+            });
```
fedemkr, line 134: Even though this won't crash, we need to log the exception if something happens inside. Right now if `_platformUtilsService.ShowDialogAsync` throws the exception, it will be swollen by the `Task`.
You can try adding `FireAndForget()` at the end and check if it works as expected (it should work but maybe I'm not considering something and it changes the thread execution).

fedemkr (changes requested):

fedemkr (approved):

PR #2027, opened 2022-08-04
[SG-416] Authenticator toggle remake 

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

After demo the authenticator feature, it was decided that the way we were showing the TOTP codes needed a rework. If the user is premium, It now has a vault entry that navigates to a new page only for TOTP codes.

## Code changes

* **src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs:306:**
When the user only has one Cipher with TOTP and removes the configuration in the edit page, when he comes back from edit, it will be redirected to the main vault page.

## Screenshots

image
image

## Before you submit

### DIFF
--- a/src/Android/Properties/AndroidManifest.xml
+++ b/src/Android/Properties/AndroidManifest.xml
@@ -11,7 +11,6 @@
 	<uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
     <uses-permission android:name="android.permission.VIBRATE" />
 	<uses-permission android:name="com.samsung.android.providers.context.permission.WRITE_USE_APP_FEATURE_SURVEY" />
-	<uses-permission android:name="android.permission.VIBRATE" />
 	<uses-feature android:name="android.hardware.camera" android:required="false" />
 	<uses-feature android:name="android.hardware.camera.autofocus" android:required="false" />
 	<application android:label="Bitwarden" android:theme="@style/LaunchTheme" android:allowBackup="false" tools:replace="android:allowBackup" android:icon="@mipmap/ic_launcher" android:roundIcon="@mipmap/ic_launcher_round" android:networkSecurityConfig="@xml/network_security_config">
--- a/src/Android/Properties/AndroidManifest.xml
+++ b/src/Android/Properties/AndroidManifest.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools" android:versionCode="1" android:versionName="2022.6.3" android:installLocation="internalOnly" package="com.x8bit.bitwarden">
 	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="30" />
 	<uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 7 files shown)


src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
```diff
@@ -222,6 +223,10 @@ private async void RowSelected(object sender, SelectionChangedEventArgs e)
             {
                 await _vm.SelectTrashAsync();
             }
+            else if (item.IsTotpCode)
+            {
+                await _vm.SelectTotpCodesAsync();
+            }
```
fedemkr, line 229: Given this is called from `async void RowSelected`, if an exception is raised it will crash the app. IMO, this should be refactored to have a command that takes the `item` and in the VM calls the right method.
For now, at least a try...catch should be added to the whole `RowSelected` to avoid potential crashes and to log the exception

fedemkr (changes requested):

src/Android/Properties/AndroidManifest.xml
```diff
@@ -9,8 +9,8 @@
 	<uses-permission android:name="android.permission.USE_BIOMETRIC" />
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 	<uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
+    <uses-permission android:name="android.permission.VIBRATE" />
 	<uses-permission android:name="com.samsung.android.providers.context.permission.WRITE_USE_APP_FEATURE_SURVEY" />
-  	<uses-permission android:name="android.permission.VIBRATE" />
```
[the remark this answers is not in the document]
andrebispo5: Both vscode and vsm were showing the manifest formatted correctly, fixed 👍

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
```diff
@@ -202,46 +202,53 @@ protected override async void OnDisappearing()
...
+                {
+                    await _vm.SelectTypeAsync(item.Type.Value);
+                }
             }
-            else if (item.Type != null)
+            catch (Exception ex)
             {
-                await _vm.SelectTypeAsync(item.Type.Value);
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
             }
```
fedemkr, line 252: we should also let the user know that something went wrong, if not they might think that the action takes a lot to complete or something like that

src/App/Pages/Vault/GroupingsPage/GroupingsPageViewModel.cs
```diff
@@ -394,6 +382,23 @@ public async Task LoadAsync()
             }
         }
 
+        private void AddTotpGroupItem(bool canAccessPremium, List<GroupingsPageListGroup> groupedItems, bool uppercaseGroupNames)
+        {
+            if (canAccessPremium && (TOTPCiphers?.Any() ?? false))
```
fedemkr, line 387: I think that `(TOTPCiphers?.Any() ?? false)` can be written as `TOTPCiphers?.Any() == true` to be less confusing to read and you can remove the parenthesis

fedemkr (changes requested):

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs
andrebispo5: true, i'll add a generic error dialog.

fedemkr (approved):

PR #2037, opened 2022-08-15
[SG-381] Passwordless - Add setting to Mobile

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add a setting to the Bitwarden mobile app under the “Security” section: “Approve login requests”
Default state is Disabled.
When user Enables the option, check to see if the Bitwarden app has notification permissions. If not, request them. Even if the user does not enable notifications, the Approve login requests feature is enabled.

## Code changes

* **src/iOS/Services/iOSPushNotificationService.cs:** Added new method because iOS has different approach on notifications status. I needed a way to tell if the user had blocked the "alert notifications" in order to redirect to the settings.
 
## Screenshots

New entry in the security section of settings tab:
image
Dialog to provide additional information and enable/disable the feature:
image
If the user doesn't have notification "Alert" enable, show this dialog to open settings directly:
image

## Before you submit

### DIFF
--- a/src/Android/Services/DeviceActionService.cs
+++ b/src/Android/Services/DeviceActionService.cs
@@ -964,5 +964,14 @@ namespace Bit.Droid.Services
             }
             activity.RunOnUiThread(() => activity.Window.AddFlags(WindowManagerFlags.Secure));
         }
+
+        public void OpenAppSettings()
+        {
+            var intent = new Intent(Android.Provider.Settings.ActionApplicationDetailsSettings);
+            intent.AddFlags(ActivityFlags.NewTask);
+            var uri = Android.Net.Uri.FromParts("package", Application.Context.PackageName, null);
+            intent.SetData(uri);
+            Application.Context.StartActivity(intent);
+        }
     }
 }
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -348,7 +348,7 @@ namespace Bit.App.Pages
             await _stateService.SetApprovePasswordlessLoginsAsync(_approvePasswordlessLoginRequests);
             if(_approvePasswordlessLoginRequests && !_pushNotificationService.IsRegisteredForPush)
             {
-                var openAppSettingsResult = await _platformUtilsService.ShowDialogAsync(AppResources.OpenSettingsAlert,title: "", confirmText: AppResources.OpenSettings, cancelText: AppResources.Cancel);
+                var openAppSettingsResult = await _platformUtilsService.ShowDialogAsync(AppResources.ReceivePushNotificationsForNewLoginRequests, title: string.Empty, confirmText: AppResources.Settings, cancelText: AppResources.NoThanks);
 
                 if (openAppSettingsResult)
                 {
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -337,7 +337,7 @@ namespace Bit.App.Pages
                     CreateSelectableOption(AppResources.No, !_approvePasswordlessLoginRequests),
             };
 
-            var selection = await Page.DisplayActionSheet(AppResources.SubmitCrashLogsDescription, AppResources.Cancel, null, options);
+            var selection = await Page.DisplayActionSheet(AppResources.UseThisDeviceToApproveLoginRequests, AppResources.Cancel, null, options);
 
             if (selection == null || selection == AppResources.Cancel)
             {
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -346,18 +346,26 @@ namespace Bit.App.Pages
 
             _approvePasswordlessLoginRequests = CompareSelection(selection, AppResources.Yes);
             await _stateService.SetApprovePasswordlessLoginsAsync(_approvePasswordlessLoginRequests);
-            if(_approvePasswordlessLoginRequests && !_pushNotificationService.IsRegisteredForPush)
-            {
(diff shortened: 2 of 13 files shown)


src/Android/Services/DeviceActionService.cs
```diff
@@ -964,5 +964,14 @@ public async Task SetScreenCaptureAllowedAsync()
...
         }
+
+        public void OpenAppSettings()
+        {
+            var intent = new Intent(Android.Provider.Settings.ActionApplicationDetailsSettings);
+            intent.AddFlags(ActivityFlags.NewTask);
+            var uri = Android.Net.Uri.FromParts("package", Application.Context.PackageName, null);
+            intent.SetData(uri);
+            Application.Context.StartActivity(intent);
+        }
```
fedemkr, line 975: Could you check if we can also use this method on `OpenAccessibilityOverlayPermissionSettings()` in the same file? it seems to be opening the app settings as well in there so maybe we could reuse this method there as well

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -83,6 +84,7 @@ public SettingsPageViewModel()
             _keyConnectorService = ServiceContainer.Resolve<IKeyConnectorService>("keyConnectorService");
             _clipboardService = ServiceContainer.Resolve<IClipboardService>("clipboardService");
             _loggerService = ServiceContainer.Resolve<ILogger>("logger");
+            _pushNotificationService = ServiceContainer.Resolve<IPushNotificationService>("pushNotificationService");
```
fedemkr, line 87: there's no need for the string key anymore

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -326,6 +329,45 @@ public async Task LoggerReportingAsync()
...
+            {
+                return;
+            }
+
+            var openAppSettingsResult = await _platformUtilsService.ShowDialogAsync(AppResources.ReceivePushNotificationsForNewLoginRequests, title: string.Empty, confirmText: AppResources.Settings, cancelText: AppResources.NoThanks);
+            if (openAppSettingsResult)
+            {
+                _deviceActionService.OpenAppSettings();
+            }
+
```
fedemkr, line 368: Remove this empty line

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -326,6 +329,45 @@ public async Task LoggerReportingAsync()
...
+            if (notificationsSettingsEnabled)
+            {
+                return;
+            }
+
+            var openAppSettingsResult = await _platformUtilsService.ShowDialogAsync(AppResources.ReceivePushNotificationsForNewLoginRequests, title: string.Empty, confirmText: AppResources.Settings, cancelText: AppResources.NoThanks);
+            if (openAppSettingsResult)
+            {
+                _deviceActionService.OpenAppSettings();
+            }
```
fedemkr, line 367: you can put everything in one if:
```c#
if (_approvePasswordlessLoginRequests
    &&
    !await _pushNotificationService.GetNotificationsSettingsEnabledAsync()
    &&
    await _platformUtilsService.ShowDialogAsync(AppResources.ReceivePushNotificationsForNewLoginRequests, title: string.Empty, confirmText: AppResources.Settings, cancelText: AppResources.NoThanks))
{
    _deviceActionService.OpenAppSettings();
}
```
or at least the first two if you think the third condition is two lengthy to put it on the if:
```c#
if (!_approvePasswordlessLoginRequests
    ||
    await _pushNotificationService.GetNotificationsSettingsEnabledAsync())
{
    return;
}
```

src/App/Resources/AppResources.resx
```diff
@@ -2317,4 +2317,19 @@
   <data name="LogInDenied" xml:space="preserve">
     <value>Log in denied</value>
   </data>
+  <data name="ApproveLoginRequests" xml:space="preserve">
+    <value>Approve login requests</value>
+  </data>
+  <data name="UseThisDeviceToApproveLoginRequests" xml:space="preserve">
+    <value>Use this device to approve login requests made from other devices.</value>
+  </data>
```
fedemkr, line 2325: IMO, the key should be the same as the value unless very lengthy value, so here is not that long so: `UseThisDeviceToApproveLoginRequestsMadeFromOtherDevices `

src/iOS/Services/iOSPushNotificationService.cs
```diff
@@ -19,6 +19,15 @@ public Task<string> GetTokenAsync()
...
         public bool IsRegisteredForPush => UIApplication.SharedApplication.IsRegisteredForRemoteNotifications;
 
+        public async Task<bool> GetNotificationsSettingsEnabledAsync()
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            UNUserNotificationCenter.Current.GetNotificationSettings((settings) => {
+                tcs.SetResult(settings.AlertSetting == UNNotificationSetting.Enabled);
+            });
+            return await tcs.Task;
+        }
```
fedemkr, line 29: Use the async version of the method: `UNUserNotificationCenter.Current.GetNotificationSettingsAsync()` so that you don't have to use the `TaskCompletionSource<bool>`.
And just FYI, in here you don't gain anything by async awaiting so the `async` is not needed and you should return directly `tcs.Task`.
Nonetheless, it's better to use directly the async version of the method:
``` c#
public async Task<bool> GetNotificationsSettingsEnabledAsync()
{
    var settings = await UNUserNotificationCenter.Current.GetNotificationSettingsAsync();
    return settings.AlertSetting == UNNotificationSetting.Enabled;
}
```

fedemkr (changes requested):

src/Android/Services/DeviceActionService.cs
andrebispo5: It's not the same thing, it is android specific and it's to add "on top" permissions.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
andrebispo5: i'll change for the second option. I usually don't like to include method lengthy calls into if statements but i guess it cleans up a couple of lines in this case.

fedemkr (approved):

PR #2038, opened 2022-08-15
[PS-1009] Changed keyboard on Passphrase generator to not allow emojis

aj-rosado:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Remove the possibility of adding emojis on the word separator at Generator when passphrase is selected

## Code changes

* **NoEmojiKeyboardEffect.cs:** Created effect with settings to not show emojis on the keyboard, both on iOS and Android.
* **GeneratorPage.cs:** Added NoEmojiKeyboardEffect to the passphrase generator Entry

## Screenshots

Screenshot_1660582307
Simulator Screen Shot - iPhone 11 Pro Max - 2022-08-15 at 17 52 36

## Before you submit

### DIFF
--- a/src/Android/Effects/NoEmojiKeyboardEffect.cs
+++ b/src/Android/Effects/NoEmojiKeyboardEffect.cs
@@ -1,17 +1,17 @@
-﻿using Bit.Droid.Effects;
+﻿using Android.Widget;
+using Bit.Droid.Effects;
 using Xamarin.Forms;
 using Xamarin.Forms.Platform.Android;
 
-[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), "NoEmojiKeyboardEffect")]
+[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), nameof(NoEmojiKeyboardEffect))]
 namespace Bit.Droid.Effects
 {
     public class NoEmojiKeyboardEffect : PlatformEffect
     {
         protected override void OnAttached()
         {
-            if (Element is Entry && Control is FormsEditText editText)
+            if (Control != null && Control is EditText editText)
             {
-                editText.ImeOptions = Android.Views.InputMethods.ImeAction.Done;
                 editText.InputType = Android.Text.InputTypes.ClassText | Android.Text.InputTypes.TextVariationVisiblePassword | Android.Text.InputTypes.TextFlagMultiLine;
             }
         }
--- a/src/Android/Effects/NoEmojiKeyboardEffect.cs
+++ b/src/Android/Effects/NoEmojiKeyboardEffect.cs
@@ -1,0 +1,24 @@
+﻿using Bit.Droid.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+
+[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), "NoEmojiKeyboardEffect")]
+namespace Bit.Droid.Effects
+{
+    public class NoEmojiKeyboardEffect : PlatformEffect
+    {
+        protected override void OnAttached()
+        {
+            if (Element is Entry && Control is FormsEditText editText)
+            {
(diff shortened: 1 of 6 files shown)


src/Android/Effects/NoEmojiKeyboardEffect.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Bit.Droid.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+
+[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), "NoEmojiKeyboardEffect")]
```
fedemkr, line 5: Use `nameof(NoEmojiKeyboardEffect)`, in case the name of the effect changes the refactor will be easier

src/Android/Effects/NoEmojiKeyboardEffect.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class NoEmojiKeyboardEffect : PlatformEffect
+    {
+        protected override void OnAttached()
+        {
+            if (Element is Entry && Control is FormsEditText editText)
+            {
+                editText.ImeOptions = Android.Views.InputMethods.ImeAction.Done;
+                editText.InputType = Android.Text.InputTypes.ClassText | Android.Text.InputTypes.TextVariationVisiblePassword | Android.Text.InputTypes.TextFlagMultiLine;
+            }
```
fedemkr, line 16: Why do you need to check whether if the `Element` is an `Entry`? I'd just check the `Control` so that the effect can be applied to anything that has a control that is `FormsEditText`. As a matter of fact, I'd just change it to the base class `EditText`.
Aside from that, why does this effect changes the `ImeOptions`? IMO, this is like adding a "hidden" change in the effect because unless you read this code you don't know that it also changes it. Furthermore, you can control the `ImeOptions` [directly on the xaml](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/android/entry-ime-options) so it's not needed here.

src/App/Pages/Generator/GeneratorPage.xaml
```diff
@@ -128,7 +129,12 @@
                                 Text="{Binding WordSeparator}"
                                 IsSpellCheckEnabled="False"
                                 IsTextPredictionEnabled="False"
-                                StyleClass="box-value" />
+                                StyleClass="box-value"
+                                >
```
fedemkr, line 133: Inline the `>`, i.e. `StyleClass="box-value">`

fedemkr (changes requested):

fedemkr:
Forgot to add it to the PR review comments but please update the PR objective

src/Android/Effects/NoEmojiKeyboardEffect.cs
aj-rosado: Removed the ImeOptions as they are not necessary for this

src/Android/Effects/NoEmojiKeyboardEffect.cs
```diff
@@ -1,17 +1,17 @@
...
-[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), "NoEmojiKeyboardEffect")]
+[assembly: ExportEffect(typeof(NoEmojiKeyboardEffect), nameof(NoEmojiKeyboardEffect))]
 namespace Bit.Droid.Effects
 {
     public class NoEmojiKeyboardEffect : PlatformEffect
     {
         protected override void OnAttached()
         {
-            if (Element is Entry && Control is FormsEditText editText)
+            if (Control != null && Control is EditText editText)
```
fedemkr, line 13: By telling `Control is EditText`, if it's then it's not `null` so the first condition can be removed

fedemkr (changes requested):

fedemkr (approved):

PR #2053, opened 2022-08-29
[PS-191] Improve support for larger fonts Android

aj-rosado:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
The fonts on Android are too big on some devices. Removed Android native scaling as xamarin `NamedSize` is already scaling the fonts.

## Code changes

* **MainActivity.cs:** Set the native Android font scaling to 1. So that it doesn't scale as the Xamarin `NamedSize` is already scaling the text

## Screenshots

Max font size with Android scaling:
scaling

Max font size without Android scaling:
no-scaling

## Before you submit

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -38,6 +38,7 @@ namespace Bit.Droid
         private IStateService _stateService;
         private IAppIdService _appIdService;
         private IEventService _eventService;
+        private ILogger _logger;
         private PendingIntent _eventUploadPendingIntent;
         private AppOptions _appOptions;
         private string _activityKey = $"{nameof(MainActivity)}_{Java.Lang.JavaSystem.CurrentTimeMillis().ToString()}";
@@ -59,6 +60,7 @@ namespace Bit.Droid
             _stateService = ServiceContainer.Resolve<IStateService>("stateService");
             _appIdService = ServiceContainer.Resolve<IAppIdService>("appIdService");
             _eventService = ServiceContainer.Resolve<IEventService>("eventService");
+            _logger = ServiceContainer.Resolve<ILogger>("logger");
 
             TabLayoutResource = Resource.Layout.Tabbar;
             ToolbarResource = Resource.Layout.Toolbar;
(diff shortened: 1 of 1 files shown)


src/Android/MainActivity.cs
```diff
@@ -18,7 +19,9 @@
 using Bit.Core.Utilities;
 using Bit.Droid.Receivers;
 using Bit.Droid.Utilities;
+using Xamarin.Essentials;
 using ZXing.Net.Mobile.Android;
+using FileProvider = AndroidX.Core.Content.FileProvider;
```
fedemkr, line 24: I don't see this being used, if that's the case remove it

fedemkr (changes requested):

src/Android/MainActivity.cs
aj-rosado: When added the Xamarin.Essentials the FileProvider now had 2 references. Xamarin.Essentials and AndroidX.Core.Content. Keeping it as the one that was being used before

src/Android/MainActivity.cs
```diff
@@ -401,5 +404,12 @@ private async Task StopEventAlarmAsync()
             alarmManager.Cancel(_eventUploadPendingIntent);
             await _eventService.UploadEventsAsync();
         }
+
+        private void DisableAndroidFontScale()
+        {
+            //As we are using NamedSizes the xamarin will change the font size. So we are disabling the Android scaling.
+            Resources.Configuration.FontScale = 1f;
+            BaseContext.Resources.DisplayMetrics.ScaledDensity = Resources.Configuration.FontScale * (float)DeviceDisplay.MainDisplayInfo.Density;
+        }
```
fedemkr, line 413: Not 100% sure if this is safe enough that doesn't throw exception at some case or device specifics. Just in case, could you wrap this in a try...catch that logs the exception to AppCenter? So that if something happens the app doesn't crash and the users can use the app nonetheless given that you're calling it from `OnCreate`

fedemkr (changes requested):

fedemkr (approved):

PR #2056, opened 2022-08-31
[SG-472] Add configuration for new notification type

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Configure the mobile app to be able to receive and process a new notification type associated with passwordless login. 
Added configuration to be able to show local notifications to the user and dismiss them. 

## Code changes

* **src/App/App.xaml.cs:**
Added subscription in the broadcast service for `unlock` and `passwordlessLoginRequest`. Every time the user unlocks its vault, if there is any pending request, then the app should show the accept/decline modal. If the user in within the app and a new passwordless login notification arrives, it will broadcast a `passwordlessLoginRequest` message, showing the  the accept/decline modal.
* **src/Android/Services/IPushNotificationService.cs:**  
`SendLocalNotification` `DismissLocalNotification` added methods to the interface and following platform implementation. These methods are used to manage device notifications.  

## Screenshots
image

## Before you submit

### DIFF
--- a/src/Android/Services/AndroidPushNotificationService.cs
+++ b/src/Android/Services/AndroidPushNotificationService.cs
@@ -58,16 +58,23 @@ namespace Bit.Droid.Services
 
         public void DismissLocalNotification(string notificationId)
         {
-            var notificationManager = NotificationManagerCompat.From(Android.App.Application.Context);
-            notificationManager.Cancel(int.Parse(notificationId));
+            if (int.TryParse(notificationId, out int intNotificationId))
+            {
+                var notificationManager = NotificationManagerCompat.From(Android.App.Application.Context);
+                notificationManager.Cancel(intNotificationId);
+            }
         }
 
-        public void SendLocalNotification(string title, string message, string notificationId = null)
+        public void SendLocalNotification(string title, string message, string notificationId)
         {
-            notificationId = notificationId ?? new Random().Next(1000, 999999).ToString();
+            if (string.IsNullOrEmpty(notificationId))
+            {
+                throw new ArgumentNullException("notificationId cannot be null or empty.");
+            }
+            
             var context = Android.App.Application.Context;
             var intent = new Intent(context, typeof(MainActivity));
-            var pendingIntent = PendingIntent.GetActivity(context, new Random().Next(1000, 999999), intent, PendingIntentFlags.OneShot);
+            var pendingIntent = PendingIntent.GetActivity(context, 20220801, intent, PendingIntentFlags.OneShot);
             var builder = new NotificationCompat.Builder(context, Constants.AndroidNotificationChannelId)
                .SetContentIntent(pendingIntent)
                .SetContentTitle(title)
@@ -75,10 +82,8 @@ namespace Bit.Droid.Services
                .SetSmallIcon(Resource.Mipmap.ic_launcher)
                .SetAutoCancel(true);
 
-            // Build the notification:
-            var notification = builder.Build();
             var notificationManager = NotificationManagerCompat.From(context);
-            notificationManager.Notify(int.Parse(notificationId), notification);
+            notificationManager.Notify(int.Parse(notificationId), builder.Build());
         }
     }
 }
--- a/src/Android/Services/AndroidPushNotificationService.cs
+++ b/src/Android/Services/AndroidPushNotificationService.cs
@@ -1,8 +1,11 @@
 ﻿#if !FDROID
 using System;
 using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
 using AndroidX.Core.App;
 using Bit.App.Abstractions;
+using Bit.Core;
 using Bit.Core.Abstractions;
 using Xamarin.Forms;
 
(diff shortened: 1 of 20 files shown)


src/App/Abstractions/IPushNotificationService.cs
```diff
@@ -9,5 +10,7 @@ public interface IPushNotificationService
         Task<string> GetTokenAsync();
         Task RegisterAsync();
         Task UnregisterAsync();
+        void SendLocalNotification(string title, string message, string notificationId = null);
```
fedemkr, line 13: I think we could avoid having a `null` default value on `notificationId`. For now at least is not being used and I see that you're generating a random number for the id in the implementations, the problem with that is that the id you generate won't be known by the caller so the caller won't be able to dismiss the notification.
So for now I think we don't need the `null`, and if it's null or empty in the implementation you can just throw an `ArgumentNullException`. If in the future we need this we should change the method to return the newly created `notificationId`

src/Android/Services/AndroidPushNotificationService.cs
```diff
@@ -52,6 +55,31 @@ public Task UnregisterAsync()
             // Do we ever need to unregister?
             return Task.FromResult(0);
         }
+
+        public void DismissLocalNotification(string notificationId)
+        {
+            var notificationManager = NotificationManagerCompat.From(Android.App.Application.Context);
+            notificationManager.Cancel(int.Parse(notificationId));
+        }
```
fedemkr, line 63: I'd add a try parse check here to avoid potential future crashes because if the `notificationId` is `null` or something that is not an `int` here the Parse will crash:
```c#
public void DismissLocalNotification(string notificationId)
{
    if (int.TryParse(notificationId, out int intNotificationId)
    {
        var notificationManager = NotificationManagerCompat.From(Android.App.Application.Context);
        notificationManager.Cancel(intNotificationId);
    }
}
```

src/Android/Services/AndroidPushNotificationService.cs
```diff
@@ -52,6 +55,31 @@ public Task UnregisterAsync()
...
+
+        public void SendLocalNotification(string title, string message, string notificationId = null)
+        {
+            notificationId = notificationId ?? new Random().Next(1000, 999999).ToString();
+            var context = Android.App.Application.Context;
+            var intent = new Intent(context, typeof(MainActivity));
+            var pendingIntent = PendingIntent.GetActivity(context, new Random().Next(1000, 999999), intent, PendingIntentFlags.OneShot);
```
fedemkr, line 70: I believe you can use a const `requestCode` for `GetActivity` or did you have any issues with being it constant?

src/App/App.xaml.cs
```diff
@@ -140,6 +136,10 @@ public App(AppOptions appOptions)
                                 new NavigationPage(new RemoveMasterPasswordPage()));
                         });
                     }
+                    else if (message.Command == "passwordlessLoginRequest" || message.Command == "unlocked")
+                    {
+                        Device.BeginInvokeOnMainThread(() => CheckPasswordlessLoginRequestsAsync().FireAndForget());
```
fedemkr, line 141: I don't think you need the whole method to be invoked on the main thread, do you?. Could you check if you can move that to the specific part where you need it inside `CheckPasswordlessLoginRequestsAsync`? Beware that if you use that inside the async method use `Device.InvokeOnMainThreadAsync` so the inner exceptions bubble up.

fedemkr (changes requested):

src/iOS/AppDelegate.cs
```diff
@@ -304,6 +304,7 @@ public void InitApp()
             iOSCoreHelpers.InitLogger();
             _pushHandler = new iOSPushNotificationHandler(
                 ServiceContainer.Resolve<IPushNotificationListenerService>("pushNotificationListenerService"));
+            UNUserNotificationCenter.Current.Delegate = _pushHandler;
```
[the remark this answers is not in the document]
andrebispo5: You are totally right. Was testing some stuff out and forgot to delete this.

src/iOS/Services/iOSPushNotificationHandler.cs
```diff
@@ -83,8 +83,6 @@ private static string DictionaryToJson(NSDictionary dictionary)
         public void WillPresentNotification(UNUserNotificationCenter center, UNNotification notification, Action<UNNotificationPresentationOptions> completionHandler)
         {
             Debug.WriteLine($"{TAG} WillPresentNotification {notification?.Request?.Content?.UserInfo}");
-
-            OnMessageReceived(notification?.Request?.Content?.UserInfo);
```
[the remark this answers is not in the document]
andrebispo5: oh shoot! Good catch this was by mistake! i actually wanted to delete on the response method below, to avoid infinite looping, but since i decided not to inject any info in the local notification, it won't trigger the `OnMessageReceived` anyway.

src/iOS/Services/iOSPushNotificationService.cs
```diff
@@ -64,5 +67,36 @@ public Task UnregisterAsync()
...
+            };
+
+            notificationId = notificationId ?? new Random().Next(1000, 999999).ToString();
+
+            var request = UNNotificationRequest.FromIdentifier(notificationId, content, null);
+            UNUserNotificationCenter.Current.AddNotificationRequest(request, (err) =>
+            {
+                if (err != null)
+                {
+                    throw new Exception($"Failed to schedule notification: {err}");
```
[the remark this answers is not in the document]
andrebispo5: i will log it to appcenter instead.

src/Android/Services/AndroidPushNotificationService.cs
andrebispo5: Yes the intent is always the same and we actually only the keep the most up to date, so we can use the `requestCode` as a constant.

src/App/App.xaml.cs
```diff
@@ -173,7 +173,7 @@ private async Task CheckPasswordlessLoginRequestsAsync()
                 DeviceType = loginRequestData.RequestDeviceType
             });
             await _stateService.SetPasswordlessLoginNotificationAsync(null);
-            await Application.Current.MainPage.Navigation.PushModalAsync(new NavigationPage(page));
+            Device.BeginInvokeOnMainThread(async () => await Application.Current.MainPage.Navigation.PushModalAsync(new NavigationPage(page)));
```
fedemkr, line 176: Use `await Device.InvokeOnMainThreadAsync(...)` instead. It allows bubbling up the exceptions and prevents crashing the app due to the async void code block inside that could potentially throw an exception. Here is the implementation of [InvokeOnMainThreadAsync](https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Core/Device.cs#L109) if you want to look at it.

fedemkr (changes requested):

fedemkr (approved):

PR #2079, opened 2022-09-16
[SG-169] Passwordless demo change requests

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Change requests from async demo of passwordless feature.

## Code changes

Forced modal to appear after app came into foreground. Android was having erratic behaviour when the user had the app in background and clicked the received notification.

## Screenshots

image
image

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -166,6 +166,8 @@ namespace Bit.App
             var loginRequestData = await _authService.GetPasswordlessLoginRequestByIdAsync(notification.Id);
             var page = new LoginPasswordlessPage(new LoginPasswordlessDetails()
             {
+                PubKey = loginRequestData.PublicKey,
+                Id = loginRequestData.Id,
                 IpAddress = loginRequestData.RequestIpAddress,
                 Email = await _stateService.GetEmailAsync(),
                 FingerprintPhrase = loginRequestData.RequestFingerprint,
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -31,6 +31,7 @@ namespace Bit.App
         private readonly IAccountsManager _accountsManager;
         private readonly IPushNotificationService _pushNotificationService;
         private static bool _isResumed;
+        private static bool _pendingCheckPasswordlessLoginRequests;
 
         public App(AppOptions appOptions)
         {
@@ -152,16 +153,18 @@ namespace Bit.App
 
         private async Task CheckPasswordlessLoginRequestsAsync()
         {
+            if (!_isResumed)
+            {
+                _pendingCheckPasswordlessLoginRequests = true;
+                return;
+            }
+
+            _pendingCheckPasswordlessLoginRequests = false;
             if (await _vaultTimeoutService.IsLockedAsync())
             {
                 return;
             }
 
-            // Wait for the app to be in the foreground to show the pop up modal
-            while (!_isResumed)
-            {
-                await Task.Delay(500);
-            }
 
             var notification = await _stateService.GetPasswordlessLoginNotificationAsync();
             if (notification == null)
@@ -205,6 +208,10 @@ namespace Bit.App
                     SyncIfNeeded();
                 }
             }
+            if (_pendingCheckPasswordlessLoginRequests)
+            {
+                CheckPasswordlessLoginRequestsAsync().FireAndForget();
+            }
             if (Device.RuntimePlatform == Device.Android)
             {
                 await _vaultTimeoutService.CheckVaultTimeoutAsync();
@@ -237,6 +244,10 @@ namespace Bit.App
         {
             System.Diagnostics.Debug.WriteLine("XF App: OnResume");
             _isResumed = true;
+            if (_pendingCheckPasswordlessLoginRequests)
+            {
+                CheckPasswordlessLoginRequestsAsync().FireAndForget();
+            }
             if (Device.RuntimePlatform == Device.Android)
             {
                 ResumedAsync().FireAndForget();
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -31,6 +31,7 @@ namespace Bit.App
         private readonly IAccountsManager _accountsManager;
         private readonly IPushNotificationService _pushNotificationService;
         private static bool _isResumed;
+        // this variable is static because the app is launching new activities on notification click, creating new instances of App. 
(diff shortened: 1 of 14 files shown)


src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -45,11 +43,9 @@ public LoginPasswordlessViewModel()
 
         public ICommand RejectRequestCommand { get; }
 
-        public FormattedString FingerprintPhraseFormatted => CreateFingerprintPhrase(LoginRequest.FingerprintPhrase);
+        public string LogInAttemptByLabel => string.Format(AppResources.LogInAttemptByXOnY, LoginRequest?.Email, LoginRequest?.Origin);
```
fedemkr, line 46: If `LoginRequest` is `null` then the string would not make much sense: "Login attempt by on".
So if it can be `null` I'd show a different string or put a loading indicator if it needs to wait to load or show an error or empty if it's just because of the order on the bindings that it's `null` the first time and then it completes the second time the property is changed.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -60,56 +56,30 @@ public LoginPasswordlessDetails LoginRequest
...
             _platformUtilsService.ShowToast("info", null, approveRequest ? AppResources.LogInAccepted : AppResources.LogInDenied);
         }
 
-        private string CreateRequestDate(DateTime requestDate)
+        private string CreateRequestDate(DateTime? requestDate)
         {
-            var minutesSinceRequest = requestDate.ToUniversalTime().Minute - DateTime.UtcNow.Minute;
+            if (!requestDate.HasValue)
+            {
+                return String.Empty;
```
fedemkr, line 79: Use lowercase `string`

src/App/App.xaml.cs
```diff
@@ -155,6 +157,12 @@ private async Task CheckPasswordlessLoginRequestsAsync()
                 return;
             }
 
+            // Wait for the app to be in the foreground to show the pop up modal
+            while (!_isResumed)
+            {
+                await Task.Delay(500);
+            }
```
fedemkr, line 164: I'd like to avoid having this Delay thing, what do you think of having something like using a flag to indicate that there is a pending operation to check the passwordless login request and trigger this method again?

```c#
// maybe it needs to be volatile, I'm not sure if there could be a thread update issue, not at first sight but wanted to point it out just in case
private bool _pendingCheckPasswordlessLoginRequests;

...

private async Task CheckPasswordlessLoginRequestsAsync()
{
    if(!_isResumed)
    {
        return;
    }

    _pendingCheckPasswordlessLoginRequests = false;

    if (await _vaultTimeoutService.IsLockedAsync())
    {
        return;
    }

...

else if (message.Command == "passwordlessLoginRequest" || message.Command == "unlocked")
{
    _pendingCheckPasswordlessLoginRequests = true;
    CheckPasswordlessLoginRequestsAsync().FireAndForget();
}

...

protected async override void OnStart()
{
    System.Diagnostics.Debug.WriteLine("XF App: OnStart");
    _isResumed = true;
    if(_pendingCheckPasswordlessLoginRequests)
    {
        CheckPasswordlessLoginRequestsAsync().FireAndForget();
    }

...

protected override void OnResume()
{
    System.Diagnostics.Debug.WriteLine("XF App: OnResume");
    _isResumed = true;
    // dunno if it's necessary here
    if(_pendingCheckPasswordlessLoginRequests)
    {
        CheckPasswordlessLoginRequestsAsync().FireAndForget();
    }
```

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessPage.xaml
```diff
@@ -19,62 +19,66 @@
...
-                FontSize="Title"
-                FontAttributes="Bold"
-                Margin="0,14,0,21"/>
-            <Label
+    <StackLayout 
+            Padding="7, 0, 7, 20">
+        <ScrollView
+            VerticalOptions="FillAndExpand">
+            <StackLayout
+                VerticalOptions="FillAndExpand">
```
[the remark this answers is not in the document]
andrebispo5: Was moving things around and trying different approaches to avoid nesting and forgot about it. In the end needed to add one more level to the tree anyway.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: it should have value, will make it empty until the data is loaded and property raised.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -43,12 +43,14 @@ public LoginPasswordlessViewModel()
 
         public ICommand RejectRequestCommand { get; }
 
-        public string LogInAttemptByLabel => string.Format(AppResources.LogInAttemptByXOnY, LoginRequest?.Email, LoginRequest?.Origin);
+        public string LogInAttemptByLabel => hasEmailAndOrigin ? string.Format(AppResources.LogInAttemptByXOnY, LoginRequest.Email, LoginRequest.Origin) : string.Empty;
```
fedemkr, line 46: Can you have a `LoginRequest` that doesn't have an `Email` or `Origin` and then those properties are updated?
Because unless that scenario could happen I think that here you could have just 
`public string LogInAttemptByLabel => LoginRequest != null ? string.Format(...) : string.Empty;`
and remove `hasEmailAndOrigin`, btw if you were to keep the property, it needs first letter uppercase `HasEmailAndOrigin`

src/App/App.xaml.cs
```diff
@@ -31,6 +31,7 @@ public partial class App : Application, IAccountsManagerHost
         private readonly IAccountsManager _accountsManager;
         private readonly IPushNotificationService _pushNotificationService;
         private static bool _isResumed;
+        private static bool _pendingCheckPasswordlessLoginRequests;
```
fedemkr, line 34: why is this `static`?

fedemkr (changes requested):

src/App/App.xaml.cs
andrebispo5: otherwise the value gets lost on sleep on android.

PR #2097, opened 2022-09-23
[SG-671] OTP Menu Screen causes Crash on Android

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

User was not able to enter the ´Verification codes´ screen because the app was crashing. The crash was due to otpauth url having a bad format making the function to calculate the period throw an exception.

## Code changes

Added protection against exception on the calc of the period, defaulting it to 30 seconds rotation.
**src/App/Pages/Vault/GroupingsPage/GroupingsPageTOTPListItem.cs:** removed calculation and variable that wasn't being used.

## Before you submit


src/Core/Utilities/CoreHelpers.cs
```diff
@@ -183,24 +184,31 @@ public static Uri GetUri(string uriString)
...
+                    {
+                        dict.Add(key, parts[1] == null ? string.Empty : System.Net.WebUtility.UrlDecode(parts[1]));
+                    }
                 }
             }
+            catch (Exception ex)
+            {
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
+            }
             return dict;
```
fedemkr, line 212: I wonder if [HttpUtility.ParseQueryString(...) ](https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.parsequerystring?redirectedfrom=MSDN&view=net-6.0#System_Web_HttpUtility_ParseQueryString_System_String_) would be the same for this.

Just to check that it's specific for query strings so you can't pass the full uri or there may some problems.
```c#
string uriString = "https://www.google.com/search?q=bitwarden&oq=bitwarden&aqs=chrome.0.02&sourceid=chrome&ie=UTF-8";
var queryStringParams = uriString.Split("?").Last();
var queryStringNameValueCollection = HttpUtility.ParseQueryString(queryStringParams);
```

andrebispo5: Tried it out and it's for query strings, if you pass the full uri it will not parse the first item well.
To be honest i couldn't find the reason why the client wasn't able to parse the query params. Tried different combination to break it but without success. Ended up protecting with a try catch so at least he could see the 'verification codes' screen, and the rotation value default to 30s.

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -191,9 +191,9 @@ public static Uri GetUri(string uriString)
                 var queryStringNameValueCollection = HttpUtility.ParseQueryString(queryStringParams);
                 foreach(var key in queryStringNameValueCollection.AllKeys)
                 {
-                    if (!dict.ContainsKey(key))
+                    if (key!= null && !dict.ContainsKey(key))
                     {
-                        dict.Add(key, queryStringNameValueCollection[key] == null ? string.Empty : System.Net.WebUtility.UrlDecode(queryStringNameValueCollection[key]));
+                        dict.Add(key, queryStringNameValueCollection[key]);
                     }
                 }
```
fedemkr, line 198: As it states in the docs, `ParseQueryString` already takes care of repeating keys that are consolidated into one entry on the `NameValueCollection` with comma separated values, so the if check is not necessary and there can't be `null` keys. So this can be written as:
```c#
return queryStringNameValueCollection.AllKeys.ToDictionary(k => k, k => queryStringNameValueCollection[k]);
```
Then remove creating the dict at the top and return an empty Dictionary in the catch just in case

fedemkr (changes requested):

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -182,26 +184,17 @@ public static Uri GetUri(string uriString)
 
         public static Dictionary<string, string> GetQueryParams(string urlString)
         {
-            var dict = new Dictionary<string, string>();
-            if (!Uri.TryCreate(urlString, UriKind.Absolute, out var uri) || string.IsNullOrWhiteSpace(uri.Query))
```
fedemkr, line 186: I'd re-add this check again to check that the uri is valid and if not just early return an empty Dict

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -182,26 +184,17 @@ public static Uri GetUri(string uriString)
...
+            try
             {
-                return dict;
+                var queryStringParams = urlString.Split('?').Last();
+                var queryStringNameValueCollection = HttpUtility.ParseQueryString(queryStringParams);
```
fedemkr, line 190: Sorry I missed the above uri creation, but we could even simplify this to use 
```c#
var queryStringNameValueCollection = HttpUtility.ParseQueryString(uri.Query);
```
I made a few tests and it seems to work, keep in mind that `uri.Query` includes a "?" at beginning.

fedemkr (changes requested):

PR #2102, opened 2022-09-27
Added accessibility service disclosure prompt for Android

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Now that we're targeting API 32 we're required to comply with [this policy regarding non-accessibility apps using accessibility services](https://support.google.com/googleplay/android-developer/answer/10964491).  This PR adds the disclosure and consent prompt as described on the help page.

## Code changes

* **AutofillServicesPage.xaml/xaml.cs/ViewModel.cs:** Added support for showing disclosure dialog when the user enables the accessibility switch, and requires acceptance before continuing to the Android accessibility settings screen to enable the service.
* **accessibilityservice.xml:** Added `isAccessibilityTool="false"` per documentation (already defaults to `false` but added for completeness)

## Before you submit


src/App/Pages/Settings/AutofillServicesPage.xaml
```diff
@@ -83,7 +83,7 @@
                             StyleClass="box-value"
                             HorizontalOptions="End" />
                         <Button
-                            Clicked="ToggleAccessibility"
+                            Clicked="ToggleAccessibilityAsync"
```
fedemkr, line 86: Can we move this to a command?

src/App/Pages/Settings/AutofillServicesPage.xaml.cs
```diff
@@ -55,11 +55,11 @@ private void ToggleInlineAutofill(object sender, EventArgs e)
...
 
-        private void ToggleAccessibility(object sender, EventArgs e)
+        private async void ToggleAccessibilityAsync(object sender, EventArgs e)
         {
             if (DoOnce())
             {
-                _vm.ToggleAccessibility();
+                await _vm.ToggleAccessibilityAsync();
             }
         }
```
fedemkr, line 64: This can be removed if we take the command approach. Furthermore, if we keep it like this it will crash the app if an exception is thrown.

src/App/Pages/Settings/AutofillServicesPageViewModel.cs
```diff
@@ -176,8 +178,19 @@ public void ToggleInlineAutofill()
...
+            if (!_deviceActionService.AutofillAccessibilityServiceRunning())
+            {
+                var accept = await _platformUtilsService.ShowDialogAsync(AppResources.AccessibilityDisclosureText,
+                    AppResources.AccessibilityDisclosureTitle, AppResources.AcceptAccessibilityDisclosure,
+                    AppResources.DeclineAccessibilityDisclosure);
+                if (accept)
+                {
+                    _deviceActionService.OpenAccessibilitySettings();
+                }
+                return;
```
fedemkr, line 192: You can simplify it by:
```c#
if (!accept)
{
    return;
}
```
and if it's accepted it will just use the call outside the if to open the a11y settings

fedemkr (changes requested):

fedemkr (approved):

PR #2108, opened 2022-09-30
SG-687 Request time not updating

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If you wait more than 5 minutes to open the login request, then the “Time” field should say “# minutes ago” instead of “Just now”. However, it still says “Just Now”.

## Code changes

Added `TaskTimer` to update the values with a period of 5 mins. After 15mins the request expires showing a dialog.

## Before you submit


src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -63,8 +70,30 @@ public LoginPasswordlessDetails LoginRequest
...
+        {
+            TriggerPropertyChanged(nameof(TimeOfRequestText));
+            if (DateTime.UtcNow > LoginRequest?.RequestDate.ToUniversalTime().AddMinutes(15))
+            {
+                StopRequestTimeUpdater();
+                await _platformUtilsService.ShowDialogAsync(AppResources.LoginRequestHasAlreadyExpired);
+                await Page.Navigation.PopModalAsync();
+                return;
+            }
+        }
```
fedemkr, line 92: Given that this is `async void` if an exception is thrown inside, it'll crash the application so a try...catch should be added here or convert this method to be a `Task` and either add the possibility of passing a `Func<Task>` to the `TimerTask` or just do a `FireAndForget` of the task depending on the desired behavior.
Also the `return` is not necessary given that it's the end of the method

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -63,8 +70,30 @@ public LoginPasswordlessDetails LoginRequest
...
+
         private async Task PasswordlessLoginAsync(bool approveRequest)
         {
+            StopRequestTimeUpdater();
```
fedemkr, line 96: If this method fails throwing an exception at some point shouldn't the timer start again?

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -63,8 +70,30 @@ public LoginPasswordlessDetails LoginRequest
...
+        {
+            _requestTimeCts?.Cancel();
+            _requestTimeCts = new CancellationTokenSource();
+            _requestTimeTask = new TimerTask(_logger, UpdateRequestTime, _requestTimeCts).RunPeriodic(TimeSpan.FromMinutes(5));
+        }
+
+        private async void UpdateRequestTime()
+        {
+            TriggerPropertyChanged(nameof(TimeOfRequestText));
+            if (DateTime.UtcNow > LoginRequest?.RequestDate.ToUniversalTime().AddMinutes(15))
```
fedemkr, line 85: Could you extract the `15` as a constant in this class?

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -63,8 +70,30 @@ public LoginPasswordlessDetails LoginRequest
             }
         }
 
+        public void StopRequestTimeUpdater() => _requestTimeCts?.Cancel();
+
+        private void StartRequestTimeUpdater()
+        {
+            _requestTimeCts?.Cancel();
+            _requestTimeCts = new CancellationTokenSource();
+            _requestTimeTask = new TimerTask(_logger, UpdateRequestTime, _requestTimeCts).RunPeriodic(TimeSpan.FromMinutes(5));
```
fedemkr, line 79: Could you extract the `5` as a constant in this class? so if one wants to change the frequency they don't have to read the code and just update the constant.
And since you're here, could you change the parameter of the `TimerTask` to be the `CancellationToken` instead of the `CancellationTokenSource`? There is no need to cancel the request from inside the `TimerTask` so the token should suffice.
[Quoting](https://stackoverflow.com/a/54386551) the great Jon Skeet:

> The rule of thumb is:
> - Any code that needs to cancel the token should have access to the CancellationTokenSource
> - Any code that needs to respond to cancellation should have access to the CancellationToken

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -86,17 +115,12 @@ private string CreateRequestDate(DateTime? requestDate)
                 return string.Empty;
             }
 
-            var minutesSinceRequest = requestDate.Value.ToUniversalTime().Minute - DateTime.UtcNow.Minute;
-            if (minutesSinceRequest < 5)
+            if (DateTime.UtcNow < requestDate.Value.ToUniversalTime().AddMinutes(5))
```
fedemkr, line 118: Could you extract the `5` as a constant in this class? (or a general one if the `Just now` is being used elsewhere so that it standardizes it)

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -63,8 +70,30 @@ public LoginPasswordlessDetails LoginRequest
             }
         }
 
+        public void StopRequestTimeUpdater() => _requestTimeCts?.Cancel();
```
fedemkr, line 73: You should dispose the `CancellationTokenSource` when you finish using it

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: Adding the possibility to pass a Task to the `TimerTask` makes more sense.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: Moved to the end of the method so that problem doesn't happen.

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
```diff
@@ -70,30 +72,32 @@ public LoginPasswordlessDetails LoginRequest
...
+            _requestTimeCts?.Cancel();
+            _requestTimeCts?.Dispose();
+        }
 
         private void StartRequestTimeUpdater()
         {
             _requestTimeCts?.Cancel();
             _requestTimeCts = new CancellationTokenSource();
-            _requestTimeTask = new TimerTask(_logger, UpdateRequestTime, _requestTimeCts).RunPeriodic(TimeSpan.FromMinutes(5));
+            _requestTimeTask = new TimerTask(_logger, UpdateRequestTime, _requestTimeCts.Token).RunPeriodic(TimeSpan.FromMinutes(REQUEST_TIME_UPDATE_PERIOD));
```
fedemkr, line 85: I see that this task is getting cancelled only when the user taps on "Cancel".
Can the user go away from this view in some other way? Like swiping back on iOS? if so you should also cancel this if not there will be a constant waste of resources.
OTOH, you could use `OnAppearing()` and `OnDisappearing()` methods to start/stop the timer if that suits your needs

src/App/Utilities/TimerTask.cs
```diff
@@ -10,15 +10,23 @@ public class TimerTask
     {
         private readonly ILogger _logger;
         private readonly Action _action;
-        private readonly CancellationTokenSource _cancellationToken;
+        private readonly Func<Task> _func;
```
fedemkr, line 13: I'd rename this to `_asyncAction` or `_actionTask` or something like that, func is not very descriptive IMHO

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessViewModel.cs
andrebispo5: had it on close and accept / deny request, but will also add it on `OnDisappearing` just to be sure.

src/App/Pages/Accounts/LoginPasswordlessPage.xaml.cs
```diff
@@ -24,8 +24,21 @@ private async void Close_Clicked(object sender, System.EventArgs e)
         {
             if (DoOnce())
             {
+                _vm.StopRequestTimeUpdater();
```
fedemkr, line 27: If I'm not mistaken we don't need this anymore given that closing the view will call `OnDisappearing` but test it just in case I'm missing something

fedemkr (changes requested):

fedemkr (approved):

PR #2109, opened 2022-10-03
[SG-666][SG-667] Email is not prefilled and username isn't generated automatically

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
With this changes email is now prefilled when selecting Plus Adressed Email username type, the username is generated automatically upon navigation and if the chosen item has an URI we pass it correctly as a field for Plus Adressed Email or Catch-all Email username types.

## Code changes
* **GeneratorPageViewModel.cs:** Now we fetch the user's email and generate automatically the username for all username types except UsernameType.ForwardedEmailAlias.
* **CipherAddEditPageViewModel.cs:** Fixed the field sent to the generator page.

## Before you submit


src/App/Pages/Vault/CipherAddEditPageViewModel.cs
```diff
@@ -605,6 +605,8 @@ public async Task GenerateUsernameAsync()
                 return;
             }
 
+            var website = (bool)(Cipher?.Login?.HasUris) ? Cipher?.Login?.Uris[0].Host : null;
```
fedemkr, line 608: If `Cipher` or `Login` is `null` then when you cast it to `bool` it will throw an exception, so it should be something like:
```c#
var website = (Cipher?.Login?.HasUris ?? false) ? Cipher.Login.Uris[0].Host : null;
```
or another even better way:
```c#
var website = Cipher?.Login?.Uris?.FirstOrDefault()?.Host;
```

src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -564,26 +567,28 @@ public bool IncludeNumberRandomWordUsername
 
         public UsernameEmailType PlusAddressedEmailTypeSelected
         {
-            get => _plusAddressedEmailTypeSelected;
```
fedemkr, line 567: Remove the field given that it's no longer used

src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -564,26 +567,28 @@ public bool IncludeNumberRandomWordUsername
...
-                    SaveUsernameOptionsAsync(false).FireAndForget();
+                    TriggerPropertyChanged(nameof(PlusAddressedEmailTypeSelected));
+                    SaveUsernameOptionsAsync().FireAndForget();
                 }
             }
         }
 
         public UsernameEmailType CatchAllEmailTypeSelected
         {
-            get => _catchAllEmailTypeSelected;
```
fedemkr, line 580: Remove the field given that it's no longer used

fedemkr (changes requested):

src/App/Pages/Generator/GeneratorPageViewModel.cs
LRNcardozoWDF: Field removed.

src/App/Pages/Generator/GeneratorPageViewModel.cs
LRNcardozoWDF: Field removed.

src/App/Pages/Vault/CipherAddEditPageViewModel.cs
LRNcardozoWDF: Fixed with second suggestion, thank you for your help.

PR #2111, opened 2022-10-03
[SG-691] Login request is not displayed after changing accounts

andrebispo5:
… to trigger a check for login requests.

## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When the user have more than one account logged in and sends a login request for the not active (still unlocked) account , The user sees the push notification but when switches to the account that received the request the user is not being redirected to the login request screen.

## Code changes

Passwordless request check was not being triggered because there was no "unlock" message being broadcasted. Added new broadcast message when account is switched to trigger passwordless login requests check.

## Before you submit

### DIFF
--- a/src/App/Utilities/AccountManagement/AccountsManager.cs
+++ b/src/App/Utilities/AccountManagement/AccountsManager.cs
@@ -215,8 +215,7 @@ namespace Bit.App.Utilities.AccountManagement
                 }
                 await Task.Delay(50);
                 await _accountsManagerHost.UpdateThemeAsync();
-
-                _messagingService.Send("accountSwitched");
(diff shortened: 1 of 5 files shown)


src/App/Utilities/AccountManagement/AccountsManager.cs
```diff
@@ -213,6 +215,8 @@ private async Task SwitchedAccountAsync()
                 }
                 await Task.Delay(50);
                 await _accountsManagerHost.UpdateThemeAsync();
+
+                _messagingService.Send("accountSwitched");
```
fedemkr, line 219: There is a `AccountsManagerMessageCommands` that have these consts, could you pass it there? so we avoid hardcoding

fedemkr (changes requested):

fedemkr (approved):

PR #2112, opened 2022-10-03
[SG-702] Tapping Push Notification does not open the account the request is for

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Tapping Push Notification does not take you to account the request is for.

## Code changes

Added new method in `src/App/Services/PushNotificationListenerService.cs` to handle notification tap logic.

## Before you submit

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -153,7 +153,7 @@ namespace Bit.Droid
             if (Intent?.GetStringExtra(Constants.NotificationData) is string notificationDataJson)
             {
                 var notificationType = JToken.Parse(notificationDataJson).SelectToken(Constants.NotificationDataType);
-                if (notificationType!= null && (string)notificationType == PasswordlessNotificationData.TYPE)
+                if (notificationType != null && (string)notificationType == PasswordlessNotificationData.TYPE)
                 {
                     _pushNotificationListenerService.OnNotificationTapped(PasswordlessNotificationData.TYPE, JsonConvert.DeserializeObject<PasswordlessNotificationData>(notificationDataJson)).FireAndForget();
                 }
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -20,6 +20,8 @@ using Bit.Core.Enums;
 using Bit.Core.Utilities;
 using Bit.Droid.Receivers;
 using Bit.Droid.Utilities;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Xamarin.Essentials;
 using ZXing.Net.Mobile.Android;
 using FileProvider = AndroidX.Core.Content.FileProvider;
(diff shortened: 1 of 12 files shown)


src/App/Services/PushNotificationListenerService.cs
```diff
@@ -213,6 +218,29 @@ public void OnError(string message, string deviceType)
...
+                    if (notificationUserId != null)
+                    {
+                        await _stateService.SetActiveUserAsync(notificationUserId);
+                        _messagingService.Send(AccountsManagerMessageCommands.SWITCHED_ACCOUNT);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
```
fedemkr, line 240: If possible inject the `ILogger` and use it instead of the helper

fedemkr (changes requested):

src/Android/MainActivity.cs
```diff
@@ -145,6 +147,11 @@ protected override void OnResume()
             AndroidHelpers.SetPreconfiguredRestrictionSettingsAsync(this)
                 .GetAwaiter()
                 .GetResult();
+
+            if (Intent?.GetStringExtra(Constants.PasswordlessNotificationType) is string notificationDataJson)
+            {
+                _pushNotificationListenerService.OnNotificationTapped(Constants.PasswordlessNotificationType, notificationDataJson).FireAndForget();
+            }
```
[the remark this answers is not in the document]
andrebispo5: Yes, if the app is in the background and the user clicks the notification it won’t pass on `OnCreate` or `NewIntent`

src/App/Abstractions/IPushNotificationService.cs
```diff
@@ -10,7 +10,7 @@ public interface IPushNotificationService
         Task<string> GetTokenAsync();
         Task RegisterAsync();
         Task UnregisterAsync();
-        void SendLocalNotification(string title, string message, string notificationId);
+        void SendLocalNotification(string title, string message, string notificationId, string notificationType, Dictionary<string, string> notificationData = null, int notificationTimeoutMinutes = 0);
```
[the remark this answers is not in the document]
andrebispo5: Did a big rework for this one.

src/App/Models/NotificationData.cs
```diff
@@ -3,21 +3,20 @@ namespace Bit.App.Models
 {
     public abstract class BaseNotificationData
     {
-        public abstract string Type { get; }
+        public abstract string NotificationType { get; }
 
+        public string NotificationId { get; set; }
```
fedemkr, line 8: IMHO there is no need to add the prefix "Notification" to the properties given that the class is called notification data so adding the prefix to anything in there is redundant. Or was is confusing on some context?
Also, for now we can leave it as is but in the future if there are notifications that don't require additional data we should extract the `Id` out of this class (which is the only thing required).

src/App/Abstractions/IPushNotificationListenerService.cs
```diff
@@ -9,7 +10,7 @@ public interface IPushNotificationListenerService
         Task OnRegisteredAsync(string token, string device);
         void OnUnregistered(string device);
         void OnError(string message, string device);
-        Task OnNotificationTapped(string type, string data);
+        Task OnNotificationTapped(string type, BaseNotificationData data);
```
fedemkr, line 13: if I'm not mistaken, `type` is no longer needed

fedemkr (changes requested):

fedemkr (approved):

PR #2125, opened 2022-10-11
[SG-703] Login request is not removed after dismissing push notification

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fixe bug when the user dismissed the passwordless login notification by swiping, the data related to the login request was not erased. 

## Code changes

**src/App/Services/PushNotificationListenerService.cs:** added new method to handle notification dismiss events from platforms. This removes from preferences the data stored for that request.
**src/iOS/Services/iOSPushNotificationHandler.cs:** code to handle the event of dismissing notification on iOS
**src/Android/Receivers/NotificationDismissReceiver.cs:** new broadcast receiver to handle the dismiss event on Android.
**src/Core/Constants.cs:** fixed a bug on android where the constant wasn't matching with the property name.

## Before you submit

### DIFF
--- a/src/Android/Receivers/NotificationDismissReceiver.cs
+++ b/src/Android/Receivers/NotificationDismissReceiver.cs
@@ -15,8 +15,8 @@ namespace Bit.Droid.Receivers
     [BroadcastReceiver(Name = Constants.PACKAGE_NAME + "." + nameof(NotificationDismissReceiver), Exported = false)]
     public class NotificationDismissReceiver : BroadcastReceiver
     {
-        private readonly LazyResolve<IPushNotificationListenerService> _pushNotificationListenerService = new LazyResolve<IPushNotificationListenerService>("pushNotificationListenerService");
-        private readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>("logger");
+        private readonly LazyResolve<IPushNotificationListenerService> _pushNotificationListenerService = new LazyResolve<IPushNotificationListenerService>();
+        private readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>();
 
         public override void OnReceive(Context context, Intent intent)
         {
--- a/src/Android/Receivers/NotificationDismissReceiver.cs
+++ b/src/Android/Receivers/NotificationDismissReceiver.cs
@@ -19,7 +19,7 @@ namespace Bit.Droid.Receivers
         private IPushNotificationListenerService _pushNotificationListenerService;
         private ILogger _logger;
 
-        public override async void OnReceive(Context context, Intent intent)
+        public override void OnReceive(Context context, Intent intent)
         {
             try
             {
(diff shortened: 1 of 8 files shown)


src/Android/Receivers/NotificationDismissReceiver.cs
```diff
@@ -0,0 +1,54 @@
...
+using Bit.Core.Abstractions;
+using Bit.Core.Services;
+using Bit.Core.Utilities;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using static AndroidX.Concurrent.Futures.CallbackToFutureAdapter;
+
+namespace Bit.Droid.Receivers
+{
+    [BroadcastReceiver(Name = "com.x8bit.bitwarden.NotificationDismissReceiver", Exported = false)]
```
fedemkr, line 15: Change this to use the new `Constants.PACKAGE_NAME` prefix and if possible use nameof:
```c#
[BroadcastReceiver(Name = Constants.PACKAGE_NAME + "." + nameof(NotificationDismissReceiver), Exported = false)]
```

src/Android/Receivers/NotificationDismissReceiver.cs
```diff
@@ -0,0 +1,54 @@
...
+        private void Resolve()
+        {
+            if (_resolved)
+            {
+                return;
+            }
+            _pushNotificationListenerService = ServiceContainer.Resolve<IPushNotificationListenerService>();
+            _logger = ServiceContainer.Resolve<ILogger>();
+            _resolved = true;
+        }
```
fedemkr, line 51: Remove this and use `LazyResolve<>`, e.g.
```c#
private readonly LazyResolve<IPushNotificationListenerService> _pushNotificationListenerService = new LazyResolve<IPushNotificationListenerService>("pushNotificationListenerService"); // the change to avoid passing the string key is in an open PR so for now it's needed
...
// and when you need to use it:
_pushNotificationListenerService.Value.OnNotificationDismissed(...)
```

src/App/Services/PushNotificationListenerService.cs
```diff
@@ -245,6 +245,30 @@ public async Task OnNotificationTapped(BaseNotificationData data)
             }
         }
 
+        public async Task OnNotificationDismissed(BaseNotificationData data)
+        {
+            Resolve();
```
fedemkr, line 250: Since you're here, can you remove the `Resolve()` method and use `LazyResolve<>` instead for the services?

fedemkr (changes requested):

src/App/Services/PushNotificationListenerService.cs
fedemkr: As far as I see this has not been addressed @andrebispo5

fedemkr (changes requested):

src/App/Services/PushNotificationListenerService.cs
andrebispo5: sorry i though it was on the `NotificationDismissReceiver` class didn't noticed it was in another file. on it!

fedemkr (approved):

PR #2135, opened 2022-10-14
[SG-705] Popup when a request for authentication comes in on a logged-in account that is not active

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Add popup notification when a request for authentication comes in on a logged-in account that is not active.

## Code changes
Changed the `GetPasswordlessLoginNotificationAsync` to not be user specific to make it easier to check the user the request was from instead of having to iterate between all available users in the device and perform checks. 

## Screenshots

image

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -11,6 +11,7 @@ using Bit.Core;
 using Bit.Core.Abstractions;
 using Bit.Core.Enums;
 using Bit.Core.Models.Data;
+using Bit.Core.Models.Response;
 using Bit.Core.Services;
 using Bit.Core.Utilities;
 using Xamarin.Forms;
@@ -168,10 +169,26 @@ namespace Bit.App
                 return;
             }
 
-
             var notification = await _stateService.GetPasswordlessLoginNotificationAsync();
             if (notification == null)
             {
+                return;
+            }
+
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
+            if (notification.UserId != activeUserId)
+            {
+                var notificationUserEmail = await _stateService.GetEmailAsync(activeUserId);
+                Device.BeginInvokeOnMainThread(async () =>
+                {
+                    var result = await _deviceActionService.DisplayAlertAsync(AppResources.LogInRequested, string.Format(AppResources.LoginAttemptFromXDoYouWantToSwitch, notificationUserEmail), AppResources.Cancel, AppResources.Ok);
+                    if (result == AppResources.Ok)
+                    {
+                        await _stateService.SetActiveUserAsync(notification.UserId);
+                        _messagingService.Send(AccountsManagerMessageCommands.SWITCHED_ACCOUNT);
+                    }
+                });
+
                 return;
             }
 
--- a/src/App/Resources/AppResources.resx
+++ b/src/App/Resources/AppResources.resx
@@ -2464,4 +2464,7 @@ select Add TOTP to store the key safely</value>
   <data name="LoginRequestHasAlreadyExpired" xml:space="preserve">
     <value>Login request has already expired.</value>
   </data>
+  <data name="LoginAttemptFromXDoYouWantToSwitch" xml:space="preserve">
+    <value>Login attempt from {0}. Do you want to switch to this account?</value>
+  </data>
 </root>
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -30,6 +30,7 @@
         public static string EventCollectionKey = "eventCollection";
         public static string RememberedEmailKey = "rememberedEmail";
         public static string RememberedOrgIdentifierKey = "rememberedOrgIdentifier";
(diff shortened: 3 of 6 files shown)


src/App/App.xaml.cs
```diff
@@ -168,13 +169,29 @@ private async Task CheckPasswordlessLoginRequestsAsync()
...
+                    var result = await _deviceActionService.DisplayAlertAsync(AppResources.LogInRequested, string.Format(AppResources.LoginAttemptFromXDoYouWantToSwitch, notificationUserEmail), AppResources.Cancel, AppResources.Ok);
+                    if (result == AppResources.Ok)
+                    {
+                        await _stateService.SetActiveUserAsync(notification.UserId);
+                        _messagingService.Send(AccountsManagerMessageCommands.SWITCHED_ACCOUNT);
+                    }
+                });
+
+                return;
+            }
```
fedemkr, line 193: This method is getting big, please extract this into a different method and remember to use the awaitable method of `Device` => `await Device.InvokeOnMainThreadAsync` (in your case, in an exception happens in there it'll crash the app).

src/App/Resources/AppResources.resx
```diff
@@ -2464,4 +2464,7 @@ select Add TOTP to store the key safely</value>
   <data name="LoginRequestHasAlreadyExpired" xml:space="preserve">
     <value>Login request has already expired.</value>
   </data>
+  <data name="LoginAttemptFromXDoYouWantToSwitch" xml:space="preserve">
```
fedemkr, line 2467: It's not that long to add the last three words, please do so:
`LoginAttemptFromXDoYouWantToSwitch` => `LoginAttemptFromXDoYouWantToSwitchToThisAccount`

src/Core/Constants.cs
```diff
@@ -30,6 +30,7 @@ public static class Constants
         public static string EventCollectionKey = "eventCollection";
         public static string RememberedEmailKey = "rememberedEmail";
         public static string RememberedOrgIdentifierKey = "rememberedOrgIdentifier";
+        public static string PasswordlessLoginNofiticationKey = "passwordlessLoginNofiticationKey";
```
fedemkr, line 33: typo `PasswordlessLoginNofiticationKey` => `PasswordlessLoginNotificationKey` and also on the string

fedemkr (changes requested):
I have a few questions besides the requested changes:
1. How does this works if you receive multiple login notification requests from different logged in accounts "at the same time" and the user taps in all of them? Does it cache only the last one?
2. If the active user denies switching account and the next minute they switch to the other account, should the passwordless login be presented?

src/App/App.xaml.cs
andrebispo5: 1. Only the most request request is saved for now.
2. If the user cancels the automatic account switching, he can manually change the account and the request will pop.

src/App/Resources/AppResources.resx
andrebispo5: Yup it is better to show it in separate lines.

fedemkr (changes requested):

src/App/App.xaml.cs
```diff
@@ -168,13 +169,20 @@ private async Task CheckPasswordlessLoginRequestsAsync()
...
                 return;
             }
 
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
+            var notificationUserEmail = await _stateService.GetEmailAsync(activeUserId);
+            if (notification.UserId != activeUserId)
+            {
+                await DisplaySwitchAccountAlert(notification.UserId, notificationUserEmail);
+                return;
+            }
```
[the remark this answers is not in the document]
andrebispo5: done 👍

src/App/App.xaml.cs
```diff
@@ -168,13 +169,18 @@ private async Task CheckPasswordlessLoginRequestsAsync()
...
             {
                 return;
             }
 
+            if (!await CheckShouldSwitchActiveUserAsync(notification))
+            {
+                return;
+            }
+
+
```
fedemkr, line 183: Extra space not needed (I think it should be caught by the dotnet format)

src/App/App.xaml.cs
```diff
@@ -197,6 +203,27 @@ private async Task CheckPasswordlessLoginRequestsAsync()
             }
         }
 
+        private async Task<bool> CheckShouldSwitchActiveUserAsync(PasswordlessRequestNotification notification)
+        {
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
+            var notificationUserEmail = await _stateService.GetEmailAsync(activeUserId);
```
fedemkr, line 209: This should be after the `if`. You might be doing async work that could not be necessary in the scenario the notification user id is the same as the active one

fedemkr (changes requested):

fedemkr (dismissed):

fedemkr (approved):

PR #2140, opened 2022-10-18
[SG 547] Mobile username generator iOS.Extension UI changes

LRNcardozoWDF:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
Add the username generator feature to iOS.Extension

## Code changes

* **IDeviceActionService.cs 
     Android/Services/DeviceActionService.cs
     iOS.Core/Services/DeviceActionService.cs :** Added support to close new modal pop up.
* **ITokenService.cs
      TokenService.cs :** Added new method to refresh token without returning it.
* **GeneratorPage.xaml.cs :** Added flag to identify iOS.Extension flow and support to close a Xamarin.Forms page as View Controller.
* **GeneratorPageViewModel.cs :** Made sure token is refreshed before using it to get user's email.
* **LoginAddViewController.cs :** Refactored button for generate password, added button for generate username that performs navigation to Username Generator page. Added button eye/eye slash to control password field visibility.
* **FormEntryTableViewCell.cs :** Added configuration for a second button and adjusted size when the second button is enabled.

## Screenshots

Add item page in iOS.Extension:

**Before**
image

**Now**
image

## Before you submit


src/App/Pages/Generator/GeneratorPage.xaml.cs
```diff
@@ -15,17 +16,21 @@ public partial class GeneratorPage : BaseContentPage, IThemeDirtablePage
 
         private GeneratorPageViewModel _vm;
         private readonly bool _fromTabPage;
+        private readonly bool _fromExtension;
```
fedemkr, line 19: Instead of this flag, I'd use the same approach as in other places where we use `AppOptions` and in it you find the flag

src/App/Pages/Generator/GeneratorPage.xaml.cs
```diff
@@ -138,7 +143,14 @@ private async void Close_Clicked(object sender, EventArgs e)
...
-                await Navigation.PopModalAsync();
+                if (_fromExtension)
+                {
+                    _deviceActionService.ClosePopUpPage();
+                }
+                else
+                {
+                    await Navigation.PopModalAsync();
+                }
             }
```
fedemkr, line 154: Move the whole Close to a command in the ViewModel and bind to it. Thus move the logic to the ViewModel as well and remove the initialization of the `_deviceActionService` from here.

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -13,6 +13,12 @@
 using Bit.Core.Models.View;
 using Bit.Core.Utilities;
 using Bit.Core.Exceptions;
+using Bit.App.Models;
+using Bit.App.Pages;
+using Bit.App.Utilities;
+using Xamarin.Forms;
+using Bit.Core;
+using GameController;
```
fedemkr, line 21: I don't think `GameController` is necessary here

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -75,17 +85,35 @@ public override void ViewDidLoad()
...
+                {
+                    PasswordCell.SecondButton.SetTitle(BitwardenIcons.EyeSlash, UIControlState.Normal);
+                    PasswordCell.TextField.SecureTextEntry = false;
+                }
+                else
+                {
+                    PasswordCell.SecondButton.SetTitle(BitwardenIcons.Eye, UIControlState.Normal);
+                    PasswordCell.TextField.SecureTextEntry = true;
+                }
+            };
```
fedemkr, line 109: This is repeated in several places like `BaseLockPasswordViewController`, `LockPasswordViewController`. Made a class that can configure the cell and button to have password behavior or add a Builder to create cells with different behaviors/uses

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -12,14 +12,15 @@ public class FormEntryTableViewCell : ExtendedUITableViewCell, ISelectable
...
+        public UIButton SecondButton { get; set; }
         public event EventHandler ValueChanged;
 
-
         public FormEntryTableViewCell(
             string labelName = null,
             bool useTextView = false,
             nfloat? height = null,
             bool useButton = false,
+            bool useTwoButtons = false,
```
fedemkr, line 23: Change this to an enum internal to this class:
```c#
public enum ButtonsConfig : byte
{
     None = 0,
     One = 1,
     Two = 2
}
```
So we avoid the flags, if not it's kind of confusing to have possibilities to pass parameters like `useButton=false` and `useTwoButtons=true` unless you throw an exception if that happens or something like that.
However, I think that the enum way is clearer for usage

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -161,6 +161,27 @@ public class FormEntryTableViewCell : ExtendedUITableViewCell, ISelectable
...
+
+                ContentView.Add(SecondButton);
+                ContentView.Add(Button);
+
+                ContentView.BottomAnchor.ConstraintEqualTo(Button.BottomAnchor, 10f).Active = true;
+                ContentView.BottomAnchor.ConstraintEqualTo(SecondButton.BottomAnchor, 10f).Active = true;
+                SecondButton.LeadingAnchor.ConstraintEqualTo(TextField.TrailingAnchor, 9f).Active = true;
+                Button.LeadingAnchor.ConstraintEqualTo(SecondButton.TrailingAnchor, 10f).Active = true;
+                ContentView.TrailingAnchor.ConstraintEqualTo(Button.TrailingAnchor, 10f).Active = true;
+            }
```
fedemkr, line 184: Refactor this and the part above of using one button to use the enum and extract the logic away from this method:
```c#
if (buttonsConfig != ButtonsConfig.None)
{
     AddButtons(buttonsConfig);
}
```
Also reuse code when possible, like here that you're duplicating code to configure the first button on the one button and two buttons scenarios, as well as some constraints that are the same.

Oh and btw, you can use `NSLayoutConstraint.ActivateConstraints(new NSLayoutConstraint[] { });` to pass all the constraints and it will activate all of them instead of having to set `.Active = true` on each of them

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -112,7 +112,7 @@ public class FormEntryTableViewCell : ExtendedUITableViewCell, ISelectable
                 ContentView.Add(TextField);
                 ContentView.AddConstraints(new NSLayoutConstraint[] {
                     NSLayoutConstraint.Create(TextField, NSLayoutAttribute.Leading, NSLayoutRelation.Equal, ContentView, NSLayoutAttribute.Leading, 1f, leadingConstant),
-                    NSLayoutConstraint.Create(ContentView, NSLayoutAttribute.Trailing, NSLayoutRelation.Equal, TextField, NSLayoutAttribute.Trailing, 1f, useButton ? 55f : 15f),
+                    NSLayoutConstraint.Create(ContentView, NSLayoutAttribute.Trailing, NSLayoutRelation.Equal, TextField, NSLayoutAttribute.Trailing, 1f, useTwoButtons ? 95f : useButton ? 55f : 15f),
```
fedemkr, line 115: I'm not fond of using nested ternary operator, I'd just use a local function with a switch and return the proper size depending on the case

fedemkr (changes requested):

src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -91,6 +95,19 @@ public GeneratorPageViewModel()
...
+        {
+            if (_iosExtension)
+            {
+                _deviceActionService.CloseExtensionPopUp();
+            }
+            else
+            {
+                await Page.Navigation.PopModalAsync();
+            }
         }
```
fedemkr, line 111: Methods should be below properties, you can move this to the end of the file if you want

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -195,5 +165,75 @@ public void Select()
...
+        public void BuildSecureCell(UIButton button)
+        {
+            button.TitleLabel.Font = UIFont.FromName("bwi-font", 28f);
+            button.SetTitle(BitwardenIcons.Eye, UIControlState.Normal);
+            button.TouchUpInside += (sender, e) =>
+            {
+                TextField.SecureTextEntry = !TextField.SecureTextEntry;
+                button.SetTitle(TextField.SecureTextEntry ? BitwardenIcons.Eye : BitwardenIcons.EyeSlash, UIControlState.Normal);
+            };
+        }
```
fedemkr, line 178: Change to something like
```c#
public void ConfigureToggleSecureTextButton(bool useSecondaryButton = false)
{
    var button = useSecondaryButton ? SecondButton : Button;
    ...
```
Using it like this is less confusing for a caller to do so, given that in this clear it's clear that the button is the one already in the cell and in the other way you can pass any button as a parameter

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -195,5 +165,75 @@ public void Select()
...
+
+        private void AddButtons(ButtonsConfig buttonsConfig)
+        {
+            Button = new UIButton(UIButtonType.System);
+            Button.Frame = ContentView.Bounds;
```
fedemkr, line 183: You shouldn't need to set a `Frame` when you use constraints, except on certain occasions. Were you getting some kind of error without it?

src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -91,6 +95,19 @@ public GeneratorPageViewModel()
             RegenerateUsernameCommand = new AsyncCommand(RegenerateUsernameAsync, onException: ex => OnSubmitException(ex), allowsMultipleExecutions: false);
             ToggleForwardedEmailHiddenValueCommand = new AsyncCommand(ToggleForwardedEmailHiddenValueAsync, onException: ex => OnSubmitException(ex), allowsMultipleExecutions: false);
             CopyCommand = new AsyncCommand(CopyAsync, onException: ex => OnSubmitException(ex), allowsMultipleExecutions: false);
+            CloseCommand = new AsyncCommand(CloseAsync, onException: ex => OnSubmitException(ex), allowsMultipleExecutions: false);
```
fedemkr, line 98: Change the `onException` method, you're using the same as OnSubmit so if the value of `UsernameTypeSelected` is `UsernameType.ForwardedEmailAlias` you're going to show a confusing error to the user if an exception is raised on close. You can just log the exception and that's it

src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -140,6 +158,12 @@ public bool IsUsername
             set => SetProperty(ref _isUsername, value);
         }
 
+        public bool IosExtension
+        {
+            get => _iosExtension;
+            set => SetProperty(ref _iosExtension, value);
+        }
```
fedemkr, line 165: Do you need this to fire property changes? If not, remove the backing field as well

fedemkr (changes requested):
Much better now, only a few things I've spotted to improve

src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -174,5 +165,74 @@ public void Select()
...
+                        NSLayoutConstraint.Create(ContentView, NSLayoutAttribute.Bottom, NSLayoutRelation.Equal, SecondButton, NSLayoutAttribute.Bottom, 1f, 10f),
+                        NSLayoutConstraint.Create(SecondButton, NSLayoutAttribute.Leading, NSLayoutRelation.Equal, TextField, NSLayoutAttribute.Trailing, 1f, 9f),
+                        NSLayoutConstraint.Create(Button, NSLayoutAttribute.Leading, NSLayoutRelation.Equal, SecondButton, NSLayoutAttribute.Trailing, 1f, 10f),
+                        NSLayoutConstraint.Create(ContentView, NSLayoutAttribute.Trailing, NSLayoutRelation.Equal, Button, NSLayoutAttribute.Trailing, 1f, 10f)
+                    });
+                    break;
+            }
+        }
+
+        private float GetTrailingConstant(ButtonsConfig buttonsConfig)
```
fedemkr, line 216: Even though it's not wrong, I'd improve it by calling it:
`GetTextFieldToContainerTrailingConstant(ButtonsConfig buttonsConfig)`
Given that it's very specific for that case. But will leave that to you if you want to make the change or not

fedemkr (dismissed):

fedemkr (approved):

PR #2144, opened 2022-10-20
[SG-747] Add Android QA build to mobile build pipeline

andrebispo5:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [X] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Workflow needs to create 2 android apks, which can be installed side-by-side (`prod` and `qa`) and each receives push notifications from the respective environment.

## Code changes

* **Decrypt Secrets step:** Separated into two steps. Google-Services file is only decrypted in `prod` to allow the `qa` build to receive notifications from the QA Cloud.
* **Deploy to PlayStore step:** Before fix, if condition was always returning true.

## Before you submit


.github/workflows/build.yml
```diff
@@ -135,44 +149,44 @@ jobs:
         shell: pwsh
 
       - name: Verify Format
+        if: ${{ matrix.variant == 'prod' }}
```
vvolkgang, line 152: Any reason why we need this to be prod specific?

vvolkgang (changes requested):

PR #2151, opened 2022-10-25
Updated avatar color selection logic

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Updated the way background and text color are established for the account avatar to sync with recent changes in other clients.

**_Note_**: Work was also done to replace `SkiaSharp` with a simple layout w/ text (required for Emoji support, plus it's another MAUI roadblock) but I ran into issues with iOS.  I'm adding this first to get color sync in before the next release, and will continue with the layout conversion next.

## Code changes

* **CoreHelpers.cs:** Added/moved color selection methods here (`StringToColor` and `TextColorFromBgColor`)
* **AvatarImageSource.cs:** Added `userId` arg and appropriate support, also renamed/simplified some vars to sync with other clients
* **AccountViewCellViewModel.cs/AvatarImageSourcePool.cs/BaseContentPage.cs/AccountSwitchingOverlayHelper.cs:** Added `userId` arg

## Before you submit

### DIFF
--- a/src/Core/Utilities/CoreHelpers.cs
+++ b/src/Core/Utilities/CoreHelpers.cs
@@ -264,5 +264,38 @@ namespace Bit.Core.Utilities
         {
             return JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(obj));
         }
+
+        public static string TextColorFromBgColor(string hexColor, int threshold = 166)
+        {
+            var bgColorHexRgb = hexColor.Substring(0, 1) == "#" ? hexColor.Substring(3, 6) : hexColor.Substring(2, 6);
+            var r = Convert.ToUInt32(bgColorHexRgb.Substring(0, 2), 16);
+            var g = Convert.ToUInt32(bgColorHexRgb.Substring(2, 2), 16);
+            var b = Convert.ToUInt32(bgColorHexRgb.Substring(4, 2), 16);
+            var blackColor = "#ff000000";
+            var whiteColor = "#ffffffff";
+            var comp = r * 0.299 + g * 0.587 + b * 0.114;
+            return r * 0.299 + g * 0.587 + b * 0.114 > threshold ? blackColor : whiteColor;
+        }
+
+        public static string StringToColor(string str)
+        {
+            if (str == null)
+            {
+                return "#33ffffff";
+            }
+            var hash = 0;
+            for (var i = 0; i < str.Length; i++)
+            {
+                hash = str[i] + ((hash << 5) - hash);
+            }
+            var color = "#FF";
+            for (var i = 0; i < 3; i++)
+            {
+                var value = (hash >> (i * 8)) & 0xff;
+                var base16 = "00" + Convert.ToString(value, 16);
+                color += base16.Substring(base16.Length - 2);
+            }
+            return color;
+        }
     }
 }
--- a/src/Core/Utilities/CoreHelpers.cs
+++ b/src/Core/Utilities/CoreHelpers.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Web;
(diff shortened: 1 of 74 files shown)


src/Core/Utilities/CoreHelpers.cs
```diff
@@ -264,5 +264,38 @@ public static T Clone<T>(T obj)
...
+        {
+            var bgColorHexRgb = hexColor.Substring(0, 1) == "#" ? hexColor.Substring(3, 6) : hexColor.Substring(2, 6);
+            var r = Convert.ToUInt32(bgColorHexRgb.Substring(0, 2), 16);
+            var g = Convert.ToUInt32(bgColorHexRgb.Substring(2, 2), 16);
+            var b = Convert.ToUInt32(bgColorHexRgb.Substring(4, 2), 16);
+            var blackColor = "#ff000000";
+            var whiteColor = "#ffffffff";
+            var comp = r * 0.299 + g * 0.587 + b * 0.114;
+            return r * 0.299 + g * 0.587 + b * 0.114 > threshold ? blackColor : whiteColor;
+        }
```
fedemkr, line 278: This can be simply written by:
```c#
var bgColor = (Color)new ColorConverter().ConvertFromString(hexColor);
var luminance = bgColor.R * 0.299 + bgColor.G * 0.587 + bgColor.B * 0.114;
return luminance > threshold ? "#ff000000" : "#ffffffff";
```

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -264,5 +264,38 @@ public static T Clone<T>(T obj)
...
+            for (var i = 0; i < str.Length; i++)
+            {
+                hash = str[i] + ((hash << 5) - hash);
+            }
+            var color = "#FF";
+            for (var i = 0; i < 3; i++)
+            {
+                var value = (hash >> (i * 8)) & 0xff;
+                var base16 = "00" + Convert.ToString(value, 16);
+                color += base16.Substring(base16.Length - 2);
```
fedemkr, line 296: In order to avoid adding and then performing a substring, you can write it as:
```c#
color += Convert.ToString(value, 16).PadLeft(2, '0');
```

src/Core/Utilities/CoreHelpers.cs
mp-bw: This required a null check before casting - let me know if you're ok with the adjustment

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -264,5 +265,37 @@ public static T Clone<T>(T obj)
...
+        {
+            var bgColorObj = new ColorConverter().ConvertFromString(hexColor);
+            if (bgColorObj == null)
+            {
+                return "#ff000000";
+            }
+            var bgColor = (Color)bgColorObj;
+            var luminance = bgColor.R * 0.299 + bgColor.G * 0.587 + bgColor.B * 0.114;
+            return luminance > threshold ? "#ff000000" : "#ffffffff";
+        }
```
fedemkr, line 279: I'd write this as:
```c#
if(new ColorConverter().ConvertFromString(hexColor) is Color bgColor)
{
    var luminance = bgColor.R * 0.299 + bgColor.G * 0.587 + bgColor.B * 0.114;
    return luminance > threshold ? "#ff000000" : "#ffffffff";
}

return "#ff000000";
```
Bear in mind that the conversion [can throw FormatException](https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.colorconverter.convertfromstring?view=windowsdesktop-6.0#exceptions) but I'm hesitant to catch the exception on this method, I think it should be the responsability of the caller to provide a `hexColor` correctly formatted

src/Core/Utilities/CoreHelpers.cs
```diff
@@ -264,5 +265,37 @@ public static T Clone<T>(T obj)
...
+
+        public static string StringToColor(string str)
+        {
+            if (str == null)
+            {
+                return "#33ffffff";
+            }
```
fedemkr, line 286: I think that given that now this method is in `CoreHelpers` (and could be reused) we shouldn't hardcode the default value inside but return null directly and let the caller handle the default value:
```c#
public static string StringToColor(string str, string default)
{
    if (str == null)
    {
        return null;
    }
```
so when calling it like in `AvatarImageSource`, we should do so like this:
```c#
var bgColor = CoreHelpers.StringToColor(_id ?? upperCaseText) ?? "#33ffffff";
```

fedemkr (changes requested):

src/Core/Utilities/CoreHelpers.cs
mp-bw: Agreed, good call

fedemkr (approved):

PR #2156, opened 2022-10-28
SG-786 - Fix 400 error code log outs without invalid_grant

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Be more precise to force log out when receiving 400 error code by ensuring that the response error is invalid grant.

## Code changes

This is related to the clients PR [400s only log out on invalid grant error #3924](https://github.com/bitwarden/clients/pull/3924)

* **ApiService.cs:** After parsing the json, log out the user if the response error is "invalid_grant".

## Before you submit

### DIFF
--- a/src/Core/Services/ApiService.cs
+++ b/src/Core/Services/ApiService.cs
@@ -787,19 +787,6 @@ namespace Bit.Core.Services
         private async Task<ErrorResponse> HandleErrorAsync(HttpResponseMessage response, bool tokenError,
             bool authed, bool logoutOnUnauthorized = true)
         {
-            if (authed
-                &&
-                (
-                    (tokenError && response.StatusCode == HttpStatusCode.BadRequest)
-                    ||
-                    (logoutOnUnauthorized && response.StatusCode == HttpStatusCode.Unauthorized)
-                    ||
-                    response.StatusCode == HttpStatusCode.Forbidden
-                ))
-            {
-                await _logoutCallbackAsync(new Tuple<string, bool, bool>(null, false, true));
-                return null;
-            }
             try
             {
                 JObject responseJObject = null;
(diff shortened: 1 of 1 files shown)


src/Core/Services/ApiService.cs
```diff
@@ -808,6 +795,21 @@ private ErrorResponse HandleWebError(Exception e)
...
+                        (tokenError && response.StatusCode == HttpStatusCode.BadRequest && responseJObject?["error"]?.ToString() == "invalid_grant")
+                        ||
+                        (logoutOnUnauthorized && response.StatusCode == HttpStatusCode.Unauthorized)
+                        ||
+                        response.StatusCode == HttpStatusCode.Forbidden
+                    ))
+                {
+                    await _logoutCallbackAsync(new Tuple<string, bool, bool>(null, false, true));
+                    return null;
+                }
```
fedemkr, line 811: I'd split this in two:
In the same place as it was before I'd leave:
```c#
if (authed
    &&
    (
        (logoutOnUnauthorized && response.StatusCode == HttpStatusCode.Unauthorized)
        ||
        response.StatusCode == HttpStatusCode.Forbidden
    ))
{
    await _logoutCallbackAsync(new Tuple<string, bool, bool>(null, false, true));
    return null;
}
```
and in on the new place:
```c#
if (authed && tokenError 
    &&
    response.StatusCode == HttpStatusCode.BadRequest 
    && 
    responseJObject?["error"]?.ToString() == "invalid_grant")
{
    await _logoutCallbackAsync(new Tuple<string, bool, bool>(null, false, true));
    return null;
}
```
The reason behind this is that if let's say that the status code is `HttpStatusCode.Unauthorized` then it will waste time reading the response content; so splitting it like this we gain performance on those cases.

fedemkr (changes requested):

fedemkr (approved):

PR #2166, opened 2022-11-04
[SG-800] Selecting notification for inactive account on iOS while app is open does not display the request

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix bug where If the user never answers the switch accounts pop up, the lock is never released. This was problematic because when tapping a notification while pop-up was showing, the user navigates to the other account automatically and the pop-up is never dismissed.

## Code changes

Removed await to allow the execution flow and lock to be released.

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -221,7 +221,7 @@ namespace Bit.App
             }
 
             var notificationUserEmail = await _stateService.GetEmailAsync(notification.UserId);
-            await Device.InvokeOnMainThreadAsync(async () =>
+            Device.BeginInvokeOnMainThread(async () =>
             {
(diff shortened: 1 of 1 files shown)


src/App/App.xaml.cs
```diff
@@ -221,7 +221,7 @@ private async Task<bool> CheckShouldSwitchActiveUserAsync(PasswordlessRequestNot
             }
 
             var notificationUserEmail = await _stateService.GetEmailAsync(notification.UserId);
-            await Device.InvokeOnMainThreadAsync(async () =>
+            Device.BeginInvokeOnMainThread(async () =>
```
fedemkr, line 224: You need to add a try...catch inside otherwise if an exception occurs inside, it'll crash the app.
OTOH, given that you have removed the `await` now the lock will last less cause the call to the main thread won't wait until the inside finishes (unless I'm seeing something wrong). So you could potentially have the main thread block running and another `CheckShouldSwitchActiveUserAsync(...)` called (you can test this by having a delay of 10 seconds inside the main thread block and calling again the check)

fedemkr (changes requested):

fedemkr (approved):

PR #2167, opened 2022-11-04
[SG-174] Login with Device Request - Mobile

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add a new login option to the methods list. The login with another device will allow users to log into their accounts using a passwordless method. 

## Code changes

* **src/App/Pages/CaptchaProtectedViewModel.cs:** New base method to handle captcha. Older method didn't have any changes to protect other parts of the code that were using it. May be target of some tech dept to change to use the newer method.
* **src/Core/Services/CryptoService.cs:** Added parameter to allow to rsa decrypt with a custom private key.

## Screenshots

image
image
image
image

## Before you submit


src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
```diff
@@ -0,0 +1,75 @@
...
+            _vm.Page = this;
+            _vm.Email = email;
+            _vm.StartTwoFactorAction = () => Device.BeginInvokeOnMainThread(async () => await StartTwoFactorAsync());
+            _vm.LogInSuccessAction = () => Device.BeginInvokeOnMainThread(async () => await LogInSuccessAsync());
+            _vm.UpdateTempPasswordAction = () => Device.BeginInvokeOnMainThread(async () => await UpdateTempPasswordAsync());
+            _vm.CloseAction = async () => { await Navigation.PopModalAsync(); };
+
+            _vm.CreatePasswordlessLoginCommand.Execute(null);
+
+            ToolbarItems.Add(_closeItem);
```
fedemkr, line 29: If the toolbar item is always added, then do it on xaml and move it away from the `ResourceDictionary` and therefore remove this line as well

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+        private IStateService _stateService;
+        private ICryptoFunctionService _cryptoFunctionService;
+        private ICryptoService _cryptoService;
+        private IPasswordGenerationService _passwordGenerationService;
+        private IPlatformUtilsService _platformUtilsService;
+        private IEnvironmentService _environmentService;
+        private ILogger _logger;
+        private CancellationTokenSource _checkLoginRequestStatusCts;
+        private Task _checkLoginRequestStatusTask;
+        private const int REQUEST_TIME_UPDATE_PERIOD_IN_SECONDS = 4;
```
fedemkr, line 38: Consts should be at the top

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+        private ISyncService _syncService;
+        private II18nService _i18nService;
+        private IStateService _stateService;
+        private ICryptoFunctionService _cryptoFunctionService;
+        private ICryptoService _cryptoService;
+        private IPasswordGenerationService _passwordGenerationService;
+        private IPlatformUtilsService _platformUtilsService;
+        private IEnvironmentService _environmentService;
+        private ILogger _logger;
+        private CancellationTokenSource _checkLoginRequestStatusCts;
```
fedemkr, line 36: Make an space between services and other class fields

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+        private Task _checkLoginRequestStatusTask;
+        private const int REQUEST_TIME_UPDATE_PERIOD_IN_SECONDS = 4;
+        private string _fingerprintPhrase;
+        private string _email;
+        private string _requestId;
+        private string _requestAccessCode;
+        private Tuple<byte[], byte[]> _requestKeyPair;
+        public Action StartTwoFactorAction { get; set; }
+        public Action LogInSuccessAction { get; set; }
+        public Action UpdateTempPasswordAction { get; set; }
```
fedemkr, line 46: Properties should be below the constructor

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+            _platformUtilsService = ServiceContainer.Resolve<IPlatformUtilsService>();
+            _environmentService = ServiceContainer.Resolve<IEnvironmentService>();
+            _authService = ServiceContainer.Resolve<IAuthService>();
+            _syncService = ServiceContainer.Resolve<ISyncService>();
+            _i18nService = ServiceContainer.Resolve<II18nService>();
+            _stateService = ServiceContainer.Resolve<IStateService>();
+            _cryptoFunctionService = ServiceContainer.Resolve<ICryptoFunctionService>();
+            _cryptoService = ServiceContainer.Resolve<ICryptoService>();
+            _passwordGenerationService = ServiceContainer.Resolve<IPasswordGenerationService>();
+            _logger = ServiceContainer.Resolve<ILogger>("logger");
```
fedemkr, line 60: No need for the string key. You can remove `"logger"`

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+
+            PageTitle = AppResources.LogInWithAnotherDevice;
+
+            CreatePasswordlessLoginCommand = new AsyncCommand(async () => await Device.InvokeOnMainThreadAsync(CreatePasswordlessLoginAsync),
```
fedemkr, line 64: No need for the `async/await` keywords.

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        public string Email
+        {
+            get => _email;
+            set
+            {
+                SetProperty(ref _email, value);
+            }
+        }
```
fedemkr, line 87: inline the setters, e.g.
```c#
public string FingerprintPhrase
{
    get => _fingerprintPhrase;
    set => SetProperty(ref _fingerprintPhrase, value);
}
```

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        protected override II18nService i18nService => _i18nService;
+        protected override IEnvironmentService environmentService => _environmentService;
+        protected override IDeviceActionService deviceActionService => _deviceActionService;
+        protected override IPlatformUtilsService platformUtilsService => _platformUtilsService;
```
fedemkr, line 92: Move this below the other services with an space between the first ones and these ones

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+            try
+            {
+                _checkLoginRequestStatusCts?.Cancel();
+                _checkLoginRequestStatusCts?.Dispose();
+            }
+            catch (Exception ex)
+            {
+                _logger.Exception(ex);
+            }
+        }
```
fedemkr, line 105: After disposing set it to `null`

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        public void StartCheckLoginRequestStatus()
```
fedemkr, line 107: I'd reorder the methods, first the `Start...` and after it the `Stop...`; I think it's easier to read when checking the file

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+            var accessCode = await _passwordGenerationService.GeneratePasswordAsync(new PasswordGenerationOptions(true) { Length = 25 });
+            var response = await _authService.PasswordlessCreateLoginRequestAsync(_email, publicB64, accessCode, fingerprintPhrase);
+
+            if (response != null)
+            {
+                FingerprintPhrase = fingerprintPhrase;
+                _requestAccessCode = accessCode;
+                _requestId = response.Id;
+                _requestKeyPair = keyPair;
+            }
```
fedemkr, line 196: If possible, could you move this to a service that makes all the operations and returns an struct with the things needed here: `fingerprintPhrase, accessCode, response.Id, keyPair`?
Is the `Length = 25` something fixed and we can move to a const somewhere or will it change to some setting?
I believe there's already a const for the fingerprint phrase separator

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    return;
+                }
+
+                StopCheckLoginRequestStatus();
+
+                var decKey = await _cryptoFunctionService.RsaDecryptAsync(response.Key, _requestKeyPair.Item2);
+                var decPasswordHash = await _cryptoFunctionService.RsaDecryptAsync(response.MasterPasswordHash, _requestKeyPair.Item2);
+
+                var authResult = await _authService.LogInPasswordlessAsync(Email, _requestAccessCode, _requestId, decKey, decPasswordHash);
```
fedemkr, line 142: If possible, I'd like to reduce the complexity on the ViewModel and move this into the `LogInPasswordlessAsync(...)` method that can take the raw values and the key pair and make the decryption inside. Thus we can remove the dependency on `ICryptoFunctionService` on the VM

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+                if (authResult.CaptchaNeeded)
+                {
+                    if (await HandleCaptchaAsync(authResult.CaptchaSiteKey))
+                    {
+                        await CheckLoginRequestStatus();
+                        _captchaToken = null;
+                    }
+                    return;
+                }
+                _captchaToken = null;
```
fedemkr, line 154: I don't like much setting base fields here when it can be donde above. Move this to a method on the base class, so here you can have:
```c#
if(!await HandleCaptchaAsync(authResult.CaptchaNeeded, authResult.CaptchaSiteKey, CheckLoginRequestStatus))
{
     return;
}
```

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    StartTwoFactorAction?.Invoke();
+                }
+                else if (authResult.ForcePasswordReset)
+                {
+                    UpdateTempPasswordAction?.Invoke();
+                }
+                else
+                {
+                    var task = Task.Run(async () => await _syncService.FullSyncAsync(true));
```
fedemkr, line 168: There's no need for setting the task to a local var. Also you don't have any exception handling there. This could be done simply by:
```c#
_syncService.FullSyncAsync(true).FireAndForget();
```

src/App/Resources/AppResources.resx
```diff
@@ -2491,4 +2491,22 @@ Do you want to switch to this account?</value>
...
   </data>
+  <data name="LogInInitiated" xml:space="preserve">
+    <value>Log in initiated</value>
+  </data>
+  <data name="ANotificationHasBeenSentToYourDevice" xml:space="preserve">
+    <value>A notification has been sent to your device.</value>
+  </data>
+  <data name="PleaseMakeSureYourVaultIsUnlockedAnd" xml:space="preserve">
+    <value>Please make sure your vault is unlocked and the Fingerprint phrase matches on the other device.</value>
+  </data>
```
fedemkr, line 2502: It's not that long to make the key like the value, I'd do so. Also, by reading the key is like your lacking something or you don't know if it needs a parameter or something else.

src/Core/Abstractions/IAuthService.cs
```diff
@@ -26,9 +27,12 @@ public interface IAuthService
         Task<AuthResult> LogInSsoAsync(string code, string codeVerifier, string redirectUrl, string orgId);
         Task<AuthResult> LogInCompleteAsync(string email, string masterPassword, TwoFactorProviderType twoFactorProvider, string twoFactorToken, bool? remember = null);
         Task<AuthResult> LogInTwoFactorAsync(TwoFactorProviderType twoFactorProvider, string twoFactorToken, string captchaToken, bool? remember = null);
+        Task<AuthResult> LogInPasswordlessAsync(string email, string accessCode, string authRequestId, byte[] key, byte[] localHashedPassword);
```
fedemkr, line 30: Modify as in the comment on the VM

src/Core/Abstractions/IAuthService.cs
```diff
@@ -26,9 +27,12 @@ public interface IAuthService
...
 
         Task<PasswordlessLoginResponse> GetPasswordlessLoginRequestByIdAsync(string id);
+        Task<PasswordlessLoginResponse> GetPasswordlessLoginResponse(string id, string accessCode);
```
fedemkr, line 33: Add async suffix

src/Core/Abstractions/IApiService.cs
```diff
@@ -84,7 +84,9 @@ public interface IApiService
         Task<SendResponse> PutSendRemovePasswordAsync(string id);
         Task DeleteSendAsync(string id);
         Task<PasswordlessLoginResponse> GetAuthRequestAsync(string id);
+        Task<PasswordlessLoginResponse> GetAuthResponseAsync(string id, string accessCode);
         Task<PasswordlessLoginResponse> PutAuthRequestAsync(string id, string key, string masterPasswordHash, string deviceIdentifier, bool requestApproved);
+        Task<PasswordlessLoginResponse> PostCreateRequestAsync(string email, string publicKey, string deviceIdentifier, string accessCode, AuthRequestType? type, string fingerprintPhrase);
```
fedemkr, line 89: Change the signature to have one only parameter for the request `PasswordlessCreateLoginRequest`. It's simpler to refactor afterwards if another parameter is needed or removed or modified, you just simply change the model of the request and you don't have to change the signatures.

src/Core/Services/AuthService.cs
```diff
@@ -476,6 +485,11 @@ public async Task<PasswordlessLoginResponse> GetPasswordlessLoginRequestByIdAsyn
             return await _apiService.GetAuthRequestAsync(id);
         }
 
+        public async Task<PasswordlessLoginResponse> GetPasswordlessLoginResponse(string id, string accessCode)
+        {
+            return await _apiService.GetAuthResponseAsync(id, accessCode);
+        }
```
fedemkr, line 491: Is this really needed, can't the caller directly use the `IApiService` ?

src/Core/Services/PclCryptoFunctionService.cs
```diff
@@ -172,6 +172,55 @@ public Task<byte[]> RsaDecryptAsync(byte[] data, byte[] privateKey, CryptoHashAl
...
+                    break;
+                case EncryptionType.Rsa2048_OaepSha1_B64:
+                case EncryptionType.Rsa2048_OaepSha1_HmacSha256_B64:
+                    break;
+                default:
+                    throw new Exception("encType unavailable.");
+            }
+
+            return await RsaDecryptAsync(data, privateKey, alg);
+        }
```
fedemkr, line 222: This is the same as `CryptoService -> RsaDecryptAsync(string encValue)` but with an extra parameter. Please don't duplicate code and instead in the one in `CryptoService`, add the parameter `byte[] privateKey` with `null` as default in the parameter, i.e. `byte[] privateKey = null`; and inside you can have:
```c#
...
var data = Convert.FromBase64String(encPieces[0]);
if(privateKey is null)
{
    privateKey = await GetPrivateKeyAsync();
}
if (privateKey is null)
{
    throw new Exception("No private key.");
}
...
```
Then you can call it with or without the `privateKey` and everything works with this minimum change on the already existing method.
Also if you can, add more verbosity to the exception messages

fedemkr (changes requested):

src/Core/Services/AuthService.cs
andrebispo5: I left it to be easier for a dev who is using Passwordless methods to have them all in one place and with intuitive naming.

src/Core/Services/PclCryptoFunctionService.cs
andrebispo5: You are totally right i copied the code to test decryption and forgot to refactor it, good catch!

src/App/Pages/Accounts/LoginPage.xaml.cs
```diff
@@ -26,9 +26,9 @@ public LoginPage(string email = null, AppOptions appOptions = null)
             _vm.Page = this;
             _vm.StartTwoFactorAction = () => Device.BeginInvokeOnMainThread(async () => await StartTwoFactorAsync());
             _vm.LogInSuccessAction = () => Device.BeginInvokeOnMainThread(async () => await LogInSuccessAsync());
+            _vm.LogInWithDeviceAction = async () => await StartLoginWithDeviceAsync();
```
fedemkr, line 29: No need to add `async/await`, just do `FireAndForget()` or invoke on main thread if needed

src/App/Pages/Accounts/LoginPage.xaml.cs
```diff
@@ -121,6 +121,19 @@ private void LogInSSO_Clicked(object sender, EventArgs e)
                 _vm.StartSsoLoginAction();
             }
         }
+        private void LogInWithDevice_Clicked(object sender, EventArgs e)
+        {
+            if (DoOnce())
+            {
+                _vm.LogInWithDeviceAction();
+            }
+        }
```
fedemkr, line 130: If I'm not mistaken, this is no longer used, please remove it

src/App/Pages/Accounts/LoginPageViewModel.cs
```diff
@@ -54,6 +54,7 @@ public LoginPageViewModel()
             TogglePasswordCommand = new Command(TogglePassword);
             LogInCommand = new Command(async () => await LogInAsync());
             MoreCommand = new AsyncCommand(MoreAsync, onException: _logger.Exception, allowsMultipleExecutions: false);
+            LogInWithDeviceCommand = new AsyncCommand(async () => await Device.InvokeOnMainThreadAsync(LogInWithDeviceAction), onException: _logger.Exception, allowsMultipleExecutions: false);
```
fedemkr, line 57: No need for the `async/await` keywords here

src/App/Pages/Accounts/LoginPageViewModel.cs
```diff
@@ -114,11 +115,13 @@ public bool IsKnownDevice
         public Command LogInCommand { get; }
         public Command TogglePasswordCommand { get; }
         public ICommand MoreCommand { get; internal set; }
+        public AsyncCommand LogInWithDeviceCommand { get; }
```
fedemkr, line 118: If possible, just use `ICommand` for the definition of the command property unless you need some access to something specific on the `AsyncCommand`. This way, if then you want to change the command implementation you only have to do it once, like when we migrate to MAUI

src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml
```diff
@@ -0,0 +1,76 @@
...
+    xmlns:u="clr-namespace:Bit.App.Utilities"
+    x:DataType="pages:LoginPasswordlessRequestViewModel"
+    Title="{Binding PageTitle}">
+
+    <ContentPage.BindingContext>
+        <pages:LoginPasswordlessRequestViewModel />
+    </ContentPage.BindingContext>
+
+    <ContentPage.ToolbarItems>
+        <ToolbarItem Text="{u:I18n Close}" Clicked="Close_Clicked" Order="Primary" Priority="-1" x:Name="_closeItem"/>
```
fedemkr, line 17: Convert `Clicked` to `Command`

src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml
```diff
@@ -0,0 +1,76 @@
...
+                    VerticalTextAlignment="End"/>
+                <Label
+                    Text="{u:I18n ViewAllLoginOptions}"
+                    StyleClass="text-md"
+                    VerticalTextAlignment="End"
+                    VerticalOptions="CenterAndExpand"
+                    Margin="5, 0"
+                    TextColor="{DynamicResource HyperlinkColor}">
+                    <Label.GestureRecognizers>
+                        <TapGestureRecognizer Tapped="Close_Clicked" />
```
fedemkr, line 69: Convert `Tapped` to `Command`

src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
```diff
@@ -0,0 +1,73 @@
...
+            _vm.StopCheckLoginRequestStatus();
+        }
+
+        private void Close_Clicked(object sender, System.EventArgs e)
+        {
+            if (DoOnce())
+            {
+                _vm.CloseAction();
+            }
+        }
```
fedemkr, line 48: Move this to a `Command` in the VM

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,196 @@
...
+        private const int REQUEST_TIME_UPDATE_PERIOD_IN_SECONDS = 4;
+
+        private IDeviceActionService _deviceActionService;
+        private IAuthService _authService;
+        private ISyncService _syncService;
+        private II18nService _i18nService;
+        private IStateService _stateService;
+        private ICryptoFunctionService _cryptoFunctionService;
+        private ICryptoService _cryptoService;
+        private IPasswordGenerationService _passwordGenerationService;
```
fedemkr, line 34: These are no longer used, please remove

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,196 @@
...
+
+                var authResult = await _authService.LogInPasswordlessAsync(Email, _requestAccessCode, _requestId, _requestKeyPair.Item2, response.Key, response.MasterPasswordHash);
+                await AppHelpers.ResetInvalidUnlockAttemptsAsync();
+
+                if (!await HandleCaptchaAsync(authResult.CaptchaSiteKey, authResult.CaptchaNeeded, CheckLoginRequestStatus))
+                {
+                    return;
+                }
+
+                await _deviceActionService.HideLoadingAsync();
```
fedemkr, line 146: Not quite sure where the loading is starting, but if you hide it here you need to hide it as well before the `return` in the if

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,196 @@
...
+            _cryptoFunctionService = ServiceContainer.Resolve<ICryptoFunctionService>();
+            _cryptoService = ServiceContainer.Resolve<ICryptoService>();
+            _passwordGenerationService = ServiceContainer.Resolve<IPasswordGenerationService>();
+            _logger = ServiceContainer.Resolve<ILogger>();
+
+            PageTitle = AppResources.LogInWithAnotherDevice;
+
+            CreatePasswordlessLoginCommand = new AsyncCommand(() => Device.InvokeOnMainThreadAsync(CreatePasswordlessLoginAsync),
+                onException: ex => HandleException(ex),
+                allowsMultipleExecutions: false);
```
fedemkr, line 70: If possible, move the invoking to the main thread on the parts actually needed inside `CreatePasswordlessLoginAsync`. E.g. the call to `await _authService.PasswordlessCreateLoginRequestAsync(...)` shouldn't happen in the main thread

src/App/Pages/CaptchaProtectedViewModel.cs
```diff
@@ -16,6 +16,23 @@ public abstract class CaptchaProtectedViewModel : BaseViewModel
...
+            }
+
+            if(await HandleCaptchaAsync(captchaSiteKey))
+            {
+                await onSuccess();
+                _captchaToken = null;
+                return false;
+            }
+            return false;
+        }
```
fedemkr, line 34: IMO, given the name, it should return `true` if the captcha is needed and `false` if not.
Also, given that the last two `return` are the same then you only need the last one.

src/App/Resources/AppResources.resx
```diff
@@ -2491,4 +2491,22 @@ Do you want to switch to this account?</value>
...
   </data>
+  <data name="LogInInitiated" xml:space="preserve">
+    <value>Log in initiated</value>
+  </data>
+  <data name="ANotificationHasBeenSentToYourDevice" xml:space="preserve">
+    <value>A notification has been sent to your device.</value>
+  </data>
+  <data name="PleaseMakeSureYourVaultIsUnlockedAndTheFingerprintPhraseMatchesOnTheOtherDevice" xml:space="preserve">
+    <value>Please make sure your vault is unlocked and the Fingerprint phrase matches on the other device.</value>
+  </data>
```
fedemkr, line 2502: Are the periods in the end needed? I don't usually put them

src/Core/Models/Request/TokenRequest.cs
```diff
@@ -15,13 +15,14 @@ public class TokenRequest
...
         public string Token { get; set; }
+        public string AuthRequestId { get; set; }
         public TwoFactorProviderType? Provider { get; set; }
         public bool? Remember { get; set; }
         public string CaptchaToken { get; set; }
         public DeviceRequest Device { get; set; }
 
         public TokenRequest(string[] credentials, string[] codes, TwoFactorProviderType? provider, string token,
-            bool? remember, string captchaToken, DeviceRequest device = null)
+            bool? remember, string captchaToken, string authRequestId = null, DeviceRequest device = null)
```
fedemkr, line 25: Put the new parameter to the end, I see you had to add explicit parameter naming on several places because of this change. If you move it to the end, then all places stay as is except the ones where you need to pass the `authRequestId`. When you do that remove the explicit parameter naming of `device` when there's no need of it

fedemkr (changes requested):

src/App/Resources/AppResources.resx
andrebispo5: ye the designs had periods. 
image

src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
fedemkr, line 48: Remove this method given that it's no longer used, or is it?

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -0,0 +1,193 @@
...
+        protected override IDeviceActionService deviceActionService => _deviceActionService;
+        protected override IPlatformUtilsService platformUtilsService => _platformUtilsService;
+
+        private CancellationTokenSource _checkLoginRequestStatusCts;
+        private Task _checkLoginRequestStatusTask;
+        private string _fingerprintPhrase;
+        private string _email;
+        private string _requestId;
+        private string _requestAccessCode;
+        private Tuple<byte[], byte[]> _requestKeyPair;
```
fedemkr, line 47: It could be good to have a comment here to state what is the Item1 and what is the Item2 of the Tuple

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
fedemkr: @andrebispo5 did you mark this as resolved because it's not possible to move the invocation to the main thread to inside the method's parts where it's needed?

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
andrebispo5: I'm sorry, i was doing some trial and error changes and ended up confusing myself. It's resolved now, promise!

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -165,7 +166,7 @@ private async Task CheckLoginRequestStatus()
 
         private async Task CreatePasswordlessLoginAsync()
         {
-            await _deviceActionService.ShowLoadingAsync(AppResources.Loading);
+            await Device.InvokeOnMainThreadAsync(async () => await _deviceActionService.ShowLoadingAsync(AppResources.Loading));
```
fedemkr, line 169: No need for `async/await` inside.
Also don't you need invoking on main thread when setting the `FingerprintPhrase`, are you getting the property changed on the view?

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
andrebispo5: Yes the Fingerprint Phrase is updating and working fine.

fedemkr (approved):

PR #2178, opened 2022-11-09
Remove 'Remove Account' option from login page

differsthecat:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Remove options from the login page altogether as the items in it are redundant with the updates.

## Code changes

- **src/App/Pages/Accounts/LoginPage.xaml:** Remove options
- **src/App/Pages/Accounts/LoginPage.xaml.cs:** Remove Android specific check on the removeAccount option
- **src/App/Pages/Accounts/LoginPageViewModel.cs:** Remove options

## Screenshots
Screen Shot 2022-11-09 at 10 34 36 AM

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPage.xaml
+++ b/src/App/Pages/Accounts/LoginPage.xaml
@@ -31,15 +31,6 @@
     <ContentPage.Resources>
         <ResourceDictionary>
             <u:InverseBoolConverter x:Key="inverseBool" />
-            <ToolbarItem Icon="more_vert.png" Clicked="More_Clicked" Order="Primary"
-                x:Name="_moreItem" x:Key="moreItem"
-                 AutomationProperties.IsInAccessibleTree="True"
-                 AutomationProperties.Name="{u:I18n Options}" />
-            <ToolbarItem Text="{u:I18n GetPasswordHint}"
-                x:Key="getPasswordHint"
-                x:Name="_getPasswordHint"
-                Clicked="Hint_Clicked" 
-                Order="Secondary" />
 
             <ScrollView x:Name="_mainLayout" x:Key="mainLayout">
                 <StackLayout Spacing="0">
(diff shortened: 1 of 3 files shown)


andrebispo5 (changes requested):

differsthecat:
> `Remove Account` is for accounts that are already logged in (it's part of account switching). I'm actually a bit confused as to how we're able to get to this screen for a user that's already logged in though, which is why the original "remember me" functionality was removed when account switching was added.

I could be wrong here, but users should have still seen this screen before when adding an account. Does that sound right @andrebispo5 ?

mp-bw:
> > `Remove Account` is for accounts that are already logged in (it's part of account switching). I'm actually a bit confused as to how we're able to get to this screen for a user that's already logged in though, which is why the original "remember me" functionality was removed when account switching was added.
> 
> I could be wrong here, but users should have still seen this screen before when adding an account. Does that sound right @andrebispo5 ?

On other clients yes, but mobile handled it differently to support quick account switching back to an account that was logged out via timeout vs manually logging out (which is how those accounts remained in the account switching list when they timed out).

differsthecat:
> > > `Remove Account` is for accounts that are already logged in (it's part of account switching). I'm actually a bit confused as to how we're able to get to this screen for a user that's already logged in though, which is why the original "remember me" functionality was removed when account switching was added.
> > 
> > 
> > I could be wrong here, but users should have still seen this screen before when adding an account. Does that sound right @andrebispo5 ?
> 
> On other clients yes, but mobile handled it differently to support quick account switching back to an account that was logged out via timeout vs manually logging out (which is how those accounts remained in the account switching list when they timed out).

I see what you are saying now, I just did some testing with it. If we remove the 'Remove Account' option here, a user can't remove the user unless they log in and then log back out. @andrebispo5 This will require more thought.

PR #2184, opened 2022-11-10
[SG-813] Not You? crashes app after vault logout timeout

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If your account is set to logout on vault time-out, once your vault times out, if you click on “not you?” from the master password screen, the app crashes.

### Steps to Reproduce:

1. Login to account
2. Set “vault timeout action” to “Logout”
3. Let your vault timeout so it logs you out
4. Click “not you?”

**Expected:** Taken to email screen

**Actual:** App crashes

## Code changes

Renamed HomePage parameter `checkRememberedEmail` to `checkNavigateToLogin`, also made the change on iOS Extensions.

**src/App/Utilities/AccountManagement/AccountsManager.cs:**
- Add account now navigates to `HomePage`. 
- `VaultTimeout` actions now navigates to the `HomePage` with a set remembered email as the VaultTimeout account, and with redirect to login page. This allows the user to click `Not you?` and go back to the `HomePage`.
## Screenshots
IMG_5730
IMG_5731
IMG_2CC606322E06-1

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/HomePageViewModel.cs
+++ b/src/App/Pages/Accounts/HomePageViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
+using Bit.App.Abstractions;
 using Bit.App.Controls;
 using Bit.App.Resources;
 using Bit.App.Utilities;
@@ -25,6 +26,7 @@ namespace Bit.App.Pages
         private IPlatformUtilsService _platformUtilsService;
         private ILogger _logger;
         private IEnvironmentService _environmentService;
+        private IAccountsManager _accountManager;
 
         public HomeViewModel()
         {
@@ -33,6 +35,7 @@ namespace Bit.App.Pages
             _platformUtilsService = ServiceContainer.Resolve<IPlatformUtilsService>();
             _logger = ServiceContainer.Resolve<ILogger>();
             _environmentService = ServiceContainer.Resolve<IEnvironmentService>();
+            _accountManager = ServiceContainer.Resolve<IAccountsManager>();
 
             PageTitle = AppResources.Bitwarden;
 
@@ -70,7 +73,7 @@ namespace Bit.App.Pages
 
         public bool CanContinue => !string.IsNullOrEmpty(Email);
 
-        public bool CheckNavigateToLogin { get; set; }
+        public bool ShouldCheckRememberEmail { get; set; }
 
         public FormattedString CreateAccountText
         {
@@ -109,11 +112,11 @@ namespace Bit.App.Pages
 
         public void CheckNavigateLoginStep()
         {
-            if (CheckNavigateToLogin && RememberEmail)
+            if (ShouldCheckRememberEmail && RememberEmail)
             {
                 StartLoginAction();
             }
-            CheckNavigateToLogin = false;
+            ShouldCheckRememberEmail = false;
         }
 
         public async Task ContinueToLoginStepAsync()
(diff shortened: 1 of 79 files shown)


src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -140,5 +152,17 @@ public async Task ContinueToLoginStepAsync()
...
+        {
+            var switchToAccount = await _platformUtilsService.ShowDialogAsync(
+                AppResources.SwitchToAlreadyAddedAccountConfirmation,
+                AppResources.AccountAlreadyAdded, AppResources.Yes, AppResources.Cancel);
+            if (switchToAccount)
+            {
+                await _stateService.SetActiveUserAsync(userId);
+                _messagingService.Send("switchedAccount");
+            }
+        }
```
fedemkr, line 166: Please don't duplicate code, this is the same as in `LoginPageViewModel`. Move this to `AccountsManager` so that it can be reused in both places

src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -107,11 +109,11 @@ public async Task InitAsync()
...
         public void CheckNavigateLoginStep()
         {
-            if (CheckHasRememberedEmail && RememberEmail)
+            if (CheckNavigateToLogin && RememberEmail)
             {
                 StartLoginAction();
             }
-            CheckHasRememberedEmail = false;
+            CheckNavigateToLogin = false;
         }
```
fedemkr, line 117: I'd change this and the `CheckNavigateToLogin` property and home parameter to something more meaningful to the business like `ShouldCheckRememberEmail` and the method could be `CheckRememberEmail()` or something like that so we focus on the business rules rather than on the outcome (which is navigate to login).
Also it gives the reader a better idea of what the check is about.
What do you think?

src/App/Pages/Accounts/LoginPageViewModel.cs
```diff
@@ -249,8 +253,7 @@ public async Task LogInAsync(bool showLoading = true, bool checkForExistingAccou
 
         private async Task MoreAsync()
         {
-            var emailExists = _stateService.AccountViews != null && _stateService.AccountViews.Any(e => e.Email == Email);
-            var buttons = IsEmailEnabled || !emailExists
+            var buttons = IsEmailEnabled || string.IsNullOrEmpty(await _stateService.GetUserIdAsync(Email))
```
fedemkr, line 256: Can't you use `EmailIsInSavedAccounts` here? instead of making the `GetUserIdAsync` call

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPageViewModel.cs
andrebispo5: Actually what i wanted was the **Active**UserEmail

src/Core/Services/StateService.cs
```diff
@@ -46,6 +46,19 @@ public async Task<string> GetActiveUserIdAsync()
...
+            await CheckStateAsync();
+
+            var activeUserId = _state?.ActiveUserId;
+            if (activeUserId == null)
+            {
+                var state = await GetStateFromStorageAsync();
+                activeUserId = state?.ActiveUserId;
+            }
+            return await GetEmailAsync(activeUserId);
+        }
```
fedemkr, line 60: Don't duplicate code. Either call from the caller:
```c#
var activeUserId = await _stateService.GetActiveUserIdAsync();
var activeUserEmail = await _stateService.GetEmailAsync(activeUserId);
```
or here just do:
```c#
public async Task<string> GetActiveUserEmailAsync()
{
    var activeUserId = await GetActiveUserIdAsync();
    return await GetEmailAsync(activeUserId);
}
```

fedemkr (changes requested):

src/Core/Services/StateService.cs
andrebispo5: my bad...

fedemkr (approved):

fedemkr (dismissed):

fedemkr (approved):

PR #2189, opened 2022-11-13
[SG-823] Set push token on login

trmartin4:
https://bitwarden.atlassian.net/browse/SG-823

## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
One of the ways that a push token can be sent to the server (to subsequently be passed to the Azure Notification Hub) is on a login request.  The server-side code will check for the Device information on the `POST` body and update the push token for the given user and device, if it exists.

This change is to populate the `PushToken` property on that object before it is sent on the request.  Previously, all properties were set on the object except for the push token.

## Code changes

**DeviceRequest.cs:** Added parameter to constructor for push token.
**AuthService.cs:** Looked up current push token from state service and passed that in to the object that is used to send the push token on the login request.
**PushNotificationListenerService.cs:**: Previously, `SetCurrentPushTokenAsync()` was only called if we were on an Android device, as it was only used for handling Android's re-generation of tokens (we compare Registered vs. Current tokens and update when they don't match).  However, now we need a way to keep the current token in state even for iOS devices, as we'll be using it on login requests.

https://github.com/bitwarden/server/pull/2404 is also associated with this same problem.

## Before you submit

### DIFF
--- a/src/App/Services/PushNotificationListenerService.cs
+++ b/src/App/Services/PushNotificationListenerService.cs
@@ -187,10 +187,8 @@ namespace Bit.App.Services
                 Debug.WriteLine($"{TAG} Registered device with server.");
 
                 await _stateService.Value.SetPushLastRegistrationDateAsync(DateTime.UtcNow);
-                if (deviceType == Device.Android)
-                {
-                    await _stateService.Value.SetPushCurrentTokenAsync(token);
-                }
+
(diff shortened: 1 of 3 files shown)


gbubemismith (approved):
This looks good

andrebispo5 (approved):

PR #2191, opened 2022-11-15
[SG-816] Get all login requests and pick the most recent

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If the user for some reason did not receive the notification to login with the device, sync the vault will fetch all login requests and pick the most recent and valid one.

## Code changes

* **SyncService.cs:** when vault sync is triggered, get all login requests and set in the stateService the most recent. After that triggers a normal passwordless login request flow.

## Before you submit

### DIFF
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -147,7 +147,6 @@ namespace Bit.App
                     }
                     else if (message.Command == Constants.PasswordlessLoginRequestKey
                         || message.Command == "unlocked"
-                        || message.Command == "syncCompleted"
                         || message.Command == AccountsManagerMessageCommands.ACCOUNT_SWITCH_COMPLETED)
                     {
                         lock (_processingLoginRequestLock)
@@ -156,12 +155,42 @@ namespace Bit.App
                             CheckPasswordlessLoginRequestsAsync().Wait();
                         }
                     }
+                    else if (message.Command == "syncCompleted")
+                    {
+                        await SyncPasswordlessLoginRequestsAsync();
+                    }
                 }
                 catch (Exception ex)
                 {
                     LoggerHelper.LogEvenIfCantBeResolved(ex);
                 }
             });
+        }
+
+        private async Task SyncPasswordlessLoginRequestsAsync()
+        {
+            var loginRequests = await _authService.GetPasswordlessLoginRequestsAsync();
+            if (loginRequests == null || !loginRequests.Any())
+            {
+                return;
+            }
+
+            var validLoginRequests = loginRequests.Where(l => l.RequestApproved == null && l.ResponseDate == null
+                && l.CreationDate.ToUniversalTime().AddMinutes(Constants.PasswordlessNotificationTimeoutInMinutes) > DateTime.UtcNow).ToList();
+
+            if (validLoginRequests == null)
+            {
+                return;
+            }
+
+            var validLoginRequest = validLoginRequests.OrderByDescending(x => x.CreationDate).First();
+            await _stateService.SetPasswordlessLoginNotificationAsync(new PasswordlessRequestNotification()
+            {
+                Id = validLoginRequest.Id,
+                UserId = await _stateService.GetActiveUserIdAsync()
+            });
+
+            _messagingService.Send(Constants.PasswordlessLoginRequestKey);
         }
 
         private async Task CheckPasswordlessLoginRequestsAsync()
--- a/src/App/App.xaml.cs
+++ b/src/App/App.xaml.cs
@@ -169,6 +169,13 @@ namespace Bit.App
 
         private async Task SyncPasswordlessLoginRequestsAsync()
         {
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
+            // if the user has not enabled passwordless logins ignore requests
+            if (!await _stateService.GetApprovePasswordlessLoginsAsync(activeUserId))
+            {
+                return;
+            }
+
             var loginRequests = await _authService.GetPasswordlessLoginRequestsAsync();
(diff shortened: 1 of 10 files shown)


src/App/App.xaml.cs
```diff
@@ -164,6 +167,39 @@ public App(AppOptions appOptions)
...
+
+            var validLoginRequests = loginRequests.Where(l => l.RequestApproved == null && l.ResponseDate == null
+                && l.CreationDate.ToUniversalTime().AddMinutes(Constants.PasswordlessNotificationTimeoutInMinutes) > DateTime.UtcNow).ToList();
+
+            if (validLoginRequests == null || !validLoginRequests.Any())
+            {
+                return;
+            }
+
+            var validLoginRequest = validLoginRequests.OrderByDescending(x => x.CreationDate).First();
```
fedemkr, line 193: This can be written as:
```c#
var validLoginRequest = loginRequests.Where(l => l.RequestApproved == null 
                                                 && l.ResponseDate == null
                                                 && l.CreationDate.ToUniversalTime().AddMinutes(Constants.PasswordlessNotificationTimeoutInMinutes) > DateTime.UtcNow)
                                     .OrderByDescending(x => x.CreationDate)
                                     .FirstOrDefault();

if (validLoginRequest is null)
{
    return;
}
```

src/App/App.xaml.cs
```diff
@@ -156,6 +155,10 @@ public App(AppOptions appOptions)
                             CheckPasswordlessLoginRequestsAsync().Wait();
                         }
                     }
+                    else if (message.Command == "syncCompleted")
+                    {
+                        await SyncPasswordlessLoginRequestsAsync();
+                    }
```
fedemkr, line 161: I think this should be directly placed on the `SyncService` so everything related to syncing stuff is there and also the `App.xaml.cs` should not be the place to do this logic

fedemkr (changes requested):

src/App/App.xaml.cs
andrebispo5: Agreed will move it there.

src/Core/Services/SyncService.cs
```diff
@@ -382,5 +383,39 @@ private async Task SyncSendsAsync(string userId, List<SendResponse> response)
...
+                return;
+            }
+
+            var loginRequests = await _apiService.GetAuthRequestAsync();
+            if (loginRequests == null || !loginRequests.Any())
+            {
+                return;
+            }
+
+            var validLoginRequest = loginRequests.Where(l => l.RequestApproved == null
```
trmartin4, line 401: Is this the only place in the codebase where this logic is done?  It seems like it would be prone to possibly change in the future (if we end up needing another way to validate requests).  If the logic to determine what a "valid un-accepted request" is is duplicated, we might want to abstract that into a single method on a service somewhere.

src/Core/Services/SyncService.cs
```diff
@@ -382,5 +383,39 @@ private async Task SyncSendsAsync(string userId, List<SendResponse> response)
...
+            }
+
+            await _stateService.SetPasswordlessLoginNotificationAsync(new PasswordlessRequestNotification()
+            {
+                Id = validLoginRequest.Id,
+                UserId = userId
+            });
+
+            _messagingService.Send(Constants.PasswordlessLoginRequestKey);
+        }
```
fedemkr, line 419: We need to revise the `FullSyncAsync` method regarding exception handling and logging.
But for now could you add try...catch to your method that logs the exception and doesn't rethrow so that if an error happens syncing the passwordless login requests it doesn't break the full sync and given that this one doesn't rely, afaik, on the date of the last sync we can do it

fedemkr (changes requested):

src/Core/Services/SyncService.cs
fedemkr: IMO it can be moved directly into the login request model and have a `IsRequestValid` property or method

andrebispo5: Agree with @fedemkr the model can handle it on its own.

andrebispo5: But this is valid/invalid for the current context because we want unanswered requests. I'll create two properties, one for expired and another for answered.

src/Core/Models/Response/PasswordlessLoginResponse.cs
```diff
@@ -20,6 +20,11 @@ public class PasswordlessLoginResponse
         public string Origin { get; set; }
         public string RequestAccessCode { get; set; }
         public Tuple<byte[], byte[]> RequestKeyPair { get; set; }
+
+        public bool IsAnswered => RequestApproved != null
+                                    && ResponseDate != null;
```
fedemkr, line 25: There are just 2 properties and is a short line so inline the whole condition if possible

src/Core/Services/SyncService.cs
```diff
@@ -386,36 +386,41 @@ private async Task SyncSendsAsync(string userId, List<SendResponse> response)
...
+                    Id = validLoginRequest.Id,
+                    UserId = userId
+                });
 
-            _messagingService.Send(Constants.PasswordlessLoginRequestKey);
+                _messagingService.Send(Constants.PasswordlessLoginRequestKey);
+            }
+            catch (Exception ex)
+            {
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
```
fedemkr, line 422: if it can resolve the `_stateService` and the `_messagingService` it can resolve the logger, so inject it through the interface and use the instance here instead of the static helper method

fedemkr (changes requested):

fedemkr (approved):

PR #2200, opened 2022-11-19
Set push token state values to be user-specific

trmartin4:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Currently, we run into problems with push notifications with multiple accounts on a mobile device.  Namely, the second and subsequent accounts are not associated with a push token in Azure Notification Hub.  This means they won't get push notifications.

To solve this, my approach was to change two of the state values concerning push notifications to be stored at the user level.  

- The first one is `PushLastRegistrationDateKey`.  Setting this for each user ensures that each user will request a token at application startup if it has been more than a day since the last push registration, instead of only doing it for the first user to open the app.
- The second one is `PushCurrentToken`.  The "current" token represents the token that has been sent (or will be sent) to the back-end API to tie to the current user.  Storing this at the user level allows us to update push tokens at the user level.

## Code changes

[src/Android/Services/AndroidPushNotificationService.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-5b4c20a27ab74adb487b94798fc08bc8b3167af2f716b99035f8a9c7731dbab6): Updated methods to use the current active user to retrieve the token and set the registration time.
[src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-bce7881e23e99baec99687aac256ad3693130709ab596dfa2e4ff75550aca8e0): Updated the settings menu used in `DEBUG` to show the user-specific last push registration date.
[src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-5405b2eb635ec5f08e97f08bc37a02138037a37c5bcb2f14805968d0c93176ba): Changed the code that runs on application launch to use the last push registration time for the current user.  This will allow each individual user to send their own individual token when they run the application.
[src/App/Services/MobileStorageService.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-2251882deae3df7c0dc316404d76f08b562f3e20f990579e1dd8e8b63f9391b2): Removed the constant values for the constants that are now user-level.
[src/App/Services/PushNotificationListenerService.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-f341b17c68e50eb5721c6ede71f0b15be72ffa3a6deca14cf9f4c3e6d5c43efc): In `OnRegisteredAsync()`, which fires after the registration for push notifications, the method change to only set the last push registration date for the current user, since the update to `/devices/{identifier}/token` is only updating the token for that user.
[src/Core/Abstractions/IStateService.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-40fa428b50a59435a4892cd559f8fba6e5cfd6798a3239c853ff67107e142773): Updated the interface to reflect the fact that the two state values are now user-level.
[src/Core/Constants.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-94998af117f486ce85a16234b1bf36dd8c918651ff4d5c851927d203f5120e4e)
[src/Core/Services/StateService.cs](https://github.com/bitwarden/mobile/compare/SG-823-push-registrations-by-user?expand=1#diff-c97a70998c406bbdf1908944e4a13377f0ec8180fecefbf2fdb88f58cbc54377): Updated the state definition to reflect the fact that the two state values are now user-level.

## Before you submit


src/Android/Services/AndroidPushNotificationService.cs
```diff
@@ -41,11 +41,13 @@ public Task<bool> AreNotificationsSettingsEnabledAsync()
...
         public async Task<string> GetTokenAsync()
         {
-            return await _stateService.GetPushCurrentTokenAsync();
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
+            return await _stateService.GetPushCurrentTokenAsync(activeUserId);
         }
 
         public async Task RegisterAsync()
         {
+            var activeUserId = await _stateService.GetActiveUserIdAsync();
```
fedemkr, line 50: The `activeUserId` is only used in the `else` so unless I'm missing something it's better to move this line inside the `else` to improve performance on the other condition

fedemkr (changes requested):
Except for the requested change, the other things are fine.
However, I think it can be improved so if you want you can use the same approach as in other places like in `StateService -> CanAccessPremiumAsync(string userId = null)` where the parameter can be `null` and if it's, it gets the active user id inside so that you don't have to get it on every place the push related methods are called. I think it's cleaner this way and increases code reuse

src/Android/Services/AndroidPushNotificationService.cs
trmartin4: I refactored to have a `null` default value for the state service methods, so this is no longer needed.

fedemkr (dismissed):

src/Core/Services/StateService.cs
```diff
@@ -1001,17 +1001,25 @@ public async Task SetPushInitialPromptShownAsync(bool? value)
             await SetValueAsync(key, value, options);
         }
 
-        public async Task<DateTime?> GetPushLastRegistrationDateAsync()
+        public async Task<DateTime?> GetPushLastRegistrationDateAsync(string userId = null)
         {
+            if (userId == null)
```
mp-bw, line 1006: These getters should follow the established pattern from other account-specific getters like so:

`var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync());`
`var key = Constants.PushLastRegistrationDateKey(reconciledOptions.UserId);`
`return await GetValueAsync<DateTime?>(key, reconciledOptions);`

src/Core/Services/StateService.cs
```diff
@@ -1001,17 +1001,25 @@ public async Task SetPushInitialPromptShownAsync(bool? value)
...
+                userId = await GetActiveUserIdAsync();
+            }
             var options = await GetDefaultStorageOptionsAsync();
-            var key = Constants.PushLastRegistrationDateKey;
+            var key = Constants.PushLastRegistrationDateKey(userId);
             return await GetValueAsync<DateTime?>(key, options);
         }
 
-        public async Task SetPushLastRegistrationDateAsync(DateTime? value)
+        public async Task SetPushLastRegistrationDateAsync(DateTime? value, string userId = null)
```
mp-bw, line 1015: And here's the pattern for account-specific setters:

`var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync());`
`var key = Constants.PushLastRegistrationDateKey(reconciledOptions.UserId);`
`await SetValueAsync(key, value, reconciledOptions);`

src/Core/Services/StateService.cs
```diff
@@ -1029,17 +1037,25 @@ public async Task SetPushInstallationRegistrationErrorAsync(string value)
             await SetValueAsync(key, value, options);
         }
 
-        public async Task<string> GetPushCurrentTokenAsync()
+        public async Task<string> GetPushCurrentTokenAsync(string userId = null)
         {
+            if (userId == null)
```
mp-bw, line 1042: Same here

src/Core/Services/StateService.cs
```diff
@@ -1029,17 +1037,25 @@ public async Task SetPushInstallationRegistrationErrorAsync(string value)
...
             var options = await GetDefaultStorageOptionsAsync();
-            var key = Constants.PushCurrentTokenKey;
+            var key = Constants.PushCurrentTokenKey(userId);
             return await GetValueAsync<string>(key, options);
         }
 
-        public async Task SetPushCurrentTokenAsync(string value)
+        public async Task SetPushCurrentTokenAsync(string value, string userId = null)
         {
+            if (userId == null)
```
mp-bw, line 1053: And here

src/Core/Services/StateService.cs
trmartin4: @mp-bw, the other examples for account-specific getters don't set the `userId` to be the active user if it comes in as `null`, right?  This was a recommendation that @fedemkr had, to take the lookup for active user out of the calling code.  I originally had the same pattern, but it required doing the check for active user outside of each call to these methods.

I'm fine either way, I just want to make sure it fits with best practices.

mp-bw: By design all calls to StateService methods automatically use the active userId if a `null` userId is sent from the calling code (this happens in `ReconcileOptions` via the object returned from `GetDefaultStorageOptionsAsync`).  I think Fede's suggestion was to follow that pattern, but I'm not sure if I'm misreading something.  @fedemkr ?

mp-bw: > I'd keep the userId = null parameter on the methods though.

Yes, that is required for this design to work.

trmartin4: Thank you both, and apologies for misunderstanding the recommendation @fedemkr .  I've corrected this, and it is much cleaner now.

fedemkr (changes requested):
Instead of the line
```c#
var options = await GetDefaultStorageOptionsAsync();
```
you should use what Matt suggested:
```c#
var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync());
```

and afterwards use `reconciledOptions.UserId` when the user id is needed.

So if a user is indeed passed through the parameter it's taken into account when getting the value

src/Core/Services/StateService.cs
```diff
@@ -1001,17 +1001,17 @@ public async Task SetPushInitialPromptShownAsync(bool? value)
             await SetValueAsync(key, value, options);
         }
 
-        public async Task<DateTime?> GetPushLastRegistrationDateAsync()
+        public async Task<DateTime?> GetPushLastRegistrationDateAsync(string userId = null)
         {
-            var options = await GetDefaultStorageOptionsAsync();
-            var key = Constants.PushLastRegistrationDateKey;
+            var options = ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync());
```
mp-bw, line 1006: This should be named `reconciledOptions` to distinguish it from a default options object

src/Core/Services/StateService.cs
```diff
@@ -1001,17 +1001,17 @@ public async Task SetPushInitialPromptShownAsync(bool? value)
...
+            var key = Constants.PushLastRegistrationDateKey(userId);
```
mp-bw, line 1007: You have to reference `reconciledOptions.UserId` here otherwise you'll never get the current active userId

mp-bw (dismissed):

fedemkr (approved):

trmartin4:
@fedemkr can you take a look and re-approve if you're still OK with the change?  I had a merge conflict from `master` from an additional constant added for the Apple Watch integration.

fedemkr (approved):

PR #2205, opened 2022-11-25
[SG-601] Enhance experience when user denies camera access in Authenticator

andrebispo5:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

When a user denies camera access for the QR code scan, they are neither prompted again nor are they given a modal that says ‘You must enable camera access…’ etc. when they tap the icon to add an authenticator code to an item that does not yet have one. This occurs in Prod as well; the camera icon/Set up TOTP buttons do not respond when selected if you have previously denied the setting.This poses an issue in the Authenticator UI because there is no way to manually add a new TOTP without accessing those screens, however, you can edit an existing one.

Expected behavior: When the user has camera permissions turned off, clicking the “add new TOTP” button launches the “Enter Key Manually” mode. If the user then clicks the “scan QR Code” link (attempting to switch to camera mode) we show the user the dialog requesting camera permissions.

## Code changes

**ScanPage.xaml**: Changed ScannerView to a container. The scanner is only added if the user has permissions to use the camera.
**ScanPage.xaml.cs**: Added couple of code protections agains a null scanner. Also added a command to init the scanner from the ViewModel.
**ScanViewModel.cs**: When executing the ToggleScanModeCommand it now checks if the user has camera permissions. If not it will prompt the user to go to settings and enable them. If it has permissions tries to init the scanner if needed and changes the UI.

## Screenshots

https://user-images.githubusercontent.com/4648522/204020241-c9501aa7-6569-4f4a-b9d1-e9fb83684f60.mov

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/ScanPage.xaml
+++ b/src/App/Pages/Vault/ScanPage.xaml
@@ -36,8 +36,6 @@
         </Grid.RowDefinitions>
         <ContentView 
             x:Name="_scannerContainer"
-            HorizontalOptions="FillAndExpand"
-            VerticalOptions="FillAndExpand"
             AutomationId="zxingScannerView"
             IsVisible="{Binding ShowScanner}"
             Grid.Column="0"
--- a/src/App/Pages/Vault/ScanPage.xaml
+++ b/src/App/Pages/Vault/ScanPage.xaml
@@ -34,16 +34,15 @@
             <RowDefinition Height="*" />
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
-        <zxing:ZXingScannerView
-            x:Name="_zxing"
+        <ContentView 
+            x:Name="_scannerContainer"
             HorizontalOptions="FillAndExpand"
             VerticalOptions="FillAndExpand"
             AutomationId="zxingScannerView"
             IsVisible="{Binding ShowScanner}"
             Grid.Column="0"
             Grid.Row="0"
-            Grid.RowSpan="3"
-            OnScanResult="OnScanResult"/>
+            Grid.RowSpan="3"/>
         <StackLayout
             VerticalOptions="Center"
             HorizontalOptions="FillAndExpand"
--- a/src/App/Pages/Vault/ScanPage.xaml.cs
+++ b/src/App/Pages/Vault/ScanPage.xaml.cs
@@ -55,64 +55,70 @@ namespace Bit.App.Pages
         protected override void OnAppearing()
         {
             base.OnAppearing();
-            if (_zxing != null)
+            if (_zxing == null)
             {
-                _zxing.IsScanning = true;
-
-                // Fix for Autofocus, now it's done every 2 seconds so that the user does't have to do it
-                // https://github.com/Redth/ZXing.Net.Mobile/issues/414
-                _autofocusCts?.Cancel();
-                _autofocusCts = new CancellationTokenSource(TimeSpan.FromMinutes(3));
-
-                var autofocusCts = _autofocusCts;
-                // this task is needed to be awaited OnDisappearing to avoid some crashes
-                // when changing the value of _zxing.IsScanning
-                _continuousAutofocusTask = Task.Run(async () =>
-                {
-                    try
-                    {
-                        while (!autofocusCts.IsCancellationRequested)
-                        {
-                            await Task.Delay(TimeSpan.FromSeconds(2), autofocusCts.Token);
-                            await Device.InvokeOnMainThreadAsync(() =>
-                            {
-                                if (!autofocusCts.IsCancellationRequested)
-                                {
-                                    try
-                                    {
-                                        _zxing.AutoFocus();
-                                    }
-                                    catch (Exception ex)
-                                    {
-                                        _logger.Value.Exception(ex);
-                                    }
-                                }
-                            });
-                        }
-                    }
-                    catch (TaskCanceledException) { }
-                    catch (Exception ex)
-                    {
-                        _logger.Value.Exception(ex);
-                    }
-                }, autofocusCts.Token);
-                _pageIsActive = true;
-                AnimationLoopAsync();
+                return;
             }
+
+            _zxing.OnScanResult += OnScanResult;
+            _zxing.IsScanning = true;
+
+            // Fix for Autofocus, now it's done every 2 seconds so that the user does't have to do it
+            // https://github.com/Redth/ZXing.Net.Mobile/issues/414
+            _autofocusCts?.Cancel();
+            _autofocusCts = new CancellationTokenSource(TimeSpan.FromMinutes(3));
+
+            var autofocusCts = _autofocusCts;
+            // this task is needed to be awaited OnDisappearing to avoid some crashes
+            // when changing the value of _zxing.IsScanning
+            _continuousAutofocusTask = Task.Run(async () =>
+            {
+                try
+                {
+                    while (!autofocusCts.IsCancellationRequested)
+                    {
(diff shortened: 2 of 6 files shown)


src/App/Pages/Vault/ScanPage.xaml.cs
```diff
@@ -50,66 +49,98 @@ public ScanPage(Action<string> callback)
...
 
+        private void InitScanner()
+        {
+            try
+            {
+                if (ViewModel.HasCameraPermission && ViewModel.ShowScanner && _zxing == null)
+                {
+                    _zxing = new ZXingScannerView();
+                    _zxing.OnScanResult -= OnScanResult;
+                    _zxing.OnScanResult += OnScanResult;
```
fedemkr, line 127: Here you are initializing a new `ZXingScannerView` so `_zxing.OnScanResult -= OnScanResult;` does nothing, given that it's a fresh new instance. Thus you're subscribing to `OnScanResult` but you never unsubscribe from that, thus you might have a leak here.

src/App/Pages/Vault/ScanPage.xaml.cs
```diff
@@ -26,20 +28,17 @@ public partial class ScanPage : BaseContentPage
         private bool _pageIsActive;
         private bool _qrcodeFound;
         private float _scale;
-
+        private ZXingScannerView _zxing;
         private readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>("logger");
 
-        public ScanPage(Action<string> callback)
+        public ScanPage(Action<string> callback, bool hasCameraPermission)
```
fedemkr, line 34: I don't think the caller should have the responsibility of telling this page whether the app has camera permission. IMO this Page/ViewModel should be the responsible of checking whether the camera is allowed, so I'd remove this parameter from the ctor and check the permission on the spot that is needed.

src/App/Resources/AppResources.resx
```diff
@@ -2512,4 +2512,7 @@ Do you want to switch to this account?</value>
   <data name="ThisRequestIsNoLongerValid" xml:space="preserve">
     <value>This request is no longer valid</value>
   </data>
+  <data name="EnableCamerPermissionsToUseTheScanner" xml:space="preserve">
+    <value>Enable camera permissions to use the scanner</value>
+  </data>
```
fedemkr, line 2517: Typo (Camer -> Camera) and I think permissions should be singular: `EnableCameraPermissionToUseTheScanner`

src/App/Pages/Vault/ScanPageViewModel.cs
```diff
@@ -39,6 +53,23 @@ public bool ShowScanner
...
+            {
+                var openAppSettingsResult = await _platformUtilsService.ShowDialogAsync(AppResources.EnableCamerPermissionsToUseTheScanner, title: string.Empty, confirmText: AppResources.Settings, cancelText: AppResources.NoThanks);
+                if (openAppSettingsResult)
+                {
+                    _deviceActionService.OpenAppSettings();
+                    return;
+                }
+            }
+            ShowScanner = !ShowScanner;
+            InitScannerCommand.Execute(null);
```
fedemkr, line 70: I don't like much having a command that has its action on the view; I think it might cause to never dispose correctly the ViewModel because of the strong reference to its action.
We could have something like [MvxInteraction](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/ViewModels/MvxInteraction.cs) of MvvmCross but we'll need to discuss it and address this throughout the app. For now we can leave it like that given that there are already other things that hold strong references

fedemkr (changes requested):

src/App/Pages/Vault/ScanPage.xaml
```diff
@@ -34,16 +34,15 @@
...
+            x:Name="_scannerContainer"
             HorizontalOptions="FillAndExpand"
             VerticalOptions="FillAndExpand"
             AutomationId="zxingScannerView"
             IsVisible="{Binding ShowScanner}"
             Grid.Column="0"
             Grid.Row="0"
-            Grid.RowSpan="3"
-            OnScanResult="OnScanResult"/>
+            Grid.RowSpan="3"/>
```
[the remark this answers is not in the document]
andrebispo5: Oh ye, `LazyView` might do the trick! Let me try it out.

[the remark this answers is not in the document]
andrebispo5: After trying it out i don't think the `LazyView` will be very useful in this case. I would have to do a custom implementation with the initialization just for this one view. If we were to use the scanner in other places it might be useful but in this case i don't think so.

src/App/Pages/Vault/ScanPage.xaml.cs
```diff
@@ -55,6 +52,53 @@ public ScanPage(Action<string> callback)
...
+                    TryInverted = true,
+                    DelayBetweenAnalyzingFrames = 5,
+                    DelayBetweenContinuousScans = 5
+                };
+                _scannerContainer.Content = _zxing;
+                StartScanner();
+            }
+            catch (Exception ex)
+            {
+                _logger?.Value.Exception(ex);
```
fedemkr, line 89: `_logger` is never `null`

src/App/Pages/Vault/ScanPage.xaml.cs
```diff
@@ -55,6 +52,53 @@ public ScanPage(Action<string> callback)
         protected override void OnAppearing()
         {
             base.OnAppearing();
+            StartScanner();
+        }
+
+        protected override void OnDisappearing()
+        {
+            StopScanner();
```
fedemkr, line 60: Add `FireAndForget()` otherwise if an exception happens is swallowed by the `Task` and we don't have the feedback

src/App/Pages/Vault/ScanPageViewModel.cs
```diff
@@ -9,13 +17,46 @@ public class ScanPageViewModel : BaseViewModel
...
         }
 
-        public Command ToggleScanModeCommand { get; set; }
+        public async Task InitAsync()
+        {
+            try
+            {
+                await Device.InvokeOnMainThreadAsync(async () =>
+                {
+                    var hasCameraPermission = await PermissionManager.CheckAndRequestPermissionAsync(new Permissions.Camera());
```
fedemkr, line 39: does this need to be done on main thread?

src/App/Pages/Vault/ScanPageViewModel.cs
```diff
@@ -9,13 +17,46 @@ public class ScanPageViewModel : BaseViewModel
...
+
+                if (!HasCameraPermission)
+                {
+                    return;
+                }
+                InitScannerCommand.Execute(null);
+            }
+            catch (System.Exception ex)
+            {
+                HandleException(ex);
```
fedemkr, line 52: In case an exception happens, does it make sense to stay on the screen? can the user retry somehow (if not it should go back showing some message)?

fedemkr (changes requested):

src/App/Pages/Vault/ScanPageViewModel.cs
andrebispo5: yes otherwise the dialog to ask for permissions won't show.

src/App/Pages/Vault/ScanPageViewModel.cs
andrebispo5: you can always enter the code manually by typing even though the scanner is not working.

fedemkr (approved):

PR #2209, opened 2022-11-29
[PS-1707] improve dark layout colors

aj-rosado:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
The color of callouts and switch hints (for android) weren't using theme colors, instead were using the platform defined color for light or dark theme and now they are using Theme Colors. 
The switch hint text on Android wasn't visible when using a dark Bitwarden theme on a Light OS Theme.
Changed the Color.Default to defines colors as they weren't returning always the expected color and this will be an obsolete feature for MAUI.

## Code changes

* **SetPasswordPage.xaml,UpdatePasswordPage.xaml, GeneratorPage.xaml, SendAddEditPage.xaml, SendAddOnlyPage.xaml, ExportVaultPage.xaml, SettingsPage.xaml, CipherAddEditPage.xaml:** Changed Accent color to use the PrimaryColor from DynamicResources.

* **CircularProgressbarView.cs, ExtendedSlider.cs, ExtendedStepper.cs:** Changed Color.Default to a defined color as Color.Default wasn't returning a color that would keep layout integrity.

* **GeneratorPage.xaml:** Added effect to change the switch hint color

* **ThemeHelpers.cs:** Validation for dark themes was missing the SolarizedDark theme

* **HintColorSwitchEffect.cs:** Effect that changes the hint color of a switch to the Theme MutedColor

## Screenshots
Android:
Screenshot_1669748794

iOS:
Simulator Screen Shot - iPhone SE (3rd generation) - 2022-11-29 at 19 05 49

## Before you submit


src/Android/Effects/HintColorSwitchEffect.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.ComponentModel;
+using Android.Graphics.Drawables;
+using Bit.Droid.Effects;
+using Bit.Droid.Utilities;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+
+[assembly: ExportEffect(typeof(HintColorSwitchEffect), "HintColorSwitchEffect")]
```
fedemkr, line 9: use `nameof(HintColorSwitchEffect)` instead of hardcoding

src/Android/Effects/HintColorSwitchEffect.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Bit.Droid.Effects
+{
+    public class HintColorSwitchEffect : PlatformEffect
+    {
+        protected override void OnAttached()
+        {
+            if (Control is Android.Widget.Switch switchView)
+            {
+                switchView.SetHintTextColor(ThemeHelpers.MutedColor);
```
fedemkr, line 18: If you change themes, does this update correctly dynamically?

src/App/Pages/Generator/GeneratorPage.xaml
```diff
@@ -445,7 +445,11 @@
                                 StyleClass="box-value"
                                 HorizontalOptions="End" 
                                 AutomationProperties.IsInAccessibleTree="True"
-                                AutomationProperties.Name="{u:I18n UppercaseAtoZ}"/>
+                                AutomationProperties.Name="{u:I18n UppercaseAtoZ}">
+                                <Switch.Effects>
+                                    <effects:HintColorSwitchEffect/>
+                                </Switch.Effects>
```
fedemkr, line 451: If this effect needs to be applied to all switches on the app, make an implicit style defined globally and apply the effect to the style. If it's only for this page you can define the implicit style on this specific page so you don't set the effect on each control individually (unless there is some switch that doesn't require it)

src/App/Controls/CircularProgressbarView.cs
```diff
@@ -21,13 +21,13 @@ public class CircularProgressbarView : SKCanvasView
             nameof(StrokeWidth), typeof(float), typeof(CircularProgressbarView), 3f);
 
         public static readonly BindableProperty ProgressColorProperty = BindableProperty.Create(
-            nameof(ProgressColor), typeof(Color), typeof(CircularProgressbarView), Color.Default);
+            nameof(ProgressColor), typeof(Color), typeof(CircularProgressbarView), Color.FromHex("175DDC"));
```
fedemkr, line 24: Define the Color on the styles and then use it here to avoid hardcoding and maybe this color can be reused elsewhere

src/App/Controls/CircularProgressbarView.cs
```diff
@@ -21,13 +21,13 @@ public class CircularProgressbarView : SKCanvasView
...
 
         public static readonly BindableProperty EndingProgressColorProperty = BindableProperty.Create(
-            nameof(EndingProgressColor), typeof(Color), typeof(CircularProgressbarView), Color.Default);
+            nameof(EndingProgressColor), typeof(Color), typeof(CircularProgressbarView), Color.FromHex("dd4b39"));
```
fedemkr, line 27: Define the Color on the styles and then use it here to avoid hardcoding and maybe this color can be reused elsewhere

src/App/Controls/ExtendedSlider.cs
```diff
@@ -5,7 +5,7 @@ namespace Bit.App.Controls
     public class ExtendedSlider : Slider
     {
         public static readonly BindableProperty ThumbBorderColorProperty = BindableProperty.Create(
-            nameof(ThumbBorderColor), typeof(Color), typeof(ExtendedSlider), Color.Default);
+            nameof(ThumbBorderColor), typeof(Color), typeof(ExtendedSlider), Color.FromHex("b5b5b5"));
```
fedemkr, line 8: Define the Color on the styles and then use it here to avoid hardcoding and maybe this color can be reused elsewhere

fedemkr (changes requested):

src/App/Pages/Generator/GeneratorPage.xaml
aj-rosado: This effect only affects switches that have an hint set (it is set by the AutomationProperties.Name for the Android app). There is another switch that doesn't need this, on this page, as it doesn't have a hint.

Now, we could apply this to all switches anyway as it wouldn't have a negative impact and if in the future a new switch was added and had a hint it would have the correct color, even if it is unnecessary on some. What's your opinion?

src/App/Controls/CircularProgressbarView.cs
aj-rosado: This is only necessary on a case where the style is not defined (or missing the part that affects this). So putting it on a style could reproduce the error. 

In fact this color is from the default light style, the default color was resetting to white, not having it set was resetting to white. This was a kind of "improvement" if that case happens, using a color from the style, even if hardcoded

fedemkr: But which is the error that it produces having color come from the Style here?

aj-rosado: It produces no error. This is a fail safe condition, this will only apply if not possible to get a color from styles.

src/App/Pages/Generator/GeneratorPage.xaml
aj-rosado: Removed the effect and added the content to the CustomSwitchRenderer that is applied globally

src/Android/Effects/HintColorSwitchEffect.cs
aj-rosado: Removed the effect and added the content to the CustomSwitchRenderer and now is updating dynamically on theme change

src/App/Controls/CircularProgressbarView.cs
fedemkr: Alright, we'll leave it like this for now and I'll revise it later

fedemkr (approved):

Issue #2163, opened 2022-11-01
QUERY_ALL_PACKAGES bypassed


BHydden:
### Steps To Reproduce

1. Go to #1384
2. See that `<action android:name="*" />` was introduced to bypass the official review channels associated with `QUERY_ALL_PACKAGES`

### Expected Result

Formal channels are followed to maintain integrity.

### Actual Result

Formal channels are bypassed to take app data without associated review process.

### Operating System

Android

### Build Version

2022.10.0 (5247)

### Beta

- [ ] Using a pre-release version of the application.

kspearrin:
Hey @BHydden , thanks for the report. As referenced in the original PR, this change was added to support the ability to launch other apps from within the Bitwarden vault, which is a typical path that many people follow when logging into other apps. Do you know of another way to support this use case?

BHydden:
> Do you know of another way to support this use case?

https://developer.android.com/training/package-visibility/declaring#all-apps

> In rare cases, your app might need to query or interact with all installed apps on a device, independent of the components they contain. To allow your app to see all other installed apps, the system provides the `QUERY_ALL_PACKAGES` permission.

lberrymage:
@kspearrin From the PR description:

> In order for PackageManager.GetLaunchIntentForPackage(..) to function in Android 11+, the package has to be visible to us.

This [method's documentation](https://developer.android.com/reference/android/content/pm/PackageManager#getLaunchIntentForPackage(java.lang.String)) notes the solution:

> Consider using [`getLaunchIntentSenderForPackage(java.lang.String)`](https://developer.android.com/reference/android/content/pm/PackageManager#getLaunchIntentSenderForPackage(java.lang.String)) if the caller is not allowed to query for the packageName.

This method would indeed work as a replacement.

> Returns an `IntentSender` that can be used to launch a front-door activity in a package. This is used, for example, to implement an "open" button when browsing through packages. The current implementation is the same with `getLaunchIntentForPackage(java.lang.String)`. Instead of returning the `Intent`, it returns the `IntentSender` which is not restricted by the package visibility.

This method is only available as of SDK 33, but it could at least be used on recent versions of Android with fallback to the old way based on `SDK_INT`.

mp-bw:
@lberrymage Thanks for the info.  We'll be testing `getLaunchIntentSenderForPackage` on API 33 here shortly, and if it works we'll make API 33+ a requirement for this feature going forward.

[...]

PR #2220, opened 2022-12-03
[BEEEP] Support for automatic TOTP token copy via external autofill (Android)

mp-bw:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

This adds support for automatically copying the TOTP token when selecting a cipher externally via the autofill API (inline or popup).  It also fixes the visibility issue described in #2163 (see issue for details).

## Code changes

* **AndroidManifest.xml/Android.csproj:** Bumped `targetSdkVersion` to 33 and AndroidX components to latest stable versions.  Also removed wildcard app visibility (see `DeviceActionService` changes below for additional details)
* **AutofillExternalSelectionActivity.cs:** Activity for handling external autofill requests via the Autofill API Authentication intent to pass off to our `AutofillHandler` for building/returning the correct data set and copying the TOTP to the clipboard
* **AutofillHandler.cs:** Added arg to specify if call to perform autofill is triggered externally and reference the correct activity if so, also modified calls to `BuildDataset` to include/not include the authentication intent for cipher data sets
* **AutofillHelpers.cs:** Added argument to include/not include the authentication intent in the cipher dataset (if it's included in the returned dataset the Autofill API falls into an infinite loop)
* **FilledItem.cs:** Added `Id` so we can obtain the TOTP token from the chosen cipher
* **ClipboardAlarmReceiver.cs:** Use `clipboardManager.ClearPrimaryClip()` for API 33+ (function was added in 28 but I'm seeing mixed results across devices, less so with 33)
* **ClipboardService.cs:** Switch to Application context for `ClipboardManager` functions to prevent the need for hacking proper activity selection into a service designed with a platform agnostic interface
* **DeviceActionService.cs:** Restrict external app launching from within Bitwarden to API 33+ using `GetLaunchIntentSenderForPackage` since there's no other way to do it without appearing nefarious (hence the removal of wildcard visibility)

## Screenshots

Screenshots

## Before you submit

### DIFF
--- a/src/Android/Autofill/AutofillExternalSelectionActivity.cs
+++ b/src/Android/Autofill/AutofillExternalSelectionActivity.cs
@@ -1,0 +1,36 @@
+﻿using Android.App;
+using Android.Content.PM;
+using Android.OS;
+using Bit.Core.Abstractions;
+using Bit.Core.Utilities;
+using Bit.Droid.Utilities;
+
+namespace Bit.Droid.Autofill
+{
+    [Activity(
+        NoHistory = true,
+        LaunchMode = LaunchMode.SingleTop)]
+    public class AutofillExternalSelectionActivity : Xamarin.Forms.Platform.Android.FormsAppCompatActivity
+    {
+        protected override void OnCreate(Bundle bundle)
+        {
+            Intent?.Validate();
+            base.OnCreate(bundle);
+
+            var cipherId = Intent?.GetStringExtra("autofillCipherId");
+            if (string.IsNullOrEmpty(cipherId))
+            {
+                SetResult(Result.Canceled);
+                Finish();
+                return;
+            }
+
+            var cipherService = ServiceContainer.Resolve<ICipherService>("cipherService");
+            var cipher = cipherService.GetAsync(cipherId).GetAwaiter().GetResult();
+            var decCipher = cipher?.DecryptAsync().GetAwaiter().GetResult();
+
+            var autofillHandler = ServiceContainer.Resolve<IAutofillHandler>();
+            autofillHandler.Autofill(decCipher, true);
+        }
+    }
+}
--- a/src/Android/Autofill/AutofillExternalSelectionActivity.cs
+++ b/src/Android/Autofill/AutofillExternalSelectionActivity.cs
@@ -1,4 +1,5 @@
-﻿using Android.App;
+﻿using System.Threading.Tasks;
+using Android.App;
 using Android.Content.PM;
 using Android.OS;
 using Bit.Core.Abstractions;
@@ -17,7 +18,7 @@ namespace Bit.Droid.Autofill
             Intent?.Validate();
             base.OnCreate(bundle);
 
-            var cipherId = Intent?.GetStringExtra("autofillCipherId");
+            var cipherId = Intent?.GetStringExtra(AutofillConstants.AutofillFrameworkCipherId);
             if (string.IsNullOrEmpty(cipherId))
             {
                 SetResult(Result.Canceled);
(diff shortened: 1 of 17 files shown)


src/Android/Autofill/AutofillExternalSelectionActivity.cs
```diff
@@ -0,0 +1,36 @@
...
+                Finish();
+                return;
+            }
+
+            var cipherService = ServiceContainer.Resolve<ICipherService>("cipherService");
+            var cipher = cipherService.GetAsync(cipherId).GetAwaiter().GetResult();
+            var decCipher = cipher?.DecryptAsync().GetAwaiter().GetResult();
+
+            var autofillHandler = ServiceContainer.Resolve<IAutofillHandler>();
+            autofillHandler.Autofill(decCipher, true);
```
fedemkr, line 33: To prevent blocking the main thread with `GetAwaiter().GetResult()` can't this be extracted to a `Task` that you `FireAndForget()`? Or is this activity going to be finished if you do that?

src/Android/Autofill/AutofillExternalSelectionActivity.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            var cipherId = Intent?.GetStringExtra("autofillCipherId");
+            if (string.IsNullOrEmpty(cipherId))
+            {
+                SetResult(Result.Canceled);
+                Finish();
+                return;
+            }
+
+            var cipherService = ServiceContainer.Resolve<ICipherService>("cipherService");
```
fedemkr, line 28: Could the string be extracted to a public const somewhere so it can be reused here and on `AutofillHelpers`?

src/Android/Services/AutofillHandler.cs
```diff
@@ -71,9 +71,17 @@ public bool SupportsAutofillService()
...
-            var activity = (MainActivity)CrossCurrentActivity.Current.Activity;
+            Xamarin.Forms.Platform.Android.FormsAppCompatActivity activity;
+            if (isExternalSelection)
+            {
+                activity = (AutofillExternalSelectionActivity)CrossCurrentActivity.Current.Activity;
+            }
+            else
+            {
+                activity = (MainActivity)CrossCurrentActivity.Current.Activity;
+            }
```
fedemkr, line 84: Just in case, if the current activity is not the one you're casting the method will throw an exception of invalid casting. Is that ok? if not you can cast using `as` that if it fails to cast, it returns `null`. However, we should log something if that happens to check why is not casting correctly and what is trying to cast.
Furthermore, if you want you can avoid the `bool isExternalSelection = false` given that the only casting you need is to `FormsAppCompatActivity`, you're not using anything from the explicit casting so the next thing should be enough:
```c#
var activity = CrossCurrentActivity.Current.Activity as Xamarin.Forms.Platform.Android.FormsAppCompatActivity;
```
Now, if you want to explicitly check whether the activity is one of those two, check it before, log it and return.

fedemkr (changes requested):

src/Android/Services/DeviceActionService.cs
```diff
@@ -69,14 +69,17 @@ public void Toast(string text, bool longDuration = false)
...
+                // https://developer.android.com/reference/android/content/pm/PackageManager#getLaunchIntentSenderForPackage(java.lang.String)
+                return false;
             }
-            return launchIntent != null;
+            var activity = CrossCurrentActivity.Current.Activity;
+            appName = appName.Replace("androidapp://", string.Empty);
+            var launchIntentSender = activity?.PackageManager?.GetLaunchIntentSenderForPackage(appName);
+            launchIntentSender?.SendIntent(activity, Result.Ok, null, null, null);
+            return launchIntentSender != null;
         }
```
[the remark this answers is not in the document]
fedemkr: Excellent then, I was just wondering if there was any flow that was interfering.

fedemkr (approved):

yegle:
Thank you for implementing this!

techbot2342:
Doesn't work for me. I still have to go to my vault either on pop up or keyboard bar?

techbot2342:
> > Doesn't work for me. I still have to go to my vault either on pop up or keyboard bar?
> 
> Are you using the correct build? For me it's working on 2 physical devices (Pixel & Samsung) as well as several emulators. Not sure beyond that until QA gets their hands on it.

Version: 2022.11.0 (5580) Android 13 S22 Ultra. I have all four auto fill services enabled and Copy TOTP enabled. Tried both apks from Github page. Tried with and without "Use incline autofill" enabled. Tried Gboard and Samsung keyboard to see if that was the issue. Don't know if I'm missing something

mp-bw:
@techbot2342 Build 5580 is our current production release which doesn't include this yet (this was just merged today so it won't be included until our next release).

netbymatt:
I received 2023.1.0 (5786) this morning and can confirm copy TOTP is now working with the inline autofill method. Thank you!

MrR341-GER:
This Problem still exists, or is occurring again
Android 11

TheGoodsStrau:
2023.10.0 (8257) on Pixel 7 , same issue. I have to go through the autofill (works fine) of the password,  got the bitwarden's pop up TOPT copied and can actually paste. Now, after log in, when asked for witch 2fa method I wanna use, I put auth app and I can no longer paste anything. It's like the clipboard is erased in-between these steps. Once the keyboards changes layout the code is gone. 
I tried to be as clear as possible, hope it get solved

PR #2246, opened 2022-12-15
[EC-844] Improve Apple Watch states

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

On Apple Watch there's an overlap between the "Need Log in" / "Need set up" states, so this PR goal is to fix them so that:

- Need Set up => first time of the app until log in and "Connect to Watch" is turned on or if the last account has turned off "Connect to Watch"
- Need Log in => if the last logged out account had "Connect to Watch" still enabled when logging out

## Code changes

* **BaseWatchDeviceService:** Improved state logic to include the last active account value of "Connect to Watch" to determine whether to be in "Need Log in" state
* **StateService:** Added setting for `LastUserShouldConnectToWatch` which stands for the value of `ShouldConnectToWatch` of the last active user. Also, updates it accordingly when the active user changes and when the "Connect to Watch" value changes.
* **Watch -> CipherListViewModel:** Removed logic for `NeedSetup` when no user is found on the watch, given that it's not needed anymore with the additional logic on the iOS app.

## Before you submit

### DIFF
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -97,5 +97,10 @@
         public static string PushLastRegistrationDateKey(string userId) => $"pushLastRegistrationDate_{userId}";
         public static string PushCurrentTokenKey(string userId) => $"pushCurrentToken_{userId}";
         public static string ShouldConnectToWatchKey(string userId) => $"shouldConnectToWatch_{userId}";
+        /// <summary>
+        /// This key is used to store the value of "ShouldConnectToWatch" of the last user that had logged in
+        /// which is used to handle Apple Watch state logic
+        /// </summary>
+        public const string LastUserShouldConnectToWatchKey = "lastUserShouldConnectToWatch";
     }
 }
--- a/src/Core/Services/StateService.cs
+++ b/src/Core/Services/StateService.cs
@@ -84,6 +84,8 @@ namespace Bit.Core.Services
             // Update pre-auth settings based on now-active user
             await SetRememberedOrgIdentifierAsync(await GetRememberedOrgIdentifierAsync());
             await SetPreAuthEnvironmentUrlsAsync(await GetEnvironmentUrlsAsync());
+
+            await UpdateLastUserShouldConnectToWatchAsync();
         }
 
         public async Task CheckExtensionActiveUserAndSwitchIfNeededAsync()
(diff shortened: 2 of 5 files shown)


src/Core/Constants.cs
```diff
@@ -97,5 +97,10 @@ public static class Constants
         public static string PushLastRegistrationDateKey(string userId) => $"pushLastRegistrationDate_{userId}";
         public static string PushCurrentTokenKey(string userId) => $"pushCurrentToken_{userId}";
         public static string ShouldConnectToWatchKey(string userId) => $"shouldConnectToWatch_{userId}";
+        /// <summary>
```
mp-bw, line 100: This should be moved to the not-userId-specific-prefs section above

src/Core/Services/StateService.cs
```diff
@@ -1708,6 +1710,21 @@ public async Task SetShouldConnectToWatchAsync(bool shouldConnect, string userId
                 ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync());
             var key = Constants.ShouldConnectToWatchKey(reconciledOptions.UserId);
             await SetValueAsync(key, shouldConnect, reconciledOptions);
+            await UpdateLastUserShouldConnectToWatchAsync(shouldConnect);
+        }
+
+        public async Task<bool> GetLastUserShouldConnectToWatchAsync()
+        {
+            var reconciledOptions = await GetDefaultStorageOptionsAsync();
```
mp-bw, line 1718: Since there's no userId passed in here, there's nothing to reconcile, so the name should just be `options` (see the `PushRegisteredTokenAsync` getter/setter for reference)

src/Core/Services/StateService.cs
```diff
@@ -1708,6 +1710,21 @@ public async Task SetShouldConnectToWatchAsync(bool shouldConnect, string userId
...
+            var key = Constants.LastUserShouldConnectToWatchKey;
+            return await GetValueAsync<bool?>(key, reconciledOptions) ?? false;
+        }
+
+        public async Task UpdateLastUserShouldConnectToWatchAsync(bool? shouldConnect = null)
```
mp-bw, line 1723: Should be `SetLastUserShouldConnectToWatchAsync` (instead of `Update...`)

src/Core/Services/StateService.cs
```diff
@@ -1708,6 +1710,21 @@ public async Task SetShouldConnectToWatchAsync(bool shouldConnect, string userId
...
+        {
+            var reconciledOptions = await GetDefaultStorageOptionsAsync();
```
mp-bw, line 1725: Same `options` here

src/Core/Services/StateService.cs
fedemkr: mmm I use `Update` because if the caller doesn't pass a value they are not setting anything but calling the method to update its internal value as it sees fit (in this case calling `GetLastUserShouldConnectToWatchAsync()`).
But if you still prefer `Set...` just so all the methods are the same is fine with me and I can change it.

fedemkr: Going for `Set...`

mp-bw (approved):

PR #2265, opened 2022-12-26
[EC-868] Clear console logging

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Clear console logging to be only shown in debug

## Code changes

* **...:** Changed all console logging to be only shown in debug

## Before you submit

### DIFF
--- a/src/watchOS/bitwarden/bitwarden WatchKit Extension/Services/CryptoService.swift
+++ b/src/watchOS/bitwarden/bitwarden WatchKit Extension/Services/CryptoService.swift
@@ -34,7 +34,7 @@ public class CryptoService{
             let item = try JSONDecoder().decode(type, from: decryptedData)
             return item
         } catch {
-            assertionFailure("Fail to decode item for keychain: \(error)")
+            Log.e("Fail to decode item for keychain: \(error)")
             return nil
         }
     }
--- a/src/Android/Services/LocalizeService.cs
+++ b/src/Android/Services/LocalizeService.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Globalization;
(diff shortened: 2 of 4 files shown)


src/watchOS/bitwarden/bitwarden WatchKit Extension/Services/CryptoService.swift
```diff
@@ -34,7 +34,7 @@ public class CryptoService{
             let item = try JSONDecoder().decode(type, from: decryptedData)
             return item
         } catch {
-            assertionFailure("Fail to decode item for keychain: \(error)")
+            Log.e("Fail to decode item for keychain: \(error)")
```
vvolkgang, line 37: Are we sure this won't log to the console?

vvolkgang (approved):
LGTM!

src/watchOS/bitwarden/bitwarden WatchKit Extension/Services/CryptoService.swift
fedemkr: Yes, it's an internal helper [LoggerHelper](https://github.com/bitwarden/mobile/blob/master/src/watchOS/bitwarden/bitwarden%20WatchKit%20Extension/Helpers/LoggerHelper.swift)
it has a bool indicating that it only logs on DEBUG, and also the swift print func is [wrapped](https://github.com/bitwarden/mobile/blob/master/src/watchOS/bitwarden/bitwarden%20WatchKit%20Extension/Helpers/LoggerHelper.swift#L9) so that it only works on DEBUG to avoid forgetting some print somewhere, so if it happens it will do nothing 😄 😄   (loved that trick)

PR #2277, opened 2023-01-05
[SG-912] Modify the mobile app to retrieve the user's avatar color

BrandonM-Bitwarden:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

With the introduction of custom Avatar Colors, the correct custom color must be shown.

## Code changes

## Before you submit


src/Core/Services/StateService.cs
```diff
@@ -1720,6 +1720,22 @@ public async Task<bool> GetLastUserShouldConnectToWatchAsync()
             return await GetValueAsync<bool?>(key, options) ?? false;
         }
 
+        public async Task SetAvatarColorAsync(string value, string userId = null)
+        {
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var account = await GetAccountAsync(reconciledOptions);
```
mp-bw, line 1727: Establish a key and use `SetValueAsync` here instead of working directly on the `account` object (check the other setters for reference - The account object is automatically handled by setting the value).

src/Core/Services/StateService.cs
```diff
@@ -1720,6 +1720,22 @@ public async Task<bool> GetLastUserShouldConnectToWatchAsync()
...
+            account.Profile.AvatarColor = value;
+            await SaveAccountAsync(account, reconciledOptions);
+        }
+
+        public async Task<string> GetAvatarColorAsync(string userId = null)
+        {
+            return (await GetAccountAsync(
```
mp-bw, line 1734: Same with the getter - use others for reference

mp-bw (commented):
Looks good, just need to follow the established getter/setter pattern in `StateService`

src/Core/Constants.cs
```diff
@@ -41,6 +41,7 @@ public static class Constants
         /// which is used to handle Apple Watch state logic
         /// </summary>
         public const string LastUserShouldConnectToWatchKey = "lastUserShouldConnectToWatch";
+        public const string AvatarColorKey = "avatarColor";
```
mp-bw, line 44: Since this is a user-specific pref, it should be added below (following the `_{userId}` pattern)

src/Core/Services/StateService.cs
```diff
@@ -1724,16 +1724,16 @@ public async Task SetAvatarColorAsync(string value, string userId = null)
         {
             var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
                 await GetDefaultStorageOptionsAsync());
-            var account = await GetAccountAsync(reconciledOptions);
-            account.Profile.AvatarColor = value;
-            await SaveAccountAsync(account, reconciledOptions);
+            var key = Constants.AvatarColorKey;
```
mp-bw, line 1727: Should be `Constants.AvatarColorKey(reconciledOptions.UserId);`

src/Core/Services/StateService.cs
```diff
@@ -1724,16 +1724,16 @@ public async Task SetAvatarColorAsync(string value, string userId = null)
...
         }
 
         public async Task<string> GetAvatarColorAsync(string userId = null)
         {
-            return (await GetAccountAsync(
-                ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync())
-            ))?.Profile?.AvatarColor;
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var key = Constants.AvatarColorKey;
```
mp-bw, line 1735: Same here: `Constants.AvatarColorKey(reconciledOptions.UserId);`

src/Core/Services/StateService.cs
```diff
@@ -1724,15 +1724,15 @@ public async Task SetAvatarColorAsync(string value, string userId = null)
         {
             var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
                 await GetDefaultStorageOptionsAsync());
-            var key = Constants.AvatarColorKey;
+            var key = Constants.AvatarColorKey(userId);
```
mp-bw, line 1727: `userId` should be `reconciledOptions.UserId` in order to support null fallback to the current active user

src/Core/Services/StateService.cs
```diff
@@ -1724,15 +1724,15 @@ public async Task SetAvatarColorAsync(string value, string userId = null)
...
             await SetValueAsync(key, value, reconciledOptions);
         }
 
         public async Task<string> GetAvatarColorAsync(string userId = null)
         {
             var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
                 await GetDefaultStorageOptionsAsync());
-            var key = Constants.AvatarColorKey;
+            var key = Constants.AvatarColorKey(userId);
```
mp-bw, line 1735: same here

mp-bw:
Okay time for me to eat crow - you had the getter/setter right the first time since this value is stored in the `Account` profile obj and not a separate pref.  So I apologize for the wasted time there.  Everything else is great, I just made two small changes:

1. Moved the getter/setter above the `// Helpers` block in `StateService` so it lives with the others, and
2. Moved the avatar update logic to the sync completion callback (instead of the sync click) so the avatar updates via push when the color is changed on the server (this would work for name changes too but the server doesn't appear to push those, just color changes)

mp-bw (approved):

src/App/Pages/BaseContentPage.cs
```diff
@@ -130,7 +130,7 @@ protected async Task<AvatarImageSource> GetAvatarImageSourceAsync(bool useCurren
             if (useCurrentActiveAccount)
             {
                 return new AvatarImageSource(await _stateService.GetActiveUserIdAsync(),
-                    await _stateService.GetNameAsync(), await _stateService.GetEmailAsync());
+                    await _stateService.GetNameAsync(), await _stateService.GetEmailAsync(), await _stateService.GetAvatarColorAsync());
```
fedemkr, line 133: Even though this is just fine, a little while ago I've added `GetActiveUserCustomDataAsync<T>(Func<Account, T> dataMapper)` which can be used when we need multiple data from the account/profile in one call.
So instead of doing three await calls to the state service we can just do one (and also spare the added method `GetAvatarColorAsync()`):
```c#
var t = await GetActiveUserCustomDataAsync<Tuple<string, string, string>>(a => a?.Profile != null ? new Tuple<string, string, string>(a.Profile.Name, a.Profile.Email, a.Profile.AvatarColor) : null);
```
Can't remember well if we can use concise tuples given the framework we are targeting, but if we can we can write it as:
```c#
(string name, string email, string avatarColor) = await GetActiveUserCustomDataAsync(a => a?.Profile != null ? (a.Profile.Name, a.Profile.Email, a.Profile.AvatarColor) : null);
```
Also it could be used with a custom type instead of a `Tuple`

what do you think @mp-bw ?

PR #2278, opened 2023-01-05
[EC-763] Cake build watchOS

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Implemented watchOS parts on cake build and iOS / watchOS icons update

## Code changes

* **build.cake:** Added replacement of bundle id and version on watchOS related files. I couldn't find a Cake extension to handle pbxproj files so I just replace by using a regex. Also added icons update for iOS and watchOS

## Before you submit


build.cake
```diff
@@ -240,6 +243,43 @@ private void UpdateiOSEntitlementsPlist(string entitlementsPath, VariantConfig b
...
+    var fileText = FileReadText(pbxprojPath);
+    if (string.IsNullOrEmpty(fileText))
+    {
+        throw new Exception($"Couldn't find {pbxprojPath}");
+    }
+
+    const string pattern = @"MARKETING_VERSION = [^;]*;";
+
+    var plistFile = File(Path.Combine(_slnPath, "src", "iOS", "Info.plist"));
+    dynamic plist = DeserializePlist(plistFile);
```
vvolkgang, line 273: From what I'm understanding we're only deserializing the iOS plist here to fetch the previously set version string. If that's the case, we can instead:

1. In the `UpdateiOSPlist` task, save that string in a top level var. You'll find them at the top of the script, their names start with `_` and have a comment explaining where they're set
2. Use a [Criteria](https://cakebuild.net/docs/writing-builds/tasks/criteria) in the `UpdateWatchProject` to check that var is filled. This will mostly serve as a way to raise visibility of the dependency from another task. E.g. `.WithCriteria(!string.IsNullOrEmpty(_iOSVersionName))`

vvolkgang (changes requested):
One small improvement but everything else looks great!

vvolkgang (approved):

PR #2285, opened 2023-01-10
[PS-2215] Implement scrypt key derivation function

quexten:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

This pull request implements the scrypt kdf, as the second kdf next to pbkdf2.
The corresponding pull request on the clients repository is here: 

**Edit @djsmith85**: Community forum thread: https://community.bitwarden.com/t/scrypt-kdf-support/48148/1
## Code changes

There is also a pull request for the clients implementation here: bitwarden/clients/pull/4428

- CryptoPrimitiveService.cs (iOS, Android), ICryptoPrimitiveService.cs: Implement the basic scrypt call.
- KdfType.cs: Add the second KdfType
- PclCryptoFunctionService.cs, ICryptoFunctionService.cs: Add the Scrypt functions
- CryptoService: Add the scrypt case case for the kdf type. The default iterations are set to 2^16, and the minimum iterations are set to 2^15. (https://words.filippo.io/the-scrypt-parameters/ recommends 2^15 for interactive logins, but the article is from 2017 and it is arguable that bitwarden is closer to file encryption than interactive logins, so I'm open to debate here). The rest of the parameters are the same as recommended in the linked post.
- iOS.Core.csproj: Add the BouncyCastle library (which is already used on Android) to implement Scrypt

I will note that I only tested this on Android, as I do not have a working iOS testing setup. It is the same implementation, but if it still does not work, let me know and I'm happy to fix it :)

### DIFF
--- a/src/Android/Services/CryptoPrimitiveService.cs
+++ b/src/Android/Services/CryptoPrimitiveService.cs
@@ -33,5 +33,10 @@ namespace Bit.Droid.Services
             generator.Init(password, salt, iterations);
             return ((KeyParameter)generator.GenerateDerivedMacParameters(keySize)).GetKey();
         }
+
+        public byte[] Scrypt(byte[] password, byte[] salt, int N, int r, int p, int dkLen)
+        {
+            return SCrypt.Generate(password, salt, N, r, p, dkLen);
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


quexten:
Closing, as we are focusing on argon2 instead.

PR #2286, opened 2023-01-10
Upload and process test results as an artifact and report

withinfocus:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [X] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Utilizes a popular Action to process .NET test results and upload them as a report artifact. Much easier to parse 
results than working through the commandline output.

## Code changes

* **.github/workflows/build.yml:** Additional step for report processing

## Before you submit

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,10 +4,10 @@ name: Build
 on:
   push:
     branches-ignore:
-      - 'l10n_master'
-      - 'gh-pages'
+      - "l10n_master"
+      - "gh-pages"
     paths-ignore:
-      - '.github/workflows/**'
+      - ".github/workflows/**"
   workflow_dispatch:
     inputs: {}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -4,10 +4,10 @@ name: Build
 on:
   push:
     branches-ignore:
-      - 'l10n_master'
-      - 'gh-pages'
+      - "l10n_master"
```
withinfocus, line 7: ℹ️ Prettier caught a few things.

.github/workflows/build.yml
```diff
@@ -148,7 +146,17 @@ jobs:
         shell: pwsh
 
       - name: Run Core tests
```
withinfocus, line 148: ℹ️ I don't feel that this should be within the Android build vs. something fundamental across all builds earlier but given the significant setup per platform I am gonna leave it be for now since the results will still appear regardless.

.github/workflows/build.yml
```diff
@@ -27,7 +27,6 @@ jobs:
       - name: Print lines of code
         run: cloc --vcs git --exclude-dir Resources,store,test,Properties --include-lang C#,XAML
 
-
```
joseph-flinn, line 30: These double newlines are intentional to ease maintenance burdens on large workflow files. We have a [custom linter](https://github.com/bitwarden/gh-actions/tree/master/lint-workflow) that catches most of our standards and the rest are documented in Confluence (which should probably be moved to the new public contributing docs)

joseph-flinn (changes requested):

.github/workflows/build.yml
joseph-flinn: DevOps hasn't done a whole lot with the mobile build pipeline over the past two years. Mobile has been a large growth area for DevOps that we have yet to have the capacity to grow in and improve the processes around.

I'd love to hear some thoughts on what we can improve and how to improve.

PR #2288, opened 2023-01-11
[EC-469] Improve ApiException message

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Improve `ApiException` so that we can log an enhanced message with the errors provided by the Server instead of just "An API error has occurred"

### Format that will be displayed

**With `ValidationErrors`:**

{Http status code number} {Http status code string}. [{key1}]:{validation11}; {validation12}; {validation13} ...
[{key2}]: {validation21}; {validation22}; {validation23} ...
[{key3}]:{validation31}; {validation32}; {validation33} ...
...

**Without `ValidationErrors`:**

{Http status code number} {Http status code string}. {Message}

**Examples:**

404 Bad Request. []: Premium status is required

404 Bad Request. [Token]: Invalid token

404 Bad Request. []: User verification failed
["MasterPassword"]: Invalid password

and if no `ValidationErrors` are there but there is a `Message` we should get something like:

403 Forbidden. You have no access to this resource

## Code changes

* **ApiException:** Override `Message` to use `GetFullMessage()` of the `Error` if possible or default to its base message
* **ErrorResponse:** Added `GetFullMessage()` method that produces a string message based on its inner `Message` and `ValidationErrors`

## Before you submit

### DIFF
--- a/src/Core/Exceptions/ApiException.cs
+++ b/src/Core/Exceptions/ApiException.cs
@@ -16,5 +16,7 @@ namespace Bit.Core.Exceptions
         }
 
         public ErrorResponse Error { get; set; }
+
+        public override string Message => Error?.GetFullMessage() ?? base.Message;
     }
 }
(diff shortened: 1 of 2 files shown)


andrebispo5 (approved):
Looks good.

PR #2289, opened 2023-01-12
Passwordless login 2FA not working

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When trying to login via a passwordless request to another device, if the account has 2FA enable it will be stuck in the next step and cannot successfully complete the login process.

## Code changes

Added missing parameter needed to complete 2FA login when in a login with device request.

## Before you submit


src/App/Pages/Accounts/LoginPageViewModel.cs
```diff
@@ -149,8 +149,7 @@ public async Task InitAsync()
                     Email = await _stateService.GetRememberedEmailAsync();
                 }
                 var deviceIdentifier = await _appIdService.GetAppIdAsync();
-                // TODO uncomment to enable login with device
-                //IsKnownDevice = await _apiService.GetKnownDeviceAsync(Email, deviceIdentifier);
+                IsKnownDevice = await _apiService.GetKnownDeviceAsync(Email, deviceIdentifier);
```
fedemkr, line 152: just in case, is this ok to be enabled?

fedemkr (commented):
LGTM, but could you corroborate that enabling login with device is the desired behavior? Cause I don't see it mentioned on the Objective nor the Code changes of the PR.

src/App/Pages/Accounts/LoginPageViewModel.cs
andrebispo5: It needs to be enabled in order for QA to test it here in the branch. But when the time comes to be merged, i should be good to be enabled.

fedemkr (approved):

PR #2293, opened 2023-01-13
[PS-2249] Implement Argon2

quexten:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
This PR implements basic argon2id support. For now, only memory is configurable (and stored in the user's kdfIterations). There is also a[ Community discussion thread](https://community.bitwarden.com/t/argon2-kdf-support/48825). I tested it only on Android, as I do not have a working iOS setup, but as it is the same iteration, it *should* work.

## Code changes
- Android.csproj, iOS.csproj: Add the Argon2 library. For the library I went by the one with the most stars on GitHub for C#. 
- KdfType.cs: We add argon2id as the KdfType with id 2 (since I already used id 1 in the scrypt pull request).
- CryptoPrimitiveService.cs (iOS, Android), ICryptoPrimitiveService.cs: Implement argon2 calls
- ICryptoFunctionService.cs, PclCryptoFunctionService.cs: Implement argon2 calls
- CryptoService.cs: Add the case for argon2id. For now, memory is set to the user's kdfIterations. Parallelism and iterations are constants, though this is also the place to load them from a future "kdfOptions" object (possibly added in a later pull request?).


quexten:
As discussed in the scrypt pull request, we might want to switch to argon2 bindings instead of using the pure C# implementation.

Currently, the pull request uses https://github.com/mheyman/Isopoh.Cryptography.Argon2
It is based on the original phc C implementation, but rewritten in C#/.NET Core. The advantage is that "just works" on Android and iOS without doing anything more. Wherever C# runs (which is required for the app anyways), the implementation runs.

I had a brief look at the available libraries for bindings here are my findings so far:
Bouncycastle on Java has support for argon2, but the C# bouncycastle implementation does not. There is an open issue tracking this, but no progress on this whatsoever: https://github.com/bcgit/bc-csharp/issues/202

https://github.com/CalebABG/Argon2Bindings just wraps the original C library, but it is marked as experimental and does not even have a NuGet package.

https://github.com/modernuo/Argon2.Bindings does have a NuGet package, but only supports the x64 architecture which means it is unsuitable for Android/iOS.

https://github.com/alipha/csharp-argon2 These bindings have not been updated for 6 years and also only support x86/x86_64 so far.

A few more C# based implementations (not bindings):
https://github.com/kmaragon/Konscious.Security.Cryptography
https://github.com/ron4fun/SharpHash
https://github.com/TimothyMeadows/Argon2.NetCore

To summarize, if we want to use the original C library, we would have to modify the existing C# bindings packages to compile for iOS and Android (and possibly multiple platforms on Android as there are different ARM architectures, x86_64 and possibly RISC-V in the future).
Basically, I think this is more trouble than it is worth compared to just using a C# implementation. Any thoughts @kspearrin ?

kspearrin:
@quexten what about https://github.com/ektrah/nsec ? This wraps libsodium, which I believe wraps the native argon2  library?

quexten:
> @quexten what about https://github.com/ektrah/nsec ? This wraps libsodium, which I believe wraps the native argon2 library?

Don't think this supports Xamarin builds.
https://github.com/ektrah/nsec/issues/16
Supported platforms only list Windows, Mac, Linux:
https://nsec.rocks/docs/install

kspearrin:
Thanks for looking into it. I’ll spend some time seeing if I can add some iOS and android architectures to one of these libs, or make something of our own. I really don’t like the idea of having someone reimplement the algorithm in C# like the current library is doing.

quexten:
Yeah, understandable. I can't help with iOS, as I don't have an iOS build setup, but I'm happy to help on Android.

By the way, what do you think about wrapping the java (android) wrappers from https://github.com/signalapp/Argon2 in C#? They already compile the original C implementation and wrap it, and it might be simpler to create a C# wrapper for these android/ios libraries, compared to building the C library for all platforms.

kspearrin:
Not a bad idea to explore if you like. Might make it easier, but not sure if we have any extra overhead. Looks like the same library has the bindings for iOS too.

quexten:
Ok, I have updated the Android implementation to use the native bindings instead. The shared libs compiled from the original C implementation, I have copied from https://github.com/signalapp/Argon2 for now, but we can compile these ourselves of course if necessary. The rest is just a basic binding for the argon2id_hash_raw function: https://github.com/bitwarden/mobile/blob/ed4312c46b67aea5744438d8267929c00c780fae/src/Android/Services/CryptoPrimitiveService.cs#L39-L49

I expect on iOS this should not be much more difficult.

kspearrin:
@quexten This looks great. Let me work on some stuff for iOS. We might just want to move this all to a published nuget library that we can share following something like this https://learn.microsoft.com/en-us/xamarin/cross-platform/cpp/

quexten:
> @quexten I am still looking into this but am having a difficult time trying to build the `libArgon2.a` output from source for iOS arm64 and x86_64 (iOS simulator) architectures.

@kspearrin  I set up an iOS VM. The following commands seems to work (though I haven't tested the outputted executable yet):
`
# iphone simulator
export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk
cc -std=c89 -O3 -Wall -g -Iinclude -Isrc  -pthread -march=native -target x86_64-apple-ios-simulator  -c -o src/argon2.o src/argon2.c

# iphone
export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk
cc -std=c89 -O3 -Wall -g -Iinclude -Isrc  -pthread -target armv7-apple-ios  -c -o src/argon2.o src/argon2.c`

quexten:
After compiling for iphone arm7, arm7s, arm7 and iphone simulator, creating archives from the objects, and merging them with lipo I got a .a binary. Though I'll play around a bit with [CPPSharp](https://github.com/mono/CppSharp/blob/main/docs/UsersManual.md) first as it might make this whole process simpler.

kspearrin:
I have been able to create a `.a` binary using `xcodebuild` following Xamarin tutorials such as this one: 
https://learn.microsoft.com/en-us/xamarin/ios/platform/native-interop
and
https://learn.microsoft.com/en-us/xamarin/ios/platform/binding-objective-c/walkthrough?tabs=macos#creating-a-fat-binary

But most of these docs seem outdated and I have to make tweaks. For example, the simulator is `x86_64` now instead of `i386` architecture. We only need `arm64` architecture for the iPhone. `armv7` is no longer in use with the devices we support.

Adding the `libArgon2.a` binary that I get from `lipo` to the iOS project and trying to build with the `-gcc_flags` extra build arguments is giving me errors telling me that the native `.a` binary is for macOS and not `iPhone Simulator` (configuration I am using) with arch x86_64.

I have not been able to get past this yet. It does build for `iPhone` configuration, but I do not have a device to test that with.

kspearrin:
@quexten How did you include the `libargon.a` in the Xamarin project?

quexten:
As one of the guides you previously posted stated, I just added the libargon2.a to the project and then added the build flag: -gcc_flags "-L${ProjectDir} -largon2 -force_load ${ProjectDir}/libargon2.a" in the build settings in visual studio.

quexten:
Sample Xamarin Poject:
https://github.com/quexten/argon2-test

kspearrin:
@quexten Check out https://github.com/bitwarden/mobile/pull/2309/files . If this looks good to you, I can open the PR for review and get it merged into master and you can then merge that down into your PR here.

kspearrin:
Another option is to change your PR here to target my `argon2ios` branch instead of `master`. Then we can submit my PR for review against `master`

quexten:
> @quexten Check out https://github.com/bitwarden/mobile/pull/2309/files . If this looks good to you, I can open the PR for review and get it merged into master and you can then merge that down into your PR here.

Pull request looks good to me!
Just to make sure: Does the library also need to be compiled for WatchOS, and if so is that included in the archive on the PR?
From a blog post I got the following targets:
```
    -target x86_64-apple-watchos7.0-simulator
    -target arm64-apple-watchos7.0-simulator
    -target armv7k-apple-watchos7.0
    -target arm64_32-apple-watchos7.0
```

> Another option is to change your PR here to target my `argon2ios` branch instead of `master`. Then we can submit my PR for review against `master`

Yep. Sounds good!

kspearrin:
I confirmed that our Apple Watch implementation will not need a KDF. Can you re-target this PR to my `argon2ios` branch and I can approve it?

quexten:
I have re-targeted it, but it's still missing the changes to support changing KdfIterations, KdfMemory, and KdfParallelism. Iterations and parallelism are currently fixed on this branch. I'll implement that part, and then resolve the conflicts, so you can merge.

kspearrin:
We can do a follow up PR to support KDF memory and parallelism. That requires the changes to merge into server anyways.

quexten:
@kspearrin Okay, I at least had to change it such that instead of mis-using KdfIterations for loading memory, using Iterations instead to keep it compatible. Memory is set statically to 16MiB (slightly over OWASP minimum recommendation) and parallelism is set to 1. It is now re-based on your iOS branch.

quexten:
For the kdf extra parameters I have a basic (not yet well tested) version now. I'll submit it as a separate PR once I tested it a bit more, but in the meantime it is [quexten/mobile/argon2-kdf-options](https://github.com/quexten/mobile/tree/argon2-kdf-options).

kspearrin (approved):

PR #2299, opened 2023-01-16
[EC-1002] [BEEEP] Add ability to change language in app

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add ability to change the language from the app. For now it requires app restart to apply the change.

## Code changes

* **OptionsPage.xaml:** Added language picker
* **OptionsPageViewModel:** Added possibility to select a language/locale, also added system option (with null key value). This is saved into preferences.
* **MainApplication, iOSCoreHelpers:** Load the app locale if any selected and set the app culture.
* **MobileI18NService:** Added `SetCulture(...)` so that the culture can be from away the class and fixed some caps in the languages (IDK if the others should be in caps or not)
* **IPreferencesStorageService, PreferencesStorageService:** Added sync methods for Get, Save, Remove and the interface so the caller can use the synchronous way instead of the async one of the `IStorageService`. This was done to avoid needing async code for setting the culture when the app is loading which caused some race conditions and therefore some strings weren't translated correctly because the strings were being translated before the Culture was actually updated.
* **StorageMediatorService:** Added this mediator in order to interact with the storages. With this we can have sync and async methods of interaction where the logic to choose one or the other is transparent to the caller and there are also `StorageMediatorOptions` to pass.
* **StateService:** Added get/set `Locale` to here to be alike to other global settings like theming. Additionally, now it uses the aforementioned mediator to interact with the storages. For now, it's only applied to `Locale` but afterwards it'll be refactored so that the storage services are removed from here and the mediator is used for everything. Also, marked some methods as `Obsolete` so the mediator is used in new additional features.
* **Others:** Use the `StateService` to get/set the locale

## Screenshots

Language option

Language option selector

Language selected message

Language option dark theme

## Before you submit


src/Core/Abstractions/IPreferencesStorageService.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Bit.Core.Abstractions
+{
+    public interface IPreferencesStorageService
```
mp-bw, line 3: Can we give this interface a different name? It took me a moment to wrap my head around the need for it and I think it's because the inclusion of `Preferences` led me down a weird path. Perhaps something like `SynchronousStorageService`?

fedemkr: What I like about the `Preferences` prefix is that the caller knows exactly where the data is stored, whereas with `Synchronous` it could be anywhere.
However, I see the point about causing confusion on when that should be used and your suggestion makes sense.
Either if we keep `Preferences` or we change it to `Synchronous` (or something else), I'd like to add a `summary` to the interface so that it explains it further. What do you think?
In addition, I'd like to refactor at some time the registration and interface of `IStorageService` so we don't have to keep relying on the registration key to determine which implementation is using, so we'd have additional specific interfaces for each implementation and the `IStorageService` as the base one.

mp-bw: > What I like about the Preferences prefix is that the caller knows exactly where the data is stored, whereas with Synchronous it could be anywhere.

That's exactly what I'm thinking though - we already have a preference-specific storage service that implements `IStorageService` and you're adding synchronous functionality to it via another interface that could potentially be applied to the other non-preference storage services as well if the need arises.  In addition the authority that answers the question "where is this stored" is already implemented as `_preferenceStorageKeys` in `MobileStorageService`.  Does this make sense, or is my interpretation incorrect?

src/App/Pages/Settings/OptionsPageViewModel.cs
```diff
@@ -184,19 +209,30 @@ public bool ShowAndroidAutofillSettings
...
             AutoDarkThemeSelectedIndex = AutoDarkThemeOptions.FindIndex(k => k.Key == autoDarkTheme);
+
             var defaultUriMatch = await _stateService.GetDefaultUriMatchAsync();
             UriMatchSelectedIndex = defaultUriMatch == null ? 0 :
                 UriMatchOptions.FindIndex(k => (int?)k.Key == defaultUriMatch);
+
             var clearClipboard = await _stateService.GetClearClipboardAsync();
             ClearClipboardSelectedIndex = ClearClipboardOptions.FindIndex(k => k.Key == clearClipboard);
+
+            var appLocale = _synchronousStorageService.Get<string>(Core.Constants.AppLocaleKey);
```
mp-bw, line 233: Using the storage service directly from the app without going through `_stateService` goes against the pattern established with multi-account support (even for globals).  We could add a `Synchronous` bool to `StorageOptions` and create a `GetDefaultSynchronousStorageOptionsAsync` method to `StateService` to be used from new AppLocale getter/setter methods, lining up nicely with the other app settings.  (I'm assuming this value is global since it sounds like Locale must be set on app initialization, so check out `GetThemeAsync` & `SetThemeAsync` in `StateService` for a good global reference)

fedemkr: I'll make the change, although I think that at some point we should refactor it to have the globals in some other place and leave the `StateService` for account related stuff only. I believe the `StateService` is getting gigantic and having a lot of responsabilities.

mp-bw (approved):
Nice!

PR #2304, opened 2023-01-18
[SG-516] Additional forwarded email providers for username generator - mobile

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add option to select Fastmail and Duck Duck Go as options in the forwarded email alias providers.
Default to none-selected if this is the first use of the screen. Otherwise default to the last selected.

## Code changes

[src/App/Pages/Generator/GeneratorPage.xaml](https://github.com/bitwarden/mobile/pull/2304/files#diff-6646362e986f583f0da02c8c01fad244607c0d4dd38e211dd22f7d7ac64664dd): 
New UI views to hold api keys for new services added. (This can be refactored in the future to avoid having multiple views)
[src/App/Pages/Generator/GeneratorPageViewModel.cs](https://github.com/bitwarden/mobile/pull/2304/files#diff-06d6f35a0e203bdae8fe03dade2ef549a73850f02dde233ac921a0fd7214dbc5):
Add new entries for Fastmail and Duck Duck Go to `ForwardedEmailServiceTypeOptions`.
Added properties to bind to the UI for Api Keys and Visibility.
[src/Core/Enums/ForwardedEmailServiceType.cs](https://github.com/bitwarden/mobile/pull/2304/files#diff-5c4fbe584a9869f63822190a1d2af12cad7f0f18caadb45a6da61cc65841b380):
Reorder the enum with added values. Added none to be defaulted the first the user sees the screen, after that it will always have a value saved.
[src/Core/Services/ApiService.cs](https://github.com/bitwarden/mobile/pull/2304/files#diff-5b9bd6e62fd23e1f8dbc2fd0b883357945eae927d73e88753478e3f7c2de2638): 
`HandleFastMailResponse` logic was copied from the other clients.
`CreateFastmailRequest` gets the necessary `accountId` value through a get call, after that builds a JSON object that will be used to perform the actual call to get a generated email.

## Screenshots

First time on screen, default service to empty:

https://user-images.githubusercontent.com/4648522/213505609-ced1e223-9cb0-4a2b-b42d-7756f6d71f3b.mov

Subsequent screen landings, generate new emails from Fastmail and Duck Duck Go:

https://user-images.githubusercontent.com/4648522/213505852-1f347a6d-578c-4a61-9278-541369a07fe3.mov

## Before you submit

### DIFF
--- a/src/App/Pages/Generator/GeneratorPageViewModel.cs
+++ b/src/App/Pages/Generator/GeneratorPageViewModel.cs
@@ -81,10 +81,10 @@ namespace Bit.App.Pages
 
             ForwardedEmailServiceTypeOptions = new List<ForwardedEmailServiceType> {
                 ForwardedEmailServiceType.AnonAddy,
-                ForwardedEmailServiceType.FirefoxRelay,
-                ForwardedEmailServiceType.SimpleLogin,
+                ForwardedEmailServiceType.DuckDuckGo,
                 ForwardedEmailServiceType.Fastmail,
-                ForwardedEmailServiceType.DuckDuckGo
+                ForwardedEmailServiceType.FirefoxRelay,
+                ForwardedEmailServiceType.SimpleLogin
             };
 
             UsernameEmailTypeOptions = new List<UsernameEmailType>
--- a/src/App/Pages/Generator/GeneratorPageViewModel.cs
+++ b/src/App/Pages/Generator/GeneratorPageViewModel.cs
@@ -840,6 +840,8 @@ namespace Bit.App.Pages
             TriggerPropertyChanged(nameof(FirefoxRelayApiAccessToken));
             TriggerPropertyChanged(nameof(AnonAddyDomainName));
             TriggerPropertyChanged(nameof(AnonAddyApiAccessToken));
+            TriggerPropertyChanged(nameof(DuckDuckGoApiKey));
+            TriggerPropertyChanged(nameof(FastmailApiKey));
             TriggerPropertyChanged(nameof(CatchAllEmailDomain));
             TriggerPropertyChanged(nameof(ForwardedEmailServiceSelected));
             TriggerPropertyChanged(nameof(UsernameTypeSelected));
--- a/src/App/Pages/Generator/GeneratorPageViewModel.cs
+++ b/src/App/Pages/Generator/GeneratorPageViewModel.cs
@@ -52,6 +52,7 @@ namespace Bit.App.Pages
         private bool _showFirefoxRelayApiAccessToken;
         private bool _showAnonAddyApiAccessToken;
         private bool _showSimpleLoginApiKey;
+        private bool _showDuckDuckGoApiKey;
         private bool _editMode;
 
         public GeneratorPageViewModel()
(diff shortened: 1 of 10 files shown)


src/App/Pages/Generator/GeneratorPageViewModel.cs
```diff
@@ -542,6 +546,64 @@ public bool ShowSimpleLoginApiKey
...
+            }
+            set => SetProperty(ref _showFastmailApiKey, value,
+                additionalPropertyNames: new string[]
+                {
+                    nameof(ShowFastmailHiddenValueIcon)
+                });
+        }
+
+        public string ShowFastmailHiddenValueIcon => _showFastmailApiKey ? BitwardenIcons.EyeSlash : BitwardenIcons.Eye;
+
```
fedemkr, line 606: If possible, refactor this alongside with the page xaml so we have common properties used that change their values depending on the forwarded email alias service selected.

src/Core/Enums/ForwardedEmailServiceType.cs
```diff
@@ -4,11 +4,16 @@ namespace Bit.Core.Enums
...
+        DuckDuckGo = 2,
+        [LocalizableEnum("Fastmail")]
+        Fastmail = 3,
         [LocalizableEnum("FirefoxRelay")]
-        FirefoxRelay = 1,
+        FirefoxRelay = 4,
         [LocalizableEnum("SimpleLogin")]
-        SimpleLogin = 2,
+        SimpleLogin = 5
     }
```
fedemkr, line 18: I wouldn't change the values of the types. This configuration gets saved into the device so when an already configured `UsernameGenerationOptions` gets loaded it will have a different type set up.
E.g., let's say the user configured `FirefoxRelay` so an option with type `1` will be saved. But after an app update with this code, when the options are loaded `1` will be `AnonAddy` so the configuration will be wrong.

You either change the values so that they're backwards compatible or add a migration of data.

src/Core/Services/ApiService.cs
```diff
@@ -790,12 +798,99 @@ public async Task<string> GetUsernameFromAsync(ForwardedEmailServiceType service
...
+                                        }
+                                    }
+                                },
+                                "0"
+                            }
+                        })
+                };
+                return requestJObj.ToString();
+            }
+        }
```
fedemkr, line 892: The `ApiService` is gigantic and doing a lot of stuff so please move this to a new service specific for Fastmail: `FastmailService` and extract the repeated hardcoded values into constants, like `methodResponses`.
Also, would it be a good path when `accountId` is `null`? I mean, is it worth it to send the request or should we just abort/throw an exception?
Furthermore, it'd be good to have a specific exception for the fastmail instead of just `throw new Exception(...)` so if it get's logged we can identify it easily by its type, e.g. `FastmailException`

fedemkr (changes requested):
For now, these would be my review. 
I think several parts of the username generation (not only of this PR) should be refactored in the future in order to use more OOP concepts to reduce the cyclomatic complexity and be easier to maintain, like having polymorphism applied to the options to have a base options class and then specific ones depending on the needs of each service which then we can directly use to create the needed `UsernameGeneratorConfig`.
Aside from that, have you tried generating username/password from the Autofill extension? There could be memory problems because of the added things.

Also, does the user receive any feedback on slow connection? I didn't see any loading wheel or something like that.

andrebispo5:
It was decided that the refactoring work will be done in tech debt item SG-1004. 
For PR it was fixed glyph converter and the ForwardedEmailServiceType default value / order.

src/Core/Models/Domain/UsernameGenerationOptions.cs
```diff
@@ -4,7 +4,10 @@ namespace Bit.Core.Models.Domain
 {
     public class UsernameGenerationOptions
     {
-        public UsernameGenerationOptions() { }
+        public UsernameGenerationOptions()
+        {
+            ServiceType = ForwardedEmailServiceType.None;
+        }
```
fedemkr, line 10: There's no need to have the constructor, if you want you can just inline it

```c#
public ForwardedEmailServiceType ServiceType { get; set; } = ForwardedEmailServiceType.None;
```

fedemkr (approved):
LGTM, if you want change the ctor, otherwise it's not a big deal.

PR #2328, opened 2023-01-25
[PS-2358] Add kdf configuration options

quexten:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

This PR adds configuration option support for the argon2 kdf.

## Code changes
For all files: Basically anywhere where kdfIterations / kdfType is used (in requests, stateservice) we simply add the optional variables kdfMemory and kdfParalleism. The defaults are: 64 (MiB) for memory and 4 parallelism.

This reopens mistakenly closed: https://github.com/bitwarden/mobile/pull/2322

I just applied the commit as a patch on the latest master, though there were 2 errors, which I can check in ~2 hours.

### DIFF
--- a/src/App/Pages/Accounts/LockPageViewModel.cs
+++ b/src/App/Pages/Accounts/LockPageViewModel.cs
@@ -228,10 +228,7 @@ namespace Bit.App.Pages
             }
 
             ShowPassword = false;
-            var kdf = await _stateService.GetKdfTypeAsync();
-            var kdfIterations = await _stateService.GetKdfIterationsAsync();
-            var kdfMemory = await _stateService.GetKdfMemoryAsync();
-            var kdfParallelism = await _stateService.GetKdfParallelismAsync();
+            var kdfConfig = await _stateService.GetActiveUserCustomDataAsync(a => new KdfConfig(a?.Profile));
 
             if (PinLock)
             {
@@ -241,8 +238,7 @@ namespace Bit.App.Pages
                     if (_isPinProtected)
                     {
                         var key = await _cryptoService.MakeKeyFromPinAsync(Pin, _email,
-                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000),
-                            kdfMemory, kdfParallelism,
+                            kdfConfig,
                             await _stateService.GetPinProtectedKeyAsync());
                         var encKey = await _cryptoService.GetEncKeyAsync(key);
                         var protectedPin = await _stateService.GetProtectedPinAsync();
@@ -257,8 +253,7 @@ namespace Bit.App.Pages
                     }
                     else
                     {
-                        var key = await _cryptoService.MakeKeyFromPinAsync(Pin, _email,
-                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000), kdfMemory, kdfParallelism);
+                        var key = await _cryptoService.MakeKeyFromPinAsync(Pin, _email, kdfConfig);
                         failed = false;
                         Pin = string.Empty;
                         await AppHelpers.ResetInvalidUnlockAttemptsAsync();
@@ -283,7 +278,7 @@ namespace Bit.App.Pages
             }
             else
             {
-                var key = await _cryptoService.MakeKeyAsync(MasterPassword, _email, kdf, kdfIterations, kdfMemory, kdfParallelism);
+                var key = await _cryptoService.MakeKeyAsync(MasterPassword, _email, kdfConfig);
                 var storedKeyHash = await _cryptoService.GetKeyHashAsync();
                 var passwordValid = false;
 
@@ -317,8 +312,7 @@ namespace Bit.App.Pages
                         var protectedPin = await _stateService.GetProtectedPinAsync();
                         var encKey = await _cryptoService.GetEncKeyAsync(key);
                         var decPin = await _cryptoService.DecryptToUtf8Async(new EncString(protectedPin), encKey);
-                        var pinKey = await _cryptoService.MakePinKeyAysnc(decPin, _email,
-                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000), kdfMemory, kdfParallelism);
+                        var pinKey = await _cryptoService.MakePinKeyAysnc(decPin, _email, kdfConfig);
                         await _stateService.SetPinProtectedKeyAsync(await _cryptoService.EncryptAsync(key.Key, pinKey));
                     }
                     MasterPassword = string.Empty;
--- a/src/App/Pages/Accounts/LockPageViewModel.cs
+++ b/src/App/Pages/Accounts/LockPageViewModel.cs
@@ -230,6 +230,8 @@ namespace Bit.App.Pages
             ShowPassword = false;
             var kdf = await _stateService.GetKdfTypeAsync();
             var kdfIterations = await _stateService.GetKdfIterationsAsync();
+            var kdfMemory = await _stateService.GetKdfMemoryAsync();
+            var kdfParallelism = await _stateService.GetKdfParallelismAsync();
 
             if (PinLock)
             {
@@ -240,6 +242,7 @@ namespace Bit.App.Pages
                     {
                         var key = await _cryptoService.MakeKeyFromPinAsync(Pin, _email,
                             kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000),
+                            kdfMemory, kdfParallelism,
                             await _stateService.GetPinProtectedKeyAsync());
                         var encKey = await _cryptoService.GetEncKeyAsync(key);
                         var protectedPin = await _stateService.GetProtectedPinAsync();
@@ -255,7 +258,7 @@ namespace Bit.App.Pages
                     else
                     {
                         var key = await _cryptoService.MakeKeyFromPinAsync(Pin, _email,
-                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000));
+                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000), kdfMemory, kdfParallelism);
                         failed = false;
                         Pin = string.Empty;
                         await AppHelpers.ResetInvalidUnlockAttemptsAsync();
@@ -280,7 +283,7 @@ namespace Bit.App.Pages
             }
             else
             {
-                var key = await _cryptoService.MakeKeyAsync(MasterPassword, _email, kdf, kdfIterations);
+                var key = await _cryptoService.MakeKeyAsync(MasterPassword, _email, kdf, kdfIterations, kdfMemory, kdfParallelism);
                 var storedKeyHash = await _cryptoService.GetKeyHashAsync();
                 var passwordValid = false;
 
@@ -315,7 +318,7 @@ namespace Bit.App.Pages
                         var encKey = await _cryptoService.GetEncKeyAsync(key);
                         var decPin = await _cryptoService.DecryptToUtf8Async(new EncString(protectedPin), encKey);
                         var pinKey = await _cryptoService.MakePinKeyAysnc(decPin, _email,
-                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000));
+                            kdf.GetValueOrDefault(KdfType.PBKDF2_SHA256), kdfIterations.GetValueOrDefault(5000), kdfMemory, kdfParallelism);
                         await _stateService.SetPinProtectedKeyAsync(await _cryptoService.EncryptAsync(key.Key, pinKey));
                     }
                     MasterPassword = string.Empty;
--- a/src/App/Pages/Accounts/SetPasswordPageViewModel.cs
+++ b/src/App/Pages/Accounts/SetPasswordPageViewModel.cs
@@ -163,9 +163,9 @@ namespace Bit.App.Pages
                 return;
             }
 
-            var kdf = KdfType.PBKDF2_SHA256;
+            var kdfConfig = new KdfConfig(KdfType.PBKDF2_SHA256, Constants.KdfIterations, null, null);
             var email = await _stateService.GetEmailAsync();
-            var key = await _cryptoService.MakeKeyAsync(MasterPassword, email, kdf, Constants.KdfIterations, null, null);
+            var key = await _cryptoService.MakeKeyAsync(MasterPassword, email, kdfConfig);
             var masterPasswordHash = await _cryptoService.HashPasswordAsync(MasterPassword, key, HashPurpose.ServerAuthorization);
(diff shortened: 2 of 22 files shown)


quexten:
Ok, fixed the git warnings, should now be correct.

src/Core/Services/CryptoService.cs
```diff
@@ -9,6 +9,7 @@
 using Bit.Core.Models.Domain;
 using Bit.Core.Models.Response;
 using Bit.Core.Utilities;
+using static System.Net.WebRequestMethods;
```
kspearrin, line 12: remove?

src/Core/Services/PclCryptoFunctionService.cs
```diff
@@ -49,18 +49,15 @@ public Task<byte[]> Argon2Async(string password, string salt, int iterations, in
             password = NormalizePassword(password);
```
kspearrin, line 49: revert changes to this file

kspearrin (commented):
Do we need to update the Kdf constants like was done in the clients repo?

quexten:
> Do we need to update the Kdf constants like was done in the clients repo?

Yeah I was wondering this too. Seems const in c# can't be nullable though so at least memory and parallelism can't be added unless we move the 3 parameters to a kdfConfiguration object (like in the clients repo). Should I do that?

src/Core/Services/CryptoService.cs
quexten: Yeah... seems Visual studio snuck in some unused imports.

kspearrin:
> > Do we need to update the Kdf constants like was done in the clients repo?
> 
> Yeah I was wondering this too. Seems const in c# can't be nullable though so at least memory and parallelism can't be added unless we move the 3 parameters to a kdfConfiguration object (like in the clients repo). Should I do that?

Your constant defaults are not null values on the clients repo. They don't need to be here, either.

quexten:
> > > Do we need to update the Kdf constants like was done in the clients repo?
> > 
> > 
> > Yeah I was wondering this too. Seems const in c# can't be nullable though so at least memory and parallelism can't be added unless we move the 3 parameters to a kdfConfiguration object (like in the clients repo). Should I do that?
> 
> Your constant defaults are not null values on the clients repo. They don't need to be here, either.

Ah, you're right. I was about of:
https://github.com/quexten/clients/blob/674bddb609400a8c4b8830e2c0a6d4661d040ccb/libs/common/src/enums/kdfType.ts#L14

Not
https://github.com/bitwarden/clients/blob/edea0f43c180e6ec7000de787c19a954c81dcff8/libs/common/src/enums/kdfType.ts#L6-L8

src/Core/Services/CryptoService.cs
quexten: Fixed in e765f2de659d219f34b19c6523bf92baa958f809

fedemkr (changes requested):

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -230,6 +230,8 @@ public async Task SubmitAsync()
             ShowPassword = false;
             var kdf = await _stateService.GetKdfTypeAsync();
             var kdfIterations = await _stateService.GetKdfIterationsAsync();
+            var kdfMemory = await _stateService.GetKdfMemoryAsync();
+            var kdfParallelism = await _stateService.GetKdfParallelismAsync();
```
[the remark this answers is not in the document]
quexten: I agree, this is also consistent with the PR on the clients repo. Will get started on this.

src/Core/Services/StateService.cs
```diff
@@ -371,6 +371,38 @@ public async Task SetKdfIterationsAsync(int? value, string userId = null)
...
+        }
+
+        public async Task SetKdfParallelismAsync(int? value, string userId = null)
+        {
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var account = await GetAccountAsync(reconciledOptions);
+            account.Profile.KdfParallelism = value;
+            await SaveAccountAsync(account, reconciledOptions);
+        }
```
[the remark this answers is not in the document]
quexten: Removed all of these. For setting there is now 
`public async Task SetKdfConfigurationAsync(KdfConfig config, string userId = null)`

src/Core/Services/StateMigrationService.cs
```diff
@@ -115,6 +115,8 @@ private class V2Keys
             internal const string Keys_Stamp = "securityStamp";
             internal const string Keys_Kdf = "kdf";
             internal const string Keys_KdfIterations = "kdfIterations";
+            internal const string Keys_KdfMemory = "kdfMemory";
+            internal const string Keys_KdfParallelism = "kdfParallelis";
```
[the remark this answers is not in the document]
quexten: Removed in d3fb671ac8a05e5f57d6dcddda5fbf5af6642e13.

src/Core/Services/PclCryptoFunctionService.cs
quexten: Done in 8aabd1c4f328095572b66021ed12e54a63378a5b

src/Core/Constants.cs
```diff
@@ -47,6 +47,9 @@ public static class Constants
         public const int TotpDefaultTimer = 30;
         public const int PasswordlessNotificationTimeoutInMinutes = 15;
         public const int KdfIterations = 600000;
```
kspearrin, line 49: Should this be renamed like was done on clients repo?

src/Core/Services/CryptoService.cs
```diff
@@ -389,28 +389,27 @@ public async Task ToggleKeyAsync()
...
-                    CryptoHashAlgorithm.Sha256, kdfIterations.Value);
+                    CryptoHashAlgorithm.Sha256, kdfConfig.Iterations.Value);
             }
-            else if (kdf == KdfType.Argon2id)
+            else if (kdfConfig.Type == KdfType.Argon2id)
             {
-                var iterations = kdfIterations.Value;
-                const int parallelism = 1;
-                const int memory = 1024 * 16; // 16 MiB
+                var iterations = kdfConfig.Iterations.GetValueOrDefault(Constants.Argon2Iterations);
```
kspearrin, line 410: Didn't you add some validation checks here on clients repo?

src/Core/Services/StateMigrationService.cs
```diff
@@ -162,6 +162,8 @@ private async Task MigrateFrom2To3Async()
 
             var kdfType = await GetValueAsync<int?>(Storage.LiteDb, V2Keys.Keys_Kdf);
             var kdfIterations = await GetValueAsync<int?>(Storage.LiteDb, V2Keys.Keys_KdfIterations);
+            var kdfMemory = await GetValueAsync<int?>(Storage.LiteDb, V2Keys.Keys_KdfMemory);
+            var kdfParallelism = await GetValueAsync<int?>(Storage.LiteDb, V2Keys.Keys_KdfParallelism);
```
kspearrin, line 166: @fedemkr Is this even necessary? Wasn't this for migrating data when we moved to state storage?

src/Core/Services/CryptoService.cs
quexten: Ah, yeah good catch. That's still missing.

quexten: Added checks in afd42406060d62cb0726e52af49a9df6ca5deeea

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
```diff
@@ -186,8 +186,10 @@ public async Task SubmitAsync()
                 MasterPasswordHash = masterPasswordHash,
                 Key = encKey.Item2.EncryptedString,
                 MasterPasswordHint = Hint,
-                Kdf = kdf,
-                KdfIterations = Constants.KdfIterations,
+                Kdf = kdfConfig.Type.GetValueOrDefault(KdfType.PBKDF2_SHA256),
+                KdfIterations = kdfConfig.Iterations.GetValueOrDefault(Constants.KdfIterations),
```
fedemkr, line 190: The `Type` and Iterations are already set above with the default values so no need to add `GetValueOrDefault(...)` just use `kdfConfig.Type.Value` and `kdfConfig.Iterations.Value`

src/Core/Constants.cs
fedemkr: I think it would be better to have the same naming on the constants config for `PBKDF2` and `Argon2` between here and clients repo. So it's easier to find and maintain cross-repo.

src/Core/Models/Request/SetKeyConnectorKeyRequest.cs
```diff
@@ -9,15 +9,18 @@ public class SetKeyConnectorKeyRequest
...
 
-        public SetKeyConnectorKeyRequest(string key, KeysRequest keys,
-            KdfType kdf, int? kdfIterations, string orgIdentifier)
+        public SetKeyConnectorKeyRequest(string key, KeysRequest keys, KdfConfig kdfConfig, string orgIdentifier)
         {
             this.Key = key;
             this.Keys = keys;
-            this.Kdf = kdf;
-            this.KdfIterations = kdfIterations;
+            this.Kdf = kdfConfig.Type.GetValueOrDefault(KdfType.PBKDF2_SHA256);
```
fedemkr, line 20: Previously the default value wasn't being set, is it necessary now?

src/Core/Constants.cs
```diff
@@ -47,6 +47,9 @@ public static class Constants
...
+        public const int Argon2Iterations = 2;
+        public const int Argon2Memory = 19;
```
fedemkr, line 51: In order to be more guessable I think it would be better to include the unit on the constant, if I understand correctly it would be `Argon2MemoryInMb`

src/Core/Services/CryptoService.cs
```diff
@@ -389,28 +389,57 @@ public async Task ToggleKeyAsync()
...
+                    kdfConfig.Parallelism = Constants.Argon2Parallelism;
+                } 
+                else if (kdfConfig.Parallelism < 1)
+                {
+                    throw new Exception("Argon2 parallelism minimum is 1");
+                }
+
+                var iterations = kdfConfig.Iterations.GetValueOrDefault(Constants.Argon2Iterations);
+                var memory = kdfConfig.Memory.GetValueOrDefault(Constants.Argon2Memory) * 1024;
+                var parallelism = kdfConfig.Parallelism.GetValueOrDefault(Constants.Argon2Parallelism);
```
fedemkr, line 442: There is a repetition to fill the values if the original is `null` (with the if and with `GetValueOrDefault(...)`).
So I'd just fill the local variables first and then check against them or remove the local variables and use the updated `kdfConfig` when calling the service.
Also there is a formatting issue on line 426: `} else if (kdfConfig.Memory > 1024)` the `else if` should be in the next line.

fedemkr (changes requested):

src/Core/Constants.cs
quexten: Updated in [4ff8a7e](https://github.com/bitwarden/mobile/pull/2328/commits/4ff8a7ecdb6b2e8ea4c22cb378ae9e3b226a2659)

src/Core/Services/AuthService.cs
```diff
@@ -294,7 +292,7 @@ private async Task<SymmetricCryptoKey> MakePreloginKeyAsync(string masterPasswor
                     throw;
                 }
             }
-            return await _cryptoService.MakeKeyAsync(masterPassword, email, kdf, kdfIterations);
+            return await _cryptoService.MakeKeyAsync(masterPassword, email, kdfConfig.GetValueOrDefault(new KdfConfig(KdfType.PBKDF2_SHA256, 5000, null, null)));
```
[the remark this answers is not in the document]
quexten: Addressed in [5804842](https://github.com/bitwarden/mobile/pull/2328/commits/58048427009097c3497bddd92952c27bfc5490e0)

src/Core/Services/CryptoService.cs
quexten: Fixed in [5804842](https://github.com/bitwarden/mobile/pull/2328/commits/58048427009097c3497bddd92952c27bfc5490e0)

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
quexten: This is due to type and iterations being optional in the kdfconfig struct. Changing them to non-optional in the struct breaks some other things (f.e setting default iteration count in the CryptoService in the MakeKeyAsync function.

src/Core/Services/AuthService.cs
```diff
@@ -442,7 +440,8 @@ private async Task<SymmetricCryptoKey> MakePreloginKeyAsync(string masterPasswor
                 {
                     // SSO Key Connector Onboarding
                     var password = await _cryptoFunctionService.RandomBytesAsync(64);
-                    var k = await _cryptoService.MakeKeyAsync(Convert.ToBase64String(password), _tokenService.GetEmail(), tokenResponse.Kdf, tokenResponse.KdfIterations);
+                    var kdfConfig = new KdfConfig(tokenResponse.Kdf, tokenResponse.KdfIterations, tokenResponse.KdfMemory, tokenResponse.KdfParallelism);
+                    var k = await _cryptoService.MakeKeyAsync(Convert.ToBase64String(password), _tokenService.GetEmail(), kdfConfig);
```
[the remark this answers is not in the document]
quexten: Same applies for PreloginResponse I guess?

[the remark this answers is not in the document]
quexten: Added in [9ce0134](https://github.com/bitwarden/mobile/pull/2328/commits/9ce013487b05573f1764d96bed28eb58b3b62112) for both PreloginResponse and IdentityTokenResponse

src/Core/Models/Request/SetKeyConnectorKeyRequest.cs
quexten: This is due to the same reason as above:

> This is due to type and iterations being optional in the kdfconfig struct. Changing them to non-optional in the struct breaks some other things (f.e setting default iteration count in the CryptoService in the MakeKeyAsync function.) and require changes there

src/Core/Models/Request/SetKeyConnectorKeyRequest.cs
```diff
@@ -7,7 +7,7 @@ public class SetKeyConnectorKeyRequest
     {
         public string Key { get; set; }
         public KeysRequest Keys { get; set; }
-        public KdfType Kdf { get; set; }
+        public KdfType? Kdf { get; set; }
```
fedemkr, line 10: This shouldn't be null, [server side](https://github.com/bitwarden/server/blob/cb1ba50ce26ce33b2a5acf30536a2075e4fadebd/src/Api/Models/Request/Accounts/SetKeyConnectorKeyRequestModel.cs#L16) expects always a value and it's required.
I've checked and you can return to the `GetValueOrDefault(...)` on line 20 because previously the plain parameter of the type was not null. Sorry about the confusion

fedemkr (changes requested):
Excellent, one more change and that's all 😄

src/Core/Models/Request/SetKeyConnectorKeyRequest.cs
quexten: Done in 88c554c91701ddc33ced7743ec029ef3fee9f8f0.

PR #2333, opened 2023-01-30
[SG-744] Add claimed domain logic to mobile

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Now that we have the ability for domain admins to claim domains, this task will add the logic to each of the clients - which have already had Two-Step Login added - to check the API for the provided domain name on the first step of the login process to see how to handle SSO:
When a domain has been claimed by an organization the “Enterprise Single Sign-On” button on step 2 takes the user directly to the domain specific SSO action and does not require them to enter an org identifier.

## Code changes

Add new method to check if the email has an SSO domain claimed.
On login with SSO, in the Init check if the the email has a claimed SSO Domain and automatically jump to the Login.
**Note:** Due to the difficulty of testing this locally, the code of this new feature will be tested by the QA team and fixes will be done afterwards 

## Before you submit


src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -63,9 +65,26 @@ public string OrgIdentifier
...
+                }
+
+                if (string.IsNullOrWhiteSpace(OrgIdentifier))
+                {
+                    OrgIdentifier = await _stateService.GetRememberedOrgIdentifierAsync();
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.Exception(ex);
```
fedemkr, line 83: In case an exception is raised, does it leave the view in a good state?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -207,5 +226,19 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
+            var userEmail = await _stateService.GetRememberedEmailAsync();
+            var claimedDomainOrg = await _organizationService.GetClaimedOrganizationDomainAsync(userEmail);
+            if (!string.IsNullOrEmpty(claimedDomainOrg))
+            {
+                OrgIdentifier = claimedDomainOrg;
+                await LogInAsync();
+                return true;
+            }
+
+            return false;
```
fedemkr, line 241: Usually this is done the other way around, so you check and `return false` first:
```c#
if (string.IsNullOrEmpty(claimedDomainOrg))
{
    return false;
}

OrgIdentifier = claimedDomainOrg;
await LogInAsync();
return true;
```

src/Core/Abstractions/IApiService.cs
```diff
@@ -90,5 +90,6 @@ public interface IApiService
         Task<PasswordlessLoginResponse> PostCreateRequestAsync(PasswordlessCreateLoginRequest passwordlessCreateLoginRequest);
         Task<string> GetUsernameFromAsync(ForwardedEmailServiceType service, UsernameGeneratorConfig config);
         Task<bool> GetKnownDeviceAsync(string email, string deviceIdentifier);
+        Task<OrganizationDomainSsoDetailsResponse> GetOrgDomainSsoDetails(string email);
```
fedemkr, line 93: Add `Async` suffix

src/Core/Models/Response/OrganizationDomainSsoDetailsResponse.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Bit.Core.Models.Response
+{
+    public class OrganizationDomainSsoDetailsResponse
+    {
+        public bool SsoAvailable { get; private set; }
+        public string DomainName { get; private set; }
+        public string OrganizationIdentifier { get; private set; }
+        public bool SsoRequired { get; private set; }
+        public DateTime? VerifiedDate { get; private set; }
+    }
```
fedemkr, line 11: Mmm I don't think this will work with `private set` cause the json deserializer won't be able to set the values (unless private setting is activated). Why do you need the properties to be `private`?

src/Core/Services/OrganizationService.cs
```diff
@@ -51,5 +54,19 @@ public async Task ClearAllAsync(string userId)
...
+            try
+            {
+                var response = await _apiService.GetOrgDomainSsoDetails(userEmail);
+                return response.OrganizationIdentifier;
+            }
+            catch (System.Exception ex)
+            {
+                // api throws 404 if there is no domain claimed
+                return string.Empty;
+            }
```
fedemkr, line 69: Is it ok that any exception returns empty string? or should it only check when the API exception is because of a 404 code?

fedemkr (changes requested):

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
andrebispo5: Yes. This is an auto navigation feature, if something goes wrong, it will stay in the screen for the user to manually type the organization identifier.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
andrebispo5: You are right.

src/Core/Models/Response/OrganizationDomainSsoDetailsResponse.cs
andrebispo5: It was copy paste from the server i didn't even realised, thanks. Will change to public.

src/Core/Services/OrganizationService.cs
andrebispo5: If something happens we just want to ignore it and let the user be in the screen to continue to login manually. What i missed was logging it, just in case.

src/Core/Services/OrganizationService.cs
```diff
@@ -5,13 +5,15 @@
...
 using Bit.Core.Models.Request;
+using Bit.Core.Utilities;
 
 namespace Bit.Core.Services
 {
     public class OrganizationService : IOrganizationService
     {
         private readonly IStateService _stateService;
         private readonly IApiService _apiService;
+        readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>("logger");
```
fedemkr, line 16: No need to specify the `"logger"` string key. Remove it.

fedemkr (changes requested):

fedemkr (approved):

src/Core/Abstractions/IStateService.cs
```diff
@@ -162,7 +162,7 @@ public interface IStateService
         Task<bool> GetLastUserShouldConnectToWatchAsync();
         Task SetAvatarColorAsync(string value, string userId = null);
         Task<string> GetAvatarColorAsync(string userId = null);
-
-
+        Task<string> GetPreLoginEmail();
+        Task SetPreLoginEmail(string value);
```
fedemkr, line 166: Add `Async` suffix

src/Core/Services/StateService.cs
```diff
@@ -1302,6 +1302,20 @@ public async Task<string> GetAvatarColorAsync(string userId = null)
             ))?.Profile?.AvatarColor;
         }
 
+        public async Task<string> GetPreLoginEmail()
+        {
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.PreLoginEmailKey;
+            return await GetValueAsync<string>(key, options);
```
fedemkr, line 1309: Set the constant as the parameter directly, no need to assign it to a variable

src/Core/Services/StateService.cs
```diff
@@ -1302,6 +1302,20 @@ public async Task<string> GetAvatarColorAsync(string userId = null)
...
+        }
+
+        public async Task SetPreLoginEmail(string value)
+        {
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.PreLoginEmailKey;
+            await SetValueAsync(key, value, options);
```
fedemkr, line 1316: Set the constant as the parameter directly, no need to assign it to a variable

andrebispo5:
Added new state variable to hold the pre login email to be used in the SSO login page. This was done because there was a case in `BaseLockPasswordViewController` where it was difficult to pass the email through navigation.

src/Core/Services/OrganizationService.cs
```diff
@@ -51,5 +56,20 @@ public async Task ClearAllAsync(string userId)
...
+                var response = await _apiService.GetOrgDomainSsoDetailsAsync(userEmail);
+                return response.OrganizationIdentifier;
+            }
+            catch (System.Exception ex)
+            {
+                // api throws 404 if there is no domain claimed
+                _logger.Value.Exception(ex);
+                return string.Empty;
+            }
+        }
```
fedemkr, line 73: if the `userEmail` is `null`, does it makes sense to call the service? I think we could put a null check that returns an empty string.

fedemkr (changes requested):

andrebispo5:
Updated error handling after discussion with the QA team and it was decided to fix web form validations.

src/Core/Services/OrganizationService.cs
```diff
@@ -63,13 +63,25 @@ public async Task ClearAllAsync(string userId)
...
+                if (ex is ApiException apiEx && apiEx.Error != null && apiEx.Error.StatusCode == HttpStatusCode.NotFound)
+                {
+                    // this is a valid case so there is no need to show an error
+                    return null;
+                }
 
-            var response = await _apiService.GetOrgDomainSsoDetailsAsync(userEmail);
-            return response;
+                throw ex;
+            }
```
fedemkr, line 84: You don't need to catch all exception to do this, you can just write it as:
```c#
catch (ApiException ex) when (apiEx.Error?.StatusCode == HttpStatusCode.NotFound)
{
    // this is a valid case so there is no need to show an error
    return null;
}
```

fedemkr (changes requested):

src/Core/Services/OrganizationService.cs
```diff
@@ -1,19 +1,28 @@
...
+using Bit.Core.Models.Response;
+using Bit.Core.Utilities;
 
 namespace Bit.Core.Services
 {
     public class OrganizationService : IOrganizationService
     {
         private readonly IStateService _stateService;
+        private readonly IApiService _apiService;
+        readonly LazyResolve<ILogger> _logger = new LazyResolve<ILogger>();
```
fedemkr, line 20: I don't see the logger being used, if that's the case remove it (if for some reason this would be needed, it should be injected through the ctor)

src/Core/Services/OrganizationService.cs
```diff
@@ -1,19 +1,28 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
```
fedemkr, line 2: I don't see this using being used, if that's the case remove it (use the automatic clear of using)

fedemkr (changes requested):

fedemkr (approved):

PR #2366, opened 2023-02-13
[PS-2280] Retain app settings on logout

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

These changes allow the app to retain account-specific settings after a hard logout (previous would only retain on soft logout).  In order to accomplish this without re-architecting the way soft logout works with multiple users, some stored settings (`VaultTimeout`, `VaultTimeoutAction`, and `ScreenCaptureAllowed`) had to be relocated from the state object to normal local storage with an account-specific pref name (follows existing pattern).  This requires a new state migration, which resulted in a conflict with a previous migration with 3 values that were originally account-specific (`DisableFavicon`, `Theme`, and `AutoDarkTheme` but used trickery to apply globally to match other clients) requiring migrating those to true global values (remove the trickery).

## Code changes

* **StateService.cs:** 
   * Update getter/setter for `VaultTimeout`, `VaultTimeoutAction`, and `ScreenCaptureAllowed` to support account-specific storage key
   * Update getter/setter for `DisableFavicon`, `Theme`, and `AutoDarkTheme` to remove account-specific storage key
   * Remove account-to-global trickery method
   * Remove deletion of app settings in `RemoveAccountAsync`
   * Remove `DisableFavicon`, `Theme`, and `AutoDarkTheme` getting/setting trickery in `ScaffoldNewAccountAsync`
   * Remove setting defaults for `VaultTimeout`, `VaultTimeoutAction`, and `ScreenCaptureAllowed` (using established pattern of getter fallbacks since these are no longer in the state object)
   * Remove purging of any existing data when logging into account that was previously logged in (sensitive data is already removed on logout and we want the settings to remain)
* **StateMigrationService.cs:** Added new 3to4 migration block, details in Objective section above (realized I should not be referencing `Constants` for keys in 2to3 migration as those change over time and need to be self-contained in the service - will fix this in a subsequent PR since it adds a lot of noise)
* **CoreHelpers.cs:** Removed `ForceScreenCaptureEnabled` as it made debugging the toggle impossible, now simply defaults to allowing screenshots in debug builds but still allows toggling (returns `CoreHelpers.InDebugMode()` as fallback default in `StateService.GetScreenCaptureAllowedAsync(..)`
* **Account.cs:** Marked `VaultTimeout`, `VaultTimeoutAction`, and `ScreenCaptureAllowed` as `[Obsolete]` to allow older migrations to function, will eventually remove
* **Constants.cs:** Changed `VaultTimeout`, `VaultTimeoutAction`, and `ScreenCaptureAllowed` to account specific keys, and `DisableFavicon`, `Theme`, and `AutoDarkTheme` to global keys

## Before you submit

### DIFF
--- a/src/Core/Services/StateMigrationService.cs
+++ b/src/Core/Services/StateMigrationService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Bit.Core.Abstractions;
 using Bit.Core.Enums;
@@ -329,62 +330,58 @@ namespace Bit.Core.Services
         private async Task MigrateFrom3To4Async()
         {
             var state = await GetValueAsync<State>(Storage.LiteDb, Constants.StateKey);
-            if (state?.Accounts != null)
+            if (state?.Accounts is null)
             {
-                string firstUserId = null;
+                // Update stored version
+                await SetLastStateVersionAsync(4);
+                return;
+            }
 
-                // move values from state to standalone values in LiteDB
-                foreach (var account in state.Accounts)
+            string firstUserId = null;
+
+            // move values from state to standalone values in LiteDB
+            foreach (var account in state.Accounts.Where(a => a.Value?.Profile?.UserId != null))
+            {
+                var userId = account.Value.Profile.UserId;
+                if (firstUserId == null)
                 {
-                    var userId = account.Value?.Profile?.UserId;
-                    if (userId != null)
-                    {
-                        if (firstUserId == null)
-                        {
-                            firstUserId = userId;
-                        }
-                        var vaultTimeout = account.Value?.Settings?.VaultTimeout;
-                        await SetValueAsync(Storage.LiteDb, V4Keys.VaultTimeoutKey(userId), vaultTimeout);
-
-                        var vaultTimeoutAction = account.Value?.Settings?.VaultTimeoutAction;
-                        await SetValueAsync(Storage.LiteDb, V4Keys.VaultTimeoutActionKey(userId), vaultTimeoutAction);
-
-                        var screenCaptureAllowed = account.Value?.Settings?.ScreenCaptureAllowed;
-                        await SetValueAsync(Storage.LiteDb, V4Keys.ScreenCaptureAllowedKey(userId),
-                            screenCaptureAllowed);
-                    }
+                    firstUserId = userId;
                 }
+                var vaultTimeout = account.Value.Settings?.VaultTimeout;
+                await SetValueAsync(Storage.LiteDb, V4Keys.VaultTimeoutKey(userId), vaultTimeout);
 
-                // use values from first userId to apply globals
-                if (firstUserId != null)
-                {
-                    var theme = await GetValueAsync<int?>(Storage.LiteDb, V3Keys.ThemeKey(firstUserId));
-                    await SetValueAsync(Storage.LiteDb, V4Keys.ThemeKey, theme);
+                var vaultTimeoutAction = account.Value.Settings?.VaultTimeoutAction;
+                await SetValueAsync(Storage.LiteDb, V4Keys.VaultTimeoutActionKey(userId), vaultTimeoutAction);
 
-                    var autoDarkTheme = await GetValueAsync<int?>(Storage.LiteDb, V3Keys.AutoDarkThemeKey(firstUserId));
-                    await SetValueAsync(Storage.LiteDb, V4Keys.AutoDarkThemeKey, autoDarkTheme);
+                var screenCaptureAllowed = account.Value.Settings?.ScreenCaptureAllowed;
+                await SetValueAsync(Storage.LiteDb, V4Keys.ScreenCaptureAllowedKey(userId), screenCaptureAllowed);
+            }
 
-                    var disableFavicon =
-                        await GetValueAsync<bool?>(Storage.LiteDb, V3Keys.DisableFaviconKey(firstUserId));
-                    await SetValueAsync(Storage.LiteDb, V4Keys.DisableFaviconKey, disableFavicon);
-                }
+            // use values from first userId to apply globals
+            if (firstUserId != null)
+            {
+                var theme = await GetValueAsync<int?>(Storage.LiteDb, V3Keys.ThemeKey(firstUserId));
+                await SetValueAsync(Storage.LiteDb, V4Keys.ThemeKey, theme);
+
+                var autoDarkTheme = await GetValueAsync<int?>(Storage.LiteDb, V3Keys.AutoDarkThemeKey(firstUserId));
+                await SetValueAsync(Storage.LiteDb, V4Keys.AutoDarkThemeKey, autoDarkTheme);
+
+                var disableFavicon = await GetValueAsync<bool?>(Storage.LiteDb, V3Keys.DisableFaviconKey(firstUserId));
+                await SetValueAsync(Storage.LiteDb, V4Keys.DisableFaviconKey, disableFavicon);
             }
 
             // Update stored version
             await SetLastStateVersionAsync(4);
 
             // Remove old data
-            if (state?.Accounts != null)
+            foreach (var account in state.Accounts)
             {
-                foreach (var account in state.Accounts)
+                var userId = account.Value?.Profile?.UserId;
+                if (userId != null)
                 {
-                    var userId = account.Value?.Profile?.UserId;
-                    if (userId != null)
-                    {
-                        await RemoveValueAsync(Storage.LiteDb, V3Keys.ThemeKey(userId));
-                        await RemoveValueAsync(Storage.LiteDb, V3Keys.AutoDarkThemeKey(userId));
-                        await RemoveValueAsync(Storage.LiteDb, V3Keys.DisableFaviconKey(userId));
-                    }
+                    await RemoveValueAsync(Storage.LiteDb, V3Keys.ThemeKey(userId));
+                    await RemoveValueAsync(Storage.LiteDb, V3Keys.AutoDarkThemeKey(userId));
+                    await RemoveValueAsync(Storage.LiteDb, V3Keys.DisableFaviconKey(userId));
                 }
             }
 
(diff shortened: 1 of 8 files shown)


src/Core/Services/StateMigrationService.cs
```diff
@@ -314,6 +317,90 @@ private async Task MigrateFrom2To3Async()
...
+            if (state?.Accounts != null)
+            {
+                string firstUserId = null;
+
+                // move values from state to standalone values in LiteDB
+                foreach (var account in state.Accounts)
+                {
+                    var userId = account.Value?.Profile?.UserId;
+                    if (userId != null)
+                    {
```
fedemkr, line 341: You can write this as:
```c#
foreach (var account in state.Accounts.Where(a => account.Value?.Profile?.UserId != null))
```
and you avoid the if afterwards.
Also, if you were to keep the `if`, I would write it inverted so that it doesn't create arrow code

src/Core/Services/StateService.cs
```diff
@@ -539,55 +539,50 @@ public async Task SetLastActiveTimeAsync(long? value, string userId = null)
 
         public async Task<int?> GetVaultTimeoutAsync(string userId = null)
         {
-            return (await GetAccountAsync(
-                ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync())
-            ))?.Settings?.VaultTimeout;
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var key = Constants.VaultTimeoutKey(reconciledOptions.UserId);
+            return await GetValueAsync<int?>(key, reconciledOptions) ?? 15;
```
fedemkr, line 545: I think that `15` should be extracted to some constant

src/Core/Services/StateService.cs
```diff
@@ -650,23 +645,18 @@ public async Task SetLastBuildAsync(string value)
...
-        public async Task<bool?> GetDisableFaviconAsync(string userId = null)
+        public async Task<bool?> GetDisableFaviconAsync()
         {
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.DisableFaviconKey(reconciledOptions.UserId);
-            return await GetValueAsync<bool?>(key, reconciledOptions);
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.DisableFaviconKey;
+            return await GetValueAsync<bool?>(key, options);
```
fedemkr, line 652: I would avoid assigning the `key` line and just use the `Constants.DisableFaviconKey` inline in the func. Furthermore, it has `Key` as suffix so it's obvious that it's a key in the parameter so we don't lose that either when reading it.

src/Core/Services/StateService.cs
```diff
@@ -650,23 +645,18 @@ public async Task SetLastBuildAsync(string value)
...
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.DisableFaviconKey(reconciledOptions.UserId);
-            await SetValueAsync(key, value, reconciledOptions);
-
-            // TODO remove this to restore per-account DisableFavicon support
-            SetValueGloballyAsync(Constants.DisableFaviconKey, value, reconciledOptions).FireAndForget();
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.DisableFaviconKey;
+            await SetValueAsync(key, value, options);
```
fedemkr, line 659: Ídem as `Constants.DisableFaviconKey`

src/Core/Services/StateService.cs
```diff
@@ -937,42 +927,32 @@ public async Task SetRememberedOrgIdentifierAsync(string value)
...
-        public async Task<string> GetThemeAsync(string userId = null)
+        public async Task<string> GetThemeAsync()
         {
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.ThemeKey(reconciledOptions.UserId);
-            return await GetValueAsync<string>(key, reconciledOptions);
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.ThemeKey;
+            return await GetValueAsync<string>(key, options);
```
fedemkr, line 934: Ídem as `Constants.DisableFaviconKey`

src/Core/Services/StateService.cs
```diff
@@ -937,42 +927,32 @@ public async Task SetRememberedOrgIdentifierAsync(string value)
...
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.ThemeKey(reconciledOptions.UserId);
-            await SetValueAsync(key, value, reconciledOptions);
-
-            // TODO remove this to restore per-account Theme support
-            SetValueGloballyAsync(Constants.ThemeKey, value, reconciledOptions).FireAndForget();
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.ThemeKey;
+            await SetValueAsync(key, value, options);
```
fedemkr, line 941: Ídem as `Constants.DisableFaviconKey`

src/Core/Services/StateService.cs
```diff
@@ -937,42 +927,32 @@ public async Task SetRememberedOrgIdentifierAsync(string value)
...
-        public async Task<string> GetAutoDarkThemeAsync(string userId = null)
+        public async Task<string> GetAutoDarkThemeAsync()
         {
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.AutoDarkThemeKey(reconciledOptions.UserId);
-            return await GetValueAsync<string>(key, reconciledOptions);
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.AutoDarkThemeKey;
+            return await GetValueAsync<string>(key, options);
```
fedemkr, line 948: Ídem as `Constants.DisableFaviconKey`

src/Core/Services/StateService.cs
```diff
@@ -937,42 +927,32 @@ public async Task SetRememberedOrgIdentifierAsync(string value)
...
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var key = Constants.AutoDarkThemeKey(reconciledOptions.UserId);
-            await SetValueAsync(key, value, reconciledOptions);
-
-            // TODO remove this to restore per-account Theme support
-            SetValueGloballyAsync(Constants.AutoDarkThemeKey, value, reconciledOptions).FireAndForget();
+            var options = await GetDefaultStorageOptionsAsync();
+            var key = Constants.AutoDarkThemeKey;
+            await SetValueAsync(key, value, options);
```
fedemkr, line 955: Ídem as `Constants.DisableFaviconKey`

src/Core/Services/StateService.cs
```diff
@@ -1551,28 +1480,6 @@ private async Task ScaffoldNewAccountAsync(Account account)
             {
                 state.Accounts = new Dictionary<string, Account>();
             }
-            if (state.Accounts.ContainsKey(account.Profile.UserId))
-            {
-                // Run cleanup pass on existing account before proceeding
-                await RemoveAccountAsync(account.Profile.UserId, false);
```
fedemkr, line 1557: I see that `RemoveAccountAsync(...)` does more stuff other than things related to global properties of this PR. Are you sure it's ok to remove this line?

src/Core/Services/StateMigrationService.cs
```diff
@@ -314,6 +317,90 @@ private async Task MigrateFrom2To3Async()
...
+                        await RemoveValueAsync(Storage.LiteDb, V3Keys.ThemeKey(userId));
+                        await RemoveValueAsync(Storage.LiteDb, V3Keys.AutoDarkThemeKey(userId));
+                        await RemoveValueAsync(Storage.LiteDb, V3Keys.DisableFaviconKey(userId));
+                    }
+                }
+            }
+
+            // Removal of old state data will happen organically as state is rebuilt in app
+        }
```
fedemkr, line 392: In order to avoid arrow code, I'd write it as:
```c#
if (state?.Accounts is null)
{
    // Update stored version
    await SetLastStateVersionAsync(4);
    return;
}

string firstUserId = null;
....

// Update stored version
await SetLastStateVersionAsync(4);

// Remove old data
...
```
It also helps so that you don't have to check it while removing old data

fedemkr (changes requested):

src/Core/Services/StateMigrationService.cs
```diff
@@ -314,6 +317,90 @@ private async Task MigrateFrom2To3Async()
             await RemoveValueAsync(Storage.LiteDb, V2Keys.Keys_PassGenHistory);
         }
 
+        // v3 to v4 Migration
+
+        private class V3Keys
+        {
+            internal static string ThemeKey(string userId) => $"theme_{userId}";
+            internal static string AutoDarkThemeKey(string userId) => $"autoDarkTheme_{userId}";
+            internal static string DisableFaviconKey(string userId) => $"disableFavicon_{userId}";
```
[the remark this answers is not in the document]
mp-bw: Not sure I'm following - changing these to `private` breaks the ability to access them from `MigrateFrom3To4Async`

src/Core/Services/StateService.cs
mp-bw: If it's ok I'd like to add this (and the others mentioned below) to the followup PR mentioned in the description.  It's a pattern used throughout all of `StateService` so changing it will be noisy.

src/Core/Services/StateService.cs
mp-bw: Yes, the other things were a pre-login safety measure to pave over any existing data in case a previous logout didn't execute cleanly for any reason, and that situation no longer applies.

src/Core/Services/StateMigrationService.cs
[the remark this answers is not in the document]
fedemkr: oh my bad, thought it was possible

fedemkr (approved):

PR #2331, opened 2023-01-27
[PS-2330][PS-2329][PS-2332][PS-2333] Checking for username terms on Android Autofill

aj-rosado:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective
Improving the way we recognize fields as Username and Password at the login autofill to avoid filling username fields with passwords or not detecting username fields as valid username fields

## Code changes

* **FieldCollection.cs:** Added common username terms 
On fields that identify as password we check if they don't have "Username terms" on hints to avoid filling passwords on usernames.
In cases we haven't found any Username Field, we will check if there is any Field with any hint or IdEntry that matches our "Username Terms"

## Before you submit
- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #2367, opened 2023-02-13
[PS-2486] Finally stop filling password into username field

larena1:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

The logic in #2331 is unfortunately not very reliable as it'll only detect fields that have one of "email", "phone" or "username" in their id as username fields.
This PR ensures that additonally fields that have TextVariationWebEmailAddress are also detected as username fields.

## Code changes

## Before you submit


aj-rosado:
Hello, thank you for your contribution.

I’m unable to reproduce the issue, on the Field with the `IdEntry = i0116` the `InputType` doesn’t have the flag `InputTypes.TextVariationWebEmailAddress`. Instead it is returning the `TextVariationPassword` and that’s the reason why it was filling with the password. As you can see on the image below.
field-data


Now, the implemented fix doesn’t look only to the `IdEntry` it also looks into the `Hints`, as we still have improvements to make, right now on my emulator with Android 12 it is working. That field is being identified as username.

Screenshot 2023-02-16 at 23 16 53

I'm sharing a video showing the behaviour I'm seeing on my emulator on the current master branch. As you can see, we've tried to verify your findings but we're experiencing something different than you are. Would it be possible for you to attach a video of the behaviour you're seeing both with and without this PR so we can understand better? We really appreciate your help!

https://user-images.githubusercontent.com/109146700/219656121-86b54ec3-7c26-4a1e-8636-ab610be4085b.mov

larena1:
> I’m unable to reproduce the issue, on the Field with the `IdEntry = i0116` the `InputType` doesn’t have the flag `InputTypes.TextVariationWebEmailAddress`. Instead it is returning the `TextVariationPassword` and that’s the reason why it was filling with the password. As you can see on the image below.

Do not trust the inspector on this. It wouldn't show TextVariationWebEmailAddress but it'll show if you log the value. Unfortunately your second screenshot is also cut off but can you maybe see TextVariationWebEmailAddress there?

larena1:
I understand now why it works for you. Hint seems to come from aria-label which in your language contains the term "email" and thus will be identified as username field. This is not true for other languages though.

aj-rosado:
Thank you for helping us understand! It seems like this PR would be a very positive addition to our codebase. We'll target testing this soon.

larena1:
I've also added TextVariationEmailAddress now. Not sure where it's used but you probably don't want passwords in there too.

aj-rosado:
@larena1 thank you for your contribution, the validation for `TextVariationWebEmailAddress` seems to be working fine however the `TextVariationEmailAddress` flag seems to break some password fills. Could you please remove this one?

aj-rosado (changes requested):
The `InputTypes.TextVariationEmailAddress` is affecting password fill on some cases. Please remove it

aj-rosado (approved):

aj-rosado:
Thank for your contribution, this will be part of a future release!

PR #2372, opened 2023-02-15
[EC-1045] add vault timeout action to policy

jlf0dev:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

This change locks the "Vault timeout action" in settings if the vault timeout policy is set with a specified action. It also changes the callout in settings to include the vault timeout action.

## Code changes

## Screenshots

image

image

## Before you submit


src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -374,6 +374,10 @@ public async Task ApproveLoginRequestsAsync()
 
         public async Task VaultTimeoutActionAsync()
         {
+            if (!string.IsNullOrEmpty(_policyService.GetPolicyString(_vaultTimeoutPolicy, "action")))
+            {
+                return;
+            }
```
fedemkr, line 380: I know we're already using hardcoded strings to get policies, but could we have a place where all the Policies are as constants. It could be in the `Policy` class.
This one in particular is being used more than once so IMO it makes sense to extract it to a constant somewhere.
In order not to pollute this PR, we could add a tech debt task to move all the others hardcoded policy strings into constants as well.
Also, I wanted to add that if we have all the strings as constants in one file it's easier to see which policies are being implemented in this repo and which are not.

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -598,13 +602,37 @@ public void BuildList()
...
+                var timeoutAction = _policyService.GetPolicyString(_vaultTimeoutPolicy, "action");
+                if (maximumTimeout != default && timeoutAction != default)
                 {
-                    Name = string.Format(AppResources.VaultTimeoutPolicyInEffect,
+                    securityItems.Insert(0, new SettingsPageListItem
+                    {
+                        Name = string.Format(AppResources.VaultTimeoutPolicyWithActionInEffect,
+                        Math.Floor((float)maximumTimeout / 60),
+                        maximumTimeout % 60,
+                        timeoutAction == "lock" ? AppResources.Lock : AppResources.LogOut),
```
fedemkr, line 613: IMO this `"lock"` value should be in the same constants file as the policies constants. And also IMO this should be tabbed or at first glance it might be confused with properties of the `SettingsPageListItem` instead of params of the `string.Format`

fedemkr (changes requested):

src/Core/Services/PolicyService.cs
```diff
@@ -247,6 +247,19 @@ private bool isExcemptFromPolicies(Organization organization, PolicyType policyT
...
+        {
+            if (policy.Data.ContainsKey(key))
+            {
+                var value = policy.Data[key];
+                if (value != null)
+                {
+                    return (string)value;
+                }
+            }
+            return null;
```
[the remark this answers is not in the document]
jlf0dev: Added! Been in JS land for too long 😄

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
jlf0dev: 👍

fedemkr (dismissed):

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
fedemkr: Yes, IMO they should be in the `Policy` class because I think that class is responsible of knowing which are the valid string values of policies. I think a tech debt task for this would be good so we put all the policies in one place 😄

fedemkr (approved):

PR #2395, opened 2023-03-01
[PM-1249] Clear/reset password/PIN fields on login/lock screen when app is backgrounded

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When backgrounding the app, clear any partially-filled text in the password and/or PIN fields on the login and lock screens.  Also reset the visibility toggle to default (not visible) to prevent any accidental reveals when entering password/pin after foregrounding the app.

## Code changes

* **App.xaml.cs:** Added new `ClearSensitiveFieldsAsync` method to send `ClearSensitiveFields` message, called from both `SleptAsync` and `ResumedAsync`
* **LockPage/LoginPage.xaml.cs:** Registered `BroadcasterService` to handle `ClearSensitiveFields` message and call method to set fields to `string.Empty` and `ShowPassword` to `false`
* **LockPageViewModel.cs:** Added `_masterPassword` and `_pin` backing vars to allow for programmatic clearing

## Before you submit


src/App/App.xaml.cs
```diff
@@ -335,6 +337,14 @@ public async Task UpdateThemeAsync()
             });
         }
 
+        private async Task ClearSensitiveFieldsAsync()
+        {
+            await Device.InvokeOnMainThreadAsync(() =>
+            {
+                _messagingService.Send(Constants.ClearSensitiveFields);
+            });
+        }
```
fedemkr, line 346: This shouldn't need to be invoked on the main thread, the receiver of the message should be the one who does that, or isn't the message being received because of being in another thread?

src/App/Pages/Accounts/LockPage.xaml.cs
```diff
@@ -22,6 +26,7 @@ public LockPage(AppOptions appOptions = null, bool autoPromptBiometric = true)
             _appOptions = appOptions;
             _autoPromptBiometric = autoPromptBiometric;
             InitializeComponent();
+            _broadcasterService = ServiceContainer.Resolve<IBroadcasterService>("broadcasterService");
```
fedemkr, line 29: No need to use the string key to resolve the service

src/App/Pages/Accounts/LockPage.xaml.cs
```diff
@@ -64,6 +69,13 @@ public async Task PromptBiometricAfterResumeAsync()
         protected override async void OnAppearing()
         {
             base.OnAppearing();
+            _broadcasterService.Subscribe(nameof(LockPage), async (message) =>
+            {
+                if (message.Command == Constants.ClearSensitiveFields)
+                {
+                    Device.BeginInvokeOnMainThread(ResetPinPasswordFields);
+                }
+            });
```
fedemkr, line 78: Either mark as `async` and use `await Device.InvokeOnMainThreadAsync(...)` or remove the `async` keyword. As it's now, it doesn't serve any purpose.
Also, if it's marked as `async` then the whole body should be wrapped in a `try...catch` to prevent crashes if an exception arises

src/App/Pages/Accounts/LoginPage.xaml.cs
```diff
@@ -70,6 +73,13 @@ public LoginPage(string email = null, AppOptions appOptions = null)
         protected override async void OnAppearing()
         {
             base.OnAppearing();
+            _broadcasterService.Subscribe(nameof(LoginPage), async (message) =>
+            {
+                if (message.Command == Constants.ClearSensitiveFields)
+                {
+                    Device.BeginInvokeOnMainThread(ResetPasswordField);
+                }
+            });
```
fedemkr, line 82: Idem as `LockPage`, move this and the reset method to the ViewModel

fedemkr (changes requested):

src/App/App.xaml.cs
mp-bw: The latter I believe.  I'm unable to update the UI in Android without invoking this on the main thread.

src/App/Pages/Accounts/LockPage.xaml.cs
```diff
@@ -102,6 +114,20 @@ protected override async void OnAppearing()
...
+            {
+                _vm.MasterPassword = string.Empty;
+                _vm.Pin = string.Empty;
+                _vm.ShowPassword = false;
+            }
+            catch (Exception ex)
+            {
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
+            }
+        }
```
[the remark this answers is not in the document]
mp-bw: I changed this to a public method in the ViewModel, but we seem to have a pattern where broadcast subscriptions happen on the page and not the ViewModel, so I left that as is (There is one exception (TwoFactor) but on closer inspection that appears to be something left behind and serves no purpose)

src/App/App.xaml.cs
fedemkr: Have you tried debugging it to see if the message arrives or where does it stop? Because from what I can see the receiver is already invoking on the main thread.
I'd be very helpful to see where it stops so we can provide a better way to handle that instead of the caller invoking on the main thread just to send a message.

mp-bw: Yessir, the message does not arrive in `LoginPage` or `LockPage` unless invoked on the main thread from `App`.  I'm able to follow the transmission into `BroadcasterService` and through the for loop in `Send`.  The only time `LockPage` is included as a `sub` in the resume flow is when the message is sent from the main thread (so `updatedTheme` goes to `LockPage` but `clearSensitiveFields` doesn't).  I originally assumed it was a UI related limitation (hence sending `updatedTheme` from the main thread as well) but now I'm not really sure.  Thoughts?

fedemkr: I think the problem is that the message is sent before the registration is done. And when you do it on the main thread it'll wait until the main thread finishes with the current work to send the message and that's why is working in that case.
I'm not quite sure if it'll work on all devices but even though I don't like much this solution the other one is to change the messenger/broadcast to have a special send method that queues the message if there are no subscribers at the moment for a specific id or something like that. So when the subscriber with that Id subscribes, the queue is checked and the message is sent accordingly in time.
what do you think?

mp-bw: Sounds useful, though it probably deserves it's own PR since it would require refactoring other messages sent in the same manner.

fedemkr: Agree 😄 approving this then

fedemkr (approved):

PR #2404, opened 2023-03-06
[EC-981] OTP handling - Set to selected cipher

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add the possibility to choose a cipher to set the scanned OTP handled by the Bitwarden app.

### AutofillCiphersPage -> CipherSelectionPage

Given that selecting a cipher on `Autofill` and selecting it on `OTP handling` is pretty much the same, the `AutofillCiphersPage` was taken and renamed to `CipherSelectionPage` to have a common UI to share between these flows. The context changes depending on the `AppOptions` passed to the page creating the appropriate `ViewModel` for each flow.

A diagram can illustrate this better:

```mermaid
classDiagram
    class CipherSelectionPageViewModel
    <<Abstract>> CipherSelectionPageViewModel
    CipherSelectionPageViewModel <|-- AutofillCiphersPageViewModel
    CipherSelectionPageViewModel <|-- OTPCipherSelectionPageViewModel
    CipherSelectionPage --> CipherSelectionPageViewModel
```

### Other changes

* **AccountsManager:** Added method to start the navigation flow (as in changing an account) that allows the caller to perform any action to the current `AppOptions`
* **DeepLinkContext:** Added method to handle new incoming Uris
* **App.xaml.cs:** Added otp handling logic and updated some hardcoded message commands values with consts
* **MainActivity:** Updated hardcoded message commands with consts
* **AppOptions:** Added `OtpData` to get the data from the otp uri
* **CipherSelectionPage.xaml:** Added close toolbar item and updated obsolete properties. Also, changed selection to new approach and added callout for the OTP cipher selection.
* **CipherSelectionPage.xaml.cs:** Added logic to choose between the `Autofill` and `OTP` viewmodels depending on the `AppOptions`. Also, added logic for `Close`. Something to highlight, is that the `Account Switching` toolbar item was left with the default `..` icon on iOS because there is an issue and it's not being updated correctly (it just shows a white backgrounded circle). So, for now I've just left it with the default one which IMO is better than the white circle.
* **CipherSelectionPageViewModel.cs:** Moved common things to this abstract class and added template methods to be implemented by the children
* **AutofillCiphersPageViewModel:** Add inheritance from `CipherSelectionPageVIewModel` and removed things that are now in the abstract class. Additionally, the select flow was changed to accept an `IGroupingsPageListItem` to be more generic and flexible.
* **OTPCipherSelectionPageVIewModel:** Implemented for OTP handling given the base `CipherSelectionPageVIewModel` class. The suggested ciphers to show on loading are the ones which include the `Issuer` of the OTP if exists or the `AccountName` otherwise.
* **CipherAddEditPageViewModel:** Added automatic set of OTP key when coming from OTP flow, the tooltip and the appropriate navigation when saving.
* **CiphersPage.xaml.cs:** Moved some initialization to the VM using the `Prepare(...)` method to have less logic on the view.
* **CiphersPageViewModel:** Updated `CipherOptionsCommand` to use the new async way. Added logic to show all ciphers if there's no search term and it comes from the OTP flow. Also, on selection it goes to the edit page of the cipher when coming from the OTP flow.
* **AppHelpers:** Added navigation logic for OTP flow
* **NavigationTarget:** Added new target for OTP flow
* **OTPData:** Added this struct which is the responsible to get all the attributes/parameters from an OTP Uri to be reused throughout the app.
* **TOTPService:** Updated it to use the new `OTPData` struct.
* **UriExtensions:** Added method to unescape a URI string like a OTP URI (it needs to be done multiple times to correctly unescape everything)

### Cipher Selection
Cipher selection OTP

### Cipher Search
Cipher search OTP

### OTP set on Cipher edition
Cipher edition OTP set

## Before you submit


src/App/Pages/Vault/CipherSelectionPage.xaml.cs
```diff
@@ -51,10 +68,16 @@ protected async override void OnAppearing()
                 return;
             }
 
-            _accountAvatar?.OnAppearing();
-            _vm.AvatarImageSource = await GetAvatarImageSourceAsync();
+            // TODO: There's currently an issue on iOS where the toolbar item is not getting updated
+            // as the others somehow. Removing this so at least we get the circle with ".." instead
+            // of a white circle
+            if (Device.RuntimePlatform != Device.iOS)
```
mp-bw, line 74: Is it possible the change of `Priority` of the avatar from `-1` to `-2` broke the original colorization trick you did for iOS?  (I don't see any hardcoded positions but I may be missing something)

fedemkr: It's possible and I think I didn't try that but I've created another task to track this issue so that I make progress with the whole feature. On this PR I won't be fixing this given that's something minor and can be fixed later on.

mp-bw (approved):

PR #2410, opened 2023-03-09
[AC-1070] Enforce master password policy on login/unlock

shane-melton:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
Implement new functionality to evaluate a master password on login if the new option to enforce on login is enabled. If enabled, whenever a user logs in, their master password will be evaluated against the organization's password requirements. If it fails the requirements, the user will be redirected to the update password page and forced to update to a new master password that meets the requirements. The same requirements are also checked during an a vault unlock with master password to ensure existing SSO members are also compliant with the organization's policy.

## Related PRs
Required server PR can be found here: https://github.com/bitwarden/server/pull/2714
Clients repo PR with similar functionality for other clients: https://github.com/bitwarden/clients/pull/4795

### Forcing Password Reset
* **src/App/App.xaml.cs:** Listen for new `forceUpdatePassword` command to navigate to the update password page. Similar to the `convertAccountToKeyConnector` command.

* **src/App/Pages/TabsPage.cs:** On page load, check if there is a `forceResetPasswordReason` saved to the account's state. If so, fire the `forceUpdatePassword` command to navigate to the update password page. This is to prevent users from skipping the update password page by restarting the app.

### Model Changes

* **src/Core/Models/Domain/MasterPasswordPolicyOptions.cs:** Add the new `EnforceOnLogin` flag that is now available on the master password policy data.

* **src/Core/Models/Response/IdentityTokenResponse.cs:** Add `MasterPasswordPolicy` field that is now returned in the response.

* **src/Core/Models/Response/IdentityTwoFactorResponse.cs:** Same as above.

* **src/Core/Models/Response/VerifyMasterPasswordResponse.cs:** New response model that is returned when a user verifies their master password with the server that includes the master password policy for that user (during a vault unlock).

### Service Changes

* **src/Core/Abstractions/IStateService.cs:** Add methods to get/set an accounts `forcePasswordResetReason`. The auth service will set the value after successful authentication and the password is "weak". It can then be fetched in the `TabsPage.cs` to check if the user should be forced to update their password.

* **src/Core/Services/AuthService.cs:** Bulk of the logic change. During password login, the master password is evaluated using the master password policy returned by the Identity response. If the password does not meet requirements, save the reason to the account state to force a password update. Additional logic is in place to support 2FA login flow.

* **src/Core/Abstractions/IApiService.cs:** Add method to update the user's password

### Page Changes

* **src/App/Pages/Accounts/UpdateTempPasswordPage.xaml:** Update the warning text to depend on the reason for the password reset. Conditionally show a new "Current master password" field if updating a weak master password (not required for admin password resets)

* **src/App/Pages/Accounts/UpdateTempPasswordPageViewModel.cs:** Add logic to support updating both admin reset passwords and weak master passwords depending on the reason the user is on the page.

## Screenshots
image

## Before you submit


src/App/Resources/AppResources.resx
```diff
@@ -2592,4 +2592,10 @@ Do you want to switch to this account?</value>
   <data name="OrganizationSsoIdentifierRequired" xml:space="preserve">
     <value>Organization SSO identifier required.</value>
   </data>
+  <data name="UpdateWeakMasterPasswordWarning" xml:space="preserve">
+    <value>Your master password does not meet one or more of your organization policies. In order to access the vault, you must update your master password now. Proceeding will log you out of your current session, requiring you to log back in. Active sessions on other devices may continue to remain active for up to one hour.</value>
+  </data>
+  <data name="CurrentMasterPass" xml:space="preserve">
+    <value>Current master password</value>
+  </data>
```
fedemkr, line 2600: Key and Value should be the same when the value's length is short => `CurrentMasterPassword` so it's easier to "guess" the exact value of the resource just by looking at the key

src/Core/Services/AuthService.cs
```diff
@@ -44,7 +49,7 @@ public class AuthService : IAuthService
             IVaultTimeoutService vaultTimeoutService,
             IKeyConnectorService keyConnectorService,
             IPasswordGenerationService passwordGenerationService,
-            bool setCryptoKeys = true)
+            IPolicyService policyService, bool setCryptoKeys = true)
```
fedemkr, line 52: new line for each parameter

src/Core/Services/AuthService.cs
```diff
@@ -136,11 +142,51 @@ public void Init()
...
+                return false;
+            }
+
+            var passwordStrength = _passwordGenerationService.PasswordStrength(
+                masterPassword,
+                _passwordGenerationService.GetPasswordStrengthUserInput(email)
+            ).Score;
+
+            return !await _policyService.EvaluateMasterPassword(passwordStrength, masterPassword, _masterPasswordPolicy);
         }
```
fedemkr, line 190: I see this method being also used in `LockPageViewModel`. Could we move these to a service so that the logic is shared between the callers? I think the `PolicyService` is a good candidate to have this method.
Btw don't forget to add the `Async` suffix

src/Core/Services/AuthService.cs
```diff
@@ -157,15 +203,26 @@ public async Task<AuthResult> LogInSsoAsync(string code, string codeVerifier, st
...
+            var result = await LogInHelperAsync(Email, MasterPasswordHash, LocalMasterPasswordHash, Code, CodeVerifier, SsoRedirectUrl, _key,
                 twoFactorProvider, twoFactorToken, remember, CaptchaToken, authRequestId: AuthRequestId);
+
+            // If we successfully authenticated and we have a saved _2faForcePasswordResetReason reason from LogInAsync()
+            if (!string.IsNullOrEmpty(_authedUserId) && _2faForcePasswordResetReason.HasValue)
+            {
+                // Save the forcePasswordReset reason with the state service to force a password reset for the user
+                result.ForcePasswordReset = true;
+                await _stateService.SetForcePasswordResetReasonAsync(
+                    ForcePasswordResetReason.WeakMasterPasswordOnLogin, _authedUserId);
```
fedemkr, line 222: Even though right now is the same value, shouldn't this be `_2faForcePasswordResetReason.Value` instead of the enum value directly?

src/Core/Services/PolicyService.cs
```diff
@@ -130,6 +130,12 @@ public async Task ClearAsync(string userId)
                 {
                     enforcedOptions.RequireSpecial = true;
                 }
+
+                var enforceOnLogin = GetPolicyBool(currentPolicy, "enforceOnLogin");
+                if (enforceOnLogin != null && (bool)enforceOnLogin)
```
fedemkr, line 135: This can be simplified as: `if (enforceOnLogin == true)`

fedemkr (changes requested):
Great work on this Shane, just a few comments. Also does this also work on extensions, like Autofill/Share?

src/App/Pages/Accounts/UpdateTempPasswordPageViewModel.cs
```diff
@@ -1,27 +1,64 @@
...
+            _userVerificationService = ServiceContainer.Resolve<IUserVerificationService>("userVerificationService");
         }
 
         public Command SubmitCommand { get; }
         public Command TogglePasswordCommand { get; }
         public Command ToggleConfirmPasswordCommand { get; }
         public Action UpdateTempPasswordSuccessAction { get; set; }
         public Action LogOutAction { get; set; }
+        public string CurrentMasterPassword { get; set; }
+        public ForcePasswordResetReason Reason { get; set; } = ForcePasswordResetReason.AdminForcePasswordReset;
```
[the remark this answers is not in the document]
shane-melton: Ah, good to know, thanks for the extra context as well!

src/Core/Services/AuthService.cs
shane-melton: Good catch!

shane-melton:
@fedemkr Thanks for the review! I believe I have addressed all of your individual comments.

> Also does this also work on extensions, like Autofill/Share?

I believe this should only apply during log in to main mobile application and not necessarily the extensions on their own, but I'll double check with product.

I'll also reach out to them about condensing/collapsing the warning messages in some way so they don't hide fields on smaller devices.

src/Core/Services/AuthService.cs
shane-melton: I attempted to move this to the `PolicyService` but ended up with a circular dependency with the `PasswordGenerationSerivce` - Once I started extracting the password strength portion of the method it essentially became the inverse of `EvaluateMasterPassword()` so I think it may actually be simpler to leave it as is.

I did still add the `Async` suffix.

shane-melton: Sounds good, thank you!

fedemkr (dismissed):

mp-bw (approved):

PR #2417, opened 2023-03-16
[PM-1431] [Defect] [Android] New accounts are not able to log in

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Do not clear Master password  if login is ongoing.

## Code changes

Added control variable to check if the app is currently performing a login.

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPageViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPageViewModel.cs
@@ -40,6 +40,7 @@ namespace Bit.App.Pages
         private string _masterPassword;
         private bool _isEmailEnabled;
         private bool _isKnownDevice;
+        private bool _isExecutingLogin;
 
         public LoginPageViewModel()
         {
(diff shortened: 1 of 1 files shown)


trmartin4 (dismissed):
Good catch!

mp-bw:
At a glance it looks like this could potentially break #2395 under some conditions.  Any thoughts?

fedemkr:
From what I can see, it shouldn't break the other behavior. Given that this is only stopping the fields to be cleared while the logging process is in progress. Also the #2395 goal was to clear them when the app goes to/goes back from background and if the login is in process and the app goes to background then when it goes back it shouldn't update which IMO should be fine.

The only flow (kind of an edge case) I can think of it could be up to discussion is the next one:

1. The user enters credentials and start logging in
2. While in the process of login the user moves the app to background
3. Wait here until logging finishes
4. The user goes back and the app says that the login failed because something happened (service unavailability, wrong credentials, or something else)

In this case, should the fields be cleared? 🤔 

_Note: Just in case if I'm not mistaken, with this change the fields won't be cleared on this scenario._

PR #2421, opened 2023-03-20
[PM-115] Cipher key encryption update

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

The cipher will now have its own `Key` to encrypt/decrypt its content which will be encrypted using the master password or organization key.

This needs server changes to work and clients changes to have the key assigned (given that for now the mobile client won't create the cipher key):
- Server: [https://github.com/bitwarden/server/pull/2681](https://github.com/bitwarden/server/pull/2681
)
- Clients: [https://github.com/bitwarden/clients/pull/5114](https://github.com/bitwarden/clients/pull/5114)

## Code changes

* **Cipher:** Added new properties for the key. Added logic to decrypt using the new `Key` if available, passing such key to the different item types. And updated it to not use hardcoded strings and use `nameof()`.
* **CipherService:** Updated logic to encrypt the `Cipher` using the new `Key`. 
_Note: for now the logic of the new `Key` gets applied only if the `Key` already exists on the `Cipher`. The creation of such `Key` locally will be enabled on a later release to avoid conflicts between clients._

* **CipherData:** Added new properties for the key.
* **CipherRequest, CipherResponse, CipherView:** Added new properties for the key.
* **Attachment:** Updated it to not use hardcoded strings and use `nameof()` and also added the `key` as a parameter to support the new approach.
* **Card, Field, Identity, Login, LoginUri, PasswordHistory, SecureNote:** Added the `key` as a parameter to support the new approach.
 
## Before you submit


mp-bw (approved):
How is performance when flinging through the cipher list?

fedemkr:
As far as I see, the entire cipher list gets decrypted before shown to the user so the performance while flinging is the same as before. But yes, we should have a little bit more overhead while decrypting the whole vault ciphers.

mp-bw:
Ah ok, I was thinking this could provide some performance insight to adapter-driven content if/when we take that step.

mpbw2 (approved):

mpbw2 (dismissed):

mpbw2 (dismissed):

mpbw2 (dismissed):

mpbw2 (approved):

PR #2431, opened 2023-03-23
[PM-1351][PM-190] Add a mobile service to retrieve feature flags from API

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add a new service to consume the api endpoint `/config/`. This will be used to get feature flags to later be used by the UI.

## Code changes

Added new ConfigService, models and new method in ApiService to fetch the data.

## Screenshots

image

## Before you submit


jlf0dev (dismissed):
Had a question above but thats non-blocking, looks good!

fedemkr:
Shouldn't the config be persisted locally every time you call the API? Right now it just acts as an in memory config, so if you open the app without connection you won't have the configuration available.

andrebispo5:
> Shouldn't the config be persisted locally every time you call the API? Right now it just acts as an in memory config, so if you open the app without connection you won't have the configuration available.

You are right i was only think that it should update when the user opens the app and forgot about the offline possibility, i'll add it to state.

src/Core/Services/ConfigService.cs
```diff
@@ -11,21 +11,31 @@ namespace Bit.Core.Services
...
+                {
+                    _configs = await _apiService.GetAllConfigsAsync();
+                    _configs.ExpiresOn = DateTime.UtcNow.AddMinutes(UPDATE_INTERVAL_MINS);
+                    _stateService.SetConfigs(_configs);
+                }
+            }
+            catch (Exception)
+            {
+                //ignore and return state value or null
             }
```
fedemkr: Unless it is because of connection problem, the exception should be logged (or rethrown to the caller if necessary) so we know there is a problem with that.
Also, the check can be written shorter like:
```c#
if (_configs?.ExpiresOn is null || _configs.ExpiresOn <= DateTime.UtcNow)
```

src/Core/Services/ConfigService.cs
```diff
@@ -0,0 +1,45 @@
...
+using Bit.Core.Models.Domain;
+using Bit.Core.Models.Response;
+using Bit.Core.Models.View;
+
+namespace Bit.Core.Services
+{
+    public class ConfigService : IConfigService
+    {
+        private ConfigResponse _configs;
+        private const int UPDATE_INTERVAL_MINS = 60;
```
fedemkr, line 14: Constants should be moved to the beginning of the class

src/Core/Abstractions/IStateService.cs
```diff
@@ -168,5 +168,7 @@ public interface IStateService
         Task SetPreLoginEmailAsync(string value);
         string GetLocale();
         void SetLocale(string locale);
+        ConfigResponse GetConfigs();
+        void SetConfigs(ConfigResponse value);
```
fedemkr, line 172: I think there is a naming mismatch, here `GetConfigs()` is used but on the `IApiService` it's named `GetAllConfigs...()`. IMO both should be with or without the `...All...` in between to be consistent.

src/Core/Services/StateService.cs
```diff
@@ -1234,6 +1234,17 @@ public async Task SetPreLoginEmailAsync(string value)
...
+
+        public ConfigResponse GetConfigs()
+        {
+            return _storageMediatorService.Get<ConfigResponse>(Constants.ConfigsKey);
+        }
+
+        public void SetConfigs(ConfigResponse value)
+        {
+            _storageMediatorService.Save(Constants.ConfigsKey, value);
+        }
```
fedemkr, line 1246: Are there any sensitive data on the config to be saved? These methods (synchronous version) read/write directly to the user preferences/user defaults of the platform.

fedemkr (changes requested):
Aside of the requested changes, is the call to this new service going to be added on another PR?

andrebispo5:
> Aside of the requested changes, is the call to this new service going to be added on another PR?

Yes this was split into many different task, the other PR is. https://github.com/bitwarden/mobile/pull/2454

src/Core/Abstractions/IStateService.cs
andrebispo5: Removed the "All" to be consistent with the endpoint name.

src/Core/Services/ConfigService.cs
andrebispo5: Refactored the condition and logged the error.

src/Core/Services/ConfigService.cs
```diff
@@ -10,32 +10,34 @@ namespace Bit.Core.Services
...
                     _configs.ExpiresOn = DateTime.UtcNow.AddMinutes(UPDATE_INTERVAL_MINS);
                     _stateService.SetConfigs(_configs);
                 }
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                //ignore and return state value or null
+                _logger.Exception(ex);
             }
```
fedemkr: Just to make sure, do you want to also log the exceptions produced by no connection? I think they are not necessary and will pollute AppCenter logs but lmk so we can filter them out if not needed.

fedemkr (dismissed):
LGTM, just one small comment on the exception.

src/Core/Services/ConfigService.cs
andrebispo5: Oh of course, i'll change it to not log those.

fedemkr (dismissed):

fedemkr (dismissed):

src/Core/Models/Response/ConfigResponse.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Bit.Core.Models.Response
+{
+    public class ConfigResponse
+    {
+        public string Version { get; set; }
+        public string GitHash { get; set; }
+        public ServerConfigResponse Server { get; set; }
+        public EnvironmentConfigResponse Environment { get; set; }
+        public IDictionary<string, object> FeatureStates { get; set; }
```
vvolkgang, line 12: Can we have a better default type than `object`?

src/Core/Services/ConfigService.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading.Tasks;
+using Bit.Core.Abstractions;
+using Bit.Core.Exceptions;
+using Bit.Core.Models.Domain;
+using Bit.Core.Models.Response;
+using Bit.Core.Models.View;
+
+namespace Bit.Core.Services
+{
+    public class ConfigService : IConfigService
```
vvolkgang, line 12: This service should follow `server`'s implementation of [feature flags](https://github.com/bitwarden/server/blob/master/src/Core/Services/IFeatureService.cs), exposing the different flag variant methods (`bool`, `string` and `int`)

src/Core/Models/Response/ConfigResponse.cs
andrebispo5: #2454 will be merged into this branch in order to be tested as a whole before it goes into master.

src/Core/Services/ConfigService.cs
andrebispo5: Since i've already extended the config service in #2454 i'll add the variants there to avoid merge conflicts.

andrebispo5: Updated #2454 with variants.

src/Core/Services/ApiService.cs
```diff
@@ -580,6 +580,15 @@ public Task<bool> GetKnownDeviceAsync(string email, string deviceIdentifier)
 
         #endregion
 
+        #region Configs
+
+        public async Task<ConfigResponse> GetConfigsAsync()
+        {
+            return await SendAsync<object, ConfigResponse>(HttpMethod.Get, $"/config/", null, false, true);
```
fedemkr, line 587: ⛏️ the `$` is not needed given that you're not passing any parameters to the string

src/Core/Services/ConfigService.cs
```diff
@@ -0,0 +1,72 @@
...
+        private readonly ILogger _logger;
+
+        public ConfigService(IApiService apiService, IStateService stateService, ILogger logger)
+        {
+            _apiService = apiService;
+            _stateService = stateService;
+            _logger = logger;
+        }
+
+        public async Task<ConfigResponse> GetAsync(bool forceRefresh = false)
```
fedemkr, line 28: ⚠️  `forceRefresh` is not being used in this method

fedemkr (changes requested):

src/Core/Models/Response/ConfigResponse.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Bit.Core.Models.Response
+{
+    public class ConfigResponse
+    {
+        public static string DisplayEuEnvironmentFlag = "display-eu-environment";
```
vvolkgang, line 8: @mp-bw any better place to store feature flags? Thinking `Constants.cs` or it's own static class.

mp-bw: Agreed `Constants.cs` would be the proper place for it to follow existing patterns

andrebispo5: https://github.com/bitwarden/mobile/pull/2454#discussion_r1152100129
@fedemkr suggested it was moved to the ConfigResponse.

mp-bw: @andrebispo5 @fedemkr Since it's `public` it seems like `Constants` would be the right place for it.  Can it be `private`?

andrebispo5: Nope it needs to be public as it is going to be used as the key to retrieve the flag from the dictionary. I'll move it to the Constants file.

mp-bw (approved):
Looks fine, not sure why iOS CI is failing

andrebispo5:
@mpbw2 this was already approved but was waiting for the other clients to be done to merge everything to master. In the meanwhile there are two new small commits that need review.

mpbw2 (approved):

mpbw2 (approved):

fedemkr (approved):

fedemkr (approved):

PR #2453, opened 2023-03-28
[PM-1646] Add thread safety to migration process

mp-bw:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This makes the storage migration process thread safe to cover for cases where migration is inadvertently called more than once during app initialization.

## Code changes

* **StateMigrationService.cs:** Combine checking for and performing migration into a single step behind a `SemaphoreSlim` to make sure migration will complete before executing a subsequent check.
* **StateService.cs:** Replace `NeedsMigration()` and `Migrate()` with `MigrateIfNeeded()`

## Before you submit

### DIFF
--- a/src/Core/Abstractions/IStateMigrationService.cs
+++ b/src/Core/Abstractions/IStateMigrationService.cs
@@ -4,7 +4,6 @@ namespace Bit.Core.Abstractions
 {
     public interface IStateMigrationService
     {
-        Task<bool> NeedsMigration();
-        Task Migrate();
+        Task MigrateIfNeeded();
     }
 }
--- a/src/Core/Abstractions/IStateMigrationService.cs
+++ b/src/Core/Abstractions/IStateMigrationService.cs
@@ -4,6 +4,6 @@ namespace Bit.Core.Abstractions
 {
     public interface IStateMigrationService
     {
-        Task MigrateIfNeeded();
+        Task MigrateIfNeededAsync();
     }
 }
--- a/src/Core/Services/StateMigrationService.cs
+++ b/src/Core/Services/StateMigrationService.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Bit.Core.Abstractions;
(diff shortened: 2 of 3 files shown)


src/Core/Abstractions/IStateMigrationService.cs
```diff
@@ -4,7 +4,6 @@ namespace Bit.Core.Abstractions
 {
     public interface IStateMigrationService
     {
-        Task<bool> NeedsMigration();
-        Task Migrate();
+        Task MigrateIfNeeded();
```
fedemkr, line 7: Since you're changing this, could you add the `Async` suffix?

src/Core/Services/StateMigrationService.cs
```diff
@@ -18,6 +19,8 @@ public class StateMigrationService : IStateMigrationService
         private readonly IStorageService _liteDbStorageService;
         private readonly IStorageService _secureStorageService;
 
+        private static SemaphoreSlim _semaphore;
```
fedemkr, line 22: This service Is registered and used as a Singleton. Do you need the `static` keyword because of multiple threads accessing it or something like that? Because I think it shouldn't be necessary.
Btw, if it needs to be `static` then you can directly initialize it inline here:
```c#
private static SemaphoreSlim _semaphore = new SemaphoreSlim(1);
```

src/Core/Services/StateService.cs
```diff
@@ -1538,10 +1538,7 @@ private async Task CheckStateAsync()
             if (!_migrationChecked)
             {
                 var migrationService = ServiceContainer.Resolve<IStateMigrationService>("stateMigrationService");
```
fedemkr, line 1540: Since we're changing stuff, could you remove the string key of the service Resolve here?

fedemkr (changes requested):

src/Core/Services/StateMigrationService.cs
mp-bw: You are correct, I overlooked the singleton nature of the service.

fedemkr (approved):
Looks good to me! 😄

PR #2454, opened 2023-03-29
[PM-190] Add Environments to mobile login/registration

andrebispo5:
Merged after https://github.com/bitwarden/mobile/pull/2431
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Provide the user the ability to change the environment into which he is logging in or creating an account. 
The selected environment should persist when a user logs out.
EU environment shows based on feature flag.
Environment settings cog, on top right corner of the `HomePage` was moved to the environment picker as self-host.

## Code changes

* **src/App/Pages/Accounts/HomePage.xaml.cs:** `UpdateEnvironment` added in `OnAppearing` is used to update the environment selector label when the user comes from the self-hosted screen back to the `HomePage`

## Screenshots

Screenshot 2023-03-28 at 15 40 57
Screenshot 2023-03-28 at 17 01 35
Screenshot 2023-03-28 at 15 53 23

## Before you submit


src/Core/Services/EnvironmentService.cs
```diff
@@ -125,5 +128,17 @@ private string FormatUrl(string url)
...
+        {
+            return await SetUrlsAsync(new EnvironmentUrlData()
+            { Base = DefaultBaseUsRegion });
+        }
+
+        public async Task<EnvironmentUrlData> SetEuUrlsAsync()
+        {
+            return await SetUrlsAsync(new EnvironmentUrlData()
+            { Base = DefaultBaseEuRegion });
+        }
```
fedemkr, line 142: 🤔 🎨  Not sure how likely it would be to add more environments but this could be just one method with an enum or create some value objects that can be passed, or my favorite create static `EnvironmentUrlData` like this and remove both methods and use `SetUrlsAsync` directly :
```c#
public class EnvironmentUrlData
{
    public static EnvironmentUrlData DefaultUS = new EnvironmentUrlData { Base = "https://vault.bitwarden.com" };
    public static EnvironmentUrlData DefaultEU = new EnvironmentUrlData { Base = "https://vault.bitwarden.eu" }

    public string Base { get; set; }
    ...
}

// and this can be used like:

await SetUrlsAsync(EnvironmentUrlData.DefaultUS);
await SetUrlsAsync(EnvironmentUrlData.DefaultEU);
```

src/Core/Constants.cs
```diff
@@ -41,6 +41,8 @@ public static class Constants
         public const string NotificationDataType = "Type";
         public const string PasswordlessLoginRequestKey = "passwordlessLoginRequest";
         public const string PreLoginEmailKey = "preLoginEmailKey";
+        public const string ConfigsKey = "configsKey";
+        public const string DisplayEuEnvironmentFlag = "display-eu-environment";
```
fedemkr, line 45: ⛏️ IMO this can be moved to the `ConfigResponse` given that this flag is specific to the dictionary inside. So when more feature flags are added we have all of the keys in the same place.

fedemkr (changes requested):

src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -125,6 +141,10 @@ public async Task ContinueToLoginStepAsync()
                         AppResources.Ok);
                     return;
                 }
+                if (await _stateService.GetPreAuthEnvironmentUrlsAsync() == null)
+                {
+                    await _environmentService.SetUsUrlsAsync();
+                }
```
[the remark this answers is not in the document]
andrebispo5: Not seeing a good way to do it that will work for all the existing cases.

src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -165,39 +163,47 @@ await _stateService.IsAuthenticatedAsync(userId))
...
+            });
         }
 
         public async Task UpdateEnvironment()
         {
-            var baseUrlSaved = (await _stateService.GetPreAuthEnvironmentUrlsAsync())?.Base;
-            if (baseUrlSaved == _environmentService.DefaultBaseUsRegion || string.IsNullOrEmpty(baseUrlSaved))
+            var environmentsSaved = await _stateService.GetPreAuthEnvironmentUrlsAsync();
+            if (environmentsSaved == null || environmentsSaved.Equals(new EnvironmentUrlData()))
```
fedemkr: ⚠️ I see you override `Equals(...)` to compare if the saved one is empty basically. I wouldn't do that for this requirement, I think it's a little too much, and you're allocating a new object just to compare this (also when you override `Equals(...)` you should also override `GetHashCode(...)` or a hard to look for bug can happen if the object is used in lists)
I'd remove the `Equals(...)` and would do it simpler by just adding a readonly property in `EnvironmentUrlData` -> `IsEmpty`:
```c#
public bool IsEmpty => string.IsNullOrWhiteSpace(Base)
    && string.IsNullOrWhiteSpace(Api)
    && string.IsNullOrWhiteSpace(Identity)
    && string.IsNullOrWhiteSpace(Icons)
    && string.IsNullOrWhiteSpace(Notifications)
    && string.IsNullOrWhiteSpace(WebVault)
    && string.IsNullOrWhiteSpace(Events);
```

❓ Which would be the case the `EnvironmentUrlData` is not `null` but empty? Is that when the user filled the env data and then edited it and clear it out?

src/Core/Models/Data/EnvironmentUrlData.cs
```diff
@@ -2,12 +2,36 @@
...
+                    && env.Api == this.Api
+                    && env.Identity == this.Identity
+                    && env.Icons == this.Icons
+                    && env.Notifications == this.Notifications
+                    && env.WebVault == this.WebVault
+                    && env.Events == this.Events;
+            }
+
+            return base.Equals(obj);
+        }
```
fedemkr: ⚠️ Remove per comment where it's used

src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -25,11 +27,12 @@ public class HomeViewModel : BaseViewModel
...
-        private ILogger _logger;
-        private IEnvironmentService _environmentService;
-        private IAccountsManager _accountManager;
-        private IConfigService _configService;
+        private bool _displayEuEnvironment;
+        private readonly IPlatformUtilsService _platformUtilsService;
+        private readonly ILogger _logger;
+        private readonly IEnvironmentService _environmentService;
+        private readonly IAccountsManager _accountManager;
+        private readonly IConfigService _configService;
```
fedemkr, line 35: ⛏️  Since you're here could you move the services to the top with the other services so we maintain the format as in the other files?

fedemkr (changes requested):

src/App/Pages/Accounts/HomePageViewModel.cs
```diff
@@ -144,5 +162,56 @@ await _stateService.IsAuthenticatedAsync(userId))
...
+                    await _environmentService.SetUrlsAsync(result == AppResources.EU ? EnvironmentUrlData.DefaultEU : EnvironmentUrlData.DefaultUS);
+                    SelectedEnvironmentName = result;
+                }
+            });
+        }
+
+        public async Task UpdateEnvironment()
+        {
+            var environmentsSaved = await _stateService.GetPreAuthEnvironmentUrlsAsync();
+            if (environmentsSaved == null || environmentsSaved.IsEmpty())
```
fedemkr, line 196: ❓ Which would be the case the `EnvironmentUrlData` is not `null` but empty? Is that when the user filled the env data and then edited it and clear it out?

src/Core/Models/Data/EnvironmentUrlData.cs
```diff
@@ -2,12 +2,23 @@
...
+
         public string Base { get; set; }
         public string Api { get; set; }
         public string Identity { get; set; }
         public string Icons { get; set; }
         public string Notifications { get; set; }
         public string WebVault { get; set; }
         public string Events { get; set; }
+
+        public bool IsEmpty() => string.IsNullOrEmpty(this.Base)
```
fedemkr, line 16: ⛏️  This can be a readonly property to avoid the parenthesis on the calls.

src/Core/Models/Data/EnvironmentUrlData.cs
```diff
@@ -2,12 +2,23 @@
...
+                && string.IsNullOrEmpty(this.Api)
+                && string.IsNullOrEmpty(this.Identity)
+                && string.IsNullOrEmpty(this.Icons)
+                && string.IsNullOrEmpty(this.Notifications)
+                && string.IsNullOrEmpty(this.WebVault)
+                && string.IsNullOrEmpty(this.Events);
```
fedemkr, line 22: ⛏️  Remove the `this`. They are not necessary and we don't use it unless strictly necessary.

fedemkr (changes requested):

src/App/Pages/Accounts/HomePageViewModel.cs
andrebispo5: yes, when the user saves the env page empty.

src/App/Pages/Accounts/HomePage.xaml.cs
```diff
@@ -70,7 +72,14 @@ protected override async void OnAppearing()
...
             });
-            await _vm.UpdateEnvironment();
+            try
+            {
+                await _vm.UpdateEnvironment();
+            }
+            catch (Exception ex)
+            {
+                _logger.Value?.Exception(ex);
+            }
```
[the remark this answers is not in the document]
andrebispo5: It will default to the US servers in the ApiService, but it is very unlikely that it will throw an exception at this stage.

src/Core/Services/ConfigService.cs
```diff
@@ -49,14 +49,43 @@ public async Task<ConfigResponse> GetAsync(bool forceRefresh = false)
...
+        public async Task<int> GetFeatureFlagIntAsync(string key, bool forceRefresh = false, int defaultValue = 0)
+        {
+            await GetAsync(forceRefresh);
+            if (_configs?.FeatureStates?.TryGetValue(key, out var val) == true
                     &&
-                    boolVal;
+                    val is int intVal)
+            {
+                return intVal;
+            }
```
fedemkr: 🎨  Instead of having different methods for each type, you can have just one and use generics:
```c#
public async Task<T> GetFeatureFlagAsync<T>(string key, bool forceRefresh = false, T defaultValue = default)
{
    await GetAsync(forceRefresh);
    if (_configs?.FeatureStates?.TryGetValue(key, out var val) == true
            &&
            val is T actualValue)
    {
        return actualValue;
    }

    return defaultValue;
}
```

and call it like:
```c#
_displayEuEnvironment = await _configService.GetFeatureFlagAsync<bool>(ConfigResponse.DisplayEuEnvironmentFlag, forceRefresh: true);
```

fedemkr (changes requested):

src/Core/Services/ConfigService.cs
andrebispo5: I was doing it like that but i changed it on purpose to lock it to the 3 types available server side, and for the implementation to be similar.

fedemkr: Alright I think we can make the generic method private and leave the three public calls restricted, so we expose the restricted version but we have the generic one to avoid repeating code

andrebispo5: sounds good! 👍

fedemkr (approved):

PR #2517, opened 2023-05-10
[PM-2219] Dependency Updates

mp-bw:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Dependency updates galore.

**_Note 1:_** Some Android libs (`AndroidX.AppCompat`, `AndroidX.MediaRouter`, and `Google.Material.Android`) could not be updated to their latest versions due to being compiled at a higher build target than the current public `Xamarin.Forms` (fixed in an upcoming XF release - I tested against `5.0.0.8491` and everything seemed happy).

**_Note 2:_** Updating `MessagePack*` breaks the Android build with the following error:

```
Xamarin.Android.Legacy.targets(317, 5): [XA2002] Can not resolve reference: `Microsoft.NET.StringTools`, referenced by `MessagePack`. Please add a NuGet package or assembly reference for `Microsoft.NET.StringTools`, or remove the reference to `MessagePack`.
```
I'm not sure of the correct path forward here so I'll need input from @fedemkr & @vvolkgang . Note that iOS continues to build fine with this update so /shrug

**_Note 3:_** Saving `Google.Apis.AndroidPublisher.v3` updates for another day to prevent breaking the Github publishing automation (it's a big version jump so the likelihood of it bursting into flames is high)

## Code changes

* ***.csproj:** Dependency version bumps
* **info.plist:** Change iOS minimum supported version to 11 to support MS AppCenter 5.x
* **SendRequestTests.cs:** Declare param as long to support fixed strict input in `xunit 2.4.2`

## Before you submit

### DIFF
--- a/src/Android/Android.csproj
+++ b/src/Android/Android.csproj
@@ -78,21 +78,20 @@
       <Version>1.9.0</Version>
     </PackageReference>
     <PackageReference Include="Xamarin.AndroidX.AppCompat" Version="1.5.1.1" />
-    <PackageReference Include="Xamarin.AndroidX.AutoFill" Version="1.1.0.14" />
-    <PackageReference Include="Xamarin.AndroidX.CardView" Version="1.0.0.17" />
-    <PackageReference Include="Xamarin.AndroidX.Core" Version="1.9.0.1" />
-    <PackageReference Include="Xamarin.AndroidX.Legacy.Support.V4" Version="1.0.0.15" />
+    <PackageReference Include="Xamarin.AndroidX.AutoFill" Version="1.1.0.16" />
+    <PackageReference Include="Xamarin.AndroidX.CardView" Version="1.0.0.19" />
+    <PackageReference Include="Xamarin.AndroidX.Core" Version="1.10.0" />
     <PackageReference Include="Xamarin.AndroidX.MediaRouter" Version="1.3.1.1" />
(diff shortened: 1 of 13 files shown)


mp-bw:
> @mp-bw which version of `MessagePack` did you updated to? From what I'm seeing in the library repo they [fixed](https://github.com/neuecc/MessagePack-CSharp/issues/1509) the issue making the reference private compiled. Maybe, it doesn't work in Xamarin world... 🤔 but v2.4.66 should have the issue fixed.

`2.5.108` (latest public nuget). Maybe they accidentally re-introduced it?

fedemkr:
I'd try updating specifically to `2.4.66` to see if that fixes the issue. If so, then we know it was re-introduced after that hehe, otherwise we can submit an issue to the library/dotnet team and see if we have any response.

mp-bw:
Doesn't look like they released a stable `2.4.66`, and everything after `2.4.59` gives the same result.  I think leaving it at `2.4.59` is our only option for now.

image

fedemkr:
Agree and I've left a comment in the library's issue to see if we can get more info about it.

fedemkr:
♻️ Also we could in another PR isolate all the usage of `MessagePack` to iOS, so then we can update without having to worry about Android.

PR #2535, opened 2023-05-23
[PM-1379] add DeviceTrustCryptoService with establish trust logic

jlf0dev:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds the Establish Trust logic to mobile. More information and clients equivalent in bitwarden/clients#5339

## Code changes

- **DeviceTrustCryptoService:** Will be responsible for handling the device specific key related methods, such as the trusted device ones in there.

## Before you submit


src/Core/Services/ApiService.cs
```diff
@@ -585,6 +585,16 @@ public Task<bool> GetKnownDeviceAsync(string email, string deviceIdentifier)
...
         }
 
+        public Task<DeviceResponse> GetDeviceByIdentifierAsync(string deviceIdentifier)
+        {
+            return SendAsync<object, DeviceResponse>(HttpMethod.Get, $"/devices/identifier/{deviceIdentifier}", null, true, true);
+        }
+
+        public Task<DeviceResponse> UpdateTrustedDeviceKeysAsync(string deviceIdentifier, TrustedDeviceKeysRequest trustedDeviceKeysRequest)
+        {
+            return SendAsync<TrustedDeviceKeysRequest, DeviceResponse>(HttpMethod.Get, $"/devices/{deviceIdentifier}/keys", trustedDeviceKeysRequest, true, true);
```
andrebispo5, line 595: `HttpMethod.Get` should be `POST` or `PUT`

jlf0dev: Great catch! Updated

andrebispo5 (approved):
LGTM 👍

src/Core/Services/DeviceCryptoService.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            // Send encrypted keys to server
+            var deviceIdentifier = await _appIdService.GetAppIdAsync();
+            var deviceRequest = new TrustedDeviceKeysRequest()
+            {
+                EncryptedUserKey = (await encryptUserKeyTask).EncryptedString,
+                EncryptedPublicKey = (await encryptPublicKeyTask).EncryptedString,
+                EncryptedPrivateKey = (await encryptPrivateKeyTask).EncryptedString,
+            };
+            return await _apiService.UpdateTrustedDeviceKeysAsync(deviceIdentifier, deviceRequest);
```
fedemkr, line 58: ⛏️ Remove the use of `this` (I thought this was caught by the formatting tool)

⛏️ IMO it's easier to read and understand what part is by deconstructing the tuple, so instead of `deviceKeyPair` you can have `var (devicePublicKey, devicePrivateKey)` and you can remove the comment given that it's more readable this way. And on the usage you can directly use `devicePublicKey` instead of `deviceKeyPair.Item1`.

🤔  Is there a reason to defer the encrypt task executions? Also you could run the three tasks in parallel using `Task.WhenAll(...)` and it gives back the results of all the operations. On the other hand, you can directly `await` on the spot, so instead of creating task variables you can have the awaited result there to be use afterwards in the device request.

⛏️ You can remove the parenthesis `()` from the `new TrustedDeviceKeyRequest` if you want as well.

src/Core/Services/DeviceCryptoService.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            _apiService = apiService;
+            _appIdService = appIdService;
+            _cryptoFunctionService = cryptoFunctionService;
+            _cryptoService = cryptoService;
+            _stateService = stateService;
+        }
+
+
+        public async Task<DeviceResponse> TrustDeviceAsync()
```
fedemkr, line 32: ❓ I'm not sure where this will be used, but just wanted to ask: Does this need any special exceptions handling or is it ok to just pass it through the caller as it's now?

fedemkr (changes requested):

src/Core/Services/DeviceCryptoService.cs
```diff
@@ -0,0 +1,85 @@
...
+
+
+        public async Task<SymmetricCryptoKey> GetDeviceKeyAsync()
+        {
+            // Check if device key already exists
+            var deviceKey = await this._stateService.GetDeviceKeyAsync();
+
+            return deviceKey != null
+                ? deviceKey
+                : await this.MakeDeviceKeyAsync();
```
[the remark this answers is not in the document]
jlf0dev: Done

src/Core/Services/DeviceCryptoService.cs
jlf0dev: I was attempting to execute the tasks while we do the rest of the work, but its not worth it. Removed.

src/Core/Services/DeviceCryptoService.cs
jlf0dev: Not yet, this is a helper PR to unblock other work. We'll iterate further if needed.

src/Core/Services/DeviceCryptoService.cs
```diff
@@ -0,0 +1,85 @@
...
+        private async Task<SymmetricCryptoKey> MakeDeviceKeyAsync()
+        {
+            // Create 512-bit device key
+            var randomBytes = await _cryptoFunctionService.RandomBytesAsync(64);
+            var deviceKey = new SymmetricCryptoKey(randomBytes);
+
+            // Save new device key
+            await this._stateService.SetDeviceKeyAsync(deviceKey);
+            return deviceKey;
+
```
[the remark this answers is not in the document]
jlf0dev: Do we have documentation in the docs that we don't prefer `this`? If not we need to add that for consistency.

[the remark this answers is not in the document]
fedemkr: I don't think we have docs for that, it'd really cool to add it. I was just guided on the general use (or not use hehe) of it in the app.

src/Core/Services/DeviceTrustCryptoService.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public async Task<SymmetricCryptoKey> GetDeviceKeyAsync()
+        {
+            return await _stateService.GetDeviceKeyAsync();
+        }
+
+        private async Task SetDeviceKeyAsync(SymmetricCryptoKey deviceKey)
+        {
+            await _stateService.SetDeviceKeyAsync(deviceKey);
+        }
```
fedemkr, line 42: ⛏️ IMO this method shouldn't exist, and one should directly use the state service one, given that it's not adding/modifying any of the logic, it's private and it has the same name. However, maybe you've added it as a preparation because afterwards it will have more logic, so if that's the case it's fine.

fedemkr (approved):
LGTM, just detected one small thing but nothing to block the PR.

PR #2578, opened 2023-06-21
[PM-724] Pad activity result lock delay by 1s

mpbw2:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This PR slightly modifies the behavior introduced in #2539 by setting the lock delay to 1 second instead of relying on lifecycle events to remove the delay completely.  QA has devices and/or configurations that call `ShouldTimeoutAsync()` more than once on return resulting in returning to a locked app.  Adding back a bit of time covers for these devices while still satisfying the original issue by keeping the unlock window short.

## Code changes

* **VaultTimeoutService.cs:** Change `DelayTimeoutMs` from `null` to `1000`

## Before you submit

### DIFF
--- a/src/Core/Services/VaultTimeoutService.cs
+++ b/src/Core/Services/VaultTimeoutService.cs
@@ -132,7 +132,7 @@ namespace Bit.Core.Services
             {
                 if (ResetTimeoutDelay)
                 {
-                    DelayTimeoutMs = null;
+                    DelayTimeoutMs = 1000;
                     ResetTimeoutDelay = false;
                 }
                 return false;


fedemkr:
So with this change after a user chooses a file and returns to the app `DelayTimeoutMs` won't be `null` again until the app is killed and restarted (or tombstoned). Is that the desired approach?
Because this means that after this operation, the vault timeout will be in the selected time by the user +1 sec. I know 1 sec is almost nothing but just wanted to make sure that is what's intended.

mpbw2:
> So with this change after a user chooses a file and returns to the app `DelayTimeoutMs` won't be `null` again until the app is killed and restarted (or tombstoned). Is that the desired approach? 

That's correct, the `null` was introduced in the prior PR which exposed the issue with QA's devices.  By giving it a value we're effectively reverting to the previous behavior but with setting a much shorter window upon returning to the app made possible with the reset flag set during `OnActivityResult`.

> Because this means that after this operation, the vault timeout will be in the selected time by the user +1 sec. I know 1 sec is almost nothing but just wanted to make sure that is what's intended.

Also correct, and I believe is what triggered the initial report that locking stopped working when in fact it was always adding _60_ seconds to the vault timeout value once it was set.  I'm not sure how to safely null it without being able to see exactly why `ShouldTimeoutAsync()` is being called more than once on the QA devices, so I went this route instead.

fedemkr:
Let me know if this is blocking QA and I'll approve it while we discuss alternative solutions.

The only thing that comes to mind now to safely set that to `null` is to have an additional property stating the time the `DelayTimeoutMs` was set to `1000` called, e.g. `LastTimeDelayTimeoutPadReset`, then:
- When setting `DelayTimeoutMs` to `1000`, save current date time into `LastTimeDelayTimeoutPadReset`
- Next time `ShouldTimeoutAsync` is called, check `LastTimeDelayTimeoutPadReset` different than `null` and time past is greater than let's say 3 or 5 seconds, set `DelayTimeoutMs` to `null` and also this new property.

Doing this, we safely `nullify` the timeout after a reasonable time has past and it avoids the problem of multiple times.
Maybe it's an overkill to the problem, but it does reset the delay completely, specially for people that don't kill the app and just leave it in the background.

mpbw2:
Closing this as QA is reporting the issue persists in this build, so something else must be happening

PR #2579, opened 2023-06-21
[PM-192] Refactor forwarded email providers

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Refactor forwarded email providers to use better patterns, reuse code and improve maintainability.

## Code changes

* **GeneratorPage.xaml:** Refactor all forwarded email providers to unify all the API token/key into one field binded which changes its value depending on the provider type selected. This makes the view tree lighter and changing from one type to the other smoother.
* **GeneratorPageViewModel.cs:** Unified all the API Token/Key into `ForwardedEmailApiSecret` alongside the label and the show toggle as well. Inside it gets/sets the correct one depending on the type selected. Simplified the api secret toggle command and improved the errors to the user when we get a 401/403 from the provider.
* **ForwardedEmailInvalidSecretException.cs:** Special exception to be raised when a 401/403 is gotten from the provider.
* **UsernameGenerationOptions.cs:** Added the "mapper" of the particular options for the forwarded email services.
* **UsernameGeneratorConfig.cs:** No longer necessary -> Removed.
* **ApiService.cs:** Moved all the particular forwarded email logic to the new `...Forwarder` classes and added a new `SendAsync(HttpRequestMessage, CancellationToken)` to be used for these kind of things where we don't call our API. Only left `GetFastmailAccountIdAsync(...)` cause I didn't want to move the `HttpClient` outside this class in this PR. ♻️  Eventually, we should refactor this class and split it, it's getting pretty big.
* **CryptoService.cs:** Moved `RandomStringAsync` here cause it made more sense to be in the same place as random number.
* **BaseForwarder.cs:** Base class for all forwarders where it makes the call to the service with custom configurations from its children.
* **AnonAddyForwarder.cs, DuckDuckGoForwarder.cs, FastmailForwarder.cs, FirefoxRelayForwarder.cs, SimpleLoginForwarder.cs:** Forwarder providers particular logic and configurations.
* **ForwarderOptions.cs:** Options to configure the forwarders.
* **UsernameGenerationService.cs:** Refactor the forwarder email to create the appropriate forwarder and generate accordingly.

## Before you submit


mpbw2 (approved):

aj-rosado (approved):
LGTM

PR #2580, opened 2023-06-22
[PM-2691] Adding AutomationIDs for Vault Page sections

ifernandezdiaz:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [X] Other

## Objective
This PR adds more AutomationIDs that will help us to improve the quality of our Mobile Automation tests

## Code changes

* **GroupingsPage.xaml:** Adding AutomationID label to filters and headers 
* **GroupingsPageHeaderListItem.cs:** Adding `AutomationId` variable.
* **GroupingsPageHeaderListGroup.cs:** Adding `AutomationId` variable.
* **GroupingsPageListItem.cs:** Adding `AutomationId` variable.

## Before you submit


src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
```diff
@@ -60,13 +60,13 @@
...
                     WebsiteIconsEnabled="{Binding BindingContext.WebsiteIconsEnabled, Source={x:Reference _page}}"
-                    TotpSec="{Binding TotpSec}"/>
+                    TotpSec="{Binding TotpSec}" />
             </DataTemplate>
 
             <DataTemplate x:Key="groupTemplate"
                           x:DataType="pages:GroupingsPageListItem">
                 <controls:ExtendedStackLayout Orientation="Horizontal" 
-                                                StyleClass="list-row, list-row-platform">
+                                                StyleClass="list-row, list-row-platform" AutomationId="{Binding AutomationId}">
```
mpbw2, line 69: Should be on the next line

src/App/Pages/Vault/GroupingsPage/GroupingsPage.xaml
```diff
@@ -96,7 +98,7 @@
                     Spacing="0"
                     Padding="0"
                     VerticalOptions="FillAndExpand"
-                    StyleClass="list-row-header-container, list-row-header-container-platform">
+                    StyleClass="list-row-header-container, list-row-header-container-platform" AutomationId="{Binding AutomationId}">
```
mpbw2, line 101: Should be on the next line

src/App/Pages/Vault/GroupingsPage/GroupingsPageHeaderListItem.cs
```diff
@@ -10,5 +10,12 @@ public GroupingsPageHeaderListItem(string title, string itemCount)
...
+        public string AutomationId
+        {
+            get
+            {
+                return $"{Title.Replace(" ", string.Empty).ToLowerInvariant()}Header";
+            }
+        }
     }
-}
+}
```
mpbw2, line 21: Revert the removal of the ending newline

src/App/Pages/Vault/GroupingsPage/GroupingsPageListItem.cs
```diff
@@ -2,13 +2,15 @@
 using Bit.Core;
 using Bit.Core.Enums;
 using Bit.Core.Models.View;
+using static Bit.Core.Utilities.DomainName.TLDRule;
```
mpbw2, line 5: Doesn't look like this is used

src/App/Pages/Vault/GroupingsPage/GroupingsPageListItem.cs
```diff
@@ -18,7 +20,7 @@ public class GroupingsPageListItem : IGroupingsPageListItem
         public bool FuzzyAutofill { get; set; }
         public bool IsTrash { get; set; }
         public bool IsTotpCode { get; set; }
-
+
```
mpbw2, line 23: Revert formatting change

mpbw2 (changes requested):
see inline

mpbw2 (commented):
I think this is ok for the groupings page but wouldn't mind a quick glance by @fedemkr in case I'm missing something re: performance and/or recycling...

src/App/Pages/Vault/GroupingsPage/GroupingsPageListGroup.cs
```diff
@@ -26,11 +26,13 @@ public GroupingsPageListGroup(string name, int count, bool doUpper = true, bool
...
             ItemCount = count.ToString("N0");
             First = first;
+            AutomationId = $"{NameShort}ListGroup";
         }
 
         public bool First { get; set; }
         public string Name { get; set; }
         public string NameShort => string.IsNullOrWhiteSpace(Name) || Name.Length == 0 ? "-" : Name[0].ToString();
         public string ItemCount { get; set; }
+        public string AutomationId { get; set; }
```
fedemkr, line 36: You can remove the public setter here. It's being set in the constructor so you don't need the `set` at all and can directly be:
```c#
public string AutomationId => $"{NameShort}ListGroup";
```

src/App/Pages/Vault/GroupingsPage/GroupingsPageListItem.cs
```diff
@@ -115,5 +116,51 @@ public string Icon
...
+                    _automationId = "CollectionFilter";
+                }
+                else if (IsTotpCode)
+                {
+                    _automationId = "TOTPListItem";
+                }
+                
+                return _automationId;
+            }
+        }
```
fedemkr, line 163: No need for the field `_automationId` return directly on each case:
```c#
public string AutomationId
{
    get
    {
        if (Type.HasValue)
        {
            switch (Type.Value)
            {
                case CipherType.Login:
                    return "LoginFilter";
                case CipherType.SecureNote:
                    return= "SecureNoteFilter";
                case CipherType.Card:
                    return "CardFilter";
                case CipherType.Identity:
                    return "IdentityFilter";
            }
        }
        
        if (IsTrash)
        {
            return "TrashFilter";
        }
        if (Folder != null)
        {
            return "FolderFilter";
        }
        if (Collection != null)
        {
            return "CollectionFilter";
        }
        if (IsTotpCode)
        {
            return "TOTPListItem";
        }

        return null;
    }
}

```

fedemkr (changes requested):

src/App/Pages/Settings/SettingsPage/SettingsPageListItem.cs
```diff
@@ -22,5 +23,41 @@ public class SettingsPageListItem : ISettingsPageListItem
...
+                    .Replace(" ", String.Empty)
+                    .Replace("-", String.Empty);
+                    return $"{idText}Cell";
+                }
+                else
+                {
+                    return "EnabledPolicyCell";
+                }
+            }
+        }
```
fedemkr, line 44: ♻️ I see that the logic is repeated in other places (like in the `AutomationIdSettingStatus` property). So extract this to an extension method or helper class that you can reuse. 
You can have a class named `AutomationIdsHelper` that could have this kind of stuff, like a static method `ToEnglishTitleCase(string)` which could hold this logic.
Also, you can add an add suffix static method like `AddSuffixFor(string, SuffixType)` with `SuffixType` being an enum of the different things to suffix like `...Cell` and inside add the suffix to the correct component and with that you have less error prone code because of hardcoded strings. 
Then for this scenario you could have something like this:
```c#
public string AutomationId
{
    get
    {
        return AutomationIdsHelper.AddSuffixFor( 
                    UseFrame ? "EnabledPolicy" 
                             : AutomationIdsHelper.ToEnglishTitleCase(Name)
                    , SuffixType.Cell);
    }
}
``` 
---
⛏️ Also, given that there are more than one `AutomationId...` properties, give a more descriptive name to this one to know how it differentiate from `AutomationIdSettingStatus`

src/App/Pages/Settings/SettingsPage/SettingsPageListItem.cs
```diff
@@ -22,5 +23,41 @@ public class SettingsPageListItem : ISettingsPageListItem
...
+                    var idText = new CultureInfo("en-US", false)
+                    .TextInfo
+                    .ToTitleCase(Name)
+                    .Replace(" ", String.Empty)
+                    .Replace("-", String.Empty);
+                    return $"{idText}SettingValue";
+                }
+                return null;
+            }
+        }
```
fedemkr, line 61: ♻️  Apply same thing as in `AutomationId` and I tend to use failure scenarios at first like here:
```c#
if (UseFrame)
{
    return null;
}

// the other logic
```

src/App/Pages/Vault/CipherDetailsPage.xaml
```diff
@@ -69,24 +69,27 @@
 
             <ScrollView x:Key="scrollView" x:Name="_scrollView">
                 <StackLayout Spacing="20" x:Name="_mainLayout">
-                    <StackLayout StyleClass="box">
+                    <StackLayout StyleClass="box" AutomationId="ItemInformationSection">
                         <StackLayout StyleClass="box-row-header">
                             <Label Text="{u:I18n ItemInformation, Header=True}"
-                                   StyleClass="box-header, box-header-platform" />
+                                   StyleClass="box-header, box-header-platform"
+                                   AutomationId="" />
```
fedemkr, line 76: Is it ok that this id is empty?

src/App/Pages/Vault/CipherDetailsPage.xaml
```diff
@@ -186,11 +192,11 @@
                                     AutomationProperties.IsInAccessibleTree="True"
                                     AutomationProperties.Name="{u:I18n CopyPassword}"
                                     IsVisible="{Binding Cipher.ViewPassword}"
-                                    AutomationId="LoginCopyPasswordButton" />
+                                    AutomationId="CopyValueButton" />
                             </Grid>
                             <BoxView StyleClass="box-row-separator"
                                      IsVisible="{Binding Cipher.Login.Password, Converter={StaticResource stringHasValue}}" />
-                            <Grid StyleClass="box-row" IsVisible="{Binding ShowTotp}">
+                            <Grid StyleClass="box-row" IsVisible="{Binding ShowTotp}" AutomationId="ItemRow">
```
fedemkr, line 199: ⛏️ I'm not a fan of more than 1 or 2 inline properties set up in the xaml. So as a suggestion I'd put this vertically, which IMO is easier to read:
```xaml
<Grid 
    StyleClass="box-row" 
    IsVisible="{Binding ShowTotp}" 
    AutomationId="ItemRow">
```

src/App/Pages/Vault/GroupingsPage/GroupingsPageListItem.cs
```diff
@@ -115,5 +116,44 @@ public string Icon
...
+                    return "FolderFilter";
+                }
+                else if (Collection != null)
+                {
+                    return "CollectionFilter";
+                }
+                else if (IsTotpCode)
+                {
+                    return "TOTPListItem";
+                }
```
fedemkr, line 153: ⛏️  Given that you are returning on each part, then the `else` keyword is not needed, you can just use `if (...)` directly cause the path will finish on each `return ...`

src/App/Pages/Vault/GroupingsPage/GroupingsPageListItem.cs
```diff
@@ -9,6 +9,7 @@ public class GroupingsPageListItem : IGroupingsPageListItem
     {
         private string _icon;
         private string _name;
+        private string _automationId;
```
fedemkr, line 12: ⛏️  Not being used, remove

src/App/Resources/AppResources.it.resx
```diff
@@ -2613,7 +2613,7 @@ Vuoi passare a questo account?</value>
     <value>US</value>
   </data>
   <data name="EU" xml:space="preserve">
-    <value>UE</value>
+    <value>EU</value>
```
fedemkr, line 2616: 🤔  I believe that if we want to do any change on resources we need to go to Crowdin and add a translation over there to be then impacted on the code. Is this correct @djsmith85 or can we also apply direct changes on translations without going through Crowdin?

fedemkr (changes requested):

src/App/Pages/Settings/SettingsPage/SettingsPageListItem.cs
ifernandezdiaz: I'm still feeling like a foreigner in this repository, and that's why I choose to add those similar methods instead of adding new files to the project.

What is the best location to store this helper class? I'm thinking about creating a new `AutomationHelpers` inside `App/Utilities` but I'd like to read your thoughts 🙌

fedemkr: `App/Utilities` is fine but I'd add an `Automation` folder in there so you can add anything related to that there. Therefore the location would be `App/Utilities/Automation`

src/Android/Renderers/CustomLabelRenderer.cs
```diff
@@ -0,0 +1,31 @@
...
+            var label = sender as CustomLabel;
+            switch (e.PropertyName)
+            {
+                case nameof(CustomLabel.AutomationId):
+                    Control.ContentDescription = label.AutomationId;
+                    break;
+            }
+            base.OnElementPropertyChanged(sender, e);
+        }
+    }
```
fedemkr, line 29: 🤔  Does this work to fix the list automation id binding issue? if not, remove it

src/App/Utilities/Automation/AutomationIdsHelper.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Globalization;
+
+namespace Bit.App.Utilities.Automation
+{
+    public static class AutomationIdsHelper
+    {
+
```
fedemkr, line 8: ⛏️  Remove space

src/App/Utilities/Automation/AutomationIdsHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+        public static string ToEnglishTitleCase(string name)
+        {
+            return new CultureInfo("en-US", false)
+                    .TextInfo
+                    .ToTitleCase(name)
+                    .Replace(" ", String.Empty)
+                    .Replace("-", String.Empty);
+
```
fedemkr, line 16: ⛏️  Remove space

src/App/Utilities/Automation/AutomationIdsHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public static string AddSuffixFor(string text, SuffixType type)
+        {
+            return $"{text}{Enum.GetName(typeof(SuffixType), type)}";
+        }
+
```
fedemkr, line 23: ⛏️  Remove space

src/App/Utilities/Automation/SuffixType.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
fedemkr, line 1: ⛏️  This using is not needed

src/iOS.Core/Renderers/CustomLabelRenderer.cs
```diff
@@ -23,7 +23,15 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
...
-			base.OnElementPropertyChanged(sender, e);
+            var label = sender as CustomLabel;
+            switch (e.PropertyName)
+            {
+                case nameof(CustomLabel.AutomationId):
+                    Control.AccessibilityIdentifier = label.AutomationId;
+                    break;
+            }
+
+            base.OnElementPropertyChanged(sender, e);
```
fedemkr, line 34: 🤔 Same as in android, if it doesn't work remove it.

fedemkr (changes requested):

src/Android/Renderers/CustomLabelRenderer.cs
ifernandezdiaz: This workaround is valid! It's working because I'm using the `CustomLabel` element instead of `ExtendedStackLayout` (exactly as it was mentioned [here](https://github.com/xamarin/Xamarin.Forms/issues/10492))

fedemkr (approved):

mpbw2 (approved):

PR #2583, opened 2023-06-23
[PM-2737] Adding AutomationIDs for Send page elements

ifernandezdiaz:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [X] Other

## Objective
This PR adds more AutomationIDs that will help us to improve the quality of our Mobile Automation tests

## Code changes

* **SendAddEditPage.xaml:** Adding IDs for each Send element
* **SendGroupingsPage.xaml:** Adding IDs for Send lists
* **SendGroupingsPageListItem.xaml:** Adding a new variable to make AutomationIDs dynamic
* **SendsPage.xaml:** Adding IDs for label displayed when send list is empty

## Before you submit

### DIFF
--- a/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
+++ b/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
@@ -49,7 +49,7 @@
             </DataTemplate>
 
             <DataTemplate x:Key="sendGroupTemplate"
-                          x:DataType="pages:SendGroupingsPageListItem" >
+                          x:DataType="pages:SendGroupingsPageListItem">
                 <controls:ExtendedStackLayout Orientation="Horizontal" 
                                               StyleClass="list-row, list-row-platform" AutomationId="{Binding AutomationId}">
                     <controls:IconLabel Text="{Binding Icon, Mode=OneWay}"
--- a/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
+++ b/src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
@@ -44,13 +44,14 @@
                 <controls:SendViewCell
                     Send="{Binding Send}"
                     ButtonCommand="{Binding BindingContext.SendOptionsCommand, Source={x:Reference _page}}"
-                    ShowOptions="{Binding BindingContext.SendEnabled, Source={x:Reference _page}}" />
+                    ShowOptions="{Binding BindingContext.SendEnabled, Source={x:Reference _page}}"
+                    AutomationId="SendCell" />
             </DataTemplate>
 
             <DataTemplate x:Key="sendGroupTemplate"
-                          x:DataType="pages:SendGroupingsPageListItem">
+                          x:DataType="pages:SendGroupingsPageListItem" >
                 <controls:ExtendedStackLayout Orientation="Horizontal" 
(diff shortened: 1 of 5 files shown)


src/App/Pages/Send/SendGroupingsPage/SendGroupingsPage.xaml
```diff
@@ -44,13 +44,14 @@
...
-                    ShowOptions="{Binding BindingContext.SendEnabled, Source={x:Reference _page}}" />
+                    ShowOptions="{Binding BindingContext.SendEnabled, Source={x:Reference _page}}"
+                    AutomationId="SendCell" />
             </DataTemplate>
 
             <DataTemplate x:Key="sendGroupTemplate"
                           x:DataType="pages:SendGroupingsPageListItem">
                 <controls:ExtendedStackLayout Orientation="Horizontal" 
-                                              StyleClass="list-row, list-row-platform">
+                                              StyleClass="list-row, list-row-platform" AutomationId="{Binding AutomationId}">
```
mpbw2, line 54: Move to next line

mpbw2 (changes requested):
See inline

mpbw2 (commented):
Looks good, waiting for input on #2580 since it follows the same approach

src/App/Pages/Send/SendGroupingsPage/SendGroupingsPageListItem.cs
```diff
@@ -66,5 +67,31 @@ public string Icon
...
+                        case SendType.File:
+                            _automationId = "SendFileFilter";
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                return _automationId;
+            }
+        }
```
fedemkr, line 95: ⛏️  Right now `_automationId` is not useful. If the `_automationId` is meant to be used as a cache you should add something like this at beginning:
```c#
if (_automationId != null)
{
    return _automationId;
}
```
However, as far as I can see this property won't be triggered more than once per item so you can directly return the value on each part to avoid the field:
```c#
if (_name != null)
{
    return "SendItem";
}
if (Type != null)
{
    switch (Type.Value)
    {
        case SendType.Text:
            return  "SendTextFilter";
        case SendType.File:
            return  "SendFileFilter";
    }
}
return null;
```

fedemkr (changes requested):

src/App/Pages/Send/SendGroupingsPage/SendGroupingsPageListItem.cs
```diff
@@ -66,5 +66,31 @@ public string Icon
...
+                    switch (Type.Value)
+                    {
+                        case SendType.Text:
+                            return "SendTextFilter";
+                            break;
+                        case SendType.File:
+                            return "SendFileFilter";
+                            break;
+                        default:
+                            break;
```
fedemkr, line 89: ⛏️ `break;` is not needed after `return ...` in a `case` and I think you can remove the `default` as well

fedemkr (changes requested):

src/App/Pages/Send/SendGroupingsPage/SendGroupingsPageListItem.cs
ifernandezdiaz: You are right! Updating it...

fedemkr (approved):

mpbw2 (approved):

PR #2594, opened 2023-06-29
[PM-2378] Display error toast on duplicate Passkey when moving cipher to an organization

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Display error toast on duplicate passkey when moving item to an organization.

## Code changes

* **SharePageViewModel:** Added check of duplicate passkey error and show the toast to the user in that case
* **ICipherService:** Added `ShareWithServerError` to return the appropriate response to the caller when sharing cipher to an org. I cannot throw an `ApiException` with the error inside given that `AppResources` is not available in the `Core` project. Also another way would be to create and throw a specific exception for this and handle it on the caller but I thought this was better and the caller can handle the different paths more directly.
* **CipherService:** Added the validation to check whether the cipher has a FIdo2Key and if so, check if the organization already has any ciphers with the same `RpId` (duplicate passkey) to return an error if that's the case.

## Before you submit

### DIFF
--- a/src/App/Resources/AppResources.resx
+++ b/src/App/Resources/AppResources.resx
@@ -2671,4 +2671,7 @@ Do you want to switch to this account?</value>
   <data name="InvalidAPIToken" xml:space="preserve">
     <value>Invalid API token</value>
   </data>
+  <data name="ThisItemCannotBeSharedWithTheOrganizationBecauseThereIsOneAlreadyWithTheSamePasskey" xml:space="preserve">
+    <value>This item cannot be shared with the organization because there is one already with the same passkey.</value>
+  </data>
 </root>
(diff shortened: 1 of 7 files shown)


mpbw2 (approved):
> ThisItemCannotBeSharedWithTheOrganizationBecauseThereIsOneAlreadyWithTheSamePasskey

perfect

src/App/Resources/AppResources.resx
```diff
@@ -2671,4 +2671,50 @@ Do you want to switch to this account?</value>
...
+  <data name="FormatXSeparateMultipleURIsWithAComma" xml:space="preserve">
+    <value>Format: {0}. Separate multiple URIs with a comma.</value>
+  </data>
+  <data name="FormatX" xml:space="preserve">
+    <value>Format: {0}</value>
+  </data>
+  <data name="InvalidURI" xml:space="preserve">
+    <value>Invalid URI</value>
+  </data>
+  <data name="URISaved" xml:space="preserve">
```
mpbw2: `URISaved` is already declared above (line 2683)

mpbw2 (approved):

PR #2604, opened 2023-07-07
[AC-1425] Fix iOS blank vault load

fedemkr:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This PR attempts to fix an issue happening from time to time, where the vault view's list doesn't load
Even though I couldn't reproduce the issue, I forced in the code the properties not to propagate changes to the view which end in this state, and based on these assumptions this fix is presented.

## Code changes

* **GroupingPageViewModel:** Added a bunch of main thread invocations for the properties expected to change from the view, so in case there is another thread executing the load, the property changed will be invoked in the main thread.

## Before you submit


mpbw2 (approved):

LRNcardozoWDF (approved):
LGTM

PR #2608, opened 2023-07-10
[PM-2287][PM-2289][PM-2293] Approval Options

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Create navigation and logic to trust a device on the 3 available device approval options: device, admin and master password.

## Code changes

Change existing `LoginPasswordlessRequestPage` to work with both types of auth requests, admin and login with device, by accepting an AuthRequestType and passing it the the service. Also added to this page a check to perform a device trust on successful login.
Master password also has a check to make the device trusted on successful login.

## Screenshots

New Screen:
image

## Before you submit


src/App/Pages/Accounts/LoginApproveDeviceViewModel.cs
```diff
@@ -118,6 +125,15 @@ public async Task InitAsync()
             {
                 HandleException(ex);
             }
+
+            // TODO: Change this expression to, Appear if the browser is trusted and shared the key with the app
+            ContinueEnabled = !RequestAdminApprovalEnabled && !ApproveWithMasterPasswordEnabled && !ApproveWithMyOtherDeviceEnabled;
```
fedemkr, line 130: ❓  Is this TODO going to be addressed in this PR or in another one?

src/App/Pages/Accounts/LoginApproveDeviceViewModel.cs
```diff
@@ -118,6 +125,15 @@ public async Task InitAsync()
...
+        }
+
+        private async Task SetDeviceTrustAndInvoke(Action action)
```
fedemkr, line 133: ⛏️ Add `Async` suffix to be consistent with other async methods in the app.

src/Core/Enums/AuthRequestType.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+namespace Bit.Core.Enums
+{
+    public enum AuthRequestType : byte
+    {
+        AuthenticateAndUnlock = 0,
+        Unlock = 1,
```
fedemkr, line 7: ❓ Just in case, I see this one not being used currently. Is this alright?

src/Core/Models/Request/PasswordlessCreateLoginRequest.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System;
+using Bit.Core.Enums;
+
```
fedemkr, line 3: ⛏️ Remove this given that it's not being used as far as I see

fedemkr (changes requested):

src/App/Pages/Accounts/LoginApproveDeviceViewModel.cs
andrebispo5: This will be address in another PR.

src/Core/Constants.cs
```diff
@@ -53,6 +53,7 @@ public static class Constants
         public const string AppLocaleKey = "appLocale";
         public const string ClearSensitiveFields = "clearSensitiveFields";
         public const string ForceUpdatePassword = "forceUpdatePassword";
+        public const string RememberDeviceTde = "rememberDeviceTde";
```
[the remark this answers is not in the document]
andrebispo5: Since i renamed the method i changed the constant as well to "ShouldTrustDevice"

src/Core/Models/Request/PasswordlessCreateLoginRequest.cs
andrebispo5: the import is being used on the constructor and property for the `AuthRequestType? type`

src/Core/Enums/AuthRequestType.cs
andrebispo5: It is the same enum as we have on the server.

andrebispo5: it is just there for consistency so far.

src/Core/Services/DeviceTrustCryptoService.cs
```diff
@@ -77,5 +77,26 @@ private async Task<SymmetricCryptoKey> MakeDeviceKeyAsync()
...
+        public async Task<DeviceResponse> TrustDeviceIfNeededAsync()
+        {
+            DeviceResponse response = null;
+            if (await GetShouldTrustDeviceAsync())
+            {
+                response = await TrustDeviceAsync();
+                await SetShouldTrustDeviceAsync(false);
+            }
+            return response;
+        }
```
fedemkr, line 100: ⛏️ I don't see the `DeviceResponse` used anywhere and this can be written as follows:
```c#
public async Task TrustDeviceIfNeededAsync()
{
    if (!await GetShouldTrustDeviceAsync())
    {
        return null;
    }

    await TrustDeviceAsync();
    await SetShouldTrustDeviceAsync(false);
}
```
When it's something to be executed conditionally, I try to put the conditions at first to early return and most of the times it simplifies the code.

fedemkr (changes requested):

src/Core/Services/DeviceTrustCryptoService.cs
andrebispo5: yup you are totally right, my bad !

fedemkr (approved):

PR #2610, opened 2023-07-10
[PM-1201] Change timeout actions available based on hasMasterPassword

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

For users without a master password, the unlock settings need to be different:
- Vault timeout action defaults to “log out”
- Vault timeout action “lock” is only available as an option if the user has enabled PIN unlock or biometrics unlock.
- If user disables PIN/Biometric unlock methods, then the vault timeout action defaults back to “log out”

## Code changes

- Get DecryptionOptions to check if the user has master password set. 
if the user doesn't have a master password set up:
- Change PIN and Biometric disable logic to reset timeout action to Logout
- Change default timeout action to logout if there is no saved action.

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/972cd706-61a7-4c35-bab6-9368848dc6d2

## Before you submit

### DIFF
--- a/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
+++ b/src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
@@ -48,6 +48,7 @@ namespace Bit.App.Pages
         private bool _reportLoggingEnabled;
         private bool _approvePasswordlessLoginRequests;
         private bool _shouldConnectToWatch;
+        private bool _hasMasterPassword;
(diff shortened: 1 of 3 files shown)


fedemkr (changes requested):
Looks good, some minor adjustments and also what do you think about adding a toast telling the user the Vault Timeout action has changed when they turn off Pin/Biometric?

fedemkr (approved):

PR #2623, opened 2023-07-17
[PM-2297] Login with trusted device (Flow 2)

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

After an SSO Login with an organisation that has TDE enabled:
- if the device is trusted, navigate to the users vault and decrypt the user's key with the device key.
- if the device is not trusted, navigate to the device approval options screen.

## Code changes

Additionally to the changes required to do the logic for the objective of the PR, navigations for the device approval options were also added to the iOS extensions.
Models were also updated with additional properties to match server responses. 

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/71ed138d-0736-4286-82c8-d955d04c3252

## Before you submit


src/Core/Services/AuthService.cs
```diff
@@ -486,7 +489,17 @@ private async Task<MasterKey> MakePreloginKeyAsync(string masterPassword, string
...
+                    if (decryptOptions?.TrustedDeviceOption != null)
+                    {
+                        var key = await _deviceTrustCryptoService.DecryptUserKeyWithDeviceKeyAsync(decryptOptions?.TrustedDeviceOption.EncryptedPrivateKey, decryptOptions?.TrustedDeviceOption.EncryptedUserKey);
+                        if (key != null)
+                        {
+                            await _cryptoService.SetUserKeyAsync(key);
+                        }
+                    }
+                    else if (masterKey != null &&
+                        (!string.IsNullOrEmpty(tokenResponse.KeyConnectorUrl) || !string.IsNullOrEmpty(decryptOptions?.KeyConnectorOption?.KeyConnectorUrl)))
```
jlf0dev, line 502: Can we add the `if (tokenResponse.KeyConnectorUrl != null)` above to this `else if` ladder? we can move the `SetMasterKeyEncryptedUserKeyAsync` above it.

We should also be checking for the `KeyConnectorUrl` on the `decyptOptions` as part of the key connector check right?

src/Core/Services/StateService.cs
```diff
@@ -515,13 +515,21 @@ public async Task SetPrivateKeyEncryptedAsync(string value, string userId = null
 
         public async Task<SymmetricCryptoKey> GetDeviceKeyAsync(string userId = null)
         {
-            var deviceKeyB64 = await _storageMediatorService.GetAsync<string>(Constants.DeviceKeyKey(userId), true);
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var deviceKeyB64 = await _storageMediatorService.GetAsync<string>(Constants.DeviceKeyKey(reconciledOptions.UserId), true);
```
jlf0dev, line 520: Do we need the `reconciledOptions` here? Can we pass the `userId` in?

jlf0dev (commented):
That `AuthService` is hard to reason with. We'll need to ask QA to do a full regression with key connector.

Overall it's looking great! Would love to hear @fedemkr's thoughts as well if he has the time.

src/Core/Services/StateService.cs
```diff
@@ -527,9 +525,7 @@ public async Task<SymmetricCryptoKey> GetDeviceKeyAsync(string userId = null)
 
         public async Task SetDeviceKeyAsync(SymmetricCryptoKey value, string userId = null)
         {
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            await _storageMediatorService.SaveAsync(Constants.DeviceKeyKey(reconciledOptions.UserId), value.KeyB64, true);
+            await _storageMediatorService.SaveAsync(Constants.DeviceKeyKey(userId), value.KeyB64, true);
```
mpbw2: same here

src/Core/Services/StateService.cs
```diff
@@ -515,9 +515,7 @@ public async Task SetPrivateKeyEncryptedAsync(string value, string userId = null
 
         public async Task<SymmetricCryptoKey> GetDeviceKeyAsync(string userId = null)
         {
-            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
-                await GetDefaultStorageOptionsAsync());
-            var deviceKeyB64 = await _storageMediatorService.GetAsync<string>(Constants.DeviceKeyKey(reconciledOptions.UserId), true);
+            var deviceKeyB64 = await _storageMediatorService.GetAsync<string>(Constants.DeviceKeyKey(userId), true);
```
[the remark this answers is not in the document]
jlf0dev: Oops that would be my fault for suggesting that. @andrebispo5 please disregard my previous comments.

src/Core/Services/AuthService.cs
```diff
@@ -479,18 +482,27 @@ private async Task<MasterKey> MakePreloginKeyAsync(string masterPassword, string
...
-                        await _keyConnectorService.GetAndSetKey(tokenResponse.KeyConnectorUrl);
-                    }
-
+                    var decryptOptions = await _stateService.GetAccountDecryptionOptions();
                     await _cryptoService.SetMasterKeyEncryptedUserKeyAsync(tokenResponse.Key);
 
-                    if (masterKey != null)
+                    if (decryptOptions?.TrustedDeviceOption != null)
+                    {
+                        var key = await _deviceTrustCryptoService.DecryptUserKeyWithDeviceKeyAsync(decryptOptions?.TrustedDeviceOption.EncryptedPrivateKey, decryptOptions?.TrustedDeviceOption.EncryptedUserKey);
```
fedemkr, line 490: ⛏️ `decryptOptions` here won't be null because of the previous if so no need to re-check it:
```suggestion
                        var key = await _deviceTrustCryptoService.DecryptUserKeyWithDeviceKeyAsync(decryptOptions.TrustedDeviceOption.EncryptedPrivateKey, decryptOptions.TrustedDeviceOption.EncryptedUserKey);
```

src/Core/Services/DeviceTrustCryptoService.cs
```diff
@@ -99,5 +99,33 @@ public async Task<DeviceResponse> TrustDeviceIfNeededAsync()
...
+
+            // Attempt to decrypt encryptedDevicePrivateKey with device key
+            var devicePrivateKey = await _cryptoService.DecryptToBytesAsync(
+              new EncString(encryptedDevicePrivateKey),
+              existingDeviceKey
+            );
+
+            // Attempt to decrypt encryptedUserDataKey with devicePrivateKey
+            var userKey = await _cryptoService.RsaDecryptAsync(encryptedUserKey, devicePrivateKey);
+            return new UserKey(userKey);
```
fedemkr, line 128: ⛏️ I think that given having objects that represent keys like `UserKey` we should use more specific variable names so we know what they represent (actual key object, bytes, encrypted string):
```suggestion
            // Attempt to decrypt encryptedDevicePrivateKey with device key
            var devicePrivateKeyBytes = await _cryptoService.DecryptToBytesAsync(
              new EncString(encryptedDevicePrivateKey),
              existingDeviceKey
            );

            // Attempt to decrypt encryptedUserDataKey with devicePrivateKey
            var userKeyBytes = await _cryptoService.RsaDecryptAsync(encryptedUserKey, devicePrivateKeyBytes);
            return new UserKey(userKeyBytes);
```

src/Core/Services/DeviceTrustCryptoService.cs
```diff
@@ -99,5 +99,33 @@ public async Task<DeviceResponse> TrustDeviceIfNeededAsync()
...
+
+        public async Task<bool> IsDeviceTrustedAsync()
+        {
+            var existingDeviceKey = await GetDeviceKeyAsync();
+            return existingDeviceKey != null;
+        }
+
+        public async Task<UserKey> DecryptUserKeyWithDeviceKeyAsync(string encryptedDevicePrivateKey, string encryptedUserKey)
+        {
+            // Get device key
```
fedemkr, line 111: ⛏️ Remove comment, it's unnecessary given that one can read it in the next line method.
```suggestion

```

fedemkr (changes requested):
Great work, overall looking good, just some minor adjustments 🎉

fedemkr (approved):

Issue #582, opened 2019-08-21
Connecting to a server with TLS Client Authentication crashes app


codingJWilliams:
Hello,

When connecting to a Bitwarden server that's behind an nginx proxy that requires a client cert, the app just crashes when pressing the Log In button. The same server works fine on Firefox, requesting access to my certificate as expected, and when I disable the requirement to have client authentication through my reverse proxy, the app works fine too. 
I see this is a known issue based on a few forum posts (https://community.bitwarden.com/t/client-certificates/427, https://community.bitwarden.com/t/mobile-app-cant-access-server-behind-reverse-proxy-with-client-cert-authentification/2071 etc) so thought I'd raise an issue.

agboom:
Hi, I have the same use case and a similar experience that you have @codingJWilliams. It would be a nice addition to the mobile app to support TLS client authentication. The added security would be beneficial for on-premise deployments.

Perhaps we can join forces and come up with an implementation that could be merged into mainline?

@kspearrin Could you give your opinion on this and maybe some pointers on where to start?

agboom:
Thanks for your quick answer and the pointer. I'll have a stab at it, but first I'll need to setup a C# dev environment on Linux. I'm quite new to C# development, so if anyone has any experience to share in C# Linux development, I'd be much obliged. My first bet is Rider from Jetbrains, let's hope this works :crossed_fingers:.

agboom:
It seems to be one of the advertised features of Rider: https://www.jetbrains.com/rider/features/

I'll let you know if it works out.

codingJWilliams:
Hello,

Thank you for the very kind offer @agboom but I'm rather hopeless at C#!
I did some basic research and this does seem to be possible with the System.Net.HttpClient but I wouldn't know where to start with implementing this - if you need any help testing or similar, however, please let me know.

agboom:
Thanks @codingJWilliams I'll give a shout if there's something to test or otherwise.

My main challenge right now is to get the dev environment working on Linux which is new to me for C#. The Jetbrains Rider IDE requires a paid license which is a bummer, because it's currently my only chance of Xamarin development on Linux AFAIK. Jetbrains does offer free licenses to open source project contributors, so maybe hope @kspearrin?

codingJWilliams:
Hello,

[...]

PR #2629, opened 2023-07-19
[PM-3089] mTLS - Authenticate with Client Certificate

oguzhane:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
This PR brings ability to login a bitwarden server by using a client certificate. The functionility has been implemented for Android.
User can import a X509 client certificate in pkcs#12 format to KeyStore or user can use existing certificates from keychain.

See: #582
## Code changes
**CertificateService.cs**: responsible to interact with platform's native key management apis.
**AdvancedPage.xaml**: page where user setup the certificate.
**AndroidHttpsHandler.cs**: http handler that send the requests with client certificate

## Screenshots
Demo

https://github.com/bitwarden/mobile/assets/4419532/a100dfc5-abfa-481b-b4fb-fe8cff1f67ed

## Before you submit


src/iOS.Core/Services/CertificateService.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Threading.Tasks;
+using Bit.Core.Abstractions;
+using Bit.Core.Models;
+
+namespace Bit.iOS.Core.Services
+{
+    public class CertificateService : ICertificateService
```
mpbw2, line 8: I incorrectly assumed this had already been implemented on iOS, and we require platform parity (where possible) for community PRs.  Are you planning to add this?

mpbw2 (changes requested):
See inline comment about iOS.

Also our design team has requested the cert handling UI be moved to the self-hosted configuration page (instead of adding an `Advanced` menu during the login flow).  Here's a mockup:

Screenshot 2023-07-23 at 10 34 25 PM

src/iOS.Core/Services/CertificateService.cs
oguzhane: @mpbw2 I would be willing to add for iOS but i've no dev/test envrionment to implement for iOS

oguzhane:
> See inline comment about iOS.
> 
> Also our design team has requested the cert handling UI be moved to the self-hosted configuration page (instead of adding an `Advanced` menu during the login flow). Here's a mockup:
> 
> Screenshot 2023-07-23 at 10 34 25 PM

Makes sense. I will be moving it to the self-hosted config page.

oguzhane:
> > See inline comment about iOS.
> > Also our design team has requested the cert handling UI be moved to the self-hosted configuration page (instead of adding an `Advanced` menu during the login flow). Here's a mockup:
> > Screenshot 2023-07-23 at 10 34 25 PM
> 
> Makes sense. I will be moving it to the self-hosted config page.

@mpbw2 i've moved it to environment page
image

biscuit-316:
I really would like this feature to be added. Especially because I notice that my Wireguard VPN does not work everywhere.

I saw here that you have an .apk file on your GitHub page. I've tried both [this](https://github.com/oguzhane/bitwarden-mobile/actions/runs/6108161757) and [this](https://github.com/oguzhane/bitwarden-mobile/actions/runs/6035756612) version. in both versions I found two problems.

The first problem is that the button to use system certificates does not work. When I press the button, I see a button pressed animation. but other than that nothing happens. The screen to select the certificate doesn't open.

The second problem is that the app does not accept the certificate if it is not encrypted with a password. To test the mutaul tls function I created a new certificate authority. For convenience I downloaded my client certificate without a password. But I kept getting the following error message. "Failed to import the certificate  exception decrypting data -java". Even though the certificate is not encrypted, the app does ask for a password. Even if I leave the password field blank, I get this message.

mzpqnxow:
> I really would like this feature to be added. Especially because I notice that my Wireguard VPN does not work everywhere.


Same here!

> I saw here that you have an .apk file on your GitHub page. I've tried both [this](https://github.com/oguzhane/bitwarden-mobile/actions/runs/6108161757) and [this](https://github.com/oguzhane/bitwarden-mobile/actions/runs/6035756612) version. in both versions I found two problems.

I have iOS here but I may be able to guess your issue(s)

> The first problem is that the button to use system certificates does not work. When I press the button, I see a button pressed animation. but other than that nothing happens. The screen to select the certificate doesn't open.

Seems like a non-ideal UX behavior, or bug. Sorry, can't help here. Do you not have any keypairs in your system store, I wonder? I imagine the behavior is still not desired- an error message would probably be better- but maybe that's an untested edge-case snd that's the behavior when there are no keypairs available. Perhaps some interface returns null or some exception/return code rather than raising the appropriate dialog, and the error is not handled in a helpful way

> The second problem is that the app does not accept the certificate if it is not encrypted with a password. To test the mutaul tls function I created a new certificate authority. For convenience I downloaded my client certificate without a password. But I kept getting the following error message. "Failed to import the certificate  exception decrypting data -java". Even though the certificate is not encrypted, the app does ask for a password. Even if I leave the password field blank, I get this message.

It sounds (based on the presence of "-java" in the error) that it's an improperly exported Java Keystore? Perhaps a raw JKS file?

I expect the application may want either PKCS12 or x509 format. It almost certainly doesn't support a JKS file

PKCS12 files (common extensions pfx, p12) are single password protected files with a standard/well-defined format that contain multiple files files within them- certs, private keys, CRLs. The files inside should be x509 files

If the app has separate boxes for client certificate, client private key, and CA, it almost certainly wants x509 format, one for certificate, one for private key, one for CA. Optionally, it may accept the certificate and private key combined into PEM encoded x509, concatenated into one file. The CA should still be separate

If you have your certs/keys in a JKS, export them using keytool since JKS is not a standard format recognized outside of Java (Google "export jks x509" or "export jks x509 pem")

Note: Creation of a PKCS12 file _will_ require you to specify a password when creating it and loading it into the app- it's required by the standard afaik. In that case, you should use the same password when loading it as when you created it. The private key inside the PKCS12 does not need a password, though

Remember to include the CA file, certificate and private key all in one PKCS12 (if using PKCS12) and do NOT create multiple PKCS12 files

This is my best guess, having not seen or tested the app yet. Hopefully it's helpful and I haven't misunderstood your issue

src/iOS.Core/Services/CertificateService.cs
mzpqnxow: > @mpbw2 I would be willing to add for iOS but i've no dev/test envrionment to implement for iOS

@oguzhane I'd be happy to send you a cheap/used iPhone from Swappa if you don't have one and you're willing/able to put the time in to iOS support

If so, you can reach me at:

github@<mygithubusername>.com

Either way, I appreciate the work you've done on this, if that's not obvious

I only need to know where to ship it; optionally, I may be able to find a gift card to send via e-mail, and you can handle ordering and shipping, if you have any privacy concerns

mzpqnxow:
First, thanks for taking the time to work on this, and to clarify on this

> Pkcs#12 is only supported format but it doesn't support chain of multiple certs in the file

Regarding the chain: it may want the CA + intermediates as a single PEM- the CA concatenated with the concatenated intermediate PEMs, all in one big PEM. Just an idea should you be interested in trying it

Or maybe you meant it didn't like having a CA *and* a client cert in the PKCS12- in which case my only idea would be to ensure the client certificate doesn't have the CA bit set

Regardless, it's no bother to me either way, I'm thrilled either way, I was just chiming in to assist @biscuit-316 who seems to have much larger issues (that sound to me like possibly pilot error, the "-java" string is a bit suspect)

> i already tried to import with CA but didn't work. Hence, i had to register CA system wide.

I wonder if iOS will be OK with this. I believe iOS prevents apps from utilizing _client certificates_ from the system store. Hopefully it's not the same with CA certificates

On the topic of iOS certificate policies- I believe the "Import System Certificate" you implemented for Android will need to be conditionally left out for iOS, because of what I previously mentioned re: app access to the client cert system store. There's no privilege the user can grant to override/relax that control, either. I know you're not set up to test iOS so this isn't immediately useful to you, I'm mentioning it only as an FYI. I can probably find a reference on that, even if only to keep me honest in case I'm mistaken

> i used openssl to generate pkcs#12. with that, i have been using the feature flawlessy over a month. 
> i tested importing from the system on emulator. Haven't tried on a real device.

👍 

Thanks again!

mzpqnxow:
[This](https://httptoolkit.com/blog/android-11-trust-ca-certificates/) may explain the CA issue- new policies in Android as of Android 11, preventing apps from adding to even the user CA store

Not exactly an authoritative source, but seems coherent and detailed enough to be trustworthy

biscuit-316:
> It seems like a non-ideal UX behavior or a bug. Sorry, I can't help here. Don't have any key pairs in your system store, I wonder? I imagine the behavior is still not desired - an error message would probably be better - but perhaps this is an untested edge case and that is the behavior when no key pairs are available. Perhaps an interface returns null or an exception/return code instead of opening the proper dialog, and the error is not handled in a useful way

I copied the client certificate to my smartphone. I then installed the certificate using the "Install certificate from device storage" button. that button is in settings > Security & privacy > Other security settings > Install certificate from device storage.

When I open the URL in Chrome (the Android app), Chrome asks me which certificate to use. Only one certificate is installed, namely the one for bitwarden. When I select that certificate, the Vaultwarden web vault appears in the browser.

Since Chrome was able to use the certificate, I assumed I had installed it correctly.
That's why I assumed the button was the problem.

This certificate was already installed when I tried to use the button in the bitwarden app.

(My smartphone is not set to English, so the path in the settings may be different.)




> It sounds (based on the presence of "-java" in the error) that it's an improperly exported Java Keystore? Perhaps a raw JKS file?
> 
> I expect the application may want either PKCS12 or x509 format. It almost certainly doesn't support a JKS file
> 
> PKCS12 files (common extensions pfx, p12) are single password protected files with a standard/well-defined format that contain multiple files within them- certs, private keys, CRLs. The files inside should be x509 files
> 
> If the app has separate boxes for client certificate, client private key, and CA, it almost certainly wants x509 format, one for certificate, one for private key, one for CA. Optionally, it may accept the certificate and private key combined into PEM encoded x509, concatenated into one file. The CA should still be separate
> 
> If you have your certs/keys in a JKS, export them using keytool since JKS is not a standard format recognized outside of Java (Google "export jks x509" or "export jks x509 pem")
> 
> Note: Creation of a PKCS12 file _will_ require you to specify a password when creating it and loading it into the app- it's required by the standard afaik. In that case, you should use the same password when loading it as when you created it. The private key inside the PKCS12 does not need a password, though
> 
> Remember to include the CA file, certificate and private key all in one PKCS12 (if using PKCS12) and do NOT create multiple PKCS12 files
> 
> This is my best guess, having not seen or tested the app yet. Hopefully it's helpful and I haven't misunderstood your issue

My client certificate file is in PCKS#12 format. The certificate is generated and signed by my PFsense Firewall. In PFsense there are two options to download the certificate.
One of the options is: "Export PCKS#12 archive without encryption". This is the option I use to create a certificate without a password.

The certificate without password can be installed within the system certificate store without any problem.

biscuit-316:
> Pkcs#12 is only supported format but it doesn't support chain of multiple certs in the file. i already tried to import with CA but didn't work. Hence, i had to register CA system wide.
> i used openssl to generate pkcs#12. with that, i have been using the feature flawlessy over a month.
> i tested importing from the system on emulator. Haven't tried on a real device.

The client certificate I used was in the format PKCS#12.
The certificate also worked with this version of the bitwarden app. When I use the "Import Certificate" button, and the certificate is password protected.

My client certificate is generated and signed by my PFsense Firewall. That's why I couldn't tell you what kind of files are in there.

sandstormkeshav:
Hi @oguzhane is there anything else that needs to be done on this PR? Can't wait to see this get merged in!

dbosompem:
Hi All, looks like lots of or community members were interested to see this feature being implemented, but we have some upcoming changes coming soon for our mobile app which is pretty significant, and unfortunately this contribution doesn’t align with that direction, now. What we are working on could fix this, or we could possibly revisit this feature once that is done. Unfortunately this PR would have to be closed. Thank you!

einot:
This is really sad to see. I was really hoping for this feature as it would greatly enhance the security of the system.

yarete03:
It may sounds like a silly question. Is there a way to run in iOS this client version in order to be able to implement the cert feature?

mzpqnxow:
May I ask why this was closed? Is it no longer being considered as a feature?

kuolemaaa:
> May I ask why this was closed? Is it no longer being considered as a feature?
> 
> EDIT: Nevermind, I see that the reason was because of migration to a new framework (though I don't necessarily see why this should be closed as a result...)

is a different programming language not enough for such a reason?

PR #2647, opened 2023-07-27
[PM-3165] Fix custom block URI dialog cancellation

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix custom block URI dialog cancellation to finish its internal task and not block the button.

## Code changes

* **DeviceActionService:** Added `BasicDialogWithResultCancelListener` to handle dialog cancellation and set the return task value to `null` so that the task gets finished when the user taps in the background to cancel the opened dialog.

## Before you submit

### DIFF
--- a/src/Android/Services/DeviceActionService.cs
+++ b/src/Android/Services/DeviceActionService.cs
@@ -3,7 +3,6 @@ using System.Threading.Tasks;
 using Android.App;
 using Android.Content;
 using Android.Content.PM;
-using Android.Media;
 using Android.Nfc;
 using Android.OS;
 using Android.Provider;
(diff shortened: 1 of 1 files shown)


src/Android/Services/DeviceActionService.cs
```diff
@@ -267,6 +261,7 @@ bool IsAlive(Java.Lang.Object @object)
             
             var result = new TaskCompletionSource<ValidatablePromptResponse?>();
 
+            alertBuilder.SetOnCancelListener(new BasicDialogWithResultCancelListener(result));
```
mpbw2, line 264: Just curious, why not just use the same `(sender, args) => result.TrySetResult(null)` as the negative listener?  Is it preventing the dialog from dismissing properly?

fedemkr: The negative listener is for the "cancel" button of the dialog and the `SetOnCancelListener` is for when the user taps in the background to dismiss the dialog. Currently, there is no `EventHandler` property to use the form `(sender, args) => result.TrySetResult(null)` so that's why I implemented a custom class for the listener.

mpbw2 (approved):

PR #2653, opened 2023-07-31
[PM-3102] Update Master password reprompt to be based on MP instead of Key Connector

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Update master password reprompt to be based on `HasMasterPassword` from the decrypt options instead of the Key connector. However, still use key connector as a default in case it's an old state.

Based on client's change https://github.com/bitwarden/clients/commit/87468a2aa65f500d5f2a24637296da91c1aa65ee

## Code changes

* **UserVerificationService:** Added `HasMasterPasswordAsync()` to check whether the user has a MP set up defaulting to check the key connector.
* **PasswordRepromptService:** Removed `Enabled()` given that it's not being used.
* **Others:** Updated all the references to `IKeyConnectorService.GetUsesKeyConnector()` to the new check on the `IUserVerificationService`, inverting the bool check given that it's the opposite now.

## Before you submit


andrebispo5 (approved):
LGTM 👍

andrebispo5 (approved):

PR #2655, opened 2023-08-02
[PM-1208] Add TDE flows for new users

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add necessary login flows to new users after SSO login using TDE.  

## Code changes

**src/App/Pages/Accounts/LockPageViewModel.cs:**
Log user out if they have a pending auth request or if they have TDE enabled and don't have a master password nor PIN or Biometrics

**src/App/Pages/Accounts/LoginSsoPageViewModel.cs:**
Auto check if the users pending auth request has been accepted.

**src/Core/Services/PasswordResetEnrollmentService.cs**
Service containing new user org auto enrolment.

## Screenshots

New user first time login:

https://github.com/bitwarden/mobile/assets/4648522/08c3e558-f290-4d95-b9f6-9873593af2aa

New user after auto enrolment:

https://github.com/bitwarden/mobile/assets/4648522/6a3104f9-0da5-4d84-ae12-cb9872668895

## Before you submit


src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -209,17 +209,12 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
+                    return;
                 }
-                else if (response.ResetMasterPassword)
-                {
-                    StartSetPasswordAction?.Invoke();
-                }
-                else if (response.ForcePasswordReset)
-                {
-                    UpdateTempPasswordAction?.Invoke();
-                }
```
jlf0dev, line 220: Do we need to leave these to support backwards compatibility? @JaredSnider-Bitwarden did we do that in the clients?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -235,16 +230,51 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
+                        var authRequest = await _authService.GetPasswordlessLoginRequestByIdAsync(pendingRequest.Id);
+                        if (authRequest != null && authRequest.RequestApproved != null && authRequest.RequestApproved.Value)
+                        {
+                            var authResult = await _authService.LogInPasswordlessAsync(await _stateService.GetActiveUserEmailAsync(), authRequest.RequestAccessCode, pendingRequest.Id, pendingRequest.PrivateKey, authRequest.Key, authRequest.MasterPasswordHash);
+                            if (authResult == null && await _stateService.IsAuthenticatedAsync())
+                            {
+                                await Xamarin.Essentials.MainThread.InvokeOnMainThreadAsync(
+                                 () => _platformUtilsService.ShowToast("info", null, AppResources.LoginApproved));
+                                await _stateService.SetPendingAdminAuthRequestAsync(null);
+                                var task = Task.Run(async () => await _syncService.FullSyncAsync(true));
```
jlf0dev, line 244: `task` isn't used here
```suggestion
_syncService.FullSyncAsync(true).FireAndForget();
```

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -235,16 +230,51 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
                         _syncService.FullSyncAsync(true).FireAndForget();
                         SsoAuthSuccessAction?.Invoke();
                     }
+                    else if (pendingRequest != null)
+                    {
+                        var authRequest = await _authService.GetPasswordlessLoginRequestByIdAsync(pendingRequest.Id);
+                        if (authRequest != null && authRequest.RequestApproved != null && authRequest.RequestApproved.Value)
+                        {
+                            var authResult = await _authService.LogInPasswordlessAsync(await _stateService.GetActiveUserEmailAsync(), authRequest.RequestAccessCode, pendingRequest.Id, pendingRequest.PrivateKey, authRequest.Key, authRequest.MasterPasswordHash);
```
jlf0dev, line 238: Does this also clear out the existing `pendingAdminAuthRequest`?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
JaredSnider-Bitwarden: Yes, in the existing flow, those can both still happen and should remain.  

Note: we replaced the usage of `ResetMasterPassword` with:
```TypeScript
const requireSetPassword =
        !acctDecryptionOpts.hasMasterPassword &&
        acctDecryptionOpts.keyConnectorOption === undefined;
 ```
 
 For the TDE routing path, we have to handle 3 routing scenarios (code taken from `sso.component.ts` in clients):
 ```TypeScript
   private async handleTrustedDeviceEncryptionEnabled(
    authResult: AuthResult,
    orgIdentifier: string,
    acctDecryptionOpts: AccountDecryptionOptions
  ): Promise<void> {
    // If user doesn't have a MP, but has reset password permission, they must set a MP
    if (
      !acctDecryptionOpts.hasMasterPassword &&
      acctDecryptionOpts.trustedDeviceOption.hasManageResetPasswordPermission
    ) {
      // Change implies going no password -> password in this case
      return await this.handleChangePasswordRequired(orgIdentifier);
    }

    if (authResult.forcePasswordReset !== ForceResetPasswordReason.None) {
      return await this.handleForcePasswordReset(orgIdentifier);
    }

    this.navigateViaCallbackOrRoute(
      this.onSuccessfulLoginTdeNavigate,
      // Navigate to TDE page (if user was on trusted device and TDE has decrypted
      //  their user key, the login-initiated guard will redirect them to the vault)
      [this.trustedDeviceEncRoute]
    );
  }
 ```

File: src/App/Pages/Accounts/LoginSsoPageViewModel.cs (hunk 3 above)
Comment (andrebispo5) [reply src/App/Pages/Accounts/LoginSsoPageViewModel.cs:238]: it's in line 243 👍 `await _stateService.SetPendingAdminAuthRequestAsync(null);`

File: src/App/Pages/Accounts/LoginSsoPageViewModel.cs (hunk 1 above)
Comment (andrebispo5) [reply src/App/Pages/Accounts/LoginSsoPageViewModel.cs:220]: just need to update the condition because i was looking for the KeyConnectorOption instead of the hasManageResetPasswordPermission

Comment (JaredSnider-Bitwarden) [reply src/App/Pages/Accounts/LoginSsoPageViewModel.cs:220]: oh, great catch. Didn't realize that we needed to support that for new clients. Wasn't thinking about new clients + older servers. I'll fix that on clients

**Review by jlf0dev — APPROVED** [review]

File: src/App/Pages/Accounts/LoginSsoPageViewModel.cs [hunk 4]
```diff
@@ -235,18 +230,53 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
                         StartDeviceApprovalOptionsAction?.Invoke();
                     }
+                    return;
                 }
-                else
+
+                // In the standard, non TDE case, a user must set password if they don't
+                // have one and they aren't using key connector.
+                // Note: TDE & Key connector are mutually exclusive org config options.
+                var requireSetPassword = !decryptOptions?.HasMasterPassword ?? false && decryptOptions?.KeyConnectorOption == null;
```
fedemkr, line 263: ❓ Is there a possibility where `decryptOptions` is `null` but the user is using `Key connector`? Or that can't happen on login with the new changes?

⛏️ Given that the information expert is the `DecryptOptions` object, can this be moved inside the `DecryptOptions` with a property `RequireSetPassword`? so then you can directly use `decryptOptions?.RequireSetPassword`

src/Core/Abstractions/IPasswordResetEnrollmentService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Bit.Core.Abstractions
+{
+    public interface IPasswordResetEnrollmentService
+    {
+        Task EnrollIfRequired(string organizationSsoId);
+        Task Enroll(string organizationId);
```
fedemkr, line 9: ⛏️  Add `Async` suffix

src/Core/Services/AuthService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
fedemkr, line 4: ❓ Is this being used?

src/Core/Services/AuthService.cs
```diff
@@ -694,5 +699,22 @@ private async Task<PasswordlessLoginResponse> PopulateFingerprintPhraseAsync(Pas
             passwordlessLogin.FingerprintPhrase = string.Join("-", await _cryptoService.GetFingerprintAsync(userEmail, CoreHelpers.Base64UrlDecode(passwordlessLogin.PublicKey)));
             return passwordlessLogin;
         }
+
+        public async Task CreateNewSsoUserAsync(string organizationSsoId)
+        {
+            var orgAutoEnrollStatusResponse = await _apiService.GetOrganizationAutoEnrollStatusAsync(organizationSsoId);
+            var randomBytes = _cryptoFunctionService.RandomBytes(64);
+            var userKey = new UserKey(randomBytes);
+            var keyPair = await _cryptoService.MakeKeyPairAsync(userKey);
```
fedemkr, line 708: ⛏️ In order to avoid having `Item1` and `Item2`, deconstruct the tuple to have more meaningful names.

src/Core/Services/PasswordResetEnrollmentService.cs
```diff
@@ -0,0 +1,62 @@
...
+            _cryptoService = cryptoService;
+            _organizationService = organizationService;
+            _stateService = stateService;
+        }
+
+        public async Task EnrollIfRequired(string organizationSsoId)
+        {
+            var orgAutoEnrollStatusResponse = await _apiService.GetOrganizationAutoEnrollStatusAsync(organizationSsoId);
+
+            if (!orgAutoEnrollStatusResponse.ResetPasswordEnabled)
```
fedemkr, line 32: ❓ Can `orgAutoEnrollStatusResponse` be `null`?

src/Core/Services/PasswordResetEnrollmentService.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                await this.Enroll(orgAutoEnrollStatusResponse.Id);
```
fedemkr, line 34: ⛏️ Remove `this`

src/Core/Services/PasswordResetEnrollmentService.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+
+        public async Task Enroll(string organizationId)
+        {
+            var orgKeyResponse = await _apiService.GetOrganizationKeysAsync(organizationId);
+            if (orgKeyResponse == null)
+            {
+                throw new Exception("Organization keys missing");
```
fedemkr, line 43: ❓ Will this need to be seen by the user with some custom message or the generic "An error has occurred is fine"?

fedemkr (changes requested):

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -172,6 +172,22 @@ public async Task InitAsync()
...
+                return;
+            }
+            if (decryptOptions.TrustedDeviceOption != null && await _stateService.IsAuthenticatedAsync())
+            {
+                if (!decryptOptions.HasMasterPassword && !(BiometricEnabled || PinEnabled))
+                {
+                    await _vaultTimeoutService.LogOutAsync();
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
andrebispo5: Fixed other bullet points but I can't find anywhere else where this is being used: 
`!decryptOptions.HasMasterPassword && !(BiometricEnabled || PinEnabled)`

src/Core/Services/PasswordResetEnrollmentService.cs
andrebispo5: This is just for dev use, a generic message is fine for the user.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
jlf0dev: Yes that is a possibility, I forgot we also need backwards compatibility for key connector, `tokenResponse.KeyConnectorUrl`. I also don't see the `_keyConnectorService.GetAndSetKey()` call, am I just missing that?

andrebispo5: Fixed by adding code for new user sso in auth service

jlf0dev (approved):
LGTM, @fedemkr will probably want to take another look as well

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -173,6 +180,14 @@ public async Task InitAsync()
 
             BiometricEnabled = await _vaultTimeoutService.IsBiometricLockSetAsync() && await _cryptoService.HasEncryptedUserKeyAsync();
 
+            var decryptOptions = await _stateService.GetAccountDecryptionOptions();
+            if (await _stateService.IsAuthenticatedAsync() && decryptOptions?.TrustedDeviceOption != null
+                && (!decryptOptions?.HasMasterPassword ?? false && !(BiometricEnabled || PinEnabled)))
```
fedemkr, line 185: ⛏️ When you are checking `decryptOptions?.TrustedDeviceOption != null` it means that after that `decryptOptions` won't be `null` so no need to check it again afterwards so `!decryptOptions?.HasMasterPassword ?? false` can be just `!decryptOptions.HasMasterPassword`

Also no need for the parenthesis in 
`&& (!decryptOptions.HasMasterPassword && !(BiometricEnabled || PinEnabled)))`.

Finally, using De Morgan's law you can invert the biometric/pin enabled to avoid the parenthesis as well:
```suggestion
            if (await _stateService.IsAuthenticatedAsync() 
                && decryptOptions?.TrustedDeviceOption != null
                && !decryptOptions.HasMasterPassword 
                && !BiometricEnabled 
                && !PinEnabled)
```
♻️  It's a big if, maybe extracting this into a method would clarify the reading.

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -235,18 +230,52 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
                         StartDeviceApprovalOptionsAction?.Invoke();
                     }
+                    return;
                 }
-                else
+
+                // In the standard, non TDE case, a user must set password if they don't
+                // have one and they aren't using key connector.
+                // Note: TDE & Key connector are mutually exclusive org config options.
+                if (response.ResetMasterPassword || decryptOptions.RequireSetPassword)
```
fedemkr, line 263: ⛏️ `decryptOptions` can be `null`

fedemkr (changes requested):

src/App/Pages/Accounts/LockPageViewModel.cs
andrebispo5: Refactored the condition but didn't extracted it to a method because couldn't get a good name for it, it's just what it is, authed with tde, no password, has no biometrics nor pin 😅

fedemkr (approved):

PR #2667, opened 2023-08-07
[PM-2713] Final merge from Key Migration branch to TDE Feature branch

jlf0dev:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Merges the final changes from the key migration to our TDE feature branch. See #2633 for background on key migration.

Most of these changes are bug fixes and refactors. There may be merge conflicts, but the goal here is to get these changes reviewed before getting lost in our feature branch.

## Code changes

## Before you submit


src/Core/Services/CryptoService.cs
```diff
@@ -157,10 +163,26 @@ public async Task<UserKey> DecryptUserKeyWithMasterKeyAsync(MasterKey masterKey,
...
+                        throw new Exception("No encrypted user key nor old encKeyEncrypted found");
+                    }
+
+                    var userKey = await DecryptUserKeyWithMasterKeyAsync(
+                        masterKey,
+                        new EncString(oldEncUserKey),
+                        userId
+                    );
+                    await SetMasterKeyEncryptedUserKeyAsync(oldEncUserKey, userId);
+                    return userKey;
```
fedemkr, line 183: 🤔  @jlf0dev I may have forgotten here to set the `encKey` to `null` after the migration succeeds. What do you think? Shall we maintain the `encKey` just in case, or just reset it before `return userKey`?

src/Core/Services/CryptoService.cs
```diff
@@ -670,29 +692,39 @@ public async Task<EncByteArray> EncryptToBytesAsync(byte[] plainValue, Symmetric
...
-                await _stateService.SetUserKeyAutoUnlockAsync(userKey.KeyB64, userId);
+                await _stateService.SetUserKeyAutoUnlockAsync(userKey, userId);
             }
             else
             {
                 await _stateService.SetUserKeyAutoUnlockAsync(null, userId);
             }
+
+            // Set, refresh, or clear the biometric unlock key
+            if ((await _stateService.GetBiometricUnlockAsync(userId)).GetValueOrDefault())
```
fedemkr, line 717: ⛏️ There were some internal discussions about what to do on nullable bools and most comments pointed to using `is true`, so I'd change that here:
```suggestion
            if (await _stateService.GetBiometricUnlockAsync(userId) is true)
```
Also, if possible I'd change a bit the naming of the method to something like `GetCanBiometricUnlockAsync(...)` (`can`, `allow`, `has`, what makes most sense) to represent the intent clearer. It was a bit confusing to me what `get biometric unlock` means.

src/Core/Services/CryptoService.cs
```diff
@@ -956,23 +988,33 @@ private class EncryptedObject
...
                 new EncString(encryptedUserKey),
                 userId);
+
             // Migrate
-            await _stateService.SetUserKeyAutoUnlockAsync(userKey.KeyB64, userId);
+            if (await _stateService.GetVaultTimeoutAsync(userId) == null)
+            {
+                await _stateService.SetUserKeyAutoUnlockAsync(userKey, userId);
+            }
+            if ((await _stateService.GetBiometricUnlockAsync(userId)).GetValueOrDefault())
```
fedemkr, line 1012: ⛏️ Same as before:
```suggestion
            if (await _stateService.GetBiometricUnlockAsync(userId) is true)
```

src/Core/Services/KeyConnectorService.cs
```diff
@@ -24,11 +24,11 @@ public class KeyConnectorService : IKeyConnectorService
...
-        public async Task GetAndSetKey(string url)
+        public async Task GetAndSetMasterKeyAsync(string url)
         {
             try
             {
-                var masterKeyResponse = await _apiService.GetMasterKeyFromKeyConnector(url);
+                var masterKeyResponse = await _apiService.GetMasterKeyFromKeyConnectorAsync(url);
                 var masterKeyArr = Convert.FromBase64String(masterKeyResponse.Key);
                 var masterKey = new MasterKey(masterKeyArr);
                 await _cryptoService.SetMasterKeyAsync(masterKey);
```
fedemkr, line 34: ⛏️ I see that in some places the suffix `...Bytes` it's being used and in others `...Arr`. We should be consistent on this and IMO `bytes` represents better what the variable type is (it's not an array of objects but a specific one of `byte`s).

⛏️ I'm not entirely convinced by the method name, given that the method doesn't return anything so `Get...` doesn't feel well to me. Maybe `UpdateMasterKeyFromUrlAsync(url)` or `UpdateMasterKeyFromAsync(url)` could be better, what do you think?

🤔 Sorry I can't get the catch section highlighted, but does the user need a special message (and localized) to be shown or is the generic one ok? Asking this because a generic `Exception` is being thrown.

fedemkr (changes requested):

src/Core/Services/CryptoService.cs
jlf0dev: Great callout! I wish I had time to refactor the biometric system a little more, but I'm worried about introducing more bugs this late. From what I've gathered:
- `BiometricUnlock` - Do we have biometrics enabled?
- `BiometricLocked` - Is our vault currently locked and requires biometrics to unlock?
- `BiometricUnlockKey` - What I've added, the biometric key

andrebispo5 (approved):
LGTM

PR #2668, opened 2023-08-07
[PM-3319] Check for admin request in Lock page

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

After requesting approval from admin or other device, trying to approve device with Master Password will not work.

## Code changes

Add parameter to ignore pending requests check in `LockPage`

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LockPageViewModel.cs
+++ b/src/App/Pages/Accounts/LockPageViewModel.cs
@@ -48,7 +48,6 @@ namespace Bit.App.Pages
         private string _biometricButtonText;
         private string _loggedInAsText;
         private string _lockedVerifyText;
-        private bool _checkPendingAuthRequests;
 
         public LockPageViewModel()
         {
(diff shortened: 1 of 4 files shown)


src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -149,6 +150,12 @@ public string LockedVerifyText
             set => SetProperty(ref _lockedVerifyText, value);
         }
 
+        public bool CheckPendingAuthRequests
+        {
+            get => _checkPendingAuthRequests;
+            set => SetProperty(ref _checkPendingAuthRequests, value);
+        }
+
```
fedemkr, line 158: ⛏️ There is no need for a property that notify changes. You can use a common property here without explicit backing field:
```suggestion
        public bool CheckPendingAuthRequests { get; set; }
```

fedemkr (changes requested):

src/App/Pages/Accounts/LockPageViewModel.cs
andrebispo5: yup indeed

fedemkr (approved):

PR #2670, opened 2023-08-07
[PM-2289] Change header title based on auth request type

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Change header title based on auth request type.

## Code changes

PageTitle is set after AuthRequestType so the label changes accordingly. 

## Screenshots

image
image

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
+++ b/src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
@@ -21,7 +21,6 @@ namespace Bit.App.Pages
             _vm.Page = this;
             _vm.Email = email;
             _vm.AuthRequestType = authRequestType;
-            _vm.PageTitle = _vm.HeaderTitle;
             _vm.StartTwoFactorAction = () => Device.BeginInvokeOnMainThread(async () => await StartTwoFactorAsync());
             _vm.LogInSuccessAction = () => Device.BeginInvokeOnMainThread(async () => await LogInSuccessAsync());
             _vm.UpdateTempPasswordAction = () => Device.BeginInvokeOnMainThread(async () => await UpdateTempPasswordAsync());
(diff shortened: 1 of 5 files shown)


src/App/Resources/AppResources.resx
```diff
@@ -2756,5 +2756,7 @@ Do you want to switch to this account?</value>
   <data name="LogInWithDeviceMustBeSetUpInTheSettingsOfTheBitwardenAppNeedAnotherOption" xml:space="preserve">
     <value>Log in with device must be set up in the settings of the Bitwarden app. Need another option?</value>
   </data>
-
+  <data name="LogInWithDevice" xml:space="preserve">
+    <value>Log In With Device</value>
+  </data>
```
fedemkr, line 2761: ⛏️ We need to make this consistent with `Log in initiated` which has its first letter in capital whereas with device has all words capitalized. I guess `Log in initiated` is the right approach but check just in case.

src/App/Pages/Accounts/LoginPasswordlessRequestPage.xaml.cs
```diff
@@ -21,6 +21,7 @@ public LoginPasswordlessRequestPage(string email, AuthRequestType authRequestTyp
             _vm.Page = this;
             _vm.Email = email;
             _vm.AuthRequestType = authRequestType;
+            _vm.PageTitle = _vm.HeaderTitle;
```
fedemkr, line 24: ⛏️ In order not to know the order in which these properties need to be set, you can move this to the `AuthRequestType` -> `set`, so after setting the backing field you can set the `PageTitle` and it reduces potential issues because of reordering properties assignation.

fedemkr (changes requested):

fedemkr (approved):

PR #2673, opened 2023-08-08
[PM-3341] Vault Timeout Action not persisted correctly

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Vault Timeout Action not persisted correctly when a user navigated out and then returned to settings.

## Code changes

Simple condition change.
Added action reset to Pin turning off.

## Before you submit


mpbw2 (dismissed):

src/App/Pages/Settings/SettingsPage/SettingsPageViewModel.cs
```diff
@@ -130,9 +130,12 @@ public async Task InitAsync()
...
 
-            var savedVaultTimeoutAction = await _vaultTimeoutService.GetVaultTimeoutAction();
-            var timeoutAction = savedVaultTimeoutAction ?? VaultTimeoutAction.Lock;
-            if (!_hasMasterPassword && timeoutAction == VaultTimeoutAction.Lock)
+
+            var pinSet = await _vaultTimeoutService.GetPinLockTypeAsync();
+            _pin = pinSet != PinLockType.Disabled;
+            _biometric = await _vaultTimeoutService.IsBiometricLockSetAsync();
+            var timeoutAction = await _vaultTimeoutService.GetVaultTimeoutAction() ?? VaultTimeoutAction.Lock;
+            if (!IsVaultTimeoutActionLockAllowed)
```
fedemkr, line 138: 🤔 I see that before it was also being considered the timeout action to be `Lock`, is that not needed anymore? Because if action is `Logout` then inside the if, it's going to reassign `Logout` and set it in the service, which I don't think it would be necessary.

andrebispo5: Yup you are right we can check for Lock to run the code inside and avoid unnecessary SetVaultTimeout

fedemkr (approved):

PR #2683, opened 2023-08-12
[PM-3381] Update login flows

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix TDE SSO login flows for users without a master password. 

## Code changes

Update code to match the logic in SSO Login Page

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/TwoFactorPageViewModel.cs
+++ b/src/App/Pages/Accounts/TwoFactorPageViewModel.cs
@@ -39,7 +39,6 @@ namespace Bit.App.Pages
         private TwoFactorProviderType? _selectedProviderType;
         private string _totpInstruction;
         private string _webVaultUrl = "https://vault.bitwarden.com";
-        private bool _authingWithSso = false;
         private bool _enableContinue = false;
         private bool _showContinue = true;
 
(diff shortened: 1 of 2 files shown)


src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -356,18 +348,34 @@ public async Task SubmitAsync(bool showLoading = true)
...
                         StartDeviceApprovalOptionsAction?.Invoke();
                     }
+                    return;
                 }
-                else
+
+                // In the standard, non TDE case, a user must set password if they don't
+                // have one and they aren't using key connector.
+                // Note: TDE & Key connector are mutually exclusive org config options.
+                if (_authingWithSso && (result.ResetMasterPassword || (decryptOptions?.RequireSetPassword ?? false)))
```
jlf0dev, line 365: Do we need to check `_authingWithSso` here? I don't think it matters if you're using sso to login or not, we'll still need to set your password if these are true.

Does the server have the logic to only send `ResetMasterPassword` or `decryptOptions.RequireSetPassword` if the user isn't using Key Connector? I just want to make sure we don't need that logic here.

jlf0dev (changes requested):

jlf0dev (approved):
👍

PR #2686, opened 2023-08-13
[PM-3394] Fix login with device for passwordless approvals

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If the approving device doesn't have a master key, users should still be able to use Login with Device. This adds a token request in that instance, and sets the user key after.

While investigating this, I discovered that we're not clearing the active user id when adding a new account. This is problematic because `isAuthenticated` simply checks if the user id is `null`.
**Edit:** I originally fixed this by setting the active user Id to `null` but we can't do that for autofill purposes. Instead I am passing the `AuthRequestType` to any auth request logic so we know whether to use the Admin Approval logic or Unlock and Authenticate.

I also discovered that our background script to log out inactive accounts was attempting to set the user key if the vault timeout was set to 'never'. This change prevents that from happening as well, simply just returning if we have an auto key without using it.

## Code changes

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
@@ -243,7 +243,7 @@ namespace Bit.App.Pages
                     response = await _authService.GetPasswordlessLoginResquestAsync(_requestId, _requestAccessCode);
                 }
 
-                if (!response?.RequestApproved ?? true)
+                if (!response?.RequestApproved != true)
                 {
                     return;
                 }
--- a/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
@@ -233,13 +233,13 @@ namespace Bit.App.Pages
             try
             {
                 PasswordlessLoginResponse response = null;
-                if (_authRequestType == AuthRequestType.AdminApproval)
-                {
-                    response = await _authService.GetPasswordlessLoginRequestByIdAsync(_requestId);
-                }
-                else if (_authRequestType == AuthRequestType.AuthenticateAndUnlock)
+                if (_authRequestType == AuthRequestType.AuthenticateAndUnlock)
                 {
                     response = await _authService.GetPasswordlessLoginResquestAsync(_requestId, _requestAccessCode);
+                }
+                else
+                {
+                    response = await _authService.GetPasswordlessLoginRequestByIdAsync(_requestId);
                 }
 
                 if (!(response?.RequestApproved ?? false))
--- a/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
@@ -233,11 +233,11 @@ namespace Bit.App.Pages
             try
             {
                 PasswordlessLoginResponse response = null;
-                if (await _stateService.IsAuthenticatedAsync())
+                if (_authRequestType == AuthRequestType.AdminApproval)
                 {
                     response = await _authService.GetPasswordlessLoginRequestByIdAsync(_requestId);
                 }
-                else
+                else if (_authRequestType == AuthRequestType.AuthenticateAndUnlock)
                 {
                     response = await _authService.GetPasswordlessLoginResquestAsync(_requestId, _requestAccessCode);
                 }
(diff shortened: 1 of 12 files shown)


src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -242,7 +242,7 @@ private async Task CheckLoginRequestStatus()
                     response = await _authService.GetPasswordlessLoginResquestAsync(_requestId, _requestAccessCode);
                 }
 
-                if (response.RequestApproved == null || !response.RequestApproved.Value)
+                if (response == null || response.RequestApproved == null || !response.RequestApproved.Value)
```
andrebispo5, line 245: Can be rewritten as 
`if (!response?.RequestApproved ?? false)`

andrebispo5 (changes requested):

jlf0dev:
> as long as any account is logged in (which also includes logged-out via vault timeout) there should always be an active user id in order for external functions to work, like autofill

@mpbw2 I'm surprised so many things work on the login screen then, any state service calls would be using the wrong user. Nevertheless, I'm not sure how to tell if the user has an access token or not in this state...

mpbw2:
> @mpbw2 I'm surprised so many things work on the login screen then, any state service calls would be using the wrong user. Nevertheless, I'm not sure how to tell if the user has an access token or not in this state...

There may be some situations where the active user id is set mid-flow for complex logins, but I can't remember for sure.  However for approving with a device, shouldn't queries to `StateService` be made with the pushed `userId` vs the active one?  (push is actually one of the major reasons `userId` is an optional param throughout `StateService` since inbound payloads me be for a non-active account)

jlf0dev:
I couldn't use the user id from the push because the logic was before we set up the auth request, but I've exchanged it out with route based logic instead. 

We might need to consider changing this in the future and instead having an autofill/extension specific `ActiveUserId` that defaults to an account. That way we can use the state service when adding new accounts.

src/Core/Services/VaultTimeoutService.cs
```diff
@@ -69,7 +76,7 @@ public async Task<bool> IsLockedAsync(string userId = null)
             {
                 if (await _cryptoService.HasAutoUnlockKeyAsync(userId))
                 {
-                    await _cryptoService.SetUserKeyAsync(await _cryptoService.GetAutoUnlockKeyAsync(userId));
+                    await _cryptoService.SetUserKeyAsync(await _cryptoService.GetAutoUnlockKeyAsync(userId), userId);
                 }
                 else
                 {
```
fedemkr, line 82: 🎨 To avoid the else (Github doesn't let me grab the other part which returns), put the if the other way around:
```c#
if (!await _cryptoService.HasAutoUnlockKeyAsync(userId))
{
    return true;
}

await _cryptoService.SetUserKeyAsync(await _cryptoService.GetAutoUnlockKeyAsync(userId));
await _cryptoService.SetUserKeyAsync(await _cryptoService.GetAutoUnlockKeyAsync(userId), userId);
```

src/Core/Services/AuthService.cs
```diff
@@ -222,10 +223,13 @@ public async Task<AuthResult> LogInPasswordlessAsync(string email, string access
                 return null;
             }
 
+            // The approval device may not have a master key hash if it authenticated with a passwordless method
             if (string.IsNullOrEmpty(masterKeyHash) && decryptionKey != null)
             {
+                var authResult = await LogInHelperAsync(email, accessCode, null, null, null, null, null, null, null, null, null, authRequestId: authRequestId);
```
fedemkr, line 229: ♻️ Just putting it here to make a note, we should definitely look into refactoring `LogInHelperAsync` to avoid this type of usage, it has too many arguments. Maybe for a tech-debt task given that it doesn't affect this PR only, but several other flows as well.

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -243,7 +243,7 @@ private async Task CheckLoginRequestStatus()
                     response = await _authService.GetPasswordlessLoginResquestAsync(_requestId, _requestAccessCode);
                 }
 
-                if (!response?.RequestApproved ?? true)
+                if (!response?.RequestApproved != true)
```
fedemkr, line 246: ⛏️ remove the ! at beginning
```suggestion
                if (response?.RequestApproved != true)
```

fedemkr (changes requested):

fedemkr (approved):

PR #2694, opened 2023-08-16
[PM-3462] Handle force password reset on mobile with TDE

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When a TDE user had their password reset by an admin, the `UpdateTempPassword` page would show between the SSO login and Login Options screen. This meant that if their device wasn't trusted, they didn't have a User Key to create their new password with.

This sets a `ForcePasswordResetReason` during the SSO login so that it can be picked up later on the tabs page and removes that navigation from the SSO page.

Other changes
- This bug was hidden by the `EncryptUserKeyWithMasterKeyAsync` method creating a new User Key. Since this is too dangerous I've made it explicitly have to be created before passing in.
- Fixed a key connector bug where we were sending the private key instead of the user key to the server
- Deprecated `ResetMasterPassword` on `AuthResult`
- Added comments to the navigation logic on the SSO page

## Code changes

## Screenshots

https://github.com/bitwarden/mobile/assets/24985544/ed493b0e-81c1-4a7e-a618-14c2c2857b43

## Before you submit


src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -268,18 +276,14 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
                 // In the standard, non TDE case, a user must set password if they don't
                 // have one and they aren't using key connector.
                 // Note: TDE & Key connector are mutually exclusive org config options.
-                if (response.ResetMasterPassword || (decryptOptions?.RequireSetPassword ?? false))
+                if (response.ResetMasterPassword || (decryptOptions?.RequireSetPassword == true))
                 {
+                    // TODO: We need to look into how to handle this when Org removes TDE
+                    // Will we have the User Key by now to set a new password?
```
fedemkr, line 282: ❓ Is there a task for this or should it be part of this PR?

src/Core/Services/CryptoService.cs
```diff
@@ -152,9 +152,13 @@ public Task ClearMasterKeyAsync(string userId = null)
...
         }
 
-        public async Task<Tuple<UserKey, EncString>> EncryptUserKeyWithMasterKeyAsync(MasterKey masterKey)
+        public async Task<Tuple<UserKey, EncString>> EncryptUserKeyWithMasterKeyAsync(MasterKey masterKey, UserKey userKey = null)
         {
-            var userKey = await GetUserKeyAsync() ?? await MakeUserKeyAsync();
+            userKey ??= await GetUserKeyAsync();
+            if (userKey == null)
+            {
+                throw new Exception("No user key found");
```
fedemkr, line 160: ⛏️ If we need to spot this in AppCenter to add the appropriate fix I'd go for a custom exception like `UserAndMasterKeysNullException` was used in another PR so it can be easily spotted on AppCenter. We could have one exception `KeysNullException(types of key)` so it changes the message depending on the keys missing. Thoughts?

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
```diff
@@ -268,18 +276,14 @@ private async Task LogIn(string code, string codeVerifier, string orgId)
...
                     StartSetPasswordAction?.Invoke();
                     return;
                 }
 
-                if (response.ForcePasswordReset)
-                {
-                    UpdateTempPasswordAction?.Invoke();
-                    return;
-                }
```
fedemkr, line 281: 🤔 Is this ok to be removed for a user that belongs to an Org without TDE enabled? I think it'll be shown later at the main page given the ForcePasswordResetReason but wanted to ask just in case

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -326,33 +326,36 @@ public async Task SubmitAsync(bool showLoading = true)
...
+                        await TwoFactorAuthSuccessAsync();
+                        return;
                     }
-                    else
-                    {
-                        StartDeviceApprovalOptionsAction?.Invoke();
-                    }
+
+                    StartDeviceApprovalOptionsAction?.Invoke();
                     return;
```
fedemkr, line 359: ♻️ Most of this logic is exactly the same as in `LoginSSOViewModel`, could we refactor it so we reuse the code and when changes are needed (like in this case) we don't have to modify both places? If this is time sensitive, we could create a tech-debt task (can't remember if one exists already)

fedemkr (changes requested):

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
andrebispo5: Yup that is exactly the case.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
andrebispo5: Thought about it as well, but i think we are on a time crunch better to create a ticket for that

andrebispo5: Created PM-3502

src/App/Pages/Accounts/LoginSsoPageViewModel.cs
andrebispo5: @jlf0dev ?

jlf0dev: We have a followup task to look into TDE unenroll and how to migrate users. Right now it doesn't work but wanted to add a note

src/Core/Services/CryptoService.cs
andrebispo5: After talking with @jlf0dev we decided to add classes for each null key to be easier to track by reference in code.

fedemkr (approved):

PR #2707, opened 2023-08-18
[PM-3513] Show error SSO policy

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Show the api error when a user tries to Login with device but there is an SSO only policy enforced.

## Code changes

Catch Api exception and show internal error message.

## Screenshots

image

## Before you submit

### DIFF
--- a/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
+++ b/src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
@@ -278,9 +278,9 @@ namespace Bit.App.Pages
                     await HandleLoginCompleteAsync();
                 }
             }
-            catch (ApiException ex) when (ex?.Error?.StatusCode == System.Net.HttpStatusCode.BadRequest)
+            catch (ApiException ex) when (ex.Error?.StatusCode == System.Net.HttpStatusCode.BadRequest)
             {
-                HandleException(ex, ex?.Error?.Message);
+                HandleException(ex);
             }
             catch (Exception ex)
             {
@@ -348,16 +348,6 @@ namespace Bit.App.Pages
             _requestId = response.Id;
             _requestAccessCode = response.RequestAccessCode;
             _requestKeyPair = response.RequestKeyPair;
-        }
-
-        private void HandleException(Exception ex)
-        {
-            Xamarin.Essentials.MainThread.InvokeOnMainThreadAsync(async () =>
-            {
-                await _deviceActionService.HideLoadingAsync();
-                await _platformUtilsService.ShowDialogAsync(AppResources.GenericErrorMessage);
-            }).FireAndForget();
-            _logger.Exception(ex);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


jlf0dev (approved):

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -277,6 +278,10 @@ private async Task CheckLoginRequestStatus()
                     await HandleLoginCompleteAsync();
                 }
             }
+            catch (ApiException ex) when (ex?.Error?.StatusCode == System.Net.HttpStatusCode.BadRequest)
+            {
+                HandleException(ex, ex?.Error?.Message);
+            }
```
fedemkr, line 284: ⛏️ `ex` is never null and inside given that you've already checked `null` in the `when` condition, is not necessary again.
```suggestion
            catch (ApiException ex) when (ex.Error?.StatusCode == System.Net.HttpStatusCode.BadRequest)
            {
                HandleException(ex, ex.Error.Message);
            }
```

fedemkr (changes requested):

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
fedemkr: 🤔 @andrebispo5  Can't remember the reason, why are we not using the base `HandleException`? which already takes care of this.

andrebispo5: Didn't noticed this was using a private HandleException class, this must have been before we had the base method done. I'll just remove the private and let it use the base one.

src/App/Pages/Accounts/LoginPasswordlessRequestViewModel.cs
```diff
@@ -277,6 +278,10 @@ private async Task CheckLoginRequestStatus()
                     await HandleLoginCompleteAsync();
                 }
             }
+            catch (ApiException ex) when (ex.Error?.StatusCode == System.Net.HttpStatusCode.BadRequest)
+            {
+                HandleException(ex);
+            }
```
fedemkr, line 284: 🤔 Any `BadRequest` should handle the exception the same or only the ones that corresponds to SSO only policy enforced?

andrebispo5: I don't think we will have more bad requests other than this one but if the request is bad we don't want to continue to run the check over and over. I'm even leaning towards not having the StartCheck running again in general, it causes more harm that good.

fedemkr: Alright I'll approve then and we can revise the start check afterwards.

fedemkr (approved):

PR #2710, opened 2023-08-21
[PM-3545] Fix biometric unlock for autofill

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Extension code was not updated for the user to be able to unlock with biometrics using new UserKey model.

## Code changes

Set user key after biometric is valid.

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/264e2619-da43-4e13-afa9-9b8ff16060e7

## Before you submit

### DIFF
--- a/src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
+++ b/src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
@@ -368,18 +368,8 @@ namespace Bit.iOS.Core.Controllers
             if (success)
             {
                 var userKey = await _cryptoService.GetBiometricUnlockKeyAsync();
-                await SetUserKeyAndContinueAsync(userKey);
+                await SetKeyAndContinueAsync(userKey);
             }
-        }
-
-        private async Task SetUserKeyAndContinueAsync(UserKey key)
-        {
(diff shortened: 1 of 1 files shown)


src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
```diff
@@ -367,10 +367,21 @@ public async Task PromptBiometricAsync()
             await _stateService.SetBiometricLockedAsync(!success);
             if (success)
             {
-                DoContinue();
+                var userKey = await _cryptoService.GetBiometricUnlockKeyAsync();
+                await SetUserKeyAndContinueAsync(userKey);
             }
         }
 
+        private async Task SetUserKeyAndContinueAsync(UserKey key)
```
jlf0dev, line 375: There's a `SetKeyAndContinue` on line 403 we could use instead of creating a new one.

jlf0dev (changes requested):

src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
andrebispo5: You are right!

jlf0dev (approved):

PR #2713, opened 2023-08-23
[PM-3543] [PM-3607] Fix password re-prompt when editing and on autofill.

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix 2 bugs on password re-prompt. On autofill and editing a cipher, the app would crash.

## Code changes

Update code to use userkey.

## Before you submit

### DIFF
--- a/src/App/Services/MobilePasswordRepromptService.cs
+++ b/src/App/Services/MobilePasswordRepromptService.cs
@@ -1,6 +1,7 @@
 ﻿using System.Threading.Tasks;
 using Bit.App.Abstractions;
 using Bit.App.Resources;
+using Bit.App.Utilities;
 using Bit.Core.Abstractions;
 using Bit.Core.Enums;
 
@@ -10,11 +11,15 @@ namespace Bit.App.Services
     {
         private readonly IPlatformUtilsService _platformUtilsService;
         private readonly ICryptoService _cryptoService;
+        private readonly IStateService _stateService;
+        private readonly IStorageService _secureStorageService;
 
-        public MobilePasswordRepromptService(IPlatformUtilsService platformUtilsService, ICryptoService cryptoService)
+        public MobilePasswordRepromptService(IPlatformUtilsService platformUtilsService, ICryptoService cryptoService, IStateService stateService, IStorageService storageService)
         {
             _platformUtilsService = platformUtilsService;
             _cryptoService = cryptoService;
+            _stateService = stateService;
+            _secureStorageService = storageService;
         }
 
         public string[] ProtectedFields { get; } = { "LoginTotp", "LoginPassword", "H_FieldValue", "CardNumber", "CardCode" };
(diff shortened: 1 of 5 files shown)


src/App/Services/MobilePasswordRepromptService.cs
```diff
@@ -42,7 +47,37 @@ private async Task<bool> ValidatePasswordAsync(string password)
                 return false;
             };
 
-            return await _cryptoService.CompareAndUpdateKeyHashAsync(password, null);
+            var email = await _stateService.GetEmailAsync();
+            var kdfConfig = await _stateService.GetActiveUserCustomDataAsync(a => new KdfConfig(a?.Profile));
+            var masterKey = await _cryptoService.MakeMasterKeyAsync(password, email, kdfConfig);
+
+            var storedPasswordHash = await _cryptoService.GetMasterKeyHashAsync();
+            if (storedPasswordHash == null)
```
jlf0dev, line 55: We built a method in `clients` on the crypto service to `getOrDeriveMasterKey`. Can you copy [that](https://github.com/bitwarden/clients/blob/5665576147a4fbf30024f56358d424ff26a1a131/libs/common/src/platform/services/crypto.service.ts#L161-L169) logic here?

src/App/Services/MobilePasswordRepromptService.cs
```diff
@@ -42,7 +47,37 @@ private async Task<bool> ValidatePasswordAsync(string password)
...
+            {
+                var oldKey = await _secureStorageService.GetAsync<string>("oldKey");
+                if (masterKey.KeyB64 == oldKey)
+                {
+                    var localPasswordHash = await _cryptoService.HashMasterKeyAsync(password, masterKey, HashPurpose.LocalAuthorization);
+                    await _secureStorageService.RemoveAsync("oldKey");
+                    await _cryptoService.SetMasterKeyHashAsync(localPasswordHash);
```
jlf0dev, line 62: @mpbw2 Do you remember why we have this `oldKey`? It seems like we could probably remove support for it at this point.

jlf0dev (changes requested):

src/App/Services/MobilePasswordRepromptService.cs
mpbw2: @jlf0dev It was from a migration that took place before I joined, so I agree it's probably safe to let it go.

andrebispo5: Awesome 👍

jlf0dev (approved):

equetzal:
Hey any new on when are we expecting to receive the fix update? 
Idk about other users but lot of us are relying and paying Bitwarden and this feature it's really needed to be deployed ad soon as possible (as a hot fix not in the regular deploy schedule).

PR #2721, opened 2023-08-25
[PM-3640] don't clear key needed for bio/auto migration in pin migration

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When migrating the pin, we were clearing the old encryption key. Unfortunately, this key is needed for the bio/auto migration that could still happen later.

This fix checks for the existence of an old encrypted master key inside the pin migration, indicating that the bio/auto migration is still needed, and doesn't clear the old encryption key if so. I have also added a null check to the bio/auto migration.

This will need to be picked to `rc`

## Code changes

## Before you submit


andrebispo5 (approved):
LGTM

PR #2723, opened 2023-08-25
[PM-3606] TDE user with 2FA isn't able to autofill on iOS

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix autofill for users with TDE and no MP and 2FA enabled.

## Code changes

Logout user and send them to HomePage when they don't have MP nor Biometrics or PIN.

## Before you submit


src/App/Pages/Accounts/LoginSsoPage.xaml.cs
```diff
@@ -90,8 +90,7 @@ private void Close_Clicked(object sender, EventArgs e)
 
         private async Task StartTwoFactorAsync()
         {
-            RestoreAppOptionsFromCopy();
-            var page = new TwoFactorPage(true, _appOptions, _vm.OrgIdentifier);
+            var page = new TwoFactorPage(false, _appOptions, _vm.OrgIdentifier);
```
mpbw2, line 93: WIll removing this cause issues on this page during normal (non-autofill) flow?

andrebispo5: It had nothing to do with this change, I reverted it and applied another fix. 
When a user doesn't have a MP nor Biometric or PIN, he needs to go back to the HomePage and login again.

src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
```diff
@@ -60,6 +60,8 @@ public BaseLockPasswordViewController(IntPtr handle)
 
         public string BiometricIntegritySourceKey { get; set; }
 
+        public bool HasLoginMethod => _hasMasterPassword || _biometricEnabled || _pinEnabled;
```
mpbw2, line 63: maybe rename this to `HasLoginOrUnlockMethod` to cover bio and pin?

andrebispo5: Agreed

mpbw2 (dismissed):

fedemkr (commented):
🤔 Shouldn't this fix be added to the other extensions as well?

andrebispo5:
> 🤔 Shouldn't this fix be added to the other extensions as well?

You are totally right @fedemkr. pushed additional code for other extension projects.

fedemkr (changes requested):
🎨 The code is being repeated in all `LockPasswordViewController` where it could be all done in `BaseLockPasswordViewController`, move the behavior to the base class. If I'm correct, no code is needed on the children.

src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
```diff
@@ -54,12 +54,15 @@ public BaseLockPasswordViewController(IntPtr handle)
         public abstract UIBarButtonItem BaseSubmitButton { get; }
         public abstract Action Success { get; }
         public abstract Action Cancel { get; }
+        public Action LaunchHomePage;
```
fedemkr, line 57: ⛏️ Transform into property instead of field
```suggestion
        public Action LaunchHomePage { get; set; }
```

src/iOS.Autofill/CredentialProviderViewController.cs
```diff
@@ -223,6 +223,7 @@ public override void PrepareForSegue(UIStoryboardSegue segue, NSObject sender)
                     else if (navController.TopViewController is LockPasswordViewController passwordViewController)
                     {
                         passwordViewController.CPViewController = this;
+                        passwordViewController.LaunchHomePage = () => DismissViewController(false, () => LaunchHomePage()); ;
```
fedemkr, line 226: ⛏️ Remove double ";"
```suggestion
                        passwordViewController.LaunchHomePage = () => DismissViewController(false, () => LaunchHomePage());
```

src/iOS.Extension/LockPasswordViewController.cs
```diff
@@ -19,12 +19,20 @@ public LockPasswordViewController(IntPtr handle)
...
             base.ViewDidLoad();
             CancelButton.TintColor = ThemeHelpers.NavBarTextColor;
             SubmitButton.TintColor = ThemeHelpers.NavBarTextColor;
+
+            if (!HasLoginOrUnlockMethod)
+            {
+                // user doesn't have a login method
+                // needs to go to homepage and login again
+                LaunchHomePage?.Invoke();
+            }
```
fedemkr, line 35: 🤔 Can we change the inheritance to use the same base lock VC so this doesn't have to be repeated here? If not also transform the `LaunchHomePage` into a property. I have to do a refactor for all extensions so I can tackle changing the base there as well.

fedemkr (changes requested):

fedemkr (approved):

PR #2724, opened 2023-08-25
[PM-3222] Migration of data from LiteDB to shared pref storage

mpbw2:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

This change performs data migration out of LiteDB and into shared preferences.  It also changes the mobile storage service to default to saving anything new to shared pref storage, and requires specifying keys for LiteDB storage.

## Code changes

* **StateMigrationService:** Added a 5to6 migration block for global and account-specific state values
* **MobileStorageService.cs:** Added a set of LiteDB storage keys to check against when accessing storage.  Also added an inline litedb-to-pref migration on the getter to cover for values that can't be handled cleanly via StateMigration (`accountBiometricIntegrityValid_{userId}_{systemBioIntegrityState}`) and as a catch-all to prevent anything sneaking into liteDB during the transition.
* **BaseContentPage.cs:** Prevent saving last active time without a userId to prevent the orphaned value from re-appearing in storage
* **Constants.cs:** Adjusted key names and removed `NeverDomainsKey`
* **AppIdService:** Removed `anonymousAppId` getter since `appId` is already anonymous
* **CipherService:** Name changes to sync with `Constants` and removal of `SaveNeverDomainAsync` method
* **StateService:** Name changes to sync with `Constants` and removal of `NeverDomains` getter/setter

## Before you submit

### DIFF
--- a/src/App/Services/MobileStorageService.cs
+++ b/src/App/Services/MobileStorageService.cs
@@ -24,12 +24,6 @@ namespace Bit.App.Services
             Constants.SettingsKey(""),
         };
 
-        private bool IsLiteDbKey(string key)
-        {
-            return _liteDbStorageKeys.Any(key.StartsWith) ||
-                   _liteDbStorageKeys.Contains(key);
-        }
-
         public MobileStorageService(
             IStorageService preferenceStorageService,
             IStorageService liteDbStorageService)
@@ -45,19 +39,7 @@ namespace Bit.App.Services
                 return await _liteDbStorageService.GetAsync<T>(key);
             }
 
-            var result = await _preferencesStorageService.GetAsync<T>(key);
-            if (result == null)
-            {
-                // If result is expected in prefs but found in LiteDB, migrate it to prefs
-                result = await _liteDbStorageService.GetAsync<T>(key);
-                if (result != null)
-                {
-                    await _preferencesStorageService.SaveAsync(key, result);
-                    await _liteDbStorageService.RemoveAsync(key);
-                }
-            }
-
-            return result;
+            return await _preferencesStorageService.GetAsync<T>(key) ?? await TryMigrateLiteDbToPrefsAsync<T>(key);
         }
 
         public Task SaveAsync<T>(string key, T obj)
@@ -88,6 +70,26 @@ namespace Bit.App.Services
             {
                 disposablePrefService.Dispose();
             }
+        }
+
+        // Helpers
+
+        private bool IsLiteDbKey(string key)
+        {
+            return _liteDbStorageKeys.Any(key.StartsWith) ||
+                   _liteDbStorageKeys.Contains(key);
+        }
+
+        private async Task<T> TryMigrateLiteDbToPrefsAsync<T>(string key)
+        {
+            var currentValue = await _liteDbStorageService.GetAsync<T>(key);
+            if (currentValue != null)
+            {
+                await _preferencesStorageService.SaveAsync(key, currentValue);
+                await _liteDbStorageService.RemoveAsync(key);
+            }
+
+            return currentValue;
         }
     }
 }
--- a/src/App/Services/MobileStorageService.cs
+++ b/src/App/Services/MobileStorageService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Bit.Core;
 using Bit.Core.Abstractions;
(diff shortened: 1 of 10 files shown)


mpbw2:
I'm aware that `StateMigrationService` has grown into a monster, and I'm evaluating a path forward.  It's important that this migration happens sooner than later though.

src/App/Services/MobileStorageService.cs
```diff
@@ -44,29 +40,42 @@ public class MobileStorageService : IStorageService, IDisposable
...
+                // If result is expected in prefs but found in LiteDB, migrate it to prefs
+                result = await _liteDbStorageService.GetAsync<T>(key);
+                if (result != null)
+                {
+                    await _preferencesStorageService.SaveAsync(key, result);
+                    await _liteDbStorageService.RemoveAsync(key);
+                }
+            }
+
+            return result;
```
fedemkr, line 60: 🎨 Even though this is totally fine I like it more when the migration stuff is not done in the same place and we can avoid the comment as well given that the method name is self-explanatory, i.e.:
```c#
public async Task<T> GetAsync<T>(string key)
{
    // ...
    return await _preferencesStorageService.GetAsync<T>(key) ?? TryMigrateLiteDBToPrefsAsync(key);
}

private async Task TryMigrateLiteDBToPrefsAsync(string key)
{
    var currentValue = await _liteDbStorageService.GetAsync<T>(key);
    if (currentValue != null)
    {
        await _preferencesStorageService.SaveAsync(key, currentValue);
        await _liteDbStorageService.RemoveAsync(key);
    }
    return currentValue;
}
```
Also it's easier to read on future versions where the path of migration is less likely to happen.
Thoughts?

src/App/Services/MobileStorageService.cs
```diff
@@ -11,29 +12,24 @@ public class MobileStorageService : IStorageService, IDisposable
...
+            Constants.SendsKey(""),
+            Constants.PassGenHistoryKey(""),
+            Constants.SettingsKey(""),
         };
 
+        private bool IsLiteDbKey(string key)
+        {
+            return _liteDbStorageKeys.Any(key.StartsWith) ||
+                   _liteDbStorageKeys.Contains(key);
+        }
```
fedemkr, line 31: ❓ `key` will never be null, right?

src/Core/Services/StateMigrationService.cs
```diff
@@ -544,6 +547,296 @@ private class V5Keys
...
+                    await MoveToPrefsAsync<bool?>(V6Keys.ShouldConnectToWatchKey(userId));
+                    await MoveToPrefsAsync<string>(V6Keys.DeviceKeyKey(userId));
+                }
+            }
+
+            await RemoveValueAsync(Storage.LiteDb, Constants.StateKey);
+            await RemoveValueAsync(Storage.LiteDb, V6Keys.LastActiveTimeKey(""));
+
+            // Update stored version
+            await SetLastStateVersionAsync(5);
```
fedemkr, line 617: ⛏️ Shouldn't this be `6`?
```suggestion
            await SetLastStateVersionAsync(6);
```

src/Core/Services/StateMigrationService.cs
```diff
@@ -612,5 +905,17 @@ private IStorageService GetStorageService(Storage storage)
...
+        {
+            var value = await GetValueAsync<T>(Storage.LiteDb, key);
+            if (value == null)
+            {
+                return;
+            }
+
+            await SetValueAsync(Storage.Prefs, key, value);
+            await RemoveValueAsync(Storage.LiteDb, key);
+        }
```
fedemkr, line 919: 🤔 Could it happen that the migration at some point fails for some reason and the user starts interacting with the app, which means that the new key/value are going to be stored in `Prefs` and then when the app starts again the migration runs smoothly overriding the last `Prefs` values that the user added with the ones with the old `LiteDB`values? 
E.g. 
v1 of the app has Solarized theme.
v2 of the app has the migration but the theme fails to be migrated for some exception. The user thought changes the theme to Nord.
v3 of the app fixes the migration and when run the theme will be Solarized again which would be wrong I think.

So what I'm getting at is, shall we check first to get the value for the key in `Prefs` and only if that value is `null` set the new value?
```suggestion
        private async Task MoveToPrefsAsync<T>(string key)
        {
            var value = await GetValueAsync<T>(Storage.LiteDb, key);
            if (value == null)
            {
                return;
            }

            if (await GetValueAsync<T>(Storage.Prefs, key) == null)
            {
                await SetValueAsync(Storage.Prefs, key, value);
            }
            await RemoveValueAsync(Storage.LiteDb, key);
        }
```

fedemkr (changes requested):

fedemkr:
🤔  Regarding the monster migration service I think that having a Chain of responsibility where each version migration has its own file/class would be the best approach. Then we can just stack them in the chain and adding a new migration is as easy as creating a new class and adding it to the stack.

src/App/Services/MobileStorageService.cs
mpbw2: Correct!

mpbw2:
> 🤔 Regarding the monster migration service I think that having a Chain of responsibility where each version migration has its own file/class would be the best approach. Then we can just stack them in the chain and adding a new migration is as easy as creating a new class and adding it to the stack. Furthermore, we can make all inner things private for each migration class so it's not available from anywhere, not even from the same assembly as it's today.

This is exactly what I want to do. 👍

src/Core/Services/StateMigrationService.cs
mpbw2: Yikes, good catch.

fedemkr (approved):

PR #2738, opened 2023-08-30
[PM-3699] fix: Add bitwarden.eu to HTTPS only resources

AChep:
## Objective
Make sure that if a user accesses https://bitwarden.eu, it's done though a HTTPS connection.


AChep:
Sorry for the change in ending line -- it was GitHub's web editor, I swear :stuck_out_tongue:

PR #2743, opened 2023-09-03
[DEVOPS-1406] - UPDATE: all workflows

Eeebru:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [x] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

- Update all warnings in the Mobile pipeline

## Code changes

* **.github/workflows/*:** Updates all workflow

## Before you submit

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -12,9 +12,6 @@ on:
       version_number:
         required: true
         type: string
-    secrets:
-      AZURE_PROD_KV_CREDENTIALS:
-        required: true
 
 jobs:
   bump_version:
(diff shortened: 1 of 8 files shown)


.github/workflows/version-bump.yml
```diff
@@ -12,32 +12,29 @@ on:
       version_number:
         required: true
         type: string
-    secrets:
```
urbinaalex17, line 15: Is there any reason why we are deleting this block?

urbinaalex17 (changes requested):

.github/workflows/version-bump.yml
Eeebru: The variable `AZURE_PROD_KV_CREDENTIALS` is passed from the `auto_version_bump` workflow, and it's not used anywhere in the workflow. The same is deleted in the `auto_version_bump` workflow too.

urbinaalex17: I still don't get it. Could you please point me to the workflow where we use the `AZURE_PROD_KV_CREDENTIALS` so we don't need to specify it here?

Eeebru: Version Auto Bump [workflow](https://github.com/bitwarden/mobile/actions/runs/5981829754) is failing when it calls Version Bump [workflow](https://github.com/bitwarden/mobile/actions/runs/6004555823) because of the secrets being passed from Version Auto Bump to Version Bump, I only removed the secret that is getting passed around, and let the workflow get the secret by itself.

urbinaalex17: @Eeebru , got it! If you are deleting the secret that is being passed from [version-auto-bump.yml](https://github.com/bitwarden/mobile/blob/master/.github/workflows/version-auto-bump.yml) to [version-bump.yml](https://github.com/bitwarden/mobile/blob/master/.github/workflows/version-bump.yml), shouldn't we keep it in [version-bump.yml](https://github.com/bitwarden/mobile/blob/master/.github/workflows/version-bump.yml) instead of deleting it from both workflows?

Eeebru: @urbinaalex17  I deleted the one that was passed to `version-bump.yml` in the `on:workflow` object

urbinaalex17 (approved):

PR #2747, opened 2023-09-07
[PM-3393] Excessive Invalid Biometric unlock attempts should automatically log out TDE users

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When the user has no master password or pin and exceeds the biometric login attempts, the app should logged out.

## Code changes

Add action to `IsBiometricIntegrityValidAsync` to invoke when the TooManyAttempts is triggered.
Add to `AppHelpers` a method to check if the user doesn't have other login methods available, show a warning message and logout.

## Before you submit


src/App/Utilities/AppHelpers.cs
```diff
@@ -593,5 +594,18 @@ public static async Task ClearServiceCacheAsync()
...
+            if (hasMasterPassword || pinEnabled)
+            {
+                return;
+            }
+
+            var platformUtilsService = ServiceContainer.Resolve<IPlatformUtilsService>();
+            var vaultTimeoutService = ServiceContainer.Resolve<IVaultTimeoutService>();
+            await platformUtilsService.ShowDialogAsync(AppResources.BiometricsLoginExceeded, AppResources.Warning, AppResources.OkGotIt);
+            await vaultTimeoutService.LogOutAsync();
+        }
```
fedemkr, line 609: 🎨 I think this should be moved to the `IBiometricsService` (to the base class) so we don't make this helper class bigger, which IMO should have several methods moved to services. Doing this, we try to concentrate everything relating to biometrics in the same place.
Edit: However, we could move it to `MobilePlatformUtilsService`, either the entire code or just a call to the `IBiometricsService` that performs this. Because as far as I see the same action is passed from the callers. Additionally, I'd change the parameters to something that is more specific to the method like `bool shouldLogOutOnBiometricsAttemptsExceeded` and then you can just put the `or` condition on the caller.
What do you think?

src/App/Resources/AppResources.resx
```diff
@@ -2762,4 +2762,7 @@ Do you want to switch to this account?</value>
   <data name="LoggingInOn" xml:space="preserve">
     <value>Logging in on</value>
   </data>
+  <data name="BiometricsLoginExceeded" xml:space="preserve">
```
fedemkr, line 2765: ⛏️ What do you think on appending the suffix `Description` to indicate that it's actually something larger than what's in the key? Btw I started using `DescriptionLong` as the key suffix for things that could be correct with just "Description" but in this case, I guess `Description` is just fine. `Explanation` could be another suffix.

andrebispo5: Sorry i was waiting for a final version and left it as a mock.

src/App/Resources/AppResources.resx
```diff
@@ -2161,7 +2161,7 @@ Scanning will happen automatically.</value>
   <data name="AccountLocked" xml:space="preserve">
     <value>Locked</value>
   </data>
-  <data name="AccountLoggedOut" xml:space="preserve">
+  <data name="LoggedOut" xml:space="preserve">
     <value>Logged out</value>
   </data>
```
fedemkr, line 2166: 🤔 We need to be careful when modifying keys on the app resources because we lose the translations (until this one is done). Can't remember if this was added in this app version or a previous one to see if it can be modified or if we should just add a new key and mark this one as obsolete to be removed in a later version.

andrebispo5: I think this one is pretty old. I'll add another name to the new one.

src/App/Services/MobilePlatformUtilsService.cs
```diff
@@ -269,6 +270,12 @@ public async Task<bool> IsBiometricIntegrityValidAsync(string bioIntegritySrcKey
                 {
                     fallback?.Invoke();
                 }
+                if (result.Status == FingerprintAuthenticationResultStatus.TooManyAttempts
+                    && logOutOnTooManyAttempts)
+                {
+                    await ShowDialogAsync(AppResources.AccountLoggedOutBiometricExceeded, AppResources.TooManyAttempts, AppResources.Ok);
+                    _messagingService.Send(AccountsManagerMessageCommands.LOGOUT, new Tuple<string, bool, bool>(null, true, false));
```
fedemkr, line 277: ⛏️ You can skip the `Tuple` parameter given that those are the default values when the extras parameter is `null` as you can see [here](https://github.com/bitwarden/mobile/blob/master/src/App/Utilities/AccountManagement/AccountsManager.cs#L153).

fedemkr (changes requested):

fedemkr (approved):

PR #2764, opened 2023-09-12
[PM-3446]

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Check if user has mp and allow autofill to use items with mp re-prompt.

## Code changes

Change method that fetches items to show and bypass mp re-prompt check.

## Before you submit

### DIFF
--- a/src/Android/Autofill/AutofillHelpers.cs
+++ b/src/Android/Autofill/AutofillHelpers.cs
@@ -20,6 +20,7 @@ using AndroidX.AutoFill.Inline.V1;
 using Bit.Core.Abstractions;
 using SaveFlags = Android.Service.Autofill.SaveFlags;
 using Bit.Droid.Utilities;
+using Bit.Core.Services;
 
 namespace Bit.Droid.Autofill
 {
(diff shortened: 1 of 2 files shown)


src/Android/Autofill/AutofillHelpers.cs
```diff
@@ -152,23 +153,24 @@ public static class AutofillHelpers
             "androidapp://com.oneplus.applocker",
         };
 
-        public static async Task<List<FilledItem>> GetFillItemsAsync(Parser parser, ICipherService cipherService)
+        public static async Task<List<FilledItem>> GetFillItemsAsync(Parser parser, ICipherService cipherService, IUserVerificationService userVerificationService)
         {
+            var allowReprompt = !await userVerificationService.HasMasterPasswordAsync();
```
mpbw2, line 158: I think `hasMasterPassword` would be a more appropriate name for this so it makes more sense during use below, thoughts?

mpbw2 (approved):

PR #2769, opened 2023-09-13
[PM-3726] prevent legacy user login

jlf0dev:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Prevent legacy users from logging in and direct them to the web vault for migration.

## Code changes

Legacy users are detected early in the `AuthService.LoginHelper` by checking for the existence of a local hashed password (meaning MP was used to login) and the absence of a `IdentityToken.Key`. We can then return early before the authentication process and prevent any state from being set.
- **LockPageViewModel.cs:** Prevents unlocks for legacy users. Most of these changes are formatting. I split out the `SubmitAsync` method into `UnlockWithPinAsync` and `UnlockWithMasterPasswordAsync`. I also am catching any `LegacyUserException` errors now in the `SubmitAsync` method in order to process them.
- **BaseLockPasswordViewController.cs:** Duplicate changes as `LockPageViewModel`
- **LoginPageViewModel.cs:** Prevent logins for legacy users. **Note:** Instead of using the `LegacyUserException` like on the LockPage, I have added `AuthResult.RequiresEncryptionKeyMigration`. The clients repo couldn't use an exception in the same way so I decided to keep the models similar.
- **AuthService.cs:** If a user is logging in with their MP, we can determine if they are a legacy user and return early.
- **CryptoService.cs:** Add `ValidateUserKeyAsync` and `IsLegacyUserAsync` for legacy user detection. Throw exceptions from migration methods if a legacy user is detected.
- **VaultTimeoutService.cs:** If a legacy user has an auto key set, we need to catch the exception and log them out before they are migrated.

## Screenshots

image

## Before you submit


mpbw2 (approved):

mpbw2 (dismissed):

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -230,166 +238,203 @@ public async Task InitAsync()
...
+
+            var masterKey = await _cryptoService.MakeMasterKeyAsync(MasterPassword, _email, kdfConfig);
+            var storedKeyHash = await _cryptoService.GetMasterKeyHashAsync();
+            var passwordValid = false;
+            MasterPasswordPolicyOptions enforcedMasterPasswordOptions = null;
+
+            if (await _cryptoService.IsLegacyUserAsync(masterKey))
+            {
+                throw new LegacyUserException();
+            }
```
fedemkr, line 374: ⛏️ Can the check be moved right after `masterKey` initialization, because there's no need to await for getting the master key hash if it's a legacy user. Unless it depends on some internal computation on that latter task to be able to say whether it's legacy or not.
🤔 Could the check be done directly in `MakeMasterKeyAsync()`? Or is it not done that way to prevent potentially breaking other places?

src/App/Pages/Accounts/LockPageViewModel.cs
```diff
@@ -493,5 +548,29 @@ private async Task DoContinueAsync()
             _messagingService.Send("unlocked");
             UnlockedAction?.Invoke();
         }
+
+        private async Task<bool> IsBiometricsEnabledAsync()
+        {
+            try
+            {
+                return await _vaultTimeoutService.IsBiometricLockSetAsync() &&
+                                   await _biometricService.CanUseBiometricsUnlockAsync();
```
fedemkr, line 557: ⛏️ Fix formatting, either one line or await aligned / one tab from return column.

src/App/Resources/AppResources.resx
```diff
@@ -954,6 +954,9 @@ Scanning will happen automatically.</value>
   <data name="UpdateKey" xml:space="preserve">
     <value>You cannot use this feature until you update your encryption key.</value>
   </data>
+  <data name="EncryptionKeyMigrationRequired" xml:space="preserve">
+    <value>Encryption key migration required. Please login through the web vault to update your encryption key.</value>
+  </data>
```
fedemkr, line 959: ⛏️ I think the string `Encryption key migration required` could be something we could use and given that this key is here we couldn't use the same key again. I'd use another key that indicates that it has a lengthier description. I tend to use `DescriptionLong` as a suffix given that is highly unlikely something like that could be used actually as a resource in the app.
```suggestion
  <data name="EncryptionKeyMigrationRequiredDescriptionLong" xml:space="preserve">
    <value>Encryption key migration required. Please login through the web vault to update your encryption key.</value>
  </data>
```

src/Core/Services/CryptoService.cs
```diff
@@ -62,6 +62,11 @@ public Task<UserKey> GetUserKeyAsync(string userId = null)
             return _stateService.GetUserKeyAsync(userId);
         }
 
+        public async Task<bool> IsLegacyUserAsync(MasterKey masterKey = null, string userId = null)
+        {
+            return await ValidateUserKeyAsync(new UserKey((masterKey ?? await GetMasterKeyAsync(userId)).Key));
+        }
```
fedemkr, line 68: ⚠️  `GetMasterKeyAsync(...)` could potentially [return null](https://github.com/bitwarden/mobile/blob/efd1671f489186e529ed9ff199b99c74b7bb9a21/src/Core/Services/CryptoService.cs#L133). Is it ok if that happens to throw a null reference exception here (when accessing `.Key`)?

src/Core/Services/CryptoService.cs
```diff
@@ -997,6 +1002,32 @@ private async Task<TKey> MakeKeyAsync<TKey>(string password, string salt, KdfCon
...
+
+                var privateKey = await DecryptToBytesAsync(new EncString(encPrivateKey), key);
+                await _cryptoFunctionService.RsaExtractPublicKeyAsync(privateKey);
+            }
+            catch (Exception)
+            {
+                return false;
+            }
+
+            return true;
```
fedemkr, line 1028: ⛏️ No need for `(Exception)` if you just want to catch everything and IMO it's better to return inside the try so when reading the code you don't need to skip or jump to know what's next just for returning:
```suggestion
                await _cryptoFunctionService.RsaExtractPublicKeyAsync(privateKey);
                return true;
            }
            catch
            {
                return false;
            }
```

src/iOS.Core/Controllers/BaseLockPasswordViewController.cs
```diff
@@ -370,20 +297,126 @@ private async Task HandleFailedCredentialsAsync()
...
+            {
+                failed = true;
+            }
+
+            if (failed)
+            {
+                await HandleFailedCredentialsAsync();
+            }
+
+
```
fedemkr, line 359: ⛏️ Remove spaces

fedemkr (changes requested):
Great work!
I see this repeated several times:
```c#
if (await _cryptoService.IsLegacyUserAsync(masterKey))
{
    throw new LegacyUserException();
}
```
If every time checking legacy should throw that exception, couldn't that be just done inside the method like: 
```c#
public async Task CheckIfUserIsLegacyAsync(MasterKey masterKey = null, string userId = null)
{
    if (await ValidateUserKeyAsync(new UserKey((masterKey ?? await GetMasterKeyAsync(userId)).Key))) // Also here see other comment to check the possibility of null getting the master key
    {
        throw new LegacyUserException();
    }
}
```
So we avoid checking it everywhere. Thoughts?

src/Core/Services/CryptoService.cs
jlf0dev: Good catch, added a check

src/App/Pages/Accounts/LockPageViewModel.cs
jlf0dev: We won't have all the pieces yet to determine if they are legacy when simply logging in. I also didn't know if we wanted to run that code every time so left it out for now.

jlf0dev:
> So we avoid checking and throwing it everywhere. Thoughts?

I actually had to add that check to the login process, where we don't want it to throw. Would definitely clean up the code but I didn't want to use exceptions for the login process.

fedemkr (approved):

fedemkr (approved):

PR #2795, opened 2023-09-25
[PM-3741] [PM-3750] Improvements to local storage handling

mpbw2:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Improvements to the way we handle file attachment intents and storage restrictions

## Code changes

* **FileService.cs / Constants.cs / filepaths.xml:** Convert to using specific path for temporary camera images using purpose-driven path restrictions
* **MainActivity.cs:** For non-camera attachments do not allow URIs pointing to our app's private storage. For camera attachments use the new temp path defined above.  In both cases a resulting filename is required to continue with the attachment process as another layer of security.

## Before you submit

### DIFF
--- a/src/Android/MainActivity.cs
+++ b/src/Android/MainActivity.cs
@@ -239,18 +239,22 @@ namespace Bit.Droid
                 string fileName = null;
                 if (data != null && data.Data != null)
                 {
-                    uri = data.Data;
-                    fileName = AndroidHelpers.GetFileName(ApplicationContext, uri);
+                    if (!data.Data.ToString()?.Contains(Constants.PACKAGE_NAME) ?? false)
+                    {
+                        uri = data.Data;
+                        fileName = AndroidHelpers.GetFileName(ApplicationContext, uri);
+                    }
                 }
                 else
                 {
                     // camera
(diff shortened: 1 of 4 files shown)


src/Android/MainActivity.cs
```diff
@@ -239,18 +239,22 @@ protected override void OnActivityResult(int requestCode, Result resultCode, Int
                 string fileName = null;
                 if (data != null && data.Data != null)
                 {
-                    uri = data.Data;
-                    fileName = AndroidHelpers.GetFileName(ApplicationContext, uri);
+                    if (!data.Data.ToString()?.Contains(Constants.PACKAGE_NAME) ?? false)
```
fedemkr, line 242: 🤔 Can `ToString()` return `null` when `data.Data` is not `null` (given the if check above)? 
🤔 If so, do you think the next would be easier to read?
```suggestion
                    if (data.Data.ToString()?.Contains(Constants.PACKAGE_NAME) != true)
```

mpbw2: Rider is telling me `Uri.ToString()` can return null so I'm trusting it.  Also agree your modification is easier to read 👍

fedemkr (approved):

PR #2798, opened 2023-09-26
[PM-2671] Update mobile client to use regions

andrebispo5:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Modify the region selector to store the region in state and use that region to look up the proper URLs.
Add “cleanup” logic on startup to look for the Vault URL in the base URL and set the region appropriately, removing the base URL from state.

## Code changes

**src/Core/Models/Data/EnvironmentUrlData.cs**: 
Add missing urls for US and EU clients.

https://github.com/bitwarden/mobile/compare/auth/pm-2671/regions?expand=1#diff-443cceacc227560ece2b5ab7a956287a7737bd46e0cf0d5e9fb081ddd1f515acR113
Add method to migrate old users that don't have region, this will set a region according to their base url saved.

## Screenshots

No visual updates.

## Before you submit


src/Core/Services/EnvironmentService.cs
```diff
@@ -96,15 +110,57 @@ public async Task SetUrlsFromStorageAsync()
...
+                {
+                    return await SetRegionAsync(Region.US);
+                }
+                urls = FormatUrls(selfHostedUrls);
+            }
+            else
+            {
+                urls = region.GetUrls();
+            }
+
```
fedemkr, line 163: 🤔 To avoid racing conditions, what do you think about changing the order this is done given that if the self hosted Uris are empty then US will be set afterwards, but in the meantime PreAuth would have been set to SelfHost which would be wrong, so I imagine something like this:
```suggestion
            EnvironmentUrlData urls;
            
            if (region == Region.SelfHosted)
            {
                // If user saves a self-hosted region with empty fields, default to US
                if (selfHostedUrls.IsEmpty)
                {
                    return await SetRegionAsync(Region.US);
                }
                urls = FormatUrls(selfHostedUrls);
            }
            else
            {
                urls = region.GetUrls();
            }
            
            await _stateService.SetPreAuthRegionAsync(region);
            SelectedRegion = region;

```
Addionally, this would be theoretically faster in the case self hosted urls are empty because we are awaiting only one time setting pre auth region.

src/Core/Services/EnvironmentService.cs
```diff
@@ -113,20 +169,19 @@ public async Task<EnvironmentUrlData> SetUrlsAsync(EnvironmentUrlData urls)
...
+        {
+            urls.Base = FormatUrl(urls.Base);
+            urls.WebVault = FormatUrl(urls.WebVault);
+            urls.Api = FormatUrl(urls.Api);
+            urls.Identity = FormatUrl(urls.Identity);
+            urls.Icons = FormatUrl(urls.Icons);
+            urls.Notifications = FormatUrl(urls.Notifications);
+            urls.Events = FormatUrl(urls.Events);
             return urls;
         }
```
fedemkr, line 186: 🎨 I think both `FormatUrls(...)` methods should be moved inside `EnvironmentUrlData` given that IMO the service shouldn't be responsible of formatting it but the object itself; additionally it improves maintainability if another url is added to the `EnvironementUrlData`, so the method is just in there to be seen by the developer and update it accordingly.

src/Core/Services/StateService.cs
```diff
@@ -1359,6 +1359,39 @@ public void SetConfigs(ConfigResponse value)
             _storageMediatorService.Save(Constants.ConfigsKey, value);
         }
 
+        public async Task<Region?> GetActiveUserRegionAsync(string userId = null)
+        {
+            return (await GetAccountAsync(
+                ReconcileOptions(new StorageOptions { UserId = userId }, await GetDefaultStorageOptionsAsync())
+            ))?.Settings?.Region;
+        }
```
fedemkr, line 1367: 🎨 Naming doesn't reflect the method given that it says **active user** and given that the `userId` is being passed, it could be a different user than the active one. If the intent is to get the active one, then remove the parameter and use:
```suggestion
        public async Task<Region?> GetActiveUserRegionAsync()
        {
            return await GetActiveUserCustomDataAsync(a => a?.Settings?.Region);
        }
```
Otherwise, rename to `GetRegionAsync(string userId = null)`

src/Core/Services/StateService.cs
```diff
@@ -1359,6 +1359,39 @@ public void SetConfigs(ConfigResponse value)
...
+
+        public async Task UpdateActiveUserEnvironmentAsync(Region region, EnvironmentUrlData urls, string userId = null)
```
fedemkr, line 1369: 🎨 Naming issue, if it's the active user then remove the `userId` parameter, otherwise change the name to `UpdateEnvironmentAsync(Region region, EnvironmentUrlData urls, string userId = null)`

src/Core/Utilities/RegionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            switch (region)
+            {
+                case Region.US:
+                    return EnvironmentUrlData.DefaultUS;
+                case Region.EU:
+                    return EnvironmentUrlData.DefaultEU;
+                case Region.SelfHosted:
+                default:
+                    return null;
```
fedemkr, line 18: 🎨 Use just `default` that will cover the `SelfHosted` as well. Apply it also on the other extensions methods of this file.
```suggestion
                default:
                    return null;
```

src/Core/Services/EnvironmentService.cs
```diff
@@ -54,38 +56,50 @@ public string GetWebSendUrl()
...
+            if (!string.IsNullOrWhiteSpace(url))
+            {
+                if (url.Contains(Region.US.Domain()) || url.Contains(Region.EU.Domain()))
+                {
+                    return CoreHelpers.GetDomain(url);
+                }
+                return CoreHelpers.GetHostname(url);
+            }
+            return string.Empty;
+        }
```
fedemkr, line 71: 🤔 Can this be reused somehow? I see the same logic applied in `AccountView.ParseEndpoint(...)`

fedemkr (changes requested):

src/Core/Services/EnvironmentService.cs
```diff
@@ -54,38 +56,50 @@ public string GetWebSendUrl()
...
-                BaseUrl = urls.Base;
-                WebVaultUrl = urls.WebVault;
-                ApiUrl = envUrls.Api = urls.Api;
-                IdentityUrl = envUrls.Identity = urls.Identity;
-                IconsUrl = urls.Icons;
-                NotificationsUrl = urls.Notifications;
-                EventsUrl = envUrls.Events = urls.Events;
-                _apiService.SetUrls(envUrls);
+                // Migrate old users to regions
+                region ??= await MigrateToRegionsAsync(urls);
```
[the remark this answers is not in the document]
andrebispo5: Yup thanks! This way i can iterate all accounts in state and update them all at once.

src/Core/Services/EnvironmentService.cs
```diff
@@ -96,15 +110,57 @@ public async Task SetUrlsFromStorageAsync()
...
+                await _stateService.UpdateActiveUserEnvironmentAsync(Region.US, Region.US.GetUrls());
+                return Region.US;
+            }
+            if (urls.Base == Region.EU.BaseUrl())
+            {
+                await _stateService.UpdateActiveUserEnvironmentAsync(Region.EU, Region.EU.GetUrls());
+                return Region.EU;
+            }
+            await _stateService.UpdateActiveUserEnvironmentAsync(Region.SelfHosted, urls);
+            return Region.SelfHosted;
```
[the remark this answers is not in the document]
andrebispo5: Added extra check to avoid settings urls to null if environment is SelfHosted.
`await _stateService.UpdateActiveUserEnvironmentAsync(region, region.GetUrls() ?? urls);`

src/Core/Services/StateService.cs
andrebispo5: This was removed and moved into the state migration service.

src/Core/Services/EnvironmentService.cs
andrebispo5: Moved logic into EnvironmentUrlData.

src/Core/Services/StateMigrationService.cs
```diff
@@ -837,6 +841,43 @@ internal enum UsernameEmailType
...
+        {
+            // account data
+            var state = await GetValueAsync<State>(Storage.Prefs, V7Keys.StateKey);
+
+            foreach (var account in state.Accounts.Where(a => a.Value?.Profile?.UserId != null && a.Value?.Settings != null))
+            {
+                var urls = account.Value.Settings.EnvironmentUrls ?? Region.US.GetUrls();
+                account.Value.Settings.Region = urls.Region;
+                account.Value.Settings.EnvironmentUrls = urls.Region.GetUrls() ?? urls;
+                continue;
```
fedemkr, line 865: ⛏️ `continue;` is not necessary, remove.
```suggestion
```

src/Core/Models/Data/EnvironmentUrlData.cs
```diff
@@ -44,5 +46,63 @@ public class EnvironmentUrlData
...
+            if (!string.IsNullOrWhiteSpace(url))
+            {
+                if (url.Contains(Region.US.Domain()) || url.Contains(Region.EU.Domain()))
+                {
+                    return CoreHelpers.GetDomain(url);
+                }
+                return CoreHelpers.GetHostname(url);
+            }
+            return string.Empty;
+        }
```
fedemkr, line 106: 🎨 Could we change the name? As a consumer, I wouldn't know what to expect when calling `ParseEndpoint()`, I'd need to enter and see what it does to know what it will return.
Also, I feel like changing the order would be beneficial for reading:
```suggestion
        public string ParseEndpoint()
        {
            var url = WebVault ?? Base ?? Api ?? Identity;
            if (string.IsNullOrWhiteSpace(url))
            {
                return string.Empty;
            }
            if (url.Contains(Region.US.Domain()) || url.Contains(Region.EU.Domain()))
            {
                return CoreHelpers.GetDomain(url);
            }
            return CoreHelpers.GetHostname(url);
        }
```

fedemkr (changes requested):
Much better 🎉  Such tiny improvements! Also, I think the migration is fine but @mpbw2 could you review it in case I'm missing something?

src/Core/Services/StateMigrationService.cs
```diff
@@ -837,6 +841,43 @@ internal enum UsernameEmailType
...
+
+        private class V7Keys
+        {
+            // global keys
+            internal const string StateKey = "state";
+            internal const string RegionEnvironmentKey = "regionEnvironment";
+            internal const string PreAuthEnvironmentUrlsKey = "preAuthEnvironmentUrls";
+        }
+
+        // Migrate environment data to use Regions
```
mpbw2, line 854: Comment should be inside `MigrateFrom6To7Async` block since additional (unrelated) migrations could be added before going to production

src/Core/Models/Data/EnvironmentUrlData.cs
fedemkr: 🎨 @andrebispo5 Could the name of `ParseEndpoint()` be changed? for the reasons explained above.

src/Core/Models/Data/EnvironmentUrlData.cs
```diff
@@ -20,5 +46,63 @@ public class EnvironmentUrlData
...
+            }
+            url = Regex.Replace(url, "\\/+$", string.Empty);
+            if (!url.StartsWith("http://") && !url.StartsWith("https://"))
+            {
+                url = string.Concat("https://", url);
+            }
+            return url.Trim();
+        }
+
+        public string GetDomainOrHost()
```
fedemkr, line 94: ⛏️ Just to be consistent with the helper naming I would change this to:
```suggestion
        public string GetDomainOrHostname()
```

fedemkr (changes requested):

fedemkr (approved):

PR #2813, opened 2023-10-04
[PM-868] Re-open app to item could crash the app

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Try fix the most reoccurring crash. If the app goes to background and there was an item open, when it comes to the foreground and tries to load the item again it could crash.
https://github.com/bitwarden/mobile/issues/2794
https://github.com/bitwarden/mobile/issues/2478

## Code changes

Check for previous opened item before trying to load the vault list.

## Before you submit

### DIFF
--- a/src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
+++ b/src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
@@ -38,7 +38,6 @@ namespace Bit.iOS.Core.Renderers.CollectionView
                     // Do nothing in here, this is temporary to get more info about the crash, if the logger fails, we want to get the info
                     // by crashing with the original exception and not the logger one
                 }
-                throw colEx;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
```diff
@@ -38,7 +38,6 @@ protected override void UpdateTemplatedCell(TemplatedCell cell, NSIndexPath inde
                     // Do nothing in here, this is temporary to get more info about the crash, if the logger fails, we want to get the info
                     // by crashing with the original exception and not the logger one
                 }
-                throw colEx;
```
fedemkr, line 41: 🤔 Could we allow throwing all exceptions except the one that is being reported in AppCenter while crashing? If I'm not mistaken it would be `IndexOutOfRangeException`, so basically adding:
```c#
if (ex is IndexOutOfRangeException)
{
     return;
}
throw colEx;
```
Because if we don't throw anything we might lose information on AppCenter on non-related issues to this one.

fedemkr (approved):

PR #2818, opened 2023-10-11
[PM-3340] Update timeout action for users without master password.

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Update settings screen vault timeout action cell UI to match designs.

## Code changes

Add disable and description properties to SettingsPageListItem.
Add new UI cell to data template.

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/e98284c8-4b9d-426f-b81b-58b7a979af65

## Before you submit

### DIFF
--- a/src/App/Controls/Settings/BaseSettingControlView.cs
+++ b/src/App/Controls/Settings/BaseSettingControlView.cs
@@ -1,8 +1,6 @@
-﻿using System.ComponentModel;
-using System.Runtime.CompilerServices;
+﻿using System.Runtime.CompilerServices;
 using Bit.App.Utilities;
 using Xamarin.Forms;
-using static PInvoke.BCrypt;
 
 namespace Bit.App.Controls
 {
@@ -14,12 +12,6 @@ namespace Bit.App.Controls
         public static readonly BindableProperty SubtitleProperty = BindableProperty.Create(
             nameof(Subtitle), typeof(string), typeof(SwitchItemView), null);
 
-        public static readonly BindableProperty ItemEnabledProperty = BindableProperty.Create(
-            nameof(ItemEnabled), typeof(bool), typeof(SwitchItemView), true);
-
-        public static readonly BindableProperty ItemTextColorProperty = BindableProperty.Create(
-            nameof(ItemTextColor), typeof(Color), typeof(SwitchItemView));
-
         public string Title
         {
             get { return (string)GetValue(TitleProperty); }
@@ -30,28 +22,6 @@ namespace Bit.App.Controls
         {
             get { return (string)GetValue(SubtitleProperty); }
             set { SetValue(SubtitleProperty, value); }
-        }
-
-        public bool ItemEnabled
-        {
-            get { return (bool)GetValue(ItemEnabledProperty); }
-            set { SetValue(ItemEnabledProperty, value); }
-        }
-
-        public Color ItemTextColor
-        {
-            get { return (Color)GetValue(ItemTextColorProperty); }
-            set { SetValue(ItemTextColorProperty, value); }
-        }
-
-        protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
-        {
-            base.OnPropertyChanged(propertyName);
-            if (propertyName == nameof(ItemEnabled))
-            {
-                ItemTextColor = ItemEnabled ? ThemeManager.GetResourceColor("TextColor") :
-            ThemeManager.GetResourceColor("MutedColor");
-            }
         }
     }
 }
--- a/src/App/Controls/Settings/BaseSettingControlView.cs
+++ b/src/App/Controls/Settings/BaseSettingControlView.cs
@@ -35,7 +35,7 @@ namespace Bit.App.Controls
         public bool ItemEnabled
         {
             get { return (bool)GetValue(ItemEnabledProperty); }
-            set {SetValue(ItemEnabledProperty, value);}
+            set { SetValue(ItemEnabledProperty, value); }
         }
 
         public Color ItemTextColor
--- a/src/App/Controls/Settings/BaseSettingControlView.cs
+++ b/src/App/Controls/Settings/BaseSettingControlView.cs
@@ -1,14 +1,24 @@
-﻿using Xamarin.Forms;
+﻿using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using Bit.App.Utilities;
+using Xamarin.Forms;
+using static PInvoke.BCrypt;
 
 namespace Bit.App.Controls
 {
     public class BaseSettingItemView : ContentView
     {
         public static readonly BindableProperty TitleProperty = BindableProperty.Create(
-            nameof(Title), typeof(string), typeof(SwitchItemView), null, BindingMode.OneWay);
+            nameof(Title), typeof(string), typeof(SwitchItemView), null);
 
         public static readonly BindableProperty SubtitleProperty = BindableProperty.Create(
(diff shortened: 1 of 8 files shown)


src/App/Controls/Settings/BaseSettingControlView.cs
```diff
@@ -1,14 +1,24 @@
...
 
         public static readonly BindableProperty SubtitleProperty = BindableProperty.Create(
-            nameof(Subtitle), typeof(string), typeof(SwitchItemView), null, BindingMode.OneWay);
+            nameof(Subtitle), typeof(string), typeof(SwitchItemView), null);
+
+        public static readonly BindableProperty ItemEnabledProperty = BindableProperty.Create(
+            nameof(ItemEnabled), typeof(bool), typeof(SwitchItemView), true);
+
+        public static readonly BindableProperty ItemTextColorProperty = BindableProperty.Create(
+            nameof(ItemTextColor), typeof(Color), typeof(SwitchItemView));
```
fedemkr, line 21: 🎨 I think this is not needed. You can use the `IsEnabled` property directly, and in the `ControlTemplate` use a converter that depends on the is enabled bool to change the color.
Just FYI, when you want to do something on some bindable property changing, you can directly use the parameter of `propertyChanged` of the `BindableProperty.Create` instead of the general one of the control so you don't have to check whether the property is the one you want.

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
```diff
@@ -76,7 +78,7 @@ public SecuritySettingsPageViewModel()
                 _logger,
                 OnVaultTimeoutActionChangingAsync,
                 AppResources.SessionTimeoutAction,
-                _ => _inited && !HasVaultTimeoutActionPolicy,
+                _ => _inited && !HasVaultTimeoutActionPolicy && IsVaultTimeoutActionLockAllowed,
```
fedemkr, line 81: 🎨 Given that now the action is enabled based on `IsvaultTimeoutActionLockAllowed`, you need to call `VaultTimeoutActionPickerViewModel.SelectOptionCommand.RaiseCanExecuteChanged();` when that changes so it gets enabled/disabled accordingly.

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
```diff
@@ -234,10 +256,11 @@ public async Task InitAsync()
             _supportsBiometric = await _platformUtilsService.SupportsBiometricAsync();
             _canUnlockWithBiometrics = await _vaultTimeoutService.IsBiometricLockSetAsync();
             _canUnlockWithPin = await _vaultTimeoutService.GetPinLockTypeAsync() != Core.Services.PinLockType.Disabled;
+            IsVaultTimeoutActionLockAllowed = _hasMasterPassword || _canUnlockWithBiometrics || _canUnlockWithPin;
 
             await LoadPoliciesAsync();
             await InitVaultTimeoutPickerAsync();
-            await InitVaultTimeoutActionPickerAsync();
+            await MainThread.InvokeOnMainThreadAsync(InitVaultTimeoutActionPickerAsync);
```
fedemkr, line 263: 🤔 The call to the main thread should be done only on the part that is needed inside the `IniVaultTimeoutActionPickerAsync()`. Where is that necessary?

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
```diff
@@ -305,6 +328,7 @@ private async Task InitVaultTimeoutPickerAsync()
             {
                 _customVaultTimeoutTime = TimeSpan.FromMinutes(vaultTimeout);
             }
+            MainThread.BeginInvokeOnMainThread(() => { TriggerPropertyChanged(nameof(IsVaultTimeoutActionLockAllowed)); });
```
fedemkr, line 331: ⛏️ No need for brackets
```suggestion
            MainThread.BeginInvokeOnMainThread(() => TriggerPropertyChanged(nameof(IsVaultTimeoutActionLockAllowed)) );
```

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
```diff
@@ -501,7 +529,7 @@ private async Task<bool> OnVaultTimeoutActionChangingAsync(VaultTimeoutAction ti
 
             await _vaultTimeoutService.SetVaultTimeoutOptionsAsync(CurrentVaultTimeout, timeoutActionKey);
             _messagingService.Send(AppHelpers.VAULT_TIMEOUT_ACTION_CHANGED_MESSAGE_COMMAND);
-
+            MainThread.BeginInvokeOnMainThread(() => { TriggerPropertyChanged(nameof(IsVaultTimeoutActionLockAllowed)); });
```
fedemkr, line 532: ⛏️ No need for brackets

fedemkr (changes requested):

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
andrebispo5: Actually i didn't needed, it was something else that had me doing it but after all the changes it works without it.

src/App/Utilities/BoolToColorConverter.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Bit.App.Utilities
+{
+    public class BoolToColorConverter : IValueConverter
```
fedemkr, line 6: ⛏️ I think the name is too general given that inside it's very specific on what this is for. Maybe we could use:
```suggestion
    public class BoolEnablementToTextColorConverter : IValueConverter
```
so it's easy to recognize what's this for without looking at the implementation.

src/App/Pages/Settings/SecuritySettingsPageViewModel.cs
```diff
@@ -467,6 +479,15 @@ private void TriggerUpdateCustomVaultTimeoutPicker()
             TriggerPropertyChanged(nameof(CustomVaultTimeoutTime));
         }
 
+        private void TriggerVaultTimeoutActionLockAllowedPropertyChanged()
+        {
+            MainThread.BeginInvokeOnMainThread(() =>
+            {
+                TriggerPropertyChanged(nameof(IsVaultTimeoutActionLockAllowed));
+                TriggerPropertyChanged(nameof(SetUpUnlockMethodLabel));
+            });
```
fedemkr, line 488: 🎨 Add a call to `VaultTimeoutActionPickerViewModel.SelectOptionCommand.RaiseCanExecuteChanged();` so the command's binded controls are updated given the `canExecute` changed.

fedemkr (changes requested):
Awesome, just two small improvements and all good 🎉

src/App/Styles/ControlTemplates.xaml
```diff
@@ -6,6 +6,7 @@
                     x:Class="Bit.App.Styles.ControlTemplates">
 
     <u:StringHasValueConverter x:Key="stringHasValueConverter" />
+    <u:BoolEnablementToTextColorConverter x:Key="boolToColorConverter" />
```
fedemkr, line 9: ⛏️ Update key naming as well and usage

fedemkr (changes requested):

fedemkr (approved):
LGTM! 🚀

PR #2837, opened 2023-10-23
[PM-3273][PM-4679] New owner/admin permission on login

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

When a user has it's privileges elevated to owner/admin, the user has to set a password.

## Code changes

Set `ForcePasswordResetReason` on login, check and launch a modal to set a password if needed on landing page. The user will be always prompted for a password set until the user has set it. 
Since we can't get the decryption option until the next login, we have to set the `HasMasterPassword` property manually in state for the current session.

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/3bd4b352-0879-4373-8da1-533d057a99e6

## Before you submit


fedemkr (changes requested):

fedemkr:
❓ What happens if a user taps "Cancel" on the modal for setting a password? Does it re-appear?

src/Core/Services/StateService.cs
```diff
@@ -1363,6 +1363,15 @@ public void SetConfigs(ConfigResponse value)
...
 
+        public async Task SetUserHasMasterPasswordAsync(bool value, string userId = null)
+        {
+            var reconciledOptions = ReconcileOptions(new StorageOptions { UserId = userId },
+                await GetDefaultStorageOptionsAsync());
+            var account = await GetAccountAsync(reconciledOptions);
+            account.Profile.UserDecryptionOptions.HasMasterPassword = value;
+            await SaveAccountAsync(account, reconciledOptions);
+        }
+
```
fedemkr, line 1374: 🤔 Can this leave the app in a state where `HasMasterPassword` is `true` but for some error/cancellation of a user, this is not cleaned to `false` again? I know this is a workaround but I'm not sure if we'd better use a specific temporary property for this instead of `HasMasterPassword` that is used in other places. What do you think @mpbw2 ?

src/App/App.xaml.cs
```diff
@@ -171,6 +171,14 @@ public App(AppOptions appOptions)
...
                         });
                     }
+                    else if (message.Command == Constants.ForceSetPassword)
+                    {
+                        Device.BeginInvokeOnMainThread(async () =>
+                        {
+                            await Application.Current.MainPage.Navigation.PushModalAsync(
+                                new NavigationPage(new SetPasswordPage(orgIdentifier: (string)message.Data)));
+                        });
+                    }
```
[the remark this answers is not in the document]
andrebispo5: With Device.InvokeOnMainThreadAsync(...) throwing an exception will leave the user on vault. It's a good exception scenario.

src/Core/Services/SyncService.cs
```diff
@@ -398,6 +399,32 @@ private async Task SyncProfileAsync(ProfileResponse response)
...
+
+            var decryptionOptions = await _stateService.GetAccountDecryptionOptions();
+            if (decryptionOptions != null &&
+                !decryptionOptions.HasMasterPassword &&
+                hasManageResetPasswordPermission)
+            {
+                // TDE user w/out MP went from having no password reset permission to having it.
+                // Must set the force password reset reason so the auth guard will redirect to the set password page.
+                await _stateService.SetForcePasswordResetReasonAsync(ForcePasswordResetReason.TdeUserWithoutPasswordHasPasswordResetPermission);
+            }
```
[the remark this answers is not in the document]
andrebispo5: >  Does this logic apply when the user is not logged in with TDE? What happens with a user with multiple organizations and one of them has ManageResetPassword?
TDE users will only have one organization, if they have more than one they are not using TDE and won't have decryption options.

andrebispo5:
> ❓ What happens if a user taps "Cancel" on the modal for setting a password? Does it re-appear?

Yes, it will trigger OnAppearing in TansPage and launch the modal again.

src/Core/Services/StateService.cs
andrebispo5: @mpbw2 agreed let's move `HasMasterPassword` checks to _userVerificationService

andrebispo5: Actually _userVerificationService.HasMasterPasswordAsync() does a couple more checks, it is replacing the previous check for keyconnector usage. 
I'm not 100% that i can use `_userVerificationService.HasMasterPasswordAsync()` instead of the direct  `_stateService.GetAccountDecryptionOptions().HasMasterPassword`.

andrebispo5:
```
public async Task<bool> HasMasterPasswordAsync(bool checkMasterKeyHash = false)
{
    async Task<bool> CheckMasterKeyHashAsync()
    {
        return !checkMasterKeyHash || await _cryptoService.GetMasterKeyHashAsync() != null;
    };

    var decryptOptions = await _stateService.GetAccountDecryptionOptions();
    if (decryptOptions != null)
    {
        return decryptOptions.HasMasterPassword && await CheckMasterKeyHashAsync();
    }

    return !await _keyConnectorService.GetUsesKeyConnectorAsync() && await CheckMasterKeyHashAsync();
}
```

src/Core/Models/Domain/Account.cs
```diff
@@ -70,6 +71,7 @@ public AccountProfile(AccountProfile copy)
             public bool? HasPremiumPersonally;
             public ForcePasswordResetReason? ForcePasswordResetReason;
             public AccountDecryptionOptions UserDecryptionOptions;
+            public bool? HasManageResetPasswordPermission;
```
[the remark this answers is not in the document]
fedemkr: @andrebispo5 Can't we provide the default value for that? So we avoid null checking in the app.

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
```diff
@@ -256,6 +247,23 @@ public async Task SubmitAsync()
...
+                keysRequest = new KeysRequest
+                {
+                    PublicKey = newPublicKey,
+                    EncryptedPrivateKey = newProtectedPrivateKey.EncryptedString
+                };
+
+            }
+
+            return keysRequest;
+        }
```
fedemkr, line 265: 🎨 The code can be improved like this to avoid defining `keysRequest`:
```suggestion
        private async Task<KeysRequest> GetKeysForSetPasswordRequestAsync(UserKey newUserKey)
        {
            if (ForceSetPasswordReason == ForcePasswordResetReason.TdeUserWithoutPasswordHasPasswordResetPermission)
            {
                return null;
            }

            var (newPublicKey, newProtectedPrivateKey) = await _cryptoService.MakeKeyPairAsync(newUserKey);
            return new KeysRequest
            {
                PublicKey = newPublicKey,
                EncryptedPrivateKey = newProtectedPrivateKey.EncryptedString
            };
        }
```

fedemkr (changes requested):

src/App/App.xaml.cs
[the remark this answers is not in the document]
fedemkr: But shouldn't it lock / log out the user? Is it a good flow to leave the user on the vault without setting a password? Or they will be prompted again for setting up the password the next time they unlock?

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
```diff
@@ -183,24 +198,27 @@ public async Task SubmitAsync()
...
                 await _cryptoService.SetMasterKeyEncryptedUserKeyAsync(newProtectedUserKey.EncryptedString);
-                await _cryptoService.SetUserPrivateKeyAsync(newProtectedPrivateKey.EncryptedString);
+
+                // Set private key only for new JIT provisioned users in MP encryption orgs
+                // Existing TDE users will have private key set on sync or on login
+                if (keysRequest != null && ForceSetPasswordReason != ForcePasswordResetReason.TdeUserWithoutPasswordHasPasswordResetPermission)
+                {
+
+                    await _cryptoService.SetUserPrivateKeyAsync(keysRequest.EncryptedPrivateKey);
+                }
```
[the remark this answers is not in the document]
fedemkr: I think both conditions overlap:
- `keysRequest` will be non-null if `ForceSetPasswordReason != ForcePasswordResetReason.TdeUserWithoutPasswordHasPasswordResetPermission`
- `keysRequest` will be null if `ForceSetPasswordReason == ForcePasswordResetReason.TdeUserWithoutPasswordHasPasswordResetPermission`
- `ForceSetPasswordReason` will not be `TdeUserWithoutPasswordHasPasswordResetPermission` if `keysRequest` is non-null
- `ForceSetPasswordReason` will be `TdeUserWithoutPasswordHasPasswordResetPermission` if `keysRequest` is null

Putting it in another way, `keysRequest` nullability directly depends on `ForceSetPasswordReason`.
So you can use any of the two, adding the second condition does nothing given that it will be evaluated the same as the first one. 
The only scenario I see it's worth is if in the near future we have another flow where the aforementioned conditions are not met the same way as it's today.

src/Core/Services/StateService.cs
fedemkr: Can *TDE* use *KeyConnector*? If not, I don't see a reason why you can't use `HasMasterPasswordAsync(false)`. If possible I'd use the same one and try to centralize everything there with the different flows / scenarios so we avoid having different methods to check the "same" thing.

src/App/App.xaml.cs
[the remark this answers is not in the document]
andrebispo5: They will always be prompt to set the password every time the TabsPage trigger OnAppearing, until they do set it.

src/App/Pages/Accounts/SetPasswordPageViewModel.cs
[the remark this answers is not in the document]
andrebispo5: Thanks for the detailed explanation i see what you mean!

src/Core/Models/Domain/Account.cs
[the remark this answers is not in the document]
andrebispo5: Actually i can remove it now since the way to get this property changed.

src/Core/Services/StateService.cs
andrebispo5: After discussing this with the auth team, and the same problem happening in the other clients, we decided it was best to create a follow up tech dept ticket to address this code fragmentation problem. 
[Ticket PM-4579](https://bitwarden.atlassian.net/browse/PM-4579)

fedemkr (approved):

fedemkr (approved):

src/App/Pages/TabsPage.cs
```diff
@@ -100,11 +102,35 @@ protected override async void OnAppearing()
...
+                        _messagingService.Send(Constants.ForceSetPassword, userOrgs.First().Value.Identifier);
+                        return;
+                    }
+                    var rememberedOrg = _stateService.GetRememberedOrgIdentifierAsync();
+                    if (rememberedOrg == null)
+                    {
+                        return;
+                    }
+                    _messagingService.Send(Constants.ForceSetPassword, rememberedOrg);
+                    return;
```
fedemkr, line 127: 🤔 Should we log anything when the user doesn't have an org?

andrebispo5: Maybe we should for future reference

fedemkr (approved):

PR #2869, opened 2023-11-15
[PM-4837] Hide TOTP seed copy button when Can view, except password permission set

mpbw2:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Hide the copy button on the TOTP seed field when editing an item where `Can view, except password` permission is set to prevent user from obtaining seed value

## Code changes

* **CipherAddEditPage.xaml:** Use `Cipher.ViewPassword` property to drive copy icon visibility

## Before you submit

### DIFF
--- a/src/App/Pages/Vault/CipherAddEditPage.xaml
+++ b/src/App/Pages/Vault/CipherAddEditPage.xaml
@@ -292,7 +292,7 @@
                             StyleClass="box-row-button, box-row-button-platform"
                             Text="{Binding Source={x:Static core:BitwardenIcons.Clone}}"
                             Command="{Binding CopyCommand}"
-                            IsVisible="{Binding HasTotpValue}"
+                            IsVisible="{Binding Cipher.ViewPassword}"
                             Grid.Row="0"
                             Grid.Column="1"
                             Grid.RowSpan="2"
--- a/src/App/Pages/Vault/CipherAddEditPage.xaml
+++ b/src/App/Pages/Vault/CipherAddEditPage.xaml
@@ -292,7 +292,7 @@
                             StyleClass="box-row-button, box-row-button-platform"
                             Text="{Binding Source={x:Static core:BitwardenIcons.Clone}}"
                             Command="{Binding CopyCommand}"
-                            IsVisible="{Binding Cipher.ViewPassword}"
+                            IsVisible="{Binding AllowTotpCopy}"
                             Grid.Row="0"
                             Grid.Column="1"
                             Grid.RowSpan="2"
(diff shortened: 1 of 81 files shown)


src/App/Pages/Vault/CipherAddEditPage.xaml
```diff
@@ -292,7 +292,7 @@
                             StyleClass="box-row-button, box-row-button-platform"
                             Text="{Binding Source={x:Static core:BitwardenIcons.Clone}}"
                             Command="{Binding CopyCommand}"
-                            IsVisible="{Binding HasTotpValue}"
+                            IsVisible="{Binding Cipher.ViewPassword}"
```
fedemkr, line 295: 🤔 Shouldn't this be `HasTotpValue && Cipher.ViewPassword` in the VM and binded here?

mpbw2: You're right, not sure how I missed that.  Fixed.

src/App/Pages/Vault/CipherAddEditPageViewModel.cs
```diff
@@ -307,6 +307,7 @@ public bool HasCollections
         public bool PasswordPrompt => Cipher.Reprompt != CipherRepromptType.None;
         public string PasswordVisibilityAccessibilityText => ShowPassword ? AppResources.PasswordIsVisibleTapToHide : AppResources.PasswordIsNotVisibleTapToShow;
         public bool HasTotpValue => IsLogin && !string.IsNullOrEmpty(Cipher?.Login?.Totp);
+        public bool AllowTotpCopy => HasTotpValue && (Cipher?.ViewPassword ?? false);
```
fedemkr, line 310: :art: No need for parenthesis and IMO equal to true is more readable.
```suggestion
        public bool AllowTotpCopy => HasTotpValue && Cipher?.ViewPassword == true;
```
Aside from this tiny fix, should we raise property changed when the `Cipher` property changes? I ask because if `Cipher` starts with `null` then if `AllowTotpCopy` gets evaluated will be `false` and when `Cipher` is loaded currently it won't trigger `AllowTotpCopy` property changed so there will be an inconsistency if `ViewPassword` is `true` afterwards. Thoughts?

fedemkr, line 310: :art: No need for parenthesis and IMO equal to true is more readable.
```suggestion
        public bool AllowTotpCopy => HasTotpValue && Cipher?.ViewPassword == true;
```
Aside from this tiny fix, should we raise property changed when the `Cipher` property changes? I ask because if `Cipher` starts with `null` value then if `AllowTotpCopy` gets evaluated will be `false` and when `Cipher` is loaded currently it won't trigger `AllowTotpCopy` property changed so there will be an inconsistency. Thoughts?

mpbw2: Upon closer inspection it looks like the null check isn't necessary so I just removed it.

fedemkr: @mpbw2 I'd add it to prepare for MAUI or we're going to have issues when we merge this change to the maui branch because of how the binding engine works.

fedemkr: @mpbw2 oh and isn't the `RaisePropertyChanged` necessary when `Cipher` changes?

mpbw2: > @mpbw2 oh and isn't raising property changed necessary when `Cipher` changes?

It doesn't appear so - I'm seeing the expected results on test items containing both scenarios (tested on both iOS & Android).  In addition I don't believe changes to `Cipher` are possible beyond the initial load while on the Add/Edit screen, but correct me if I'm missing something.

fedemkr (approved):

PR #2893, opened 2023-12-07
[PM-4739] Implement checksum uri validation

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [X] New feature development
- [ ] Other

## Objective

This change ensures that the encrypted string is the same as what was originally saved with the server through the use of a checksum.

## Code changes

Related to:
 https://github.com/bitwarden/clients/pull/6485
 https://github.com/bitwarden/server/pull/3318

* **LoginUriApi.cs/LoginUriData.cs:** Adds checksum to loginUri models for the api.
* **Cipher.cs:** Bypass validation if Cipher doesn't have a key
* **Login.cs:** Changed logic to only add valid uris
* **LoginUri.cs:** Add a method to validate the uri
* **CipherService.cs:** Add a checksum to domain LoginUri that will be used later for validation
* **CryptoService.cs:** Add a method that returns a base64 hash.

## Before you submit


src/Core/Models/Domain/Login.cs
```diff
@@ -44,7 +44,11 @@ public async Task<LoginView> DecryptAsync(string orgId, SymmetricCryptoKey key =
                 view.Uris = new List<LoginUriView>();
                 foreach (var uri in Uris)
                 {
-                    view.Uris.Add(await uri.DecryptAsync(orgId, key));
+                    var loginUriView = await uri.DecryptAsync(orgId, key);
+                    if (bypassValidation || (await uri.ValidateChecksum(loginUriView.Uri, orgId, key)))
```
fedemkr, line 48: ⛏️ No need for the extra parenthesis:
```suggestion
                    if (bypassValidation || await uri.ValidateChecksum(loginUriView.Uri, orgId, key))
```

src/Core/Models/Domain/LoginUri.cs
```diff
@@ -18,11 +20,18 @@ public class LoginUri : Domain
         public LoginUri(LoginUriData obj, bool alreadyEncrypted = false)
         {
             Match = obj.Match;
+
+            if (obj.UriChecksum != null)
+            {
+                UriChecksum = new EncString(obj.UriChecksum);
+            }
```
fedemkr, line 27: ⚠️  This is not needed, as it's handled by the `BuildDomainModel(..)` next line.
You should just add `nameof(UriChecksum)` to the `_map` HashSet above and it'll take care of creating the `EncString` for you. Also it'll handle building the data model in the `ToLoginUriData` method.

src/Core/Models/Domain/Cipher.cs
```diff
@@ -115,7 +117,7 @@ public async Task<CipherView> DecryptAsync()
             switch (Type)
             {
                 case Enums.CipherType.Login:
-                    model.Login = await Login.DecryptAsync(OrganizationId, model.Key);
+                    model.Login = await Login.DecryptAsync(OrganizationId, bypassValidation, model.Key);
```
fedemkr, line 120: 🤔 Is the `bypassValidation` needed here? Couldn't we just pass:
```suggestion
                    model.Login = await Login.DecryptAsync(OrganizationId, Key != null, model.Key);
```
Furthermore, I'm not 100% sure, but maybe you can skip the parameter and just check inside the method if `model.Key` is `null` or not.
If it's needed I'd change the name here to be more explicit of what this validation is referring to: `bypassUriChecksumValidation`.

src/Core/Services/CipherService.cs
```diff
@@ -1153,6 +1153,8 @@ private async Task EncryptCipherDataAsync(Cipher cipher, CipherView model, Symme
                                 Match = uri.Match
                             };
                             await EncryptObjPropertyAsync(uri, loginUri, new HashSet<string> { "Uri" }, key);
+                            var uriHash = await _cryptoService.HashAsync(uri.Uri, CryptoHashAlgorithm.Sha256);
+                            loginUri.UriChecksum = await _cryptoService.EncryptAsync(uriHash, key);
```
fedemkr, line 1157: 🤔 From what I see in `Cipher -> DecryptAsync(...)` the checksum is only taken account when the cipher has a Cipher key. So should a check be added here to see if a cipher key exists in order to set the checksum? Or is it set either way no matter if the cipher key exists or not?

src/Core/Models/Domain/LoginUri.cs
```diff
@@ -35,5 +44,21 @@ public LoginUriData ToLoginUriData()
...
+
+            // HACK: I don't like resolving this here but I can't see a better way without
+            // refactoring a lot of things.
+            var cryptoService = ServiceContainer.Resolve<ICryptoService>();
+            var localChecksum = await cryptoService.HashAsync(clearTextUri, CryptoHashAlgorithm.Sha256);
+
+            var remoteChecksum = await this.UriChecksum.DecryptAsync(orgId, key);
+            return remoteChecksum == localChecksum;
+        }
```
fedemkr, line 62: ⛏️ Remove the use of `this` to be consistent with the app code.

❓ In case an exception happens, should this break the whole decryption or just log and return `false`?

🤔 I don't like resolving that there neither. I was thinking about resolving it in one the parents and then pass it along here, so if the user has many Uris it doesn't have to resolve it many times; but it's also dirty. Also thought about moving it to a field and make it `LazyResolve` but I think this happens only once per instance so maybe it doesn't help much. The only thing that comes to mind to improve a bit the performance would be to move this check to the `CryptoService` and then resolve such service in the `Login` parent and call it from there in the for, so the service is resolved just once for all Uris of one login, furthermore we don't pollute even more the models.
```c#
// In `ICryptoService`
Task<bool> ValidateUriChecksumAsync(EncString remoteUriChecksum, string rawUri, string orgId, SymmetricCryptoKey key)
```
Then in `Login`
```c#
if (Uris != null)
{
    var cryptoService = ServiceContainer.Resolve<ICryptoService>();
    view.Uris = new List<LoginUriView>();
    foreach (var uri in Uris)
    {
        view.Uris.Add(await uri.DecryptAsync(orgId, key));
        var loginUriView = await uri.DecryptAsync(orgId, key);
        if (bypassValidation || await cryptoService.ValidateUriChecksumAsync(uri.UriChecksum, loginUriView.Uri, orgId, key))
            {
            view.Uris.Add(loginUriView);
        }
    }
}
```

src/Core/Models/Domain/Cipher.cs
LRNcardozoWDF: I believe there are cases where the `model.key` is different than `null` but isn't a `Key` from IVIE, so we would get false positives.

fedemkr: alright then changing the name would suffice in this case 😄

src/Core/Services/CipherService.cs
LRNcardozoWDF: Only the validation is tied to the existence of the cipher key, we can set the checksum either way.

src/Core/Services/CryptoService.cs
```diff
@@ -730,6 +730,26 @@ public async Task UpdateMasterKeyAndUserKeyAsync(MasterKey masterKey)
...
+        }
+
+        public async Task<bool> ValidateUriChecksumAsync(EncString remoteUriChecksum, string rawUri, string orgId, SymmetricCryptoKey key)
+        {
+            if (remoteUriChecksum == null)
+            {
+                return false;
+            }
+
+            //var cryptoService = ServiceContainer.Resolve<ICryptoService>();
```
fedemkr, line 746: ⛏️ Remove comment
```suggestion
```

fedemkr (changes requested):

src/Core/Models/Domain/Login.cs
```diff
@@ -31,7 +33,7 @@ public Login(LoginData obj, bool alreadyEncrypted = false)
         public EncString Totp { get; set; }
         public List<Fido2Credential> Fido2Credentials { get; set; }
 
-        public async Task<LoginView> DecryptAsync(string orgId, SymmetricCryptoKey key = null)
+        public async Task<LoginView> DecryptAsync(string orgId, bool bypassValidation, SymmetricCryptoKey key = null)
```
fedemkr, line 36: ⛏️ Rename to be explicit of what validation are you going to bypass:
```suggestion
        public async Task<LoginView> DecryptAsync(string orgId, bool bypassUriChecksumValidation, SymmetricCryptoKey key = null)
```

src/Core/Services/CryptoService.cs
```diff
@@ -29,10 +30,12 @@ public class CryptoService : ICryptoService
 
         public CryptoService(
             IStateService stateService,
-            ICryptoFunctionService cryptoFunctionService)
+            ICryptoFunctionService cryptoFunctionService,
+            ILogger logger)
```
fedemkr, line 34: ⚠️ Is this compiling?  I don't see the change on `iOSCoreHelpers` and `MainApplication` to pass the `logger` instance to the initialization of the `CryptoService`

fedemkr (changes requested):

src/Core/Services/CryptoService.cs
LRNcardozoWDF: Fixed in commit [fc8d058](https://github.com/bitwarden/mobile/pull/2893/commits/fc8d058046e23004a7c062c1f384984b959a9547)

fedemkr (approved):
🎉

fedemkr (dismissed):

fedemkr (approved):

PR #2909, opened 2023-12-19
[PM-4584] Login with Device is sending a notification to the initiating device

andrebispo5:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix bug when the user starts a login with device, the notification is appearing on the current device as well.

## Code changes

Add "Device-Identifier" header so it can be excluded server side.

## Before you submit

### DIFF
--- a/src/Core/Services/ApiService.cs
+++ b/src/Core/Services/ApiService.cs
@@ -593,7 +593,8 @@ namespace Bit.Core.Services
 
         public Task<PasswordlessLoginResponse> PostCreateRequestAsync(PasswordlessCreateLoginRequest passwordlessCreateLoginRequest, AuthRequestType authRequestType)
         {
-            return SendAsync<object, PasswordlessLoginResponse>(HttpMethod.Post, authRequestType == AuthRequestType.AdminApproval ? "/auth-requests/admin-request" : "/auth-requests", passwordlessCreateLoginRequest, authRequestType == AuthRequestType.AdminApproval, true);
+            return SendAsync<object, PasswordlessLoginResponse>(HttpMethod.Post, authRequestType == AuthRequestType.AdminApproval ? "/auth-requests/admin-request" : "/auth-requests", passwordlessCreateLoginRequest, authRequestType == AuthRequestType.AdminApproval, true,
+                (message) => message.Headers.Add("Device-Identifier", passwordlessCreateLoginRequest.DeviceIdentifier));
         }
 
         public Task<PasswordlessLoginResponse> PutAuthRequestAsync(string id, string encKey, string encMasterPasswordHash, string deviceIdentifier, bool requestApproved)


src/Core/Services/ApiService.cs
```diff
@@ -593,7 +593,8 @@ public Task<PasswordlessLoginResponse> GetAuthResponseAsync(string id, string ac
 
         public Task<PasswordlessLoginResponse> PostCreateRequestAsync(PasswordlessCreateLoginRequest passwordlessCreateLoginRequest, AuthRequestType authRequestType)
         {
-            return SendAsync<object, PasswordlessLoginResponse>(HttpMethod.Post, authRequestType == AuthRequestType.AdminApproval ? "/auth-requests/admin-request" : "/auth-requests", passwordlessCreateLoginRequest, authRequestType == AuthRequestType.AdminApproval, true);
+            return SendAsync<object, PasswordlessLoginResponse>(HttpMethod.Post, authRequestType == AuthRequestType.AdminApproval ? "/auth-requests/admin-request" : "/auth-requests", passwordlessCreateLoginRequest, authRequestType == AuthRequestType.AdminApproval, true,
+                (message) => message.Headers.Add("Device-Identifier", passwordlessCreateLoginRequest.DeviceIdentifier));
```
fedemkr, line 597: 🤔 Is it `Device-Identifier` or `X-Device-Identifier` the proper header to add? Because I see [this](https://github.com/bitwarden/mobile/blob/main/src/Core/Services/ApiService.cs#L405C15-L405C15) on `GetKnownDevice` and has an `X-` as a prefix. if we should use the same, please extract this into a general constant that can be reused.

andrebispo5: It's `Device-Identifier`, server side the `GetKnownDevice` is using an inline method annotation `FromHeader` [link](https://github.com/bitwarden/server/blob/main/src/Api/Controllers/DevicesController.cs#L209).
Here i'm taking advantage of user context. [link](https://github.com/bitwarden/server/blob/main/src/Core/Context/CurrentContext.cs#L71).

fedemkr: In both cases it is a header, I was wondering if we should always use the same name to be consistent, if it's referring to the same thing. Maybe you changed the name not to fire some logic, if that's the case, it's fine.

andrebispo5: I had to use `Device-Identifier` to use server CurrentContext functionality to exclude the current device from receiving the notification .  
The `SendMessageAsync` has a parameter to exclude current context. [link](https://github.com/bitwarden/server/blob/main/src/Core/Services/Implementations/NotificationsApiPushNotificationService.cs#L193)

fedemkr (approved):

PR #2912, opened 2023-12-21
[PM-4760] Admin Recovery Permissions prompted to set MP.

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

If elevating user’s role to Custom role with the “Manage Account Recovery” permission, the user needs to be prompted to set a master password. 
This check is made on Sync.

## Code changes

Add force password reset check on sync complete. This check also runs on TabsPage OnAppearing, we will leverage existing broadcast subscription to "synccomplete" to also check for password reset. 

## Before you submit


src/App/Pages/TabsPage.cs
```diff
@@ -94,6 +94,7 @@ protected override async void OnAppearing()
                 if (message.Command == "syncCompleted")
                 {
                     Device.BeginInvokeOnMainThread(async () => await UpdateVaultButtonTitleAsync());
+                    await ForcePasswordResetIfNeededAsync();
```
fedemkr, line 97: ⚠️ Keep in mind that given that you're in an action in the broadcaster subscription, this act as an `async void` so if `ForcePasswordResetIfNeededAsync()` throws an exception it'll crash the application. I don't know if that's what you expect, but I'd think that we should wrap this into a try...catch and log the exception and then show an alert to the user that an error happened and maybe log the user out, IDK which would be the appropriate flow if this method throws.

fedemkr (changes requested):

src/App/Pages/TabsPage.cs
andrebispo5: added a exception log if that ever occurred.

src/App/Pages/TabsPage.cs
```diff
@@ -94,7 +94,14 @@ protected override async void OnAppearing()
...
                 {
                     Device.BeginInvokeOnMainThread(async () => await UpdateVaultButtonTitleAsync());
-                    await ForcePasswordResetIfNeededAsync();
+                    try
+                    {
+                        await ForcePasswordResetIfNeededAsync();
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Value.Exception(ex);
```
fedemkr, line 103: @andrebispo5 if an error happens here, should the user be redirected somewhere or is it fine to just ignore the force password reset?

andrebispo5: Like we are doing inside the `ForcePasswordResetIfNeededAsync` method, when there is an error we will only log it.

fedemkr (dismissed):

fedemkr (approved):

PR #2933, opened 2024-01-08
[PM-5633] iOS app crashing on startup

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix app crash on iOS due to an iOS CollectionView bug. Ignoring exception to collect more data, as of now it is very hard to  reproduce.
This will help reducing errors from: https://github.com/bitwarden/mobile/issues/2478 

## Code changes

Ignore ArgumentOutOfRangeException and let the app continue execution. Since this is at startup, the bug must be happening on an empty state or a transition state and recover after the error. 

## Before you submit

### DIFF
--- a/src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
+++ b/src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
@@ -38,10 +38,11 @@ namespace Bit.iOS.Core.Renderers.CollectionView
                     // Do nothing in here, this is temporary to get more info about the crash, if the logger fails, we want to get the info
                     // by crashing with the original exception and not the logger one
                 }
-                if (ex is IndexOutOfRangeException)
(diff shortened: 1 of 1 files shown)


src/iOS.Core/Renderers/CollectionView/ExtendedGroupableItemsViewController.cs
```diff
@@ -38,10 +38,11 @@ protected override void UpdateTemplatedCell(TemplatedCell cell, NSIndexPath inde
                     // Do nothing in here, this is temporary to get more info about the crash, if the logger fails, we want to get the info
                     // by crashing with the original exception and not the logger one
                 }
-                if (ex is IndexOutOfRangeException)
+                if (ex is IndexOutOfRangeException || ex is ArgumentOutOfRangeException)
```
fedemkr, line 41: ℹ️ Bare in mind that this will apply only to `CollectionView`s that have `ExtraDataForLogging` set, if this is fine then approving.

fedemkr (approved):

PR #2948, opened 2024-01-22
[PM-5461] Fix item saving with blank URI

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix defect [PM-5461](https://bitwarden.atlassian.net/browse/PM-5461) where saving items with blank URI field would cause the mobile app to crash while creating the uri checksum value.

## Code changes

* **CipherService.cs:** Remove from the model the URI views without URI.

## Before you submit

### DIFF
--- a/src/Core/Services/CipherService.cs
+++ b/src/Core/Services/CipherService.cs
@@ -1146,6 +1146,7 @@ namespace Bit.Core.Services
                     if (model.Login.Uris != null)
                     {
                         cipher.Login.Uris = new List<LoginUri>();
+                        model.Login.Uris.RemoveAll(u => u.Uri == null);
                         foreach (var uri in model.Login.Uris)
                         {
                             var loginUri = new LoginUri


src/Core/Services/CipherService.cs
```diff
@@ -1146,6 +1146,7 @@ private async Task EncryptCipherDataAsync(Cipher cipher, CipherView model, Symme
                     if (model.Login.Uris != null)
                     {
                         cipher.Login.Uris = new List<LoginUri>();
+                        model.Login.Uris.RemoveAll(u => u.Uri == null);
```
fedemkr, line 1149: @LRNcardozoWDF do we need to remove them? can't we just filter them in the foreach?
`model.Login.Uris.Where(u => u.Uri != null)`
Does the code work correctly if `u.Uri` is empty or whitespace?

LRNcardozoWDF: The result is the same either by removing or filtering, the item is saved without the empty or whitespaced URIs, do you think filter approach would be better?
Yes, it worked correctly for `u.Uri` empty or whitespace, for IVIE turned off and on.

fedemkr: I think the code is simpler with the filtering and would have better performance as the list would have to be traversed once (with the remove, the list needs to be traversed twice, once for remove and the other one for the for loop). However, it would be really little gain performance wise unless the list is really long, which I don't think it'd be a case for login Uris. I'd go with filtering so we have less lines of code and less list mutations.
If you have any strong reason to go with `RemoveAll` over filtering please let me know, perhaps I'm missing something.

LRNcardozoWDF: Fixed in 06d1ad9587d10efe5ce3e7bcae9d1978f2e575f5

fedemkr (approved):

PR #2956, opened 2024-01-25
[PM-4615] [PM-6217] Add new DUO frameless 2fa flow

andrebispo5:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

DUO has update their SDK and is moving away from iframe implementation. In order to comply with the new frameless (redirect-based) logic, a new flow was added to the 2FA Viewmodel now leveraging `WebAuthenticator` for the callback values.

## Code changes

* **src/App/Pages/Accounts/TwoFactorPage.xaml:** Updated UI with labels and a button to start the DUO flow.
* **src/App/Pages/Accounts/TwoFactorPageViewModel.cs** Added new `DuoFramelessAuthenticateAsync` method to keep the frameless flow logic. This flow is only executed if the server sends `AuthUrl` as one of the `providerData` elements.

## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/a4a96466-995d-48f9-97fe-4f1a9272c92a

## Before you submit


djsmith85 (changes requested):
@andrebispo5 Please remove any file-changes under the `Resources`-folder other than `AppResources.Designer.cs` and `AppResources.resx`.

Any changes related to translations will be handled by Crowdin.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
...
+            }
+            catch (TaskCanceledException)
+            {
+                // user canceled
+                await _deviceActionService.HideLoadingAsync();
+                return;
+            }
+
+            string response = null;
+            if (authResult != null && authResult.Properties.TryGetValue("code", out var resultData))
```
fedemkr, line 249: ❓ Can `Properties` be `null` here? If so, add a check.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
...
+                });
+            }
+            catch (TaskCanceledException)
+            {
+                // user canceled
+                await _deviceActionService.HideLoadingAsync();
+                return;
+            }
+
+            string response = null;
```
fedemkr, line 248: 🎨 I'd improve the name `response` here to mean something more particular. Like for what I can see is something related with `code` given the logic below, right?

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
...
+            if (authResult != null && authResult.Properties.TryGetValue("code", out var resultData))
+            {
+                response = Uri.UnescapeDataString(resultData);
+            }
+            if (!string.IsNullOrWhiteSpace(response))
+            {
+                Token = response;
+                await SubmitAsync(false);
```
fedemkr, line 256: 🤔 I'm a bit concerned here on how the interaction is when the loading is being presented and then after it goes to `SubmitAsync` the user suddenly doesn't have internet connection which internally it will try to [show a dialog](https://github.com/bitwarden/mobile/blob/6eb7941d6ca4540f5be904095b9cee538cc10813/src/App/Pages/Accounts/TwoFactorPageViewModel.cs#L347) and perhaps it could be a problem with both of them trying to be shown at the same time. Could you test that? I know this may be also an issue with all the other providers as well so maybe the problem is bigger than just here.
If so, we should hide the loading before attempting to show any dialogs inside the `SubmitAsync(...)` method

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
...
+            await _deviceActionService.HideLoadingAsync();
+            if (authResult != null && authResult.Properties.TryGetValue("error", out var resultError))
+            {
+                await _platformUtilsService.ShowDialogAsync(resultError, AppResources.AnErrorHasOccurred,
+                    AppResources.Ok);
+            }
+            else
+            {
+                await _platformUtilsService.ShowDialogAsync(AppResources.AnErrorHasOccurred, AppResources.Ok);
+            }
```
fedemkr, line 269: 🤔 IMO it's easier to follow when errors are first and we let the main flow continue on the main branch of the code, so I'd invert the code and move this part above inside `if (string.IsNullOrWhiteSpace(response))` (inverting the condition) and put the next here.
```c#
Token = response;
await SubmitAsync(false);
```
❓ Additionally, is `resultError` human readable and localized?

fedemkr (changes requested):
❓ I'm fine with _DUO frameless_ but does _DUO_ have a formal name for the flow without using frame? If so we should use the official name here as well.

andrebispo5:
> ❓ I'm fine with _DUO frameless_ but does _DUO_ have a formal name for the flow without using frame? If so we should use the official name here as well.

We went with the frameless naming because of DUO uses it to refer a specific universal prompt type. The frameless naming is being used it both mobile and clients repos.

src/App/Pages/Accounts/TwoFactorPage.xaml
```diff
@@ -130,14 +130,47 @@
...
-                <StackLayout StyleClass="box" VerticalOptions="End">
+                    MinimumHeightRequest="400"
+                    IsVisible="{Binding IsDuoFrameless, Converter={StaticResource inverseBool}}"/>
+                <StackLayout
+                    StyleClass="box"
+                    VerticalOptions="End"
+                    IsVisible="{Binding IsDuoFrameless, Converter={StaticResource inverseBool}}">
                     <StackLayout StyleClass="box-row, box-row-switch">
                         <Label
                             Text="{u:I18n RememberMe}"
```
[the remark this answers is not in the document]
andrebispo5: Thank i didn't put much thought to it because my brain saw the remember me in different places of the stack and immediately assumed there was no way around it. Silly me. 
Like you suggested used the visibility property and put all elements under the same stack. Added Margin to frameless items to keep their placement where it should be.

djsmith85 (dismissed):
Translation-related changes are looking good. @andrebispo5 Thank you.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
             ShowContinue = !(SelectedProviderType == null || DuoMethod || Fido2Method);
         }
 
+        private async Task DuoFramelessAuthenticateAsync()
+        {
+            await _deviceActionService.ShowLoadingAsync(AppResources.Validating);
+            var providerData = _authService.TwoFactorProvidersData[SelectedProviderType.Value];
+            var url = providerData["AuthUrl"] as string;
```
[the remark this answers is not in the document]
andrebispo5: Used the `HandleException` method to show a generic message to the user and log an error message to appcenter.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -208,6 +224,51 @@ public void Load()
...
+            {
+                authResult = await WebAuthenticator.AuthenticateAsync(new WebAuthenticatorOptions
+                {
+                    Url = new Uri(url),
+                    CallbackUrl = new Uri("bitwarden://duo-callback")
+                });
+            }
+            catch (TaskCanceledException)
+            {
+                // user canceled
```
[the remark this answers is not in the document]
andrebispo5: The timeout is shown within the webview and the user has to manually close it.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
andrebispo5: Yup it's a problem because when it shows the internet error dialog it never dismisses the old loading dialog. Change the code to first hide the loading and then call the `SubmitAsync(true)`

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
andrebispo5: Agreed! Moved error handling up before the continue of the flow. Also added some error logging and handling. Let me know what you think.

andrebispo5: Since I can't be sure of the `resultError` readability, I logged the error message and shown the user a generic error only.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -245,28 +267,32 @@ private async Task DuoFramelessAuthenticateAsync()
...
-                await SubmitAsync(false);
+                _logger.Error(resultError);
+                await _platformUtilsService.ShowDialogAsync(AppResources.AnErrorHasOccurred, AppResources.Ok);
                 return;
             }
 
-            await _deviceActionService.HideLoadingAsync();
-            if (authResult != null && authResult.Properties.TryGetValue("error", out var resultError))
+            string code = null;
+            if (authResult.Properties.TryGetValue("code", out var resultData))
```
fedemkr, line 284: ❓ Can `Properties` be `null` here?

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -245,28 +267,32 @@ private async Task DuoFramelessAuthenticateAsync()
...
-                await _platformUtilsService.ShowDialogAsync(resultError, AppResources.AnErrorHasOccurred,
-                    AppResources.Ok);
+                code = Uri.UnescapeDataString(resultData);
             }
-            else
+
+            if (string.IsNullOrWhiteSpace(code))
             {
-                await _platformUtilsService.ShowDialogAsync(AppResources.AnErrorHasOccurred, AppResources.Ok);
+                throw new ArgumentException("Duo authentication error: response code is null");
```
fedemkr, line 291:
```suggestion
                throw new ArgumentException("Duo authentication error: response code is null or empty/whitespace");
```

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -245,28 +267,32 @@ private async Task DuoFramelessAuthenticateAsync()
                 return;
             }
 
-            string response = null;
-            if (authResult != null && authResult.Properties.TryGetValue("code", out var resultData))
+            await _deviceActionService.HideLoadingAsync();
```
fedemkr, line 270: 🤔 How does it feel from a user perspective? Because doing it like this, the loading will disappear and after a brief time it'll appear again on the screen. Maybe it's happening too fast so it's not a big deal but wanted to ask just in case.
If it's noticeable and an issue (I'd ask product just in case), then I'd only hide the loading just before showing another dialog that we know the flow stops there so the user won't notice anything weird about it. Furthermore, you could call `SubmitAsync(false)` again, and check inside whether a loading is already being presented to hide it before displaying a another dialog.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
andrebispo5: this check is already been done on ln 271. Since it was used twice i did the check before hand.

src/App/Pages/Accounts/TwoFactorPageViewModel.cs
andrebispo5: https://github.com/bitwarden/mobile/assets/4648522/50a6e218-c974-4ec6-9442-ca8779de6ba3

It happens so fast, I don't think that will be an issue.

fedemkr (approved):

fedemkr (approved):

fedemkr (approved):

PR #2967, opened 2024-01-29
[PM-5910] Workaround for for sliding elements in Duo 2FA flow

mpbw2:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

To stop the sliding bottom layout on the Duo 2FA page we're now setting a fixed height for the webview based on a simple measurement.  The gap at the bottom can vary depending on the device height (screenshots below), but thankfully Duo's new [frameless implementation](https://github.com/bitwarden/mobile/pull/2956) will demote this flow to an edge case.

## Code changes

* **TwoFactorPage/Viewmodel:** Added height property and calculation using `DeviceDisplay.MainDisplayInfo` height & density

## Screenshots

Screenshot 2024-01-29 at 12 38 59 PM
Screenshot 2024-01-29 at 12 39 11 PM

In the rare case we aren't able to obtain the available height for measuring, we default to 400dp as a safe minimum as seen here:

Screenshot 2024-01-29 at 12 57 25 PM
Screenshot 2024-01-29 at 12 58 11 PM

## Before you submit

### DIFF
--- a/src/Core/Pages/Accounts/TwoFactorPageViewModel.cs
+++ b/src/Core/Pages/Accounts/TwoFactorPageViewModel.cs
@@ -34,6 +34,7 @@ namespace Bit.App.Pages
         private string _webVaultUrl = "https://vault.bitwarden.com";
         private bool _enableContinue = false;
         private bool _showContinue = true;
+        private double _duoWebViewHeight;
 
         public TwoFactorPageViewModel()
         {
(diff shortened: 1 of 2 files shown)


src/Core/Pages/Accounts/TwoFactorPageViewModel.cs
```diff
@@ -172,6 +179,7 @@ public void Load()
                     break;
                 case TwoFactorProviderType.Duo:
                 case TwoFactorProviderType.OrganizationDuo:
+                    SetDuoWebViewHeight();
```
fedemkr, line 182: 🎨 You can do this only on Android:
```c#
 #if ANDROID
    SetDuoWebViewHeight();
#endif
```

fedemkr (approved):

PR #2973, opened 2024-02-01
[PM-5903] Fix Account Switcher visible on Android when no Accounts are logged in

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Fix Account Switcher visible on Android when no Accounts are logged in.
This is happening because on Android `App.xaml.cs.SetOptions()` can get called after `App.xaml.cs` already has Options set. In this scenario it would get replaced and specifically the property `HideAccountSwitcher` will have an incorrect value of `false`.

## Code changes
* **App.xaml.cs:** Update the `SetOptions()` method to only update the properties that we need from Android instead of replacing the entire object. Also added some comments and changed the method name to `SetAndroidOptions()` to make it more clear.

## Before you submit

### DIFF
--- a/src/Core/App.xaml.cs
+++ b/src/Core/App.xaml.cs
@@ -85,10 +85,27 @@ namespace Bit.App
             }
         }
 
-        //Allows setting Options from MainActivity before base.OnCreate
-        public void SetOptions(AppOptions appOptions)
+        /// <summary>
+        /// Allows setting Options from MainActivity before base.OnCreate
+        /// Note 1: This is only be used by Android due to way it's Lifecycle works
+        /// Note 2: This method does not replace existing Options in App.xaml.cs if it exists already.
+        ///         It only updates properties in Options related with Autofill/CreateSend/etc..
+        /// </summary>
+        /// <param name="appOptions">Options created in Android MainActivity.cs</param>
+        public void SetAndroidOptions(AppOptions appOptions)
         {
-            Options = appOptions ?? new AppOptions();
+            if (Options == null)
+            {
+                Options = appOptions ?? new AppOptions();
+            }
+            else if(appOptions != null)
+            { 
+                Options.Uri = appOptions.Uri;
+                Options.MyVaultTile = appOptions.MyVaultTile;
+                Options.GeneratorTile = appOptions.GeneratorTile;
+                Options.FromAutofillFramework = appOptions.FromAutofillFramework;
+                Options.CreateSend = appOptions.CreateSend;
+            }
         }
 
         protected override Window CreateWindow(IActivationState activationState)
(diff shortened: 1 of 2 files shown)


mpbw2 (approved):

src/Core/App.xaml.cs
```diff
@@ -85,10 +85,27 @@ public static ResumeWindow CurrentWindow
...
+            }
+            else if(appOptions != null)
+            { 
+                Options.Uri = appOptions.Uri;
+                Options.MyVaultTile = appOptions.MyVaultTile;
+                Options.GeneratorTile = appOptions.GeneratorTile;
+                Options.FromAutofillFramework = appOptions.FromAutofillFramework;
+                Options.CreateSend = appOptions.CreateSend;
+            }
         }
```
fedemkr, line 109: ⛏️ If this method is only for Android then I'd wrap it in `#if ANDROID` so this method is not added on iOS where it's not going to be used and also if for some reason someone calls it from iOS with the precompiled directive it will throw an error on compilation.
🤔 Additionally, @mpbw2 can't remember  but is there a reason why we are manually assigning the properties instead of just using `void SetAllFrom(AppOptions o)`? Like `Options.SetAllFrom(appOptions)`; are we afraid some property will be overridden with a wrong value?

mpbw2: @fedemkr That's my impression based on how the original (Xamarin) flow worked.  It was a safety measure to prevent mixing of Options "states".  Is that correct @dinisvieira ?

dinisvieira: @fedemkr @mpbw2 I did it differently only because I completely missed the `SetAllFrom()` method. 😄 
That said, it would still result in a bug in MAUI because it's going to replace the `HideAccountSwitcher` property (possibly others?).
I could still improve this by adding a new `SetAutofillOptions()` method with this code inside the `AppOptions` class wrapped in a` #if ANDROID`
What do you think?

dinisvieira: @fedemkr @mpbw2 Oh wait, I just remembered that it can be null in some initialization scenarios.
That's likely why I was using the `SetOptions` directly on `App.xaml.cs`.
It's also already wrapped in a `#if ANDROID`
Should I just rename it back to `SetOptions`? (instead of `SetAndroidOptions`)

mpbw2: > Should I just rename it back to `SetOptions`?

The more I look at it the more I think yes, `SetOptions` is better

fedemkr: Oh missed the outer `#if ANDROID` sorry. Thanks for the info! It's ok as it is, let's leave it like this to avoid introducing secondary effects.

fedemkr (approved):

src/Core/App.xaml.cs
fedemkr: About naming, whatever you feel is better is fine by me, it's only going to be used on Android so both options makes good sense to me.

mpbw2: @dinisvieira Just leave it as is 👍

PR #2981, opened 2024-02-06
[PM-5906] Fix for incorrect Send MaxAccess white text color on label when using light theme on iOS

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Fix for incorrect Send MaxAccess white text color on label when using light theme on iOS.
This fix is a workaround for the MAUI Issue https://github.com/dotnet/maui/pull/20100

## Code changes
* **SendAddEditPage.xaml.cs:** Set the TextColor to null and back to the correct color whenever we try to programatically change the entry text. Doing this avoids the bug.

## Before you submit

### DIFF
--- a/src/Core/Pages/Send/SendAddEditPage.xaml.cs
+++ b/src/Core/Pages/Send/SendAddEditPage.xaml.cs
@@ -181,6 +181,17 @@ namespace Bit.App.Pages
 
         private void OnMaxAccessCountTextChanged(object sender, TextChangedEventArgs e)
         {
+            var maxAccessEntry = (Microsoft.Maui.Controls.Entry)sender;
+
+#if IOS
+            //Workaround: To avoid a bug that incorrectly sets the TextColor when changing text
+            // programatically we need to set it to null and back to the correct color
+            // MAUI issue https://github.com/dotnet/maui/pull/20100
+            maxAccessEntry.TextColor = null;
+            var color = ThemeManager.GetResourceColor("TextColor");
+            maxAccessEntry.TextColor = color;
+#endif
+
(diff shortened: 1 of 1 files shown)


src/Core/Pages/Send/SendAddEditPage.xaml.cs
```diff
@@ -181,6 +181,17 @@ private async void FileType_Clicked(object sender, EventArgs eventArgs)
...
+            var maxAccessEntry = (Microsoft.Maui.Controls.Entry)sender;
+
+#if IOS
+            //Workaround: To avoid a bug that incorrectly sets the TextColor when changing text
+            // programatically we need to set it to null and back to the correct color
+            // MAUI issue https://github.com/dotnet/maui/pull/20100
+            maxAccessEntry.TextColor = null;
+            var color = ThemeManager.GetResourceColor("TextColor");
+            maxAccessEntry.TextColor = color;
+#endif
```
fedemkr, line 193: 🎨 No need for the extra variable, and `HACK` might be better given that it's tracked by Tasks:
```suggestion
#if IOS
            // HACK: To avoid a bug that incorrectly sets the TextColor when changing text
            // programatically we need to set it to null and back to the correct color
            // MAUI issue https://github.com/dotnet/maui/pull/20100
            maxAccessEntry.TextColor = null;
            maxAccessEntry.TextColor = ThemeManager.GetResourceColor("TextColor");
#endif
```

fedemkr (changes requested):

fedemkr (approved):

PR #2984, opened 2024-02-07
[PM-5913] Fix for MAUI not reporting the correct Theme when resuming an iOS App

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
We want to implement a fix/workaround for MAUI not reporting the correct Theme when resuming an iOS App.

## Code changes

* **ThemeManager.cs:** Added iOS/Android specific implementations for `OsDarkModeEnabled()` so that we can directly fetch the Theme on the iOS API for the iOS implementation.

## Before you submit


src/Core/Utilities/ThemeManager.cs
```diff
@@ -147,16 +150,41 @@ public static string GetAutoDarkTheme()
...
+
+                    requestedTheme = uiStyle switch
+                    {
+                        UIUserInterfaceStyle.Light => AppTheme.Light,
+                        UIUserInterfaceStyle.Dark => AppTheme.Dark,
+                        _ => AppTheme.Unspecified
+                    };
+                }
+            });
+            return requestedTheme == AppTheme.Dark;
```
fedemkr, line 185: ⚠️ `BeginInvokeOnMainThread` [dispatches the action to be executed](https://github.com/dotnet/maui/blob/main/src/Essentials/src/MainThread/MainThread.ios.tvos.watchos.macos.cs#L13) as soon as main thread is available to be executed, basically if on main thread right away if not queue it. So this can cause a race condition issue where the action to update the `requestedTheme` may take longer than executing the next line and returning... So unless I'm missing something this may fail on some occasions.
Also which part of the code needs the Main thread to be executed? If entirely necessary, we need to execute on main thread up the chain in the callers or transform everything into async/await and use `InvokeOnMainThreadAsync` which will in fact wait until the end of the action before executing the next line.
🎨 Additionally, there is no need for the whole check of TvOS, we just need to check that the iOS version is at least 13, we're already inside an `#if iOS` block.
⛏️ Finally, the methods are named the same both in Android and iOS, so you can have the `#if...#elif...#endif` right inside the method so you don't have to repeat the signature line twice.

fedemkr (changes requested):

src/Core/Utilities/ThemeManager.cs
dinisvieira: @fedemkr TVOs and using single signature changes make total sense. I'll make those.

For the MainThread I was trying to avoid using async so that we don't have to change multiple signatures to async Task.
What do you think of using the [internal method](https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Essentials/src/MainThread/MainThread.ios.tvos.watchos.macos.cs#L16) from MAUI? (which is what MAUI [itself uses in this case](https://github.com/dotnet/maui/blob/0da0a61a7e3f127bcab8a800dfeda4101a0d15bc/src/Essentials/src/AppInfo/AppInfo.ios.tvos.watchos.macos.cs#L58))
And use it also only for that specific `WindowStateManager.Default.GetCurrentUIViewController()` call.
If yes, where do you recommend adding it?

fedemkr: Yes, that's perfect; already checked the API and synchronously waits for the action to finish on the main thread so we can use it (we only have to check no long-running operation are done inside or we can cause a freeze or deadlock but I think in our case we don't have these 😄 )

dinisvieira: @fedemkr I updated code for you to take a look but still have the `InvokeOnMainThread` in there. Do you have some Extension/Helper/Util class somewhere where do you prefer to have this method?

PS: Lot's of changes and conflicts likely due to the merge from `maui-migration` to `main` right?

dinisvieira:
Closing PR to open new one. (just to avoid fixing several conflicts with main)

PR #2989, opened 2024-02-09
[PM-5900] Fix for Device Login push notifications cause the app to show HomePage

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
 Fix for Device Login push notifications causing the app to show HomePage.
 MAUI is treating the Intent on the push notification as a triggering the creation of a new Activity and we just want to open the existing one.

## Code changes
* **AndroidPushNotificationService.cs:** Changed the Intent when tapping the push notification to launch the existing activity instead of creating a new one.

## Before you submit

### DIFF
--- a/src/App/Platforms/Android/Services/AndroidPushNotificationService.cs
+++ b/src/App/Platforms/Android/Services/AndroidPushNotificationService.cs
@@ -79,7 +79,7 @@ namespace Bit.Droid.Services
             }
             
             var context = Android.App.Application.Context;
-            var intent = new Intent(context, typeof(MainActivity));
+            var intent = context.PackageManager?.GetLaunchIntentForPackage(context.PackageName ?? string.Empty);
             intent.PutExtra(Bit.Core.Constants.NotificationData, JsonConvert.SerializeObject(data));
             var pendingIntentFlags = AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.UpdateCurrent, true);
             var pendingIntent = PendingIntent.GetActivity(context, 20220801, intent, pendingIntentFlags);


fedemkr (commented):
🤔  Is there a flow where `context.PackageName` is `null`? If so, which will be the behavior of the `intent`? Will it crash? Will it do nothing? Should we ignore creating the `pendingIntent`?
⚠️ I see `context.PackageManager?....` which means it can be `null`, so potentially `intent` could be `null`. So I have same questions as above.
cc: @mpbw2

dinisvieira:
@fedemkr I have used this before with no issues when searching for the package of the app itself.
It can definitely be null if we are searching for another app package.
That said, I'm not 100% sure so I'll wait for @mpbw2 feedback. We could always have a fallback value of `new Intent(context, typeof(MainActivity));` even though it will have the original bug.

mpbw2:
@fedemkr It _shouldn't_ be null when referring to itself.  That said if we end up passing an empty string because it's null, in the past I believe an intent like that would simply bring its parent app to the forefront which is what we're shooting for anyway.  I'm not sure if that behavior has changed or if MAUI is interfering in some way though, might be worth testing @dinisvieira 

As for the rest there's not much we can do if the intent itself ends up being null.  Ideally we should just let the notification self-dismiss and do nothing, which is preferred to the user landing on the wrong auth page.  They'll end up manually opening the app anyway, then be prompted with the device login after unlocking - a vast improvement over the current broken flow.

dinisvieira:
@mpbw2 Notification is not even shown if we pass a `string.Empty` to` context.PackageManager?.GetLaunchIntentForPackage()`
But no crash or weird behavior occurs.
Should I go ahead and implement an edge case scenario where if `context.PackageManager==null` or `context.PackageName==null` or `intent==null` I just don't set "SetContentIntent()" or "SetDeleteIntent"?
(which still lets the notification to appear but tapping on it does nothing)

mpbw2:
> Should I go ahead and implement an edge case scenario where if context.PackageManager==null or context.PackageName==null or intent==null I just don't set "SetContentIntent()" or "SetDeleteIntent"?
(which still lets the notification to appear but tapping on it does nothing)

@dinisvieira Yeah I think that's a good idea.

dinisvieira:
@mpbw2 I have updated the code as discussed. Does it look good?

mpbw2 (approved):
@dinisvieira Perfect!

PR #3012, opened 2024-02-18
[PM-6291] Fix Device Login Pending Requests screen not displaying anything

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
The Device Login Pending Requests screen is not displaying anything.
If there are no items to show sometimes rotating the device would make the "No pending requests" message and image to appear which hints at a layout/arrange issue. When there are items to show the page never seems to load anything. 

## Code changes
Notes:
* As I started fixing issues with this screen I noticed there are multiple ones and they all seem to be related with the `RefreshView` not "layouting" itself properly when the `CollectionView` items change. I tried using `InvalidateArrange()` on several elements of the layout tree but it doesn't seem to help either.
* The changes that I found avoided this issue were having the layout directly on the Page at compile time instead of adding it to the `ContentView` in runtime and also switching the StackLayout's to Grids and that seems to "force" the `RefreshView` to take full width/height and allow the inner `CollectionView` to be able to update itself (at least on Android)
* For iOS I also had to use a background color on a full screen grid that has the activity indicator and hides the refreshview below because using `IsVisible` on the `RefreshView` "layout tree" causes issues with the layout in iOS. (in practice when making the refreshview visible the size doesn't seem to be calculating properly on iOS and having it there "behind" the activity indicator grid avoids this issue.

* **LoginPasswordlessRequestsListPage.xaml:** Moved the page xaml from the `Resources` to the Content of the page directly. Changed StackLayout's to Grids and added a new Grid for showing the `ActivityIndicator`
* **LoginPasswordlessRequestsListPage.xaml.cs:** Changes to no longer use the `BaseContentPage` `SetActivityIndicator()` and `LoadOnAppearedAsync()`
* **LoginPasswordlessRequestsListViewModel.cs:** Minor change to use `MainThread` instead of the old `Device`

## Before you submit

### DIFF
--- a/src/Core/Pages/Settings/LoginPasswordlessRequestsListPage.xaml
+++ b/src/Core/Pages/Settings/LoginPasswordlessRequestsListPage.xaml
@@ -76,12 +76,10 @@
         RowDefinitions="*, Auto"
         Padding="0, 10">
         <RefreshView
-            x:Name="_refreshView"
-            IsVisible="False"
             Grid.Row="0"
             IsRefreshing="{Binding IsRefreshing}"
             Command="{Binding RefreshCommand}"
-            VerticalOptions="FillAndExpand"
+            VerticalOptions="Fill"
             BackgroundColor="{DynamicResource BackgroundColor}">
             <Grid RowDefinitions="Auto, *">
                 <VerticalStackLayout Grid.Row="0" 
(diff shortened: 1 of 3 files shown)


fedemkr (approved):
LGTM 🎉 I'd wait for @andrebispo5 and @mpbw2  in case I'm missing anything.

mpbw2 (approved):

PR #3017, opened 2024-02-21
[PM-5154] Implement Passkeys on iOS

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Implement flows on iOS to Create or Use a Passkey.

_Note: This doesn't support Dark theme yet._

## Code changes

* **CredentialProviderViewController.Passkeys:** Passkeys part on the iOS Autofill implementation.
* **CredentialProviderViewController:** Added additional logic for new iOS 17 API and Passkeys integration.
* **Fido2Credential...:** Added `UserDisplayName`
* **CipherService:** Added method to create new login for passkey based on the relying party identifier.
* **Fido2AuthenticatorService:** Added `UserDisplayName` and updated code not to use C# 12 features.
* **ClipLogger:** Util to be used to "debug" iOS extensions where we cannot use .NET debugger.
* **Fido2MakeCredentialUserInterface:** Implementation for UI dependent passkey creation flow UI
* **LoginListViewController:** Added logic for Passkeys integration, like table headers, empty state, save passkey and cancellation.
* **LoginSearchViewController:** Added logic for Passkeys integration.
* **MainInterface.storyboard:** Updated "Logins" scene to have the empty state for passkeys.
* **HeaderItemView:** View to use as header on the table view for logins on passkeys.
* **Context:** Added helper properties to fulfill the Passkeys flow
* **BaseLoginListTableSource:** Moved logic to a base class to have row selection logic shared between the login list and the login search list.
* **ASHelpers:** Fixed `CredentialId` conversion.
* **ExtensionTableSource:** Changed some methods so we can override them to provide specific logic.

### Sign in with your passkey OS view
Save passkey as new login

### Create passkey OS view
Create passkey OS view

### Save passkey as new login
Save passkey as new login

### Save passkey flow to choose item from list
Save passkey flow to choose item from list

### Overwrite passkey dialog
Overwrite passkey dialog

## Before you submit


mpbw2 (approved):

src/Core/Services/CipherService.cs
```diff
@@ -1286,6 +1286,34 @@ async Task encryptAndAddHistoryAsync(PasswordHistoryView model, PasswordHistory
...
+                    }
+                },
+                Card = new CardView(),
+                Identity = new IdentityView(),
+                SecureNote = new SecureNoteView
+                {
+                    Type = SecureNoteType.Generic
+                },
+                Reprompt = CipherRepromptType.None
+            };
```
coroiu, line 1309: issue: can we also assign the username here to align with [PM-4661 Populate login item Username with passkey username](https://bitwarden.atlassian.net/browse/PM-4661)

src/Core/Services/Fido2AuthenticatorService.cs
```diff
@@ -316,12 +316,12 @@ private Fido2CredentialView CreateCredentialView(Fido2AuthenticatorMakeCredentia
             );
             authData.Add(flags);
 
-            authData.AddRange([
+            authData.AddRange(new List<byte> {
```
coroiu, line 319: thought: damn, I thought I had caught all of these, thanks for fixing 🙏

src/iOS.Autofill/CredentialProviderViewController.Passkeys.cs
```diff
@@ -25,52 +139,135 @@ private async Task ProvideCredentialWithoutUserInteractionAsync(ASPasskeyCredent
...
+                    }
+                }, new Fido2GetAssertionUserInterface(cipherId, true, EnsureUnlockedVaultAsync, () => Task.FromResult(true)));
+
+                var selectedUserHandleData = fido2AssertionResult.SelectedCredential != null
+                    ? NSData.FromArray(fido2AssertionResult.SelectedCredential.UserHandle)
+                    : (NSData)userHandleData;
+
+                var selectedCredentialIdData = fido2AssertionResult.SelectedCredential != null
+                    ? NSData.FromArray(fido2AssertionResult.SelectedCredential.Id)
+                    : credentialIdData;
```
coroiu, line 179: question: why do we need to null-check these? Credential id will always be defined, and user handle will either be defined in both places or null in both places?

fedemkr: @coroiu I did it at an early stage of this. If `SelectedCredential` must be non-null then should we just check if that's null we throw an exception?

src/Core/Services/CipherService.cs
fedemkr: I've added that [here](https://github.com/bitwarden/mobile/pull/3019/files#diff-2c6c72919b26d3ded8ed10ff92d929050a8b8087adffee3d5dff8b896d260b5bR52) when creating a login with the Add button but forgot about adding it to this method, will add that as part of that PR.

src/iOS.Autofill/CredentialProviderViewController.Passkeys.cs
coroiu: It is always defined so I think you can throw an exception (or just assume that it is defined)

PR #3020, opened 2024-02-21
WIP implementation for the app unlock flow when called from Passkey

dinisvieira:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
- Implement Unlock flow when launched from Passkey
- Make sure Window workarounds on Android work

## Code changes
- WIP implementation for the app unlock flow when called from Passkey.
- Also added a new Window workaround in App.xaml.cs to allow CredentialProviderSelectionActivity to launch separately.

## TODO
- Organize code in DeviceActionService.cs into some place where it can be shared with CredentialProviderService

## Before you submit

### DIFF
--- a/src/App/Platforms/Android/Autofill/CredentialProviderService.cs
+++ b/src/App/Platforms/Android/Autofill/CredentialProviderService.cs
@@ -9,6 +9,7 @@ using Bit.Core.Abstractions;
 using Bit.Core.Utilities;
 using AndroidX.Credentials.Exceptions;
 using AndroidX.Credentials.WebAuthn;
+using Bit.App.Droid.Utilities;
 using Bit.Core.Models.View;
 using Resource = Microsoft.Maui.Resource;
 
@@ -44,8 +45,22 @@ namespace Bit.Droid.Autofill
                 {
                     var response = await ProcessGetCredentialsRequestAsync(request);
                     callback.OnResult(response);
+                    return;
                 }
-                // TODO handle auth/unlock account flow
+
+                var intent = new Intent(ApplicationContext, typeof(MainActivity));
+                intent.PutExtra(CredentialProviderConstants.PasskeyFramework, true);
+                var pendingIntent = PendingIntent.GetActivity(ApplicationContext, UniqueRequestCode, intent,
+                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.UpdateCurrent, true));
+
+                var i18nService = ServiceContainer.Resolve<II18nService>("i18nService");
+                var unlockText = i18nService.T("Unlock");
+                var unlockAction = new AuthenticationAction(unlockText, pendingIntent);
+
+                var unlockResponse = new BeginGetCredentialResponse.Builder()
+                    .SetAuthenticationActions(new List<AuthenticationAction>() { unlockAction } )
+                    .Build();
+                callback.OnResult(unlockResponse);
             }
             catch (GetCredentialException e)
             {
--- a/src/App/Platforms/Android/Services/DeviceActionService.cs
+++ b/src/App/Platforms/Android/Services/DeviceActionService.cs
@@ -22,6 +22,7 @@ using Bit.App.Droid.Utilities;
 using Microsoft.Maui.Controls.Compatibility.Platform.Android;
 using Resource = Bit.Core.Resource;
 using Application = Android.App.Application;
+using static Android.Content.Res.Resources;
 
 namespace Bit.Droid.Services
 {
@@ -34,6 +35,10 @@ namespace Bit.Droid.Services
 
         private Toast _toast;
         private string _userAgent;
+
+        //TODO: These consts need to be moved somewhere else where they can be shared with the code in CredentialProviderService.cs
+        private const string GetPasskeyIntentAction = "PACKAGE_NAME.GET_PASSKEY";
+        private const int UniqueRequestCode = 94556023;
 
         public DeviceActionService(
             IStateService stateService,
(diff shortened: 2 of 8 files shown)


src/App/Platforms/Android/Autofill/CredentialProviderService.cs
```diff
@@ -44,8 +45,22 @@ public class CredentialProviderService : AndroidX.Credentials.Provider.Credentia
...
-                // TODO handle auth/unlock account flow
+
+                var intent = new Intent(ApplicationContext, typeof(MainActivity));
+                intent.PutExtra(CredentialProviderConstants.PasskeyFramework, true);
+                var pendingIntent = PendingIntent.GetActivity(ApplicationContext, UniqueRequestCode, intent,
+                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.UpdateCurrent, true));
+
+                var i18nService = ServiceContainer.Resolve<II18nService>("i18nService");
+                var unlockText = i18nService.T("Unlock");
+                var unlockAction = new AuthenticationAction(unlockText, pendingIntent);
```
fedemkr, line 58: 🎨 Use `AppResources.Unlock` instead of the `II18nService`.

src/App/Platforms/Android/Services/DeviceActionService.cs
```diff
@@ -553,6 +558,98 @@ public long GetActiveTime()
             // ref: https://developer.android.com/reference/android/os/SystemClock#elapsedRealtime()
             return SystemClock.ElapsedRealtime();
         }
+
+        public async Task ReturnToPasskeyAfterUnlock()
```
fedemkr, line 562: ⛏️ Add `Async` suffix

src/App/Platforms/Android/Services/DeviceActionService.cs
```diff
@@ -553,6 +558,98 @@ public long GetActiveTime()
...
+        {
+            var activity = (MainActivity)Microsoft.Maui.ApplicationModel.Platform.CurrentActivity;
```
fedemkr, line 564: ⚠️ Cast it using `as MainActivity` so it doesn't throw an exception and the next `if null` works as expected.

src/App/Platforms/Android/Services/DeviceActionService.cs
```diff
@@ -553,6 +558,98 @@ public long GetActiveTime()
...
+            if (activity == null)
+            {
+                return;
+            }
+
+            var request = AndroidX.Credentials.Provider.PendingIntentHandler.RetrieveBeginGetCredentialRequest(activity.Intent);
+            var response = new AndroidX.Credentials.Provider.BeginGetCredentialResponse();;
+            IList<AndroidX.Credentials.Provider.CredentialEntry> credentialEntries = null;
```
fedemkr, line 572: 🤔 To avoid casting afterwards, can we transform this to `List`?

src/Core/Utilities/AppHelpers.cs
```diff
@@ -434,6 +434,13 @@ public static bool SetAlternateMainPage(AppOptions appOptions)
                     App.MainPage = new NavigationPage(new CipherAddEditPage(appOptions: appOptions));
                     return true;
                 }
+                if (appOptions.FromPasskeyFramework)
+                {
+                    appOptions.FromPasskeyFramework = false;
+                    var deviceActionService = Bit.Core.Utilities.ServiceContainer.Resolve<IDeviceActionService>();
+                    deviceActionService.ReturnToPasskeyAfterUnlock().FireAndForget();
+                    return true;
+                }
```
fedemkr, line 443: 🤔 Should we wait until `ReturnToPasskeyAfterUnlock` finishes to set `FromPasskeyFramework` to `false`? I mean if the method throws an exception or something like that , is it ok the options state to be reset as it's now? Or should the user be redirected again after the error?

fedemkr (changes requested):

src/Core/Utilities/AppHelpers.cs
dinisvieira: @fedemkr I think the App will just crash anyway if something goes wrong at this stage.
And in that scenario I suspect the user can press the Passkey Unlock button which will launch the Activity again with the Intent that has the `FromPasskeyFramework="true"`

fedemkr: Perfect then, thanks for the explanation 😄

src/App/Platforms/Android/Autofill/CredentialProviderSelectionActivity.cs
```diff
@@ -46,20 +60,127 @@ private async Task GetCipherAndPerformPasskeyAuthAsync(string cipherId)
...
+            response.SetSignature(sig.Sign());
+            */
+
+            response.SetSignature(assertResult.Signature);
+
+            var credential = new FidoPublicKeyCredential
+            (
+                assertResult.RawId, 
+                response, 
+                "platform" //TODO: use "platform" vs "cross-platform"?
```
coroiu, line 135: `platform` is correct, see: https://github.com/bitwarden/clients/issues/6963

src/App/Platforms/Android/Services/DeviceActionService.cs
```diff
@@ -553,6 +558,107 @@ public long GetActiveTime()
...
+                Bit.Core.Services.LoggerHelper.LogEvenIfCantBeResolved(ex);
+
+                activity.SetResult(Result.Canceled);
+                activity.Finish();
+            }
+        }
+
+        #region CODE_THAT_NEEDS_TO_BE_MOVED_ELSEWHERE
+        //TODO: This region needs to be moved somewhere else where it can be shared with the code in CredentialProviderService.cs
+        private async Task<List<AndroidX.Credentials.Provider.CredentialEntry>> PopulatePasskeyDataAsync(AndroidX.Credentials.Provider.CallingAppInfo callingAppInfo,
```
coroiu, line 609: thought: we could consider exposing and using `Authenticator.SilentCredentialDiscoveryAsync` to implement this

dinisvieira: I have moved the "duplicated" code into CredentialHelpers.cs
But if you think it makes sense to change that code to call that we could do it anyway.

src/App/Platforms/Android/Services/AutofillHandler.cs
```diff
@@ -43,9 +43,28 @@ public bool CredentialProviderServiceEnabled()
...
+                {
+                    return false;
+                }
+
+                var credentialProviderServiceComponentName = new ComponentName(activity, Java.Lang.Class.FromType(typeof(CredentialProviderService)));
+                return credManager.IsEnabledCredentialProviderService(credentialProviderServiceComponentName);
+            }
+            catch (Java.Lang.NullPointerException) 
+            {
+                //TODO: CredentialManager API is not working fully and may return a NullPointerException even if the CredentialProviderService is working and enabled
```
vvolkgang, line 66: With the recent changes to the Settings structure, this is not a TODO anymore right?

vvolkgang (approved):
Mostly checked for loose ends and TODOs, this will go through a deeper review as part of the main passkeys branch

src/App/Platforms/Android/MainApplication.cs
```diff
@@ -91,6 +94,44 @@ public MainApplication(IntPtr handle, JniHandleOwnership transer)
...
+
+                var fido2MakeCredentialUserInterface = new Fido2MakeCredentialUserInterface(
+                    ServiceContainer.Resolve<IStateService>(),
+                    ServiceContainer.Resolve<IVaultTimeoutService>(),
+                    ServiceContainer.Resolve<ICipherService>(),
+                    ServiceContainer.Resolve<IUserVerificationMediatorService>(),
+                    ServiceContainer.Resolve<IDeviceActionService>());
+                ServiceContainer.Register<IFido2MakeCredentialConfirmationUserInterface>(fido2MakeCredentialUserInterface);
+
+                //TODO: WIP: Need to replace 'Fido2GetAssertionUserInterface' if needed
```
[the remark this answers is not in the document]
dinisvieira: Not anymore. Removed.

src/App/Platforms/Android/Services/AutofillHandler.cs
vvolkgang: We can leave it, just remove the "TODO:". While commenting this, noticed I missed another one later in this file

`// TODO - find a way to programmatically disable the provider service, or take the user to the settings page where they can do it`

dinisvieira: All done

PR #3029, opened 2024-02-24
[PM-6428] Changed Android LegacySecureStorage from Task.FromResult to Task.Run to mimic the Microsoft.Maui.Storage.SecureStorage implementation which ensures there's always a Task

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Execution flow seems different between `LegacySecureStorage` and `Microsoft.Maui.Storage.SecureStorage`
The obvious difference between these two is that for Android `LegacySecureStorage` uses `Task.FromResult` and `Microsoft.Maui.Storage.SecureStorage` uses `Task.Run`.
If I recall correctly `Task.Run` ensures we have an Async Task and `Task.FromResult` doesn't. Taking into account that AndroidKeyStore is synchronous I suspect everything will be synchronous with `Task.FromResult` which could result in the behavior we have.
The downside is that `Task.Run` is likely less performant but it is also what MAUI Team decided to use so I would hope it's fine.

What do you think @fedemkr ?
Do you think there could be scenarios where the "order" of execution is not respected even with Task.Run? (although so far it has worked)

## Code changes
Changed Android `LegacySecureStorage` from `Task.FromResult` to `Task.Run` to mimic the `Microsoft.Maui.Storage.SecureStorage` implementation which ensures there's always a Task.
This change was only done for Set and Get operations on Android. Other operations and iOS likely don't need this.

## Before you submit

### DIFF
--- a/src/Core/Services/LegacySecureStorage/LegacySecureStorage.cs
+++ b/src/Core/Services/LegacySecureStorage/LegacySecureStorage.cs
@@ -25,27 +25,31 @@ public class LegacySecureStorage
             throw new ArgumentNullException(nameof(key));
 
 #if ANDROID
-        object locker = new object();
-        string? encVal = Preferences.Get(key, null, Alias);
-
-        if (!string.IsNullOrEmpty(encVal))
+        return Task.Run(() =>
         {
-            try
+            object locker = new object();
+            string? encVal = Preferences.Get(key, null, Alias);
+
+            if (!string.IsNullOrEmpty(encVal))
             {
-                byte[] encData = Convert.FromBase64String(encVal);
-                lock (locker)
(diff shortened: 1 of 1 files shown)


fedemkr (approved):
I think that if MAUI team is using that, then we should probably do as well. I don't think this is a strong solution to a race condition problem, but at least for now it should work. I think that if you face another issue of this kind we can pivot to something stronger like using a global awaiter like [this one](https://github.com/bitwarden/mobile/blob/main/src/Core/Services/ConditionedAwaiterManager.cs) so we can wait until something finishes before executing the navigation.
So for now approving this 😄

PR #3032, opened 2024-02-25
Added ConditionedAwaiter to avoid race condition issue where HomePage navigation would run after the main navigation

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Try to ensure the main navigation always runs after the homepage navigation.

## Code changes
Using the ConditionedAwaiterManager I initially tried using SetCompleted in the Window ctor but that was apparently to early still.
Moved it to HomePage after it "fires" it's own Navigation and that does seem to work.

## Before you submit

### DIFF
--- a/src/Core/Abstractions/IConditionedAwaiterManager.cs
+++ b/src/Core/Abstractions/IConditionedAwaiterManager.cs
@@ -5,7 +5,8 @@ namespace Bit.Core.Abstractions
 {
     public enum AwaiterPrecondition
     {
-        EnvironmentUrlsInited
+        EnvironmentUrlsInited,
+        AndroidWindowCreated
     }
 
     public interface IConditionedAwaiterManager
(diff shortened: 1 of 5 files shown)


fedemkr (commented):
I think this may work. The only caveat I'd mention is that once the awaiter is set as completed and given that we have one per precondition all places will be completed for that precondition.
So if in the same process of the app (thus the `IConditionedAwaiterManager` singleton isn't cleared) the `HomePage` is opened and then the `AndroidNavigationRedirectPage` is opened, the second one will already be marked as completed because of setting it as completed on `HomePage`. Not sure if this scenario could happen, but if that's the case we should maybe have two different preconditions, one for `MainAppWindow` and another one for `AutofillWindow`.
Same can be argued about opening multiple times `HomePage` and if that needs the `CreateWindow` to run again then the first time will be marked as completed and all subsequent times won't wait given that the singleton precondition has finished already.

PR #3039, opened 2024-02-28
[PM-6506] Fix double execution of command on returnType Go

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix regression bug on Android where the app would crash if the "Submit" button was pressed on the keyboard while the server url entry was selected.

## Code changes

Needed to replace the Action with a Func<Task> because the keyboard returnType="Go" has a bug where it can execute the associated command twice.
 
## Screenshots

https://github.com/bitwarden/mobile/assets/4648522/2bbd9f67-f47b-47fc-b5ef-94c858efac66

## Before you submit

### DIFF
--- a/src/Core/Pages/Accounts/EnvironmentPageViewModel.cs
+++ b/src/Core/Pages/Accounts/EnvironmentPageViewModel.cs
@@ -73,7 +73,10 @@ namespace Bit.App.Pages
             IconsUrl = resUrls.Icons;
             NotificationsUrls = resUrls.Notifications;
 
-            await SubmitSuccessTask();
+            if (SubmitSuccessTask != null)
+            {
+                await SubmitSuccessTask();
+            }
         }
 
         public bool ValidateUrls()


src/Core/Pages/Accounts/EnvironmentPageViewModel.cs
```diff
@@ -73,7 +73,7 @@ public async Task SubmitAsync()
             IconsUrl = resUrls.Icons;
             NotificationsUrls = resUrls.Notifications;
 
-            SubmitSuccessAction?.Invoke();
+            await SubmitSuccessTask();
```
fedemkr, line 76: ⚠️ Just in case check `SubmitSuccessTask` is not `null` before invoking it:
```suggestion
            if (SubmitSuccessTask != null)
            {
                await SubmitSuccessTask();
            }
```

fedemkr (changes requested):

fedemkr (approved):
LGTM 🎉

PR #3040, opened 2024-02-28
[PM-6513] Omit creating CredentialIdentity if it throws an exception

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Omit creating CredentialIdentity if it throws an exception. E.g. if a Passkey doesn't have a UserName it will throw and it shouldn't break replacing all the other identities.

## Code changes

* **ASHelpers:** Added a try...catch when creating the `CredentialIdentity` which returns `null` if an exception happens so it's ignored when registering all identities and it doesn't break it.

## Before you submit

### DIFF
--- a/src/iOS.Core/Utilities/ASHelpers.cs
+++ b/src/iOS.Core/Utilities/ASHelpers.cs
@@ -2,6 +2,7 @@
 using Bit.Core.Abstractions;
 using Bit.Core.Enums;
 using Bit.Core.Models.View;
+using Bit.Core.Services;
 using Bit.Core.Utilities;
 using Foundation;
 using UIKit;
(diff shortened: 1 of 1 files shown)


src/iOS.Core/Utilities/ASHelpers.cs
```diff
@@ -135,21 +136,40 @@ public static async Task<bool> IsAutofillEnabledAsync()
...
+                if (string.IsNullOrWhiteSpace(userName))
+                {
+                    userName = cipher.Login.MainFido2Credential.UserDisplayName;
 
-            if (!cipher.Login.MainFido2Credential.DiscoverableValue)
+                    if (string.IsNullOrWhiteSpace(userName))
+                    {
+                        userName = cipher.Login.Username;
+                    }
+                }
```
coroiu, line 160: suggestion: I think you can skip the nesting

```suggestion
                if (string.IsNullOrWhiteSpace(userName))
                {
                    userName = cipher.Login.MainFido2Credential.UserDisplayName;
                }
                
                if (string.IsNullOrWhiteSpace(userName))
                {
                    userName = cipher.Login.Username;
                }
```

fedemkr: ℹ️ I did it on purpose to avoid checking twice for the same thing on the primary flow, i.e. if the most probable thing happens that is `UserName` not being null then first time it checks we know it's not null then it's not needed to check again afterwards the same thing. Nonetheless it's necessary in the nesting for the alternative case.
If we want to avoid nesting I could extract this into a method where we can just return the value on each case without further nesting, but it's small and just one additional level so I don't thing is so necessary.

coroiu: Oh interesting, I didn't consider that, fair point!

coroiu (approved):

PR #3049, opened 2024-03-01
[PM-6539] Fix Autofill Extension TDE without MP flow

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix Autofill iOS Extension when using TDE without MP flow.

_Note: I also fixed one NRE on the state migration from 6 to 7, also same happening on Generator view on iOS extension. Additionally, `WebAuthenticator` was copied and adapted from MAUI Essentials to get the current `UIWindow` as in Xamarin Essentials._

## Code changes

* **BaseLockPasswordViewController:** Updating PromptSSO view controller creation to be like in MAUI.
* **LoginAddViewController:** Fixed generator view controller creation to be like in MAUI.
* **StateMigrationService:** Fix NRE if `state` is `null`
* **LoginSSOViewModel:** Fix to use workaround for iOS Extension on the `WebAuthenticator`
* **WebAuthenticator:** MAUI Essentials copy of `WebAuthenticator` adapted to get the current `UIWIndow` as it was being done in Xamarin Essentials to work in iOS Extensions.

## Before you submit

### DIFF
--- a/src/iOS.Autofill/LockPasswordViewController.cs
+++ b/src/iOS.Autofill/LockPasswordViewController.cs
@@ -1,8 +1,8 @@
 using System;
 using Bit.App.Controls;
+using Bit.Core.Services;
 using Bit.Core.Utilities;
 using Bit.iOS.Core.Utilities;
-using MapKit;
 using UIKit;
 
 namespace Bit.iOS.Autofill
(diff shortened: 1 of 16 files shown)


src/iOS.Autofill/LockPasswordViewController.cs
```diff
@@ -33,22 +33,29 @@ public LockPasswordViewController(IntPtr handle)
...
+                    new UIBarButtonItem(_accountSwitchButton)
+                }, false);
 
-            _accountSwitchingOverlayView = _accountSwitchingOverlayHelper.CreateAccountSwitchingOverlayView(OverlayView);
+                _accountSwitchingOverlayView = _accountSwitchingOverlayHelper.CreateAccountSwitchingOverlayView(OverlayView);
+            }
+            catch (Exception ex)
+            {
+                LoggerHelper.LogEvenIfCantBeResolved(ex);
+            }
```
andrebispo5, line 58: If there is a crash here the user won't have an account switcher in the UI is that intended?

andrebispo5 (approved):
Looks good 👍  
only left a small comment.

src/iOS.Autofill/LockPasswordViewController.cs
fedemkr: Not intended, but I guess that's better than the Autofill extension crashing

PR #3069, opened 2024-03-09
[PM-6726] Fix for Android 14 devices crashing when using the Tiles

dinisvieira:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
When using Android Tiles on Android 14+ the TileService crashes.

## Code changes
Android now requires that TileServices using `PendingIntent` instead of Intent for `StartActivityAndCollapse()`
Doing this change avoids the crash. We still keep the "old" Intent code for older versions of Android.

There was also a background crash on AccessibilityService due to a PendingIntent being mutable which also seems to no longer being allowed in Android14 for that specific scenario.

* **AutofillTileService.cs:** Added if scenario for Android 14 to use `PendingIntent` instead of `Intent`.
* **MyVaultTileService.cs:** Added if scenario for Android 14 to use `PendingIntent` instead of `Intent`
* **GeneratorTileService.cs:** Added if scenario for Android 14 to use `PendingIntent` instead of `Intent`
* **AutofillHelpers.cs:** Changed InlinePresentationBuilder PendingItent to be Immutable to avoid crash in Android 14

## Before you submit

### DIFF
--- a/src/App/Platforms/Android/Autofill/AutofillHelpers.cs
+++ b/src/App/Platforms/Android/Autofill/AutofillHelpers.cs
@@ -347,7 +347,7 @@ namespace Bit.Droid.Autofill
                 // InlinePresentation requires nonNull pending intent (even though we only utilize one for the
                 // "my vault" presentation) so we're including an empty one here
                 pendingIntent = PendingIntent.GetService(context, 0, new Intent(),
-                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.OneShot | PendingIntentFlags.UpdateCurrent, true));
+                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.OneShot | PendingIntentFlags.UpdateCurrent, false));
             }
             var slice = CreateInlinePresentationSlice(
                 inlinePresentationSpec,
--- a/src/App/Platforms/Android/Tiles/AutofillTileService.cs
+++ b/src/App/Platforms/Android/Tiles/AutofillTileService.cs
@@ -77,7 +77,7 @@ namespace Bit.Droid.Tile
             var intent = new Intent(this, typeof(AccessibilityActivity));
             intent.SetFlags(ActivityFlags.NewTask | ActivityFlags.SingleTop | ActivityFlags.ClearTop);
             intent.PutExtra("autofillTileClicked", true);
-            intent.StartActivityAndCollapseFromTileService(this, isMutable: true);
+            this.StartActivityAndCollapseWithIntent(intent, isMutable: true);
         }
 
         private void ShowConfigErrorDialog()
--- a/src/App/Platforms/Android/Tiles/AutofillTileService.cs
+++ b/src/App/Platforms/Android/Tiles/AutofillTileService.cs
@@ -1,6 +1,7 @@
 ﻿using Android;
 using Android.App;
 using Android.Content;
+using Android.OS;
 using Android.Runtime;
 using Android.Service.QuickSettings;
 using Bit.Core.Resources.Localization;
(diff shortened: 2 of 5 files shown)


src/App/Platforms/Android/Tiles/AutofillTileService.cs
```diff
@@ -76,7 +77,24 @@ private void ScanAndFill()
...
+                    PendingIntentFlags.Mutable | PendingIntentFlags.UpdateCurrent
+                );
+                if (pendingIntent == null) { return; }
+
+                StartActivityAndCollapse(pendingIntent);
+            }
+            else
+            {
+                StartActivityAndCollapse(intent);
+            }
```
fedemkr, line 97: 🎨 Add extension method from `Intent` to reuse this code between the three services which would have the Android 14 check inside so the code is cleaner. Something like `intent.ToActivityPendingIntentIfNecessary()` or anything else more descriptive.

src/App/Platforms/Android/Autofill/AutofillHelpers.cs
```diff
@@ -347,7 +347,7 @@ public static RemoteViews BuildOverlayPresentation(string text, string subtext,
                 // InlinePresentation requires nonNull pending intent (even though we only utilize one for the
                 // "my vault" presentation) so we're including an empty one here
                 pendingIntent = PendingIntent.GetService(context, 0, new Intent(),
-                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.OneShot | PendingIntentFlags.UpdateCurrent, true));
+                    AndroidHelpers.AddPendingIntentMutabilityFlag(PendingIntentFlags.OneShot | PendingIntentFlags.UpdateCurrent, false));
```
fedemkr, line 350: 🤔 Does this work as expected on Android 14-? Specially on minimum supported versions

fedemkr: Alright sounds good 👍

src/App/Platforms/Android/Tiles/AutofillTileService.cs
dinisvieira: @fedemkr The return type is different and it's specific for TileService but it can still be shared for sure.

src/App/Platforms/Android/Utilities/AndroidHelpers.cs
```diff
@@ -64,5 +65,26 @@ public static PendingIntentFlags AddPendingIntentMutabilityFlag(PendingIntentFla
...
+                );
+                if (pendingIntent == null) { return; }
+
+                service.StartActivityAndCollapse(pendingIntent);
+            }
+            else
+            {
+                service.StartActivityAndCollapse(intent);
+            }
+        }
```
fedemkr, line 88: 🎨 Oh sorry didn't see that we needed to do the `StartActivityAndCollapse` inside, so could we change this extension method to be from the `TileService` instead so it's more related to the action? Also, I'd invert the if and return to avoid the `else` and let the old versions be the specific alternative scenario. Finally, I see you're always specifying `isMutable` on the calls so no need to have a default value here:
```suggestion
        public static void StartActivityAndCollapseFromTileService(this TileService service, Intent intent, bool isMutable)
        {
            //For Android 14+ We need to use PendingIntent instead of Intent directly. Older versions still need to use Intent.
            if (Build.VERSION.SdkInt < BuildVersionCodes.UpsideDownCake)
            {
                service.StartActivityAndCollapse(intent);
                return;
            }
            var pendingIntent = PendingIntent.GetActivity(
                service.ApplicationContext,
                0,
                intent,
                AddPendingIntentMutabilityFlag(PendingIntentFlags.UpdateCurrent, isMutable)
            );
            if (pendingIntent == null) 
            { 
                return; 
            }
            service.StartActivityAndCollapse(pendingIntent);
        }
```

fedemkr (changes requested):

fedemkr (approved):

PR #3099, opened 2024-03-20
[PM-6848] Improved User verification on passkeys creation

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Improved user verification on passkeys creation to better take into account User verification cancellation and also enforcement at the moment of creation.

## Code changes

* **MobilePlatformUtilsService:** Now `AuthenticateBiometricAsync` returns `null` when cancelled by the user.
* **UserVerificationMediatorService:** Changed to use `CancellableResult<T>` to tackle cancellation flows.
  * Same applies to its strategies
* **LoginListViewController/LoginAddViewController:** Added cancellation flow so the user can try again and also implemented user verification enforcement at this stage as well to prevent the cipher to be created if the verification fails and is required.

## Before you submit

### DIFF
--- a/src/iOS.Autofill/LoginListViewController.cs
+++ b/src/iOS.Autofill/LoginListViewController.cs
@@ -332,7 +332,18 @@ namespace Bit.iOS.Autofill
             bool? isUserVerified = null;
             if (Context?.PasskeyCreationParams?.UserVerificationPreference != Fido2UserVerificationPreference.Discouraged)
             {
-                isUserVerified = await VerifyUserAsync();
+                var verification = await VerifyUserAsync();
+                if (verification.IsCancelled)
+                {
+                    return;
+                }
+                isUserVerified = verification.Result;
+
+                if (!isUserVerified.Value && await _userVerificationMediatorService.Value.ShouldEnforceFido2RequiredUserVerificationAsync(Fido2UserVerificationOptions))
+                {
+                    await _platformUtilsService.Value.ShowDialogAsync(AppResources.ErrorCreatingPasskey, AppResources.SavePasskey);
+                    return;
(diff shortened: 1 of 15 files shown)


src/iOS.Autofill/LoginListViewController.cs
```diff
@@ -350,30 +361,31 @@ private async Task SavePasskeyAsNewLoginAsync()
...
             {
                 LoggerHelper.LogEvenIfCantBeResolved(ex);
-                return false;
+                return new CancellableResult<bool>(false);
             }
         }
 
+        private Fido2UserVerificationOptions Fido2UserVerificationOptions => new Fido2UserVerificationOptions(
+            false,
+            Context.PasskeyCreationParams.Value.UserVerificationPreference,
```
andrebispo5, line 384: ❓ There are other places in this class where `Context` can be `null` isn't the case here?

andrebispo5 (approved):
Just left a small comment, other than that looks good. 👍

src/iOS.Autofill/LoginListViewController.cs
fedemkr: Here if it's null it should just break the flow because we wouldn't have the necessary parameters to build the options and perform the user verification. It's unlikely to happen and also on the VerifyUserAsync we're checking the nullability before calling this. The only part that will be a concern is when checking the UV enforcement but I think it should just throw and it will display a message to the user.
Nonetheless, I can add an argument null check of the Context so if we get this in AppCenter we know that's the specific reason, thanks for catching this!

andrebispo5 (approved):

PR #3121, opened 2024-03-27
[PM-6798] Fix account switch on iOS Autofill extension

fedemkr:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix account switch on iOS Autofill extension

## Code changes

* **Fido2AuthenticatorService:** Added a do...while for the first part of the flows so if the account is changed an exception is raised and the flow starts over. This was the fastest thing I could come up with to fix this situation.
* **CredentialProviderViewController:** Subscribed to account switched completed to reset and set the `AccountSwitchedException` on the TaskCompletionSources necessary for passkeys. Also improved the code and refactored a bit to be able to reuse some of it in another place.
* **CredentialProviderViewController.Passkeys:** Changed how we ensure vault is unlocked to consider more account states and reuse the navigation on the `CredentialProviderViewController`
* **Ohers:** Some fixes to use `Try...` way for setting state on `TaskCompletionSource`

## Before you submit

### DIFF
--- a/src/Core/Services/Fido2AuthenticatorService.cs
+++ b/src/Core/Services/Fido2AuthenticatorService.cs
@@ -36,28 +36,42 @@ namespace Bit.Core.Services
                 throw new NotSupportedError();
             }
 
-            await userInterface.EnsureUnlockedVaultAsync();
-            await _syncService.FullSyncAsync(false);
+            string cipherId = null;
+            var userVerified = false;
+            var accountSwitched = false;
 
-            var existingCipherIds = await FindExcludedCredentialsAsync(
-                makeCredentialParams.ExcludeCredentialDescriptorList
-            );
-            if (existingCipherIds.Length > 0)
+            do
             {
-                await userInterface.InformExcludedCredentialAsync(existingCipherIds);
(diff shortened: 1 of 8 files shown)


src/Core/Services/Fido2AuthenticatorService.cs
```diff
@@ -36,28 +36,42 @@ public async Task<Fido2AuthenticatorMakeCredentialResult> MakeCredentialAsync(Fi
...
+                        UserName = makeCredentialParams.UserEntity.Name,
+                        UserVerificationPreference = makeCredentialParams.UserVerificationPreference,
+                        RpId = makeCredentialParams.RpEntity.Id
+                    });
+                }
+                catch (AccountSwitchedException)
+                {
+                    accountSwitched = true;
+                }
+            } while (accountSwitched);
```
vvolkgang, line 73: Let's move forward for now but revisit this later. Created [PM-7140 ](https://bitwarden.atlassian.net/browse/PM-7140)

vvolkgang (approved):
Makes sense overall but we should revisit this later.

PR #3150, opened 2024-04-11
[PM-7407] Implemented check for organizations with unassigned items

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Add check for organization with unassigned items to display a message to the user if it's needed.

## Code changes

* **Constants:** Added new feature flag for this check
* **ApiService:** Added new endpoint call to check whether the organization has any unassigned items
* **CipherService:** Added method to verify if an organization has any unassigned items
* **AppOptions:** Added flag to see if the user has just logged in / unlocked their vault when displaying the `GroupingsPage`
  * **Other files related with AppOptions**: Updated the new flag of `AppOptions` and passed along the options object to get to the `GroupingsPage`
* **StateService:** Added new Get/Set method for the flag of if one should or not perform this check
* **GroupingsPageViewModel:** Added logic to check the organization unassigned items based on the feature flag and the "should check" local flag (depending on what the user chose on an already shown dialog). Also, it waits if a sync is in progress to have the latest data locally as I'm checking whether the user has any organizations so not to call the endpoint on the cases where the user doesn't belong to an org.

### Unassigned items notice iOS
Unassigned items notice iOS

### Unassigned items notice Android
Unassigned items notice Android

## Before you submit

### DIFF
--- a/src/Core/Abstractions/IStateService.cs
+++ b/src/Core/Abstractions/IStateService.cs
@@ -186,6 +186,8 @@ namespace Bit.Core.Abstractions
         Task<BwRegion?> GetActiveUserRegionAsync();
         Task<BwRegion?> GetPreAuthRegionAsync();
         Task SetPreAuthRegionAsync(BwRegion value);
+        Task<bool> GetShouldCheckOrganizationUnassignedItemsAsync(string userId = null);
+        Task SetShouldCheckOrganizationUnassignedItemsAsync(bool shouldCheck, string userId = null);
         [Obsolete("Use GetPinKeyEncryptedUserKeyAsync instead, left for migration purposes")]
         Task<string> GetPinProtectedAsync(string userId = null);
         [Obsolete("Use SetPinKeyEncryptedUserKeyAsync instead, left for migration purposes")]
--- a/src/Core/Constants.cs
+++ b/src/Core/Constants.cs
@@ -46,6 +46,7 @@ namespace Bit.Core
         public const string PreLoginEmailKey = "preLoginEmailKey";
         public const string ConfigsKey = "configsKey";
         public const string DisplayEuEnvironmentFlag = "display-eu-environment";
+        public const string UnassignedItemsBannerFlag = "unassigned-items-banner";
         public const string RegionEnvironment = "regionEnvironment";
         public const string DuoCallback = "bitwarden://duo-callback";
 
@@ -136,6 +137,7 @@ namespace Bit.Core
         public static string ShouldConnectToWatchKey(string userId) => $"shouldConnectToWatch_{userId}";
         public static string ScreenCaptureAllowedKey(string userId) => $"screenCaptureAllowed_{userId}";
         public static string PendingAdminAuthRequest(string userId) => $"pendingAdminAuthRequest_{userId}";
+        public static string ShouldCheckOrganizationUnassignedItemsKey(string userId) => $"shouldCheckOrganizationUnassignedItems_{userId}";
         [Obsolete]
         public static string KeyKey(string userId) => $"key_{userId}";
         [Obsolete]
--- a/src/Core/Pages/Accounts/LockPage.xaml.cs
+++ b/src/Core/Pages/Accounts/LockPage.xaml.cs
@@ -233,6 +233,7 @@ namespace Bit.App.Pages
             }
             var previousPage = await AppHelpers.ClearPreviousPage();
 
+            _appOptions.HasJustLoggedInOrUnlocked = true;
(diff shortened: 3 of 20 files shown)


src/Core/Abstractions/IStateService.cs
```diff
@@ -186,6 +186,8 @@ public interface IStateService
         Task<BwRegion?> GetActiveUserRegionAsync();
         Task<BwRegion?> GetPreAuthRegionAsync();
         Task SetPreAuthRegionAsync(BwRegion value);
+        Task<bool> GetShouldCheckOrganizationUnassignedItemsAsync(string userId = null);
```
vvolkgang, line 189: 🤔  Naming wise, following a similar pattern to the other prompts (e.g. `Get/Set UnassignedItemsPromptShownAsync`) would make it easier to find this in the future.

src/Core/Constants.cs
```diff
@@ -136,6 +137,7 @@ public static class Constants
         public static string ShouldConnectToWatchKey(string userId) => $"shouldConnectToWatch_{userId}";
         public static string ScreenCaptureAllowedKey(string userId) => $"screenCaptureAllowed_{userId}";
         public static string PendingAdminAuthRequest(string userId) => $"pendingAdminAuthRequest_{userId}";
+        public static string ShouldCheckOrganizationUnassignedItemsKey(string userId) => $"shouldCheckOrganizationUnassignedItems_{userId}";
```
vvolkgang, line 140: ⛏️ (not really relevant but worth mentioning) other similar constants don't have verbiage like "ShouldCheck".

src/Core/Services/CipherService.cs
```diff
@@ -829,6 +829,24 @@ public async Task RestoreWithServerAsync(string id)
             await ClearCacheAsync();
         }
 
+        public async Task<bool> VerifyOrganizationHasUnassignedItemsAsync()
+        {
+            var organizations = await _stateService.GetOrganizationsAsync();
```
vvolkgang, line 834: 🤔 I'm not sure if it's possible for this to return 0 orgs and the next call returning true, I just noticed other clients aren't checking this.

src/Core/Abstractions/IStateService.cs
fedemkr: The problem is that this getter doesn't return if the prompt has been shown or not, it returns whether the check (and potentially the prompt) should be done or not, i.e. if the user chose "Remind me later" or not.
Also there are also two other that have `Should`:
- GetShouldTrustDeviceAsync
- GetShouldConnectToWatchAsync
Even though I agree having everything with the same name structure is better, but on the other hand I like having specific naming so one can exactly know what the method/property does without having to look much in other places.
What do you think?

src/Core/Constants.cs
fedemkr: IMO without the "should check" one doesn't know what the flag is for, only that it has to do with organization unassigned items. However, if having the verbiage alike others is more important I can change it.

src/Core/Services/CipherService.cs
fedemkr: This is exactly why I had to [wait for synchronization to complete](https://github.com/bitwarden/mobile/pull/3150/files#diff-d0d4f4baeed96a7ebebe5105ff70eb1bfad44ff3c0637043ed0b26eb80361b42R725).
I could remove this alongside the sync wait but it will fire a request to the server for users that don't have organizations which IMO is a waste of resources and given that we're waiting for sync to finish we should have the latest organizations state when checking it.

src/Core/Pages/Accounts/LockPage.xaml.cs
```diff
@@ -233,6 +233,7 @@ private async Task UnlockedAsync()
             }
             var previousPage = await AppHelpers.ClearPreviousPage();
 
+            _appOptions.HasJustLoggedInOrUnlocked = true;
```
andrebispo5, line 236: 💭 Instead of having this variable scattered around and risk missing a path to the vault, can't we just check once on TabsPage.OnCreate with a control variable `HasRunCheckOrganizationUnassignedItems`?

fedemkr: The problem is that `TabsPage.OnCreate` can be called when the vault has timeout never, then it's not technically after login/unlock as it will be called when opening the app and given timeout never the vault would be already unlocked.

vvolkgang (approved):
Thanks for adding the self-host message!

andrebispo5 (approved):

PR #3155, opened 2024-04-13
 [PM-7385] Fix for allowing switching accounts while creating a passkey of Android

dinisvieira:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
The objective is to allow switching accounts during the create passkey flow.

## Code changes
Updated `EnsureUnlockedVaultAsync` on Android to trigger the unlock flow like on iOS. This allows the account switching process to work.
Also had to expose the TaskCompletionSource for confirmCredential so that other parts of the app can stop navigations when a create passkey flow is already running.
A new TaskCompletionSource was also created for the unlockVault process that can get triggered during the credential creation when switching accounts.
To avoid additional erratic app behaviors the app is no longer relying on `Options.Fido2CredentialAction` and `Options.FromFido2Framework` after the Fido process starts. We use "new" TaskCompletionSource instead to know if the flow is still running.
There's also a fix included for scenarios where tapping on cipher won't do anything (in this case after switching).

* **Fido2MakeCredentialUserInterface.cs:** Updated `EnsureUnlockedVaultAsync` to trigger the vault unlocked process if it's not already unlocked. Also added a new unlock vault `TaskCompletionSource` to allow this.
* **DeviceActionService.cs:** We ensure we reset the `Options.Fido2CredentialAction` and `Options.FromFido2Framework` to avoid erratic behaviors that we could get after the Create passkey flow was not running anymore. We use the `IsConfirmingNewCredential` and `IsWaitingUnlockVault` instead to know that the flow is still running.
* **IFido2MakeCredentialConfirmationUserInterface.cs:** Added the definitions for the unlock vault and confirm credential TaskCompletionSource methods.
* **App.xaml.cs:** Added new event for Android navigations and added a check on Navigate() to avoid navigating again when we are already running the create passkey flow.
* **AccountsManager.cs:** Check to avoid restarting the create passkey flow when we are already running it.
* **AppHelpers.cs:** Added block to "inform" the `IFido2MakeCredentialConfirmationUserInterface` that the Vault was unlocked.
* **AutofillCiphersPageViewModel.cs.cs:** Changes to use `_fido2MakeCredentialConfirmationUserInterface.Value.IsConfirmingNewCredential` instead of `_appOptions.FromFido2Framework`
* **CipherAddEditPage.xaml.cs:** Changes to use `_fido2MakeCredentialConfirmationUserInterface.Value.IsConfirmingNewCredential` instead of `_appOptions.FromFido2Framework`
* **CipherAddEditPageViewModel.cs:** Changes to use `_fido2MakeCredentialConfirmationUserInterface.Value.IsConfirmingNewCredential` instead of `_appOptions.FromFido2Framework`

## Before you submit


src/App/Platforms/Android/Autofill/Fido2MakeCredentialUserInterface.cs
```diff
@@ -95,11 +98,35 @@ private async Task<(string CipherId, bool UserVerified)> CreateNewLoginForFido2C
...
             }
+
+            await NavigateAndWaitForUnlockAsync(Bit.Core.Enums.NavigationTarget.Lock);
+        }
+
+        private async Task NavigateAndWaitForUnlockAsync(Bit.Core.Enums.NavigationTarget navTarget)
+        {
+            _unlockVaultTcs?.TrySetCanceled();
+            _unlockVaultTcs = new TaskCompletionSource<bool>();
+            var messagingService = ServiceContainer.Resolve<IMessagingService>("messagingService");
```
fedemkr, line 123: ⛏️ No need for the string key for resolve; also either inject the service on the constructor or add a field with a `LazyResolve` so we don't have to resolve each time this method is called.
```suggestion
            var messagingService = ServiceContainer.Resolve<IMessagingService>();
```

src/App/Platforms/Android/Autofill/Fido2MakeCredentialUserInterface.cs
```diff
@@ -95,11 +98,35 @@ private async Task<(string CipherId, bool UserVerified)> CreateNewLoginForFido2C
...
+            await MainThread.InvokeOnMainThreadAsync(() =>
+            {
+                messagingService?.Send("navigateTo", navTarget);
+            });
```
fedemkr, line 127: 🤔 Is it necessary to send this on the main thread? Additionally, is it necessary to await this?

src/App/Platforms/Android/Autofill/Fido2MakeCredentialUserInterface.cs
```diff
@@ -109,6 +136,7 @@ public Task InformExcludedCredentialAsync(string[] existingCipherIds)
         }
 
         public void Confirm(string cipherId, bool? userVerified) => _confirmCredentialTcs?.TrySetResult((cipherId, userVerified));
+        public void ConfirmUnlockVault(bool unlocked) => _unlockVaultTcs?.TrySetResult(unlocked);
```
fedemkr, line 139: 🤔 I don't think you care about the boolean value. I'd remove it so the API is clear. Maybe change a bit the name to be more explicit although not entirely necessary.
```suggestion
        public void ConfirmVaultUnlocked() => _unlockVaultTcs?.TrySetResult(true);
```

src/Core/App.xaml.cs
```diff
@@ -331,18 +331,27 @@ void NavigateToSetPasswordPageAction()
                     || message.Command == "unlocked"
                     || message.Command == AccountsManagerMessageCommands.ACCOUNT_SWITCH_COMPLETED)
                 {
-                    if (message.Command == AccountsManagerMessageCommands.ACCOUNT_SWITCH_COMPLETED)
+                    var userVerificationMediatorService = ServiceContainer.Resolve<IFido2MakeCredentialConfirmationUserInterface>();
```
fedemkr, line 334: ⛏️ Move it to a field with `LazyResolve` so it's not resolve every time the account switches.

src/Core/App.xaml.cs
```diff
@@ -331,18 +331,27 @@ void NavigateToSetPasswordPageAction()
...
                     }
-                    
+
                     lock (_processingLoginRequestLock)
                     {
                         // lock doesn't allow for async execution
                         CheckPasswordlessLoginRequestsAsync().Wait();
                     }
                 }
+                else if (message.Command == "navigateTo" && message.Data is NavigationTarget navigationTarget)
```
fedemkr, line 346: ⛏️ If possible, move `navigateTo` into a constant somewhere to be reused when the message is sent and here in the receiver.

src/Core/App.xaml.cs
```diff
@@ -331,18 +331,27 @@ void NavigateToSetPasswordPageAction()
...
+                {
+                    ArgumentNullException.ThrowIfNull(MainPage);
```
fedemkr, line 348: 🤔 If this is plausible to happen in some weird known scenario, is the information enough on the exception to know that it happened in this message command? If not, we could add a bit more specifics in the exception message.

src/Core/App.xaml.cs
```diff
@@ -713,6 +722,16 @@ public void Navigate(NavigationTarget navTarget, INavigationParams navParams)
             // If we are in background we add the Navigation Actions to a queue to execute when the app resumes.
             // Links: https://github.com/dotnet/maui/issues/11501 and https://bitwarden.atlassian.net/wiki/spaces/NMME/pages/664862722/MainPage+Assignments+not+working+on+Android+on+Background+or+App+resume
 #if ANDROID
+            var userVerificationMediatorService = ServiceContainer.Resolve<IFido2MakeCredentialConfirmationUserInterface>();
```
fedemkr, line 725: ⛏️ Having moved this to a `LazyResolve` field before makes it available to reuse it here without resolving again.

src/Core/Utilities/AccountManagement/AccountsManager.cs
```diff
@@ -81,6 +81,8 @@ public async Task NavigateOnAccountChangeAsync(bool? isAuthed = null)
             var authed = isAuthed ?? await _stateService.IsAuthenticatedAsync();
             if (authed)
             {
+                var fido2MakeCredentialConfirmationUserInterface = ServiceContainer.Resolve<IFido2MakeCredentialConfirmationUserInterface>();
```
fedemkr, line 84: ⛏️ Move to `LazyResolve` field.

src/Core/Utilities/AccountManagement/AccountsManager.cs
```diff
@@ -100,8 +102,15 @@ await _vaultTimeoutService.ShouldLockAsync())
...
-                else if (Options.FromFido2Framework)
+                else if (Options.FromFido2Framework || fido2MakeCredentialConfirmationUserInterface.IsConfirmingNewCredential)
                 {
+                    if (fido2MakeCredentialConfirmationUserInterface.IsConfirmingNewCredential)
+                    {
+                        // If we are already confirming a credential we don't need to navigate again.
+                        // This could happen when switching accounts for example.
+                        return;
+                    }
+
```
fedemkr, line 113: 🎨 To simplify this you can write it as:
```suggestion
                else if (fido2MakeCredentialConfirmationUserInterface.IsConfirmingNewCredential)
                {
                    // If we are already confirming a credential we don't need to navigate again.
                    // This could happen when switching accounts for example.
                    return;
                }
                else if (Options.FromFido2Framework)
                {
                    
```

src/Core/Utilities/AppHelpers.cs
```diff
@@ -437,8 +437,18 @@ public static bool SetAlternateMainPage(AppOptions appOptions)
...
+                if ((appOptions.FromFido2Framework && !string.IsNullOrWhiteSpace(appOptions.Fido2CredentialAction) || userVerificationMediatorService.IsWaitingUnlockVault))
                 {
+                    // If we are waiting for an unlock vault we don't want to trigger 'ExecuteFido2CredentialActionAsync' again,
+                    // as it's already running. We just need to 'ConfirmUnlockVault' on the 'userVerificationMediatorService'.
+                    if (userVerificationMediatorService.IsWaitingUnlockVault)
+                    {
+                       userVerificationMediatorService.ConfirmUnlockVault(true);
+                       return true;
+                    }
+
```
fedemkr, line 451: 🎨 To simplify this, couldn't we write it as:
```suggestion
                var userVerificationMediatorService = ServiceContainer.Resolve<IFido2MakeCredentialConfirmationUserInterface>();
                // If we are waiting for an unlock vault we don't want to trigger 'ExecuteFido2CredentialActionAsync' again,
                // as it's already running. We just need to 'ConfirmUnlockVault' on the 'userVerificationMediatorService'.
                if (userVerificationMediatorService.IsWaitingUnlockVault)
                {
                    userVerificationMediatorService.ConfirmUnlockVault();
                    return true;
                }
                
                if (appOptions.FromFido2Framework && !string.IsNullOrWhiteSpace(appOptions.Fido2CredentialAction))
                {
                    
```

fedemkr (changes requested):
Great job, just a few thoughts

src/App/Platforms/Android/Autofill/Fido2MakeCredentialUserInterface.cs
dinisvieira: Not needed for sure.

src/App/Platforms/Android/Autofill/Fido2MakeCredentialUserInterface.cs
dinisvieira: Adding as DI

src/Core/App.xaml.cs
dinisvieira: Really not needed. It it happens it's already outside of the message command context.

src/Core/Utilities/AppHelpers.cs
dinisvieira: Yes, looks like it. Just like the one of the other scenarios above it's remnants of older logic that I had.

src/Core/App.xaml.cs
```diff
@@ -343,14 +343,13 @@ void NavigateToSetPasswordPageAction()
...
                 {
-                    ArgumentNullException.ThrowIfNull(MainPage);
-
-                    MainThread.BeginInvokeOnMainThread(() =>
+                    await MainThread.InvokeOnMainThreadAsync(NavigateToAction);
+                    void NavigateToAction()
                     {
                         Navigate(navigationTarget, null);
-                    });
+                    }
```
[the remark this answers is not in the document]
dinisvieira: Was just doing it for the "breadcrumb stack trace" but it's not needed because it's probably going to identify the Navigate method anyway.

src/Core/Utilities/AppHelpers.cs
```diff
@@ -442,20 +442,21 @@ public static bool SetAlternateMainPage(AppOptions appOptions)
...
                     return true;
                 }
+
+                // If we are waiting for an unlock vault we don't want to trigger 'ExecuteFido2CredentialActionAsync' again,
+                // as it's already running. We just need to 'ConfirmUnlockVault' on the 'userVerificationMediatorService'.
+                if (userVerificationMediatorService.IsWaitingUnlockVault)
+                {
+                    userVerificationMediatorService.ConfirmVaultUnlocked();
+                    return true;
+                }
```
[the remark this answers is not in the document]
dinisvieira: It's ok because we are now using different booleans to check which flow to use.
But I'll change it anyway as it does make more sense to be before.

fedemkr (approved):

PR #3158, opened 2024-04-15
[PM-7407] Updates the self-host release date in the unassigned items alert message

vvolkgang:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Self-hosted migration will be postponed, message updated with the new date.

## Before you submit

### DIFF
--- a/src/Core/Resources/Localization/AppResources.resx
+++ b/src/Core/Resources/Localization/AppResources.resx
@@ -2890,7 +2890,7 @@ Do you want to switch to this account?</value>
     <value>Unassigned organization items are no longer visible in the All Vaults view and only accessible via the Admin Console. Assign these items to a collection from the Admin Console to make them visible.</value>
   </data>
   <data name="OrganizationUnassignedItemsMessageSelfHostDescriptionLong" xml:space="preserve">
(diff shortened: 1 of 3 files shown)


src/Core/Resources/Localization/AppResources.resx
```diff
@@ -2890,7 +2890,7 @@ Do you want to switch to this account?</value>
     <value>Unassigned organization items are no longer visible in the All Vaults view and only accessible via the Admin Console. Assign these items to a collection from the Admin Console to make them visible.</value>
   </data>
   <data name="OrganizationUnassignedItemsMessageSelfHostDescriptionLong" xml:space="preserve">
```
djsmith85, line 2892: Please change the so all existing translations get removed, which would show the old date. Best to use a variable in the case where the content might change

djsmith85 (changes requested):

fedemkr (approved):

src/Core/Resources/Localization/AppResources.resx
fedemkr: Just in case because of timing on release, we've chosen to change the id of the string so the translations get the modification.
Tried to quickly add the variable with date time string but there is no standard format we can use for this specific date pattern that is automatically localized; thus chosen the id change.

djsmith85 (approved):

PR #3170, opened 2024-04-18
[PM-7257] android add support for web authn resident key credential property in our net mobile app 2

coroiu:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
This redirects all android fido2 calls to the `Fido2Client`, and correctly maps to/from extension objects

## Before you submit

### DIFF
--- a/src/App/Platforms/Android/Autofill/CredentialHelpers.cs
+++ b/src/App/Platforms/Android/Autofill/CredentialHelpers.cs
@@ -7,6 +7,7 @@ using AndroidX.Credentials.Provider;
 using AndroidX.Credentials.WebAuthn;
 using Bit.Core.Abstractions;
 using Bit.Core.Utilities;
+using Bit.Core.Utilities.Fido2.Extensions;
 using Bit.Droid;
 using Org.Json;
 using Activity = Android.App.Activity;
@@ -84,7 +85,7 @@ namespace Bit.App.Platforms.Android.Autofill
             var excludeCredentials = new List<Core.Utilities.Fido2.PublicKeyCredentialDescriptor>();
             foreach (var excludeCred in credentialCreationOptions.ExcludeCredentials)
             {
-                excludeCredentials.Add(new Core.Utilities.Fido2.PublicKeyCredentialDescriptor(){ Id = excludeCred.GetId(), Type = excludeCred.Type, Transports = excludeCred.Transports.ToArray() });
+                excludeCredentials.Add(new Core.Utilities.Fido2.PublicKeyCredentialDescriptor() { Id = excludeCred.GetId(), Type = excludeCred.Type, Transports = excludeCred.Transports.ToArray() });
             }
 
             var authenticatorSelection = new Core.Utilities.Fido2.AuthenticatorSelectionCriteria()
@@ -95,7 +96,7 @@ namespace Bit.App.Platforms.Android.Autofill
             };
 
             var timeout = Convert.ToInt32(credentialCreationOptions.Timeout);
-            
+
             var credentialCreateParams = new Bit.Core.Utilities.Fido2.Fido2ClientCreateCredentialParams()
             {
                 Challenge = credentialCreationOptions.GetChallenge(),
@@ -107,7 +108,7 @@ namespace Bit.App.Platforms.Android.Autofill
                 Attestation = credentialCreationOptions.Attestation,
                 AuthenticatorSelection = authenticatorSelection,
                 ExcludeCredentials = excludeCredentials.ToArray(),
-                //Extensions = // Can be improved later to add support for 'credProps'
+                Extensions = MapExtensionsFromJson(credentialCreationOptions),
                 SameOriginWithAncestors = true
             };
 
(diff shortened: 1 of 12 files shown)


src/App/Platforms/Android/Autofill/CredentialHelpers.cs
```diff
@@ -158,5 +159,37 @@ public static async Task CreateCipherPasskeyAsync(ProviderCreateCredentialReques
...
+        {
+            if (options == null || !options.Json.Has("extensions"))
+            {
+                return null;
+            }
+
+            var extensions = options.Json.GetJSONObject("extensions");
+            return new Fido2CreateCredentialExtensionsParams
+            {
+                CredProps = extensions.Has("credProps") && extensions.GetBoolean("credProps")
```
fedemkr, line 173: ⛏️ You can use `OptBoolean` to combine the two calls

src/Core/Services/Fido2ClientService.cs
```diff
@@ -109,16 +109,20 @@ public async Task<Fido2ClientCreateCredentialResult> CreateCredentialAsync(Fido2
                 throw new Fido2ClientException(Fido2ClientException.ErrorCode.NotSupportedError, "No supported algorithms found");
             }
 
-            var clientDataJSON = JsonSerializer.Serialize(new
+            byte[] clientDataJSONBytes = null;
+            if (clientDataHash == null)
             {
```
fedemkr, line 114: 🤔 Is it correct that if `clientDataHash` is **not** `null` then `clientDataJSONBytes` is `null`?
Same In `AssertCredentialAsync(...)`

src/Core/Services/Fido2ClientService.cs
```diff
@@ -33,7 +33,7 @@ public class Fido2ClientService : IFido2ClientService
             _makeCredentialUserInterface = makeCredentialUserInterface;
         }
 
-        public async Task<Fido2ClientCreateCredentialResult> CreateCredentialAsync(Fido2ClientCreateCredentialParams createCredentialParams)
+        public async Task<Fido2ClientCreateCredentialResult> CreateCredentialAsync(Fido2ClientCreateCredentialParams createCredentialParams, byte[] clientDataHash = null)
```
fedemkr, line 36: 🤔 Is there something on the Spec that doesn't allow us to put the `clientDataHash` in `Fido2ClientCreateCredentialParams`? Or you put it as an additional parameter here for another reason?

src/Core/Utilities/Fido2/Fido2SelectedCredential.cs
```diff
@@ -0,0 +1,10 @@
+using Bit.Core.Models.View;
+
+public class Fido2SelectedCredential
+{
+    public byte[] Id { get; set; }
+
+    public byte[] UserHandle { get; set; }
+
+    public CipherView Cipher { get; set; }
```
fedemkr, line 9: 🤔 I see that before this was being marked as `nullable`, is that not the case anymore?:
```suggestion
    #nullable enable
    public byte[]? UserHandle { get; set; }

    public CipherView? Cipher { get; set; }
```

src/Core/Services/Fido2ClientService.cs
coroiu: Yes because they are derived from each other `ClientDataJSON -> ClientDataJSONBytes -> ClientDataHash`. So if `Fido2Client` gets a `ClientDataHash` as a parameter then it won't know what that has was calculated from (Android/Browser does this for us)

src/Core/Services/Fido2ClientService.cs
coroiu: No nothing like that. The `CreateCredentialParams` are the "tradititonal parameters" defined by WebAuthn, while `clientDataHash` is our own construction, because according to WebAuthn it's intended to be created by the `Fido2Client`. Honestly I just copied `passkey-rs` :D

src/Core/Utilities/Fido2/Fido2SelectedCredential.cs
coroiu: Sort of, according to the spec this is optional in the response, but required in the request. We always save it from the request, and so we have no reason to allow this to be nullable in the response. 

> Its value’s [name](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialentity-name), [displayName](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-displayname) and [id](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-id) members are REQUIRED. [id](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialuserentity-id) can be returned as the [userHandle](https://www.w3.org/TR/webauthn-3/#dom-authenticatorassertionresponse-userhandle) in some future [authentication ceremonies](https://www.w3.org/TR/webauthn-3/#authentication-ceremony),
> -- https://www.w3.org/TR/webauthn-3/#dictionary-makecredentialoptions

PR #3188, opened 2024-04-24
[PM-7553] Implement native app Fido2 credentials

fedemkr:
## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective

Implement Fido2 credentials autofill and creation from native Android apps.

## Code changes

* **Fido2ClientService:** Added extra parameters objects to pass additional data like the client data hash and the android package name that is needed for the client data json. Also, added additional logic for the origin validations so they're not performed when the request comes from a native app.
* **CallingAppInfoExtensions:** Extracted the `CallingAppInfo` to Android origin here so it can be reused.
* **CredentialHelpers:** Updated the code with the new needed parameters and improved the code when the flows fails. Also updated `origin` accordingly depending on whether the `Origin` from the RP is `null` or not.
* **CredentialProviderSelectionActivity:** Updated the code with the new needed parameters and improved the code when the flows fails. Also updated `origin` accordingly depending on whether the `Origin` from the RP is `null` or not. Also changed the response build to be done manually given that the object from the Credentials package was giving some errors.

## Before you submit

### DIFF
--- a/src/App/Platforms/Android/Autofill/CredentialHelpers.cs
+++ b/src/App/Platforms/Android/Autofill/CredentialHelpers.cs
@@ -7,9 +7,11 @@ using AndroidX.Credentials.Exceptions;
 using AndroidX.Credentials.Provider;
 using AndroidX.Credentials.WebAuthn;
 using Bit.App.Abstractions;
+using Bit.App.Droid.Utilities;
 using Bit.Core.Abstractions;
 using Bit.Core.Resources.Localization;
 using Bit.Core.Utilities;
+using Bit.Core.Utilities.Fido2;
 using Bit.Core.Utilities.Fido2.Extensions;
 using Bit.Droid;
 using Org.Json;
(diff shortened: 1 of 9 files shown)


src/App/Platforms/Android/Autofill/CredentialHelpers.cs
```diff
@@ -77,14 +79,23 @@ private static PublicKeyCredentialCreationOptions GetPublicKeyCredentialCreation
         public static async Task CreateCipherPasskeyAsync(ProviderCreateCredentialRequest getRequest, Activity activity)
         {
             var callingRequest = getRequest?.CallingRequest as CreatePublicKeyCredentialRequest;
+
+            if (callingRequest is null)
```
vvolkgang, line 83: Did you see this happening or it it a _just in case_ check?

fedemkr: _Just in case check_ and Google's sample also has it and terminates the operation in case that happens.

coroiu (approved):
Looks good to me 👍

vvolkgang (approved):
Approving the Android specific changes after @coroiu reviewed the FIDO2 additions 👍🏾

PR #3196, opened 2024-04-26
[PM-7960] Revert change from passkey provider pr

andrebispo5:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix bug where users were unable to decrypt vault with MP after SSO authentication.

## Code changes

Revert change from passkey provider PR 

## Before you submit


src/Core/Pages/Accounts/LockPage.xaml.cs
```diff
@@ -168,7 +168,7 @@ private void Unlock_Clicked(object sender, EventArgs e)
                 var tasks = Task.Run(async () =>
                 {
                     await Task.Delay(50);
-                    _vm.SubmitCommand.Execute(null);
+                    MainThread.BeginInvokeOnMainThread(async () => await _vm.SubmitAsync());
```
fedemkr, line 171: 🤔 Why is this necessary to be reverted? Do we need to invoke on main thread?

andrebispo5: This was the only change in that flow and reverting it fixed the issue, also the AppCenter report and my bug reproduction was referring a invocation out of main thread issue.

fedemkr: @andrebispo5 could you check if instead of doing this you change the command on the VM to be like this:
```c#
SubmitCommand = CreateDefaultAsyncRelayCommand(() => MainThread.InvokeOnMainThreadAsync(SubmitAsync), onException: _logger.Exception, allowsMultipleExecutions: false);
```
Because there was an issue on the passkeys branch where  in some occasions the submit flow was being called multiple times (basically the `DoOnce` was not taking the appropriate effect), here you have [the commit](https://github.com/bitwarden/mobile/pull/3020/commits/7fded63a703a98b3aa7fe7600d67923bd4698ddb).
This was happening on the `Return` keyboard but if I recall correctly it was in this `Unlock_Clicked` as well. 
If the VM thing doesn't work, we can leave this as is and see if we have any issue reported because of it from QA or in the future.

fedemkr (approved):

PR #3265, opened 2024-05-28
[PM-950] Cannot save new login if personal ownership is disabled

LRNcardozoWDF:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fixes https://github.com/bitwarden/mobile/issues/1874

Add the ability to select a collection for the new item if personal ownership is disabled.

## Code changes

* **LoginAddViewController.cs:** Refactor how the number of rows and sections are calculated based on the value of the flag `_personalOwnershipPolicyApplies`. Added cells to warn the user the personal ownership is disabled and allow organisation and collection selection.
* **FormEntryTableViewCell.cs:** Allow a form entry empty.

## Screenshots

**BEFORE**
image

**AFTER**
image
image

## Before you submit


src/iOS.Core/Views/FormEntryTableViewCell.cs
```diff
@@ -128,18 +129,18 @@ public class FormEntryTableViewCell : ExtendedUITableViewCell, ISelectable
                         NSLayoutConstraint.Create(TextField, NSLayoutAttribute.Top, NSLayoutRelation.Equal, ContentView, NSLayoutAttribute.Top, 1f, 10f));
                 }
 
-                if (height.HasValue)
-                {
-                    ContentView.AddConstraint(
-                        NSLayoutConstraint.Create(TextField, NSLayoutAttribute.Height, NSLayoutRelation.Equal, null, NSLayoutAttribute.NoAttribute, 1f, height.Value));
-                }
-
```
fedemkr, line 136: ❓ why was this removed from here? Doesn't it break some case for the TextField?

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -167,7 +238,10 @@ protected virtual async Task SaveAsync()
                             null : UsernameCell.TextField.Text,
                         Password = string.IsNullOrWhiteSpace(PasswordCell.TextField.Text) ?
                             null : PasswordCell.TextField.Text,
-                    }
+                    },
+                    CollectionIds = CollectionCell.SelectedItem.ToString() == AppResources.NoCollectionsToList ?
+                            null : new HashSet<string> { _collections.ElementAtOrDefault(CollectionCell.SelectedIndex)?.Id },
+                    OrganizationId = OrganizationCell.SelectedItem != null ? _organizations.ElementAtOrDefault(CollectionCell.SelectedIndex)?.Id : null,
```
fedemkr, line 244: ⚠️  Shouldn't this be using the `SelectedIndex` of the `OrganizationCell`?

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -167,7 +238,10 @@ protected virtual async Task SaveAsync()
                             null : UsernameCell.TextField.Text,
                         Password = string.IsNullOrWhiteSpace(PasswordCell.TextField.Text) ?
                             null : PasswordCell.TextField.Text,
-                    }
+                    },
+                    CollectionIds = CollectionCell.SelectedItem.ToString() == AppResources.NoCollectionsToList ?
+                            null : new HashSet<string> { _collections.ElementAtOrDefault(CollectionCell.SelectedIndex)?.Id },
```
fedemkr, line 243: ⛏️ Format
```suggestion
                    CollectionIds = CollectionCell.SelectedItem.ToString() == AppResources.NoCollectionsToList 
                        ? null 
                        : new HashSet<string> { _collections.ElementAtOrDefault(CollectionCell.SelectedIndex)?.Id },
```

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -121,6 +179,19 @@ public override void ViewDidLoad()
             base.ViewDidLoad();
         }
 
+        private void Type_ValueChanged(object sender, EventArgs e)
```
fedemkr, line 182: 🎨 Rename to something more meaningful of what this method does, like `OrganizationCell_ValueChanged`

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -121,6 +179,19 @@ public override void ViewDidLoad()
...
+        {
+            var organizationList = _organizations.OrderBy(o => o.Name).ToList();
+            var collectionsNames = _writeableCollections.Where(c => c.OrganizationId == _organizations.ElementAt(OrganizationCell.SelectedIndex).Id).Select(s => s.Name).ToList();
+            if (collectionsNames.Count == 0)
+            {
+                collectionsNames.Insert(0, AppResources.NoCollectionsToList);
+            }
+
+            CollectionCell.Items = collectionsNames;
```
fedemkr, line 191: 🎨 Extract to method for reuse; given this is the same as in [above lines](https://github.com/bitwarden/mobile/pull/3265/files#diff-0ab7e04015c2d47f1b83c22c3f9128358377c95c1caecf8f34a416f5eec7a83fR150-R156)

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -113,6 +135,42 @@ public override void ViewDidLoad()
...
+            {
+
+                _organizations = _organizationService.GetAllAsync().GetAwaiter().GetResult().OrderBy(o => o.Name).ToList();
+
+                OrganizationCell.Items = _organizations.Select(o => o.Name).ToList();
+                OrganizationCell.ValueChanged += Type_ValueChanged;
+
+                _collections = _collectionService.GetAllDecryptedAsync().GetAwaiter().GetResult();
+                _writeableCollections = _collections.Where(c => !c.ReadOnly).OrderBy(c => c.Name).ToList();
+                var collectionsNames = _writeableCollections.Where(c => c.OrganizationId == _organizations.ElementAt(OrganizationCell.SelectedIndex).Id).Select(s => s.Name).ToList();
```
fedemkr, line 150: ⛏️ Format
```suggestion
                var collectionsNames = _writeableCollections
                        .Where(c => c.OrganizationId == _organizations.ElementAt(OrganizationCell.SelectedIndex).Id)
                        .Select(s => s.Name)
                        .ToList();
```

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -113,6 +135,42 @@ public override void ViewDidLoad()
             folderNames.Insert(0, AppResources.FolderNone);
             FolderCell.Items = folderNames;
 
+            _personalOwnershipPolicyApplies = _policyService.PolicyAppliesToUser(PolicyType.PersonalOwnership).GetAwaiter().GetResult();
+            var index = 0;
+            if (_personalOwnershipPolicyApplies)
```
fedemkr, line 140: 🎨 This method is pretty huge, IMO personal ownership logic should be extracted to a different method we can call here so it doesn't make this method even bigger than it is.

src/iOS.Core/Controllers/LoginAddViewController.cs
```diff
@@ -113,6 +135,42 @@ public override void ViewDidLoad()
...
+            {
+
+                _organizations = _organizationService.GetAllAsync().GetAwaiter().GetResult().OrderBy(o => o.Name).ToList();
+
+                OrganizationCell.Items = _organizations.Select(o => o.Name).ToList();
+                OrganizationCell.ValueChanged += Type_ValueChanged;
+
+                _collections = _collectionService.GetAllDecryptedAsync().GetAwaiter().GetResult();
```
fedemkr, line 148: 🤔 Could you test if `_collectionService.GetAllDecryptedAsync` takes too long the app crashes? You can mock this by just adding a task delay of some seconds in the first line inside `GetAllDecryptedAsync()` which will be mocking the case of a really large number of collections to decrypt.
We shouldn't be using `GetAwaiter().GerResult()` and just put some loading on the view and performing the work asynchronously updating the table view afterwards but well I know we already have other calls that use that like `Folders`; so I guess it's "fine" for now as long as this loads "fast" and doesn't crash the app because `ViewDidLoad` took to long to finish.

fedemkr (changes requested):

PR #3394, opened 2024-07-02
BIT-2442: check type before extracting autofill text

david-livefront:
## 🎟️ Tracking

[BIT-2442](https://livefront.atlassian.net/browse/BIT-2442)

## 📔 Objective

This PR adds a check to avoid an exception when attempting to parse autofill data as text.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
@@ -26,7 +26,12 @@ fun AutofillValue.extractMonthValue(
 /**
  * Extract a text value from this [AutofillValue].
  */
-fun AutofillValue.extractTextValue(): String? = this
-    .textValue
-    .takeIf { it.isNotBlank() }
-    ?.toString()
+fun AutofillValue.extractTextValue(): String? =
+    if (this.isText) {
+        this
+            .textValue
+            .takeIf { it.isNotBlank() }
+            ?.toString()
+    } else {
+        null
+    }
--- a/app/src/test/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensionsTest.kt
+++ b/app/src/test/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensionsTest.kt
@@ -79,6 +79,7 @@ class AutofillValueExtensionsTest {
     fun `extractTextValue should return textValue when not blank`() {
         // Setup
         val autofillValue: AutofillValue = mockk {
+            every { isText } returns true
             every { textValue } returns TEXT_VALUE
         }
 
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
```diff
@@ -26,7 +26,11 @@ fun AutofillValue.extractMonthValue(
 /**
  * Extract a text value from this [AutofillValue].
  */
-fun AutofillValue.extractTextValue(): String? = this
-    .textValue
-    .takeIf { it.isNotBlank() }
-    ?.toString()
+fun AutofillValue.extractTextValue(): String? =
+    if (isText) {
+        textValue
```
brian-livefront, line 31: Does the act of just calling `textValue` blow up if you don't check `isText` first?

app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
```diff
@@ -26,7 +26,11 @@ fun AutofillValue.extractMonthValue(
...
+            .takeIf { it.isNotBlank() }
```
shannon-livefront, line 32: I'm curious about the formatting of having one condition in the `if (...)` and one condition in the `.takeIf`. Why not combine them like `.takeIf { it.isNotBlank() && isText }`?

app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
david-livefront: Only if it is not text:

```kotlin
    @NonNull public CharSequence getTextValue() {
        Preconditions.checkState(isText(), "value must be a text value, not type=%d", mType);
        return (CharSequence) mValue;
    }
```

brian-livefront: Also, the pattern (in this file at least) is to explicitly use `this` where possible so we should probably stick with that.

app/src/main/java/com/x8bit/bitwarden/data/autofill/util/AutofillValueExtensions.kt
david-livefront: The `if` is required since the crash occurs as soon as you call `textValue`, so we have to cal `isText` first.

brian-livefront: Yeah I was wondering the same. Looks like it would blow up before that point unfortunately.

shannon-livefront: Got it, thanks!

brian-livefront (approved):

david-livefront:
Thanks @brian-livefront

PR #3435, opened 2024-07-10
Add support for AutoCompleteTextView as an input field

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

Update autofill to identify an `AutoCompleteTextView` as an input field. This allowed the username field to be found when logging in to the Linked In app.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt
@@ -2,13 +2,9 @@ package com.x8bit.bitwarden.data.autofill.util
 
 import android.app.assist.AssistStructure
 import android.view.View
+import android.widget.EditText
 import com.x8bit.bitwarden.data.autofill.model.AutofillView
 import com.x8bit.bitwarden.ui.platform.base.util.orNullIfBlank
-
-/**
- * The class name of the android edit text field.
- */
-private const val ANDROID_EDIT_TEXT_CLASS_NAME: String = "android.widget.EditText"
 
 /**
  * The default web URI scheme.
@@ -59,7 +55,18 @@ private val SUPPORTED_VIEW_HINTS: List<String> = listOf(
  * Whether this [AssistStructure.ViewNode] represents an input field.
  */
 private val AssistStructure.ViewNode.isInputField: Boolean
-    get() = className == ANDROID_EDIT_TEXT_CLASS_NAME || htmlInfo.isInputField
+    get() {
+        val isEditText = className
+            ?.let {
+                try {
+                    Class.forName(it)
+                } catch (e: ClassNotFoundException) {
+                    null
+                }
+            }
+            ?.let { EditText::class.java.isAssignableFrom(it) } == true
+        return isEditText || htmlInfo.isInputField
+    }
 
 /**
  * Attempt to convert this [AssistStructure.ViewNode] into an [AutofillView]. If the view node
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt
```diff
@@ -6,9 +6,12 @@ import com.x8bit.bitwarden.data.autofill.model.AutofillView
 import com.x8bit.bitwarden.ui.platform.base.util.orNullIfBlank
 
 /**
- * The class name of the android edit text field.
+ * A lost of possible class names of the android edit text field.
```
dseverns-livefront, line 9:
```suggestion
 * A list of possible class names of the android edit text field.

File: app/src/main/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt [hunk 2]
```diff
@@ -6,9 +6,12 @@ import com.x8bit.bitwarden.data.autofill.model.AutofillView
...
  */
-private const val ANDROID_EDIT_TEXT_CLASS_NAME: String = "android.widget.EditText"
+private val ANDROID_INPUT_FIELD_CLASS_NAMES: List<String> = listOf(
+    "android.widget.EditText",
+    "android.widget.AutoCompleteTextView",
```
brian-livefront, line 13: Question about this: `AutoCompleteTextView` subclasses `EditText` so I'm curious if there is a way we can just check that `EditText` is one of the superclasses of the current class. This is likely an issue with `AppCompatEditText` as well.

app/src/test/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensionsTest.kt
```diff
@@ -177,13 +177,32 @@ class ViewNodeExtensionsTest {
...
+        // Test
+        val actual = viewNode.toAutofillView()
+
+        // Verify
+        assertEquals(expected, actual)
+    }
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `toAutofillView should return AutofillView Login Username when is android text field and is AutoCompleteTextView`() {
```
dseverns-livefront, line 199: 👍 🔥

dseverns-livefront (dismissed):

app/src/main/java/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt
dseverns-livefront: hmm would something like this work? 
```kotlin
private val AssistStructure.ViewNode.isInputField: Boolean
    get() {
        val kClass: KClass<*> = className?.let { Class.forName(it).kotlin }
        return (kClass as? EditText) != null
    }
```
testing may be hell though

brian-livefront: Oh that's a pretty interesting idea if that works. Worth a shot! Otherwise I looked into this a little more and this `className` is getting derived from the `getAccessibilityClassName` of the class and so far the only subclasses of `EditText` that override that value that I can find are:

- `AutoCompleteTextView`
- `MultiAutoCompleteTextView `

But your approach would cover all the bases and be future-proof so I'd go in that direction if possible.

david-livefront: I did some testing and profiled it to make sure we weren't incurring to much reflection overhead and seems fine.

Ended up with this:

```kotlin
private val AssistStructure.ViewNode.isInputField: Boolean
    get() {
        val isEditText = className
            ?.let { Class.forName(it) }
            ?.let { EditText::class.java.isAssignableFrom(it) } == true
        return isEditText || htmlInfo.isInputField
    }
```

brian-livefront (approved):

david-livefront:
Thanks @brian-livefront & @dseverns-livefront

PR #3446, opened 2024-07-11
Update no network errors to be consistent and have a space inbetween concatenated text

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR updates the way we concatenate no network errors in the app. In many cases, we were not adding a space between the two strings which caused them to run into each other. Some spots used newline to separate the two strings and those have also been updated to maintain consistency throughout the app.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
@@ -610,7 +610,10 @@ class SearchViewModel @Inject constructor(
                         viewState = SearchState.ViewState.Error(
                             message = R.string.internet_connection_required_title
                                 .asText()
-                                .concat(R.string.internet_connection_required_message.asText()),
+                                .concat(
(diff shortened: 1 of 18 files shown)


SaintPatrck (approved):
Gets the job done. 🚀 

💭 Would it make sense to define an ext func like, `concatWithSpace()` to reduce some of this duplication and potentially prevent run-ons during concatenation in the future?

david-livefront:
@SaintPatrck I don't like the idea of concatenating string resources at all, since it does not take into account the localization and what that means for sentence structure. Adding that method would just make it easier to do and I'd prefer to not do that.

I think we would ideally just make a new string resource for this.

SaintPatrck:
That makes sense. We are now decoupled from the MAUI app strings so we can add new ones as we need. Only downside to that is we lose existing translations for those strings because we're creating a new one that doesn't match keys or values of other translations. 🤔 

Topic for another time I suppose. 

:shipit:

PR #3463, opened 2024-07-12
[PM-8137] Set initial FIDO 2 user verification state

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8137

## 📔 Objective

When a FIDO 2 intent is received the FIDO user verification state is set to unverified so that user verification settings will be evaluated and respected if necessary.

When the user unlocks their vault after the first intent is received the FIDO user verification state is set to verified.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
@@ -220,6 +220,9 @@ class MainViewModel @Inject constructor(
             }
 
             fido2CredentialRequestData != null -> {
+                // Set the user's verification status when a new FIDO 2 request is received to force
+                // explicit verification if the user's vault is unlocked when the request is
+                // received.
                 fido2CredentialManager.isUserVerified = false
                 specialCircumstanceManager.specialCircumstance =
                     SpecialCircumstance.Fido2Save(
--- a/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
@@ -7,6 +7,7 @@ import androidx.lifecycle.viewModelScope
 import com.bitwarden.vault.CipherView
 import com.x8bit.bitwarden.data.auth.repository.AuthRepository
 import com.x8bit.bitwarden.data.auth.util.getPasswordlessRequestDataIntentOrNull
+import com.x8bit.bitwarden.data.autofill.fido2.manager.Fido2CredentialManager
 import com.x8bit.bitwarden.data.autofill.fido2.util.getFido2CredentialRequestOrNull
 import com.x8bit.bitwarden.data.autofill.manager.AutofillSelectionManager
 import com.x8bit.bitwarden.data.autofill.util.getAutofillSaveItemOrNull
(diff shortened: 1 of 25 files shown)


app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -218,6 +220,7 @@ class MainViewModel @Inject constructor(
             }
 
             fido2CredentialRequestData != null -> {
+                fido2CredentialManager.isUserVerified = false
```
shannon-livefront, line 223: Should we include some documentation about why this is always set to false?

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/manager/Fido2CredentialManager.kt
```diff
@@ -11,6 +11,12 @@ import com.x8bit.bitwarden.data.autofill.fido2.model.Fido2ValidateOriginResult
  */
 interface Fido2CredentialManager {
 
+    /**
+     * Returns true when the user has performed an explicit verification action. E.g., biometric
+     * verification, device credential verification, or vault unlock.
+     */
+    var isUserVerified: Boolean
```
shannon-livefront, line 18: Dumb question, but for my own knowledge, this is only for the case where the user doesn't have screen lock enabled on their device so they haven't been verified by the time they get to the app, right?

app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockViewModelTest.kt
```diff
@@ -871,6 +876,38 @@ class VaultUnlockViewModelTest : BaseViewModelTest() {
...
+                isBiometricLogin = true,
+            ),
+        )
+
+        verify { fido2CredentialManager.isUserVerified = true }
+    }
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `on ReceiveVaultUnlockResult should set FIDO 2 user verification state to verified when result is not Success`() {
```
shannon-livefront, line 897:
```suggestion
    fun `on ReceiveVaultUnlockResult should set FIDO 2 user verification state to not verified when result is not Success`() {
```

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/manager/Fido2CredentialManager.kt
SaintPatrck: Not a dumb question at all. It's not exclusive to device verification, but is only relevant in the context of FIDO operations. It represents a broader verification state that also includes BW methods of authentication like unlocking the vault or entering their MP to view a protected cipher. The expectation is that when we add BW authentication methods to this flow we will update `isUserVerified` based on the result.

shannon-livefront: Awesome thanks!

app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -220,6 +220,9 @@ class MainViewModel @Inject constructor(
             }
 
             fido2CredentialRequestData != null -> {
+                // Set the user's verification status when a new FIDO 2 request is received to force
+                // explicit verification if the user's vault is unlocked when the request is
```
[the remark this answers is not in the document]
SaintPatrck: No. The scenario we're handling here is if the user wants to register/authenticate a passkey and their vault is already unlocked but the relying party requires/prefers explicit user verification. If the user's vault is **locked** when this intent is received they have to unlock it, which marks them verified when done successfully.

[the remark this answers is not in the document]
shannon-livefront: Ahhh, got it, thanks!

shannon-livefront (approved):
Looks good!

PR #3619, opened 2024-07-23
[PM-6702] 3# Open app from App Link to CompleteRegistration

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-6702

## 📔 Objective

The application should handle the URL to verify the user's email. Clicking the email should open the app in the complete registration screen.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -201,6 +203,14 @@ class MainViewModel @Inject constructor(
                     )
             }
 
+            completeRegistrationData != null -> {
+                specialCircumstanceManager.specialCircumstance =
+                    SpecialCircumstance.CompleteRegistration(
+                        completeRegistrationData = completeRegistrationData,
+                        timestamp = System.currentTimeMillis(),
```
dseverns-livefront, line 210: ⛏️ I think the preference has been to use `Clock.millis()` which will get the same thing, but will be easier for testing since you won't need to statically mock it. `Clock` should be able to just be injected.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
```diff
@@ -122,6 +137,14 @@ class RootNavViewModel @Inject constructor(
...
                     null,
                     -> RootNavState.VaultUnlocked(activeUserId = userState.activeAccount.userId)
+
+                    is SpecialCircumstance.CompleteRegistration ->
+                        RootNavState.CompleteOngoingRegistration(
+                            email = specialCircumstance.completeRegistrationData.email,
+                            verificationToken = specialCircumstance.completeRegistrationData.verificationToken,
+                            fromEmail = specialCircumstance.completeRegistrationData.fromEmail,
+                            timestamp = specialCircumstance.timestamp,
+                        )
```
SaintPatrck, line 147: ⚠️  This branch is unreachable. `CompleteRegistration` is being handled above, on line 70.

app/src/test/java/com/x8bit/bitwarden/MainViewModelTest.kt
```diff
@@ -306,6 +313,31 @@ class MainViewModelTest : BaseViewModelTest() {
...
+        val viewModel = createViewModel()
+        val mockIntent = mockk<Intent>()
+        val completeRegistrationData = mockk<CompleteRegistrationData>()
+        every { mockIntent.getPasswordlessRequestDataIntentOrNull() } returns null
+        every { mockIntent.getAutofillSaveItemOrNull() } returns null
+        every { mockIntent.getCompleteRegistrationDataIntentOrNull() } returns completeRegistrationData
+        every { mockIntent.getAutofillSelectionDataOrNull() } returns null
+        every { intentManager.getShareDataFromIntent(mockIntent) } returns null
+        every { mockIntent.isMyVaultShortcut } returns false
+        every { mockIntent.isPasswordGeneratorShortcut } returns false
```
SaintPatrck, line 328: ⛏️ 🎨 Could define the mock Intent behavior within the initializer lambda. Cleans it up a little, imo. Don't feel obligated to change though. It works either way. 😄 

```
val mockCompleteRegistrationData = mockk<ComlpeteRegistrationData>()
val mockIntent = mockk<Intent> {
    every { getPasswordlessRequestDataIntentOrNull() } returns null
    every { getAutofillSaveItemOrNull() } returns null
    // etc...
}
```
or even make `mockIntent` "relaxed" with 
```
val mockIntent = mockk<Intent>(relaxed = true) {
    every { getCompleteRegistrationDataIntentOrNull() } returns mockData
}
```
which will make functions return default values (`null` and `false`) unless you set them to do otherwise.

SaintPatrck (changes requested):
Looking good. Found one possible duplicate `when` branch. Left a few other nitpick suggestions.

dseverns-livefront (approved):
Thanks for making the changes, once you are able to point this to `main`, looks good to go!

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
```diff
@@ -118,6 +133,7 @@ class RootNavViewModel @Inject constructor(
                         )
                     }
 
+                    is SpecialCircumstance.CompleteRegistration,
```
david-livefront, line 136: This should never happen, right?

Maybe we should just throw for this case

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
```diff
@@ -66,6 +67,20 @@ class RootNavViewModel @Inject constructor(
 
             userState?.activeAccount?.needsPasswordReset == true -> RootNavState.ResetPassword
 
+            specialCircumstance is SpecialCircumstance.CompleteRegistration -> {
+                // When the user is on the lock screen or already in the vault
+                if (userState?.activeAccount != null && !authRepository.hasPendingAccountAddition) {
+                    authRepository.hasPendingAccountAddition = true
+                    return
```
david-livefront, line 74: We can't just return from here, as that will leave the user in a bad state. Why aren't we continuing to the `CompleteOngoingRegistration` state?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavScreen.kt
```diff
@@ -189,6 +191,18 @@ fun RootNavScreen(
...
             }
+
+            is RootNavState.CompleteOngoingRegistration -> {
+                navController.navigateToAuthGraph(rootNavOptions)
+                // TODO PR-3622: add navigation to complete registration
+                /*
+                    navController.navigateToCompleteRegistration(
+                        emailAddress = currentState.email,
+                        verificationToken = currentState.verificationToken,
+                        fromEmail = currentState.fromEmail
```
david-livefront, line 202: The TODO is good enough, we do not want to leave this commented code here.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
andrebispo5: After the set of `authRepository.hasPendingAccountAddition` to true it will trigger this code again and continue to the `CompleteOngoingRegistration`.

dseverns-livefront: 💭 I was under the impression this was to prevent an edge case navigating to the `CompleteRegistration` screen if the there was some active session in progress between the time when the registration started and this intent is handled. I think technically I could start this process and then go switch to another account before completing this. Would it make more sense to mutate the `hasPendingAccountAddition` in the `MainViewModel` before pushing the data through the circumstance manager?

andrebispo5: This is when the user is active and we click the **verify email** in the email that was sent. Setting the `hasPendingAccountAddition` simulates the click of "add new account" from the account switcher. After that it will launch the `Landing` -> `CompleteRegistration`. Having the `hasPendingAccountAddition` also allows the user to go back to `Landing` and use the account switcher and pick another account.

david-livefront: If the state is the same, the re-trigger will be ignored, we should just return the correct state everytime:

```kotlin
if (userState?.activeAccount != null) {
    authRepository.hasPendingAccountAddition = true
}
CompleteOngoingRegistration(...)
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
```diff
@@ -122,6 +137,11 @@ class RootNavViewModel @Inject constructor(
                     SpecialCircumstance.VaultShortcut,
                     null,
                     -> RootNavState.VaultUnlocked(activeUserId = userState.activeAccount.userId)
+
+                    is SpecialCircumstance.CompleteRegistration ->
+                        throw IllegalStateException(
```
david-livefront, line 142: Anything multi-line should be wrapped in curly braces

app/src/test/java/com/x8bit/bitwarden/MainViewModelTest.kt
```diff
@@ -306,6 +316,32 @@ class MainViewModelTest : BaseViewModelTest() {
...
+        every { intentManager.getShareDataFromIntent(mockIntent) } returns null
+
+        viewModel.trySendAction(
+            MainAction.ReceiveFirstIntent(
+                intent = mockIntent,
+            ),
+        )
+        assertEquals(
+            completeRegistrationData,
+            (specialCircumstanceManager.specialCircumstance as SpecialCircumstance.CompleteRegistration).completeRegistrationData,
```
david-livefront, line 341: You shouldn't need the cast here:

```kotlin
        assertEquals(
            SpecialCircumstance.CompleteRegistration(
                completeRegistrationData = completeRegistrationData,
                timestamp = FIXED_CLOCK.millis(),
            ),
            specialCircumstanceManager.specialCircumstance,
        )
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
david-livefront: Still have an issue here, we cannot return here without addressing the state.

andrebispo5: Moving the set of `hasPendingAccountAddition` to the `MainViewModel` before firing `CompleteRegistration` special circumstance the did the trick.

dseverns-livefront (approved):

SaintPatrck (approved):

PR #3631, opened 2024-07-25
[PM-10058] Non-remembered device TDE issue in same session

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-10058

## 📔 Objective

If a user, using SSO TDE  logouts out and attempts to log back in during an active app session (app has not been destroyed) they will get stuck in a state where they are unable to fully authenticate with Admin approval or other device. 

The root cause found was in debugging
```
val updatedRootNavState = when {
            userState?.activeAccount?.trustedDevice?.isDeviceTrusted == false &&
                !userState.activeAccount.isVaultUnlocked &&
                !userState.activeAccount.hasManualUnlockMechanism -> RootNavState.TrustedDevice
```
once the vault was unlocked for the user if it was never set to locked and they logged it, it would always return true.
                

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
@@ -93,6 +93,7 @@ class VaultLockManagerImpl(
         observeAppForegroundChanges()
         observeUserSwitchingChanges()
         observeVaultTimeoutChanges()
+        observeUserLogoutResults()
     }
 
     override fun isVaultUnlocked(userId: String): Boolean =
@@ -229,7 +230,7 @@ class VaultLockManagerImpl(
         )
 
         if (invalidUnlockAttempts >= MAXIMUM_INVALID_UNLOCK_ATTEMPTS) {
-            userLogoutManager.logout(userId)
+            userLogoutManager.logout(userId = userId)
         }
     }
 
@@ -370,6 +371,15 @@ class VaultLockManagerImpl(
             .launchIn(unconfinedScope)
     }
 
+    private fun observeUserLogoutResults() {
+        userLogoutManager
+            .logoutResultFlow
+            .onEach {
+                setVaultToLocked(it.loggedOutUserId)
+            }
+            .launchIn(unconfinedScope)
+    }
+
     private fun vaultTimeoutChangesForUserFlow(userId: String) =
         settingsRepository
             .getVaultTimeoutStateFlow(userId = userId)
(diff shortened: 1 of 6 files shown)


SaintPatrck (dismissed):

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -227,7 +227,8 @@ class VaultLockManagerImpl(
         )
 
         if (invalidUnlockAttempts >= MAXIMUM_INVALID_UNLOCK_ATTEMPTS) {
-            userLogoutManager.logout(userId)
+            setVaultToLocked(userId = userId)
+            userLogoutManager.logout(userId = userId)
```
brian-livefront, line 231: Is it possible for the `VaultLockManager` to just observe accounts and whenever there is one that logs out in some way, we automatically lock?

dseverns-livefront: Yeah I was thinking that, would be easy to do with adding VLM to the ULM as a dependency.

david-livefront: The `UserLogoutManager` is already injected into the `VaultLockManager`, you can just observe the login state from the `UserLogoutManager` there.

app/src/test/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerTest.kt
```diff
@@ -210,7 +215,6 @@ class VaultLockManagerTest {
                 VaultTimeout.FourHours,
                 is VaultTimeout.Custom,
                 -> {
-                    assertTrue(vaultLockManager.isVaultUnlocked(USER_ID))
```
dseverns-livefront, line 213: These assertions were removed as they are not relevant since locking the vault takes place after call to soft logout in new flow and these are decoupled and can now be tested in isolation

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -548,6 +555,11 @@ class VaultLockManagerImpl(
         )
     }
 
+    private fun isUserLoggedOut(userId: String): Boolean {
+        val accounts = authDiskSource.userAccountTokens
+        return (accounts.find { it.userId == userId }?.isLoggedIn) == false
```
david-livefront, line 560: Is this correct?

Wouldn't it be this?
```kotlin
return (accounts.find { it.userId == userId }?.isLoggedIn) != true
```

david-livefront: I suppose this is how it worked before, so maybe just leave it alone.

david-livefront (approved):

PR #3652, opened 2024-07-30
PM-10379: Update the timeout action logic to occur immediately after requirements are met

david-livefront:
## 🎟️ Tracking

[PM-10379](https://bitwarden.atlassian.net/browse/PM-10379)

## 📔 Objective

This PR enforces that the timeout action occurs as early as possible and not just when the app is foregrounded. This was causing a bug where the vault would be left unlocked after the user used the Autofill feature until either the background process was manually killed or the app was manually foregrounded.

This is a pretty big shift compared to the previous implementation since the app would store a `LastActiveTimeMillis` on disk and reference that value when foregrounding the app.

The new mechanism is based entirely on when the app is backgrounded, the user becoming inactive (change user), or on app startup. At those times we either enforce the timeout or set a timer for when the timeout should be enforced.

This `LastActiveTimeMillis` was based on the MAUI implementation but upon looking at the code history, much of that code was left behind from a previous iterations that worked very differently than it does today. This change reflects all the current requirements without any additional logic that is not needed.

Some of that code history can be seen here:
* [Restore Vault Timeout Timer for Android](https://github.com/bitwarden/mobile/pull/1220)
* [Fix for Vault Timeout Locking Issue on Android](https://github.com/bitwarden/mobile/pull/1442)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10379]: https://bitwarden.atlassian.net/browse/PM-10379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
@@ -1,6 +1,5 @@
 package com.x8bit.bitwarden.data.vault.manager
 
-import android.os.SystemClock
 import com.bitwarden.core.InitOrgCryptoRequest
 import com.bitwarden.core.InitUserCryptoMethod
 import com.bitwarden.core.InitUserCryptoRequest
@@ -32,6 +31,8 @@ import com.x8bit.bitwarden.data.vault.repository.util.toVaultUnlockResult
 import com.x8bit.bitwarden.data.vault.repository.util.update
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -48,9 +49,7 @@ import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
-
-private const val SECONDS_PER_MINUTE = 60
-private const val MILLISECONDS_PER_SECOND = 1000
+import kotlin.time.Duration.Companion.minutes
 
 /**
  * The number of times a user may fail to unlock before they are automatically logged out.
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -70,12 +69,12 @@ class VaultLockManagerImpl(
     private val userLogoutManager: UserLogoutManager,
     private val trustedDeviceManager: TrustedDeviceManager,
     dispatcherManager: DispatcherManager,
-    private val elapsedRealtimeMillisProvider: () -> Long = { SystemClock.elapsedRealtime() },
 ) : VaultLockManager {
     private val unconfinedScope = CoroutineScope(dispatcherManager.unconfined)
 
+    private val userIdTimerJobMap = mutableMapOf<String, Job>()
```
shannon-livefront, line 75: Should we have documentation for what this does?

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -434,66 +435,95 @@ class VaultLockManagerImpl(
...
-         * This is required in the case that an account has been "soft logged out" and has an
-         * immediate time interval time out. Without this check it would be automatically switch
-         * the active user back to an authenticated user if one exists.
-         */
-        if ((accounts.find { it.userId == userId }?.isLoggedIn) == false) {
-            return
-        }
+        // Check if the user is already logged out. If this is the case no need to check timeout.
+        // This is required in the case that an account has been "soft logged out" and has an
+        // immediate time interval time out. Without this check it would be automatically switch
```
shannon-livefront, line 443:
```suggestion
        // immediate time interval timeout. Without this check, it would automatically switch
```

shannon-livefront (dismissed):
I was looking really hard for anything to give feedback on, but only had two minor nitpicks - otherwise looks awesome! Great job!

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
david-livefront: Sure

shannon-livefront (approved):

david-livefront:
Thanks @shannon-livefront

PR #3657, opened 2024-07-31
PM-10242 PM-10243 PM-10244 PM-10245 PM-10246: Welcome carousel

shannon-livefront:
## 🎟️ Tracking

[PM-10242](https://bitwarden.atlassian.net/browse/PM-10242)
[PM-10243](https://bitwarden.atlassian.net/browse/PM-10243)
[PM-10244](https://bitwarden.atlassian.net/browse/PM-10244)
[PM-10245](https://bitwarden.atlassian.net/browse/PM-10245)
[PM-10246](https://bitwarden.atlassian.net/browse/PM-10246)

## 📔 Objective

Add a new welcome carousel before the landing page for users who are opening the app for the first time (ie, who have never logged in or created an account on the device before)

NOTE: For now, the old landing page and create account views are shown, just to make sure the buttons work. It looks a little strange, but the auth flow will be updated more in later tickets.

## 📸 Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/3cd6c5a9-3c59-4163-9f0e-c287f4b35b04" /> | <video src="https://github.com/user-attachments/assets/92120951-7df9-40f9-bf97-e9eb33a1959c" /> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10242]: https://bitwarden.atlassian.net/browse/PM-10242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-10243]: https://bitwarden.atlassian.net/browse/PM-10243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-10244]: https://bitwarden.atlassian.net/browse/PM-10244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-10245]: https://bitwarden.atlassian.net/browse/PM-10245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-10246]: https://bitwarden.atlassian.net/browse/PM-10246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/welcome/WelcomeScreen.kt
```diff
@@ -0,0 +1,216 @@
...
+private fun WelcomeScreenContent(
+    state: WelcomeState,
+    pagerState: PagerState,
+    onPagerSwipe: (Int) -> Unit,
+    onDotClick: (Int) -> Unit,
+    onCreateAccountClick: () -> Unit,
+    onLoginClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    LaunchedEffect(pagerState) {
```
david-livefront, line 105: Would this also work?
```kotlin
LaunchedEffect(key1 = pagerState.currentPage) {
```

shannon-livefront: It does work! Thanks! 🤩

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/welcome/WelcomeScreen.kt
```diff
@@ -0,0 +1,215 @@
...
+                .fillMaxWidth(),
+        )
+
+        BitwardenTextButton(
+            label = stringResource(id = R.string.log_in),
+            onClick = onLoginClick,
+            modifier = Modifier
+                .padding(horizontal = 16.dp)
+                .padding(bottom = 32.dp),
+        )
```
david-livefront, line 146: Be sure to add one last Spacer for the nav bar.

```kotlin
    Spacer(modifier = Modifier.navigationBarsPadding())
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavScreen.kt
```diff
@@ -76,15 +77,18 @@ fun RootNavScreen(
         popExitTransition = { toExitTransition()(this) },
     ) {
         splashDestination()
-        authGraph(navController)
+        authGraph(
+            startDestination = (state as? RootNavState.Auth)?.startDestination ?: LANDING_ROUTE,
```
david-livefront, line 81: Can you explain how this works here?

david-livefront: I think I see you are using the startDestination that you've set down below.

Would it be easier to just reuse that logic here?

```kotlin
                startDestination = if (authRepository.hasUserLoggedInOrCreatedAccount) {
                    LANDING_ROUTE
                } else {
                    WELCOME_ROUTE
                }
```

david-livefront: Hmmm, I see what you mean. The other option would be to move it into the top level of the state.

That would be a much bigger update, something like this:

```kotlin
data class RootNavState(
    val authStartDestination: String,
    val navState: NavState, // This NavState class would be the sealed class that RootNavState currently is
)
```

Curious if you or @brian-livefront have any thoughts on this?

shannon-livefront: I do like that option

brian-livefront: So I'm not entirely certain whether or not dynamically changing destinations in a graph is a "supported" thing. The way I picture this is that setting up the `NavHost` and its destinations is really more of a one-time thing and after that, all the dynamic logic has to do with actually navigating to and within those destinations. Personally this sounds to me more like we what need is a secondary navigation once we're in the graph, which is something we already do below.

For example, when dealing with external intents like creating a new send, we first navigate to the overall graph and then do another navigation within it:

```kotlin

            RootNavState.VaultUnlockedForNewSend -> {
                navController.navigateToVaultUnlock(rootNavOptions)
                navController.navigateToAddSend(
                    sendAddType = AddSendType.AddItem,
                    navOptions = rootNavOptions,
                )
            }
```

Is there something about what you need to do here that wouldn't already fit within that pattern?

brian-livefront: @shannon-livefront You would not need to recreate all that logic. Those destinations are already in the graph, so you would just navigate to them like normal. The only thing you are changing with this sort of secondary navigation is the first screen that you see. What I've posted above was done to solve exactly this problem: how do you make the "start destination" dynamic.

david-livefront: I think the only other option would be to move the Carousel out of the Auth Graph and make it it's own top level screen.

this was something that I had looked into previously:
https://github.com/livefront/bitwarden-android/pull/277/files#diff-2f90874fbae4270ef31e9a410a741ddacb2a78a51211ecf22319ac5f3d24b4dbR91

brian-livefront: Yeah that's definitely an option too. Then you have to start worrying about cross-graph transitions so it would be something you'd need to do very carefully.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/welcome/WelcomeNavigation.kt
```diff
@@ -0,0 +1,32 @@
+package com.x8bit.bitwarden.ui.auth.feature.welcome
+
+import androidx.navigation.NavController
+import androidx.navigation.NavGraphBuilder
+import androidx.navigation.NavOptions
+import com.x8bit.bitwarden.ui.platform.base.util.composableWithStayTransitions
+
+const val WELCOME_ROUTE: String = "welcome"
```
david-livefront, line 8: Can we make this private?

app/src/main/java/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -385,6 +388,13 @@ class AuthRepositoryImpl(
         mutableHasPendingAccountDeletionStateFlow.value = false
     }
 
+    override suspend fun getShowWelcomeCarousel(): Boolean =
+        !settingsRepository.hasUserLoggedInOrCreatedAccount &&
+            featureFlagManager.getFeatureFlag(
+                key = FlagKey.OnboardingCarousel,
+                forceRefresh = true,
```
david-livefront, line 395: Do we need this to refresh here?

We refresh the flag on launch already

shannon-livefront: Good point, thank you!

david-livefront: You may need this: https://github.com/bitwarden/android/pull/3692

shannon-livefront: Ah that does make things a lot easier! Thank you!! 😍

app/src/main/java/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -317,6 +320,10 @@ class AuthRepositoryImpl(
     override val organizations: List<SyncResponseJson.Profile.Organization>
         get() = activeUserId?.let { authDiskSource.getOrganizations(it) }.orEmpty()
 
+    override val showWelcomeCarousel: Boolean
+        get() = !settingsRepository.hasUserLoggedInOrCreatedAccount &&
+            featureFlagManager.getFeatureFlag(FlagKey.OnboardingCarousel)
```
david-livefront, line 325: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModel.kt
```diff
@@ -68,7 +68,13 @@ class RootNavViewModel @Inject constructor(
 
             userState == null ||
                 !userState.activeAccount.isLoggedIn ||
-                userState.hasPendingAccountAddition -> RootNavState.Auth
+                userState.hasPendingAccountAddition -> {
+                if (authRepository.showWelcomeCarousel) {
+                    RootNavState.AuthWithWelcome
+                } else {
+                    RootNavState.Auth
+                }
```
brian-livefront, line 76: Is there any risk that the conditions that produce a value of `authRepository.showWelcomeCarousel` of `true` could change while the user is within the `RootNavState.AuthWithWelcome` flow itself and then some unrelated update to `userState` triggers a switch here from `RootNavState.AuthWithWelcome` to `RootNavState.Auth`?

david-livefront: That is why I was suggesting that we put it in the state.

shannon-livefront: The conditions of `authRepository.showWelcomeCarousel` are that the feature flag be enabled and the user has logged in or created an account on that device. I don't think there's a way for the `hasLoggedInOrCreatedAccount` preference to change, so the only way `authRepository.showWelcomeCarousel` could change would be if the feature flag was disabled while the user was viewing the welcome carousel (even then, I'm not sure the flag would update, given that we're not using a synchronous get with a force refresh). But in that case, navigating away from the welcome carousel sounds like desired behavior, right?

brian-livefront: Gotcha OK. Sounds good then!

app/src/main/res/values/strings.xml
```diff
@@ -927,4 +927,12 @@ Do you want to switch to this account?</string>
...
   <string name="passkey_operation_failed_because_user_could_not_be_verified">Passkey operation failed because user could not be verified.</string>
   <string name="user_verification_direction">User verification</string>
+  <string name="welcome_title_1">Privacy, prioritized</string>
+  <string name="welcome_message_1">Save logins, cards, and identities to your secure vault. Bitwarden uses zero-knowledge, end-to-end encryption to protect what’s important to you.</string>
+  <string name="welcome_title_2">Never guess again</string>
+  <string name="welcome_message_2">Set up biometric unlock and autofill to log into your accounts without typing a single letter.</string>
+  <string name="welcome_title_3">Level up your logins</string>
+  <string name="welcome_message_3">Use the generator to create and save strong, unique passwords for all your accounts.</string>
+  <string name="welcome_title_4">Your data, when and where you need it</string>
+  <string name="welcome_message_4">Save unlimited passwords across unlimited devices with Bitwarden mobile, browser, and desktop apps.</string>
```
SaintPatrck, line 937: This deviates from the established pattern of resource keys and their value aligning. @vvolkgang @fedemkr do you have any concerns with us starting to use descriptive, context based resource keys like this instead? This approach has implications on Crowdin sync, since we would no longer be changing keys when their value is changed.

Aligning keys and values like we've been doing means we lose translations when the key+value changes.

Decoupling keys and values means we retain translations when values change, but also means we have the potential to continue showing old values in other languages until corresponding translations are updated.

app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/welcome/WelcomeScreenTest.kt
```diff
@@ -0,0 +1,124 @@
...
+            .assertIsDisplayed()
+
+        mutableStateFlow.update { it.copy(pages = listOf(WelcomeState.WelcomeCard.CardThree)) }
+        composeTestRule
+            .onNodeWithText("Level up your logins")
+            .assertExists()
+            .assertIsDisplayed()
+    }
+
+    @Config(qualifiers = "land")
```
SaintPatrck, line 62: 👍

app/src/main/res/values/strings.xml
SaintPatrck: @shannon-livefront After some internal discussion we decided it's best to continue the same naming pattern for titles. Since the messages are so long we can keep them named as `welcome_message_n`.

shannon-livefront: Thank you for the guidance!

app/src/main/res/values/strings.xml
```diff
@@ -927,12 +927,12 @@ Do you want to switch to this account?</string>
   <string name="self_hosted_server_url">Self-hosted server URL</string>
   <string name="passkey_operation_failed_because_user_could_not_be_verified">Passkey operation failed because user could not be verified.</string>
   <string name="user_verification_direction">User verification</string>
-  <string name="welcome_title_1">Privacy, prioritized</string>
+  <string name="privacy_protected">Privacy, prioritized</string>
```
david-livefront, line 930: Should this be `privacy_ prioritized`?

shannon-livefront: It absolutely should be, and I should get some caffeine! 👍

SaintPatrck (approved):

PR #3660, opened 2024-07-31
PM-10094: Disable double-navigation by default

shannon-livefront:
## 🎟️ Tracking

[PM-10094](https://bitwarden.atlassian.net/browse/PM-10094)

## 📔 Objective

Nearly everywhere that navigation is possible in the app, rapidly double-clicking on the navigation button would cause the navigation action to happen twice, which would either stack two new views on top of each other or, in the case of rapidly navigating _back_ too many times, break the view entirely. This PR fixes that issue by ignoring events that occur when the view is not resumed by default.

IMPORTANT NOTE: There is a new interface called `BackgroundEvent` that non-navigation events should conform to if there is a chance they might occur (and should be processed) before the view is resumed. So far only the passkey management related events fall into this category that we know of.

## 📸 Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f512b02c-f586-4a28-9482-3cdb6d1d41d8" /> | <video src="https://github.com/user-attachments/assets/a8be77de-6769-4fcc-ad2e-c0de9ebab55a" /> |

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/d21fe21f-af27-4540-ae5a-9dde15235d9a" /> | <video src="https://github.com/user-attachments/assets/d5d001ca-50a0-47c0-8e0d-0828a3f480eb" /> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
```diff
@@ -2364,8 +2365,7 @@ sealed class VaultAddEditEvent {
     /**
      * Navigate the user to the tooltip URI.
      */
-    data object NavigateToTooltipUri :
-        VaultAddEditEvent()
+    data object NavigateToTooltipUri : VaultAddEditEvent()
```
david-livefront, line 2368: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/EventsEffect.kt
```diff
@@ -14,9 +17,21 @@ fun <E> EventsEffect(
     viewModel: BaseViewModel<*, E, *>,
     handler: suspend (E) -> Unit,
 ) {
+    val lifecycleOwner = LocalLifecycleOwner.current.lifecycle
```
david-livefront, line 20: Can we make this a function param, if you add it after `viewModel` but before `handler`, it should have no breaking changes.

david-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/EventsEffect.kt
```diff
@@ -12,11 +15,23 @@ import kotlinx.coroutines.flow.onEach
...
             .launchIn(this)
     }
 }
+
+/**
+ * Almost all the events in the app involve navigation or toasts. To prevent accidentally
+ * navigating to the same view twice, by default, events are ignored if the view is not currently
+ * resumed. To avoid that restriction, specific events can conform to [BackgroundEvent].
+ */
+interface BackgroundEvent
```
brian-livefront, line 37: Would it make more sense for this to live in its own file maybe? I'm just not sure we have other examples like this where an interface is defined in the same file as a composable function, but maybe I'm wrong about that.

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
```diff
@@ -2376,7 +2376,7 @@ sealed class VaultAddEditEvent {
      */
     data class NavigateToGeneratorModal(
         val generatorMode: GeneratorMode.Modal,
-    ) : VaultAddEditEvent()
+    ) : BackgroundEvent, VaultAddEditEvent()
```
brian-livefront, line 2379: Why does this one need to happen in the background?

shannon-livefront: That is a great catch - this was left in by accident after I was testing some things. Thank you!

app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/EventsEffect.kt
shannon-livefront: Sure, we can do that! I only put it here because I didn't know if there was a precedent for creating a file with only one line of code in it

app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/BackgroundEvent.kt
```diff
@@ -0,0 +1,8 @@
+package com.x8bit.bitwarden.ui.platform.base.util
+
+/**
+ * Almost all the events in the app involve navigation or toasts. To prevent accidentally
+ * navigating to the same view twice, by default, events are ignored if the view is not currently
+ * resumed. To avoid that restriction, specific events can conform to [BackgroundEvent].
+ */
+interface BackgroundEvent
```
brian-livefront, line 8: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/EventsEffect.kt
```diff
@@ -12,10 +15,15 @@ import kotlinx.coroutines.flow.onEach
 @Composable
 fun <E> EventsEffect(
     viewModel: BaseViewModel<*, E, *>,
+    lifecycleOwner: Lifecycle = LocalLifecycleOwner.current.lifecycle,
     handler: suspend (E) -> Unit,
 ) {
     LaunchedEffect(key1 = Unit) {
         viewModel.eventFlow
+            .filter {
+                it is BackgroundEvent ||
```
brian-livefront, line 24: Can we document at the function-level the restrictions on observing events? Something like:

```kotlin
/**
 * Convenience method for observing event flow from [BaseViewModel].
 * 
 * By default, events will only be consumed when the associated screen is 
 * resumed to avoid things like duplicate navigation calls. In order to override
 * this behavior, a given event type can implement `BackgroundEvent`.
 */
@Composable
fun <E> EventsEffect(
```

david-livefront (approved):

brian-livefront (approved):
Looks good!

PR #3683, opened 2024-08-06
[PM-10071] Sort search items with same logic as displayed items

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-10071

## 📔 Objective
Show search results in the same sort order as the items appear when displayed.

## 📸 Screenshots
| before | after |
|--------|--------|
| https://github.com/user-attachments/assets/3494ddbe-34cd-4799-a71c-f93226078f25 | https://github.com/user-attachments/assets/6157db12-9223-4c4b-9678-efcf1940708e | 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
@@ -20,6 +20,7 @@ import com.x8bit.bitwarden.data.platform.repository.SettingsRepository
 import com.x8bit.bitwarden.data.platform.repository.model.DataState
 import com.x8bit.bitwarden.data.platform.repository.util.baseIconUrl
 import com.x8bit.bitwarden.data.platform.repository.util.baseWebSendUrl
+import com.x8bit.bitwarden.data.platform.util.CompareStringSpecialCharWithPrecedence
 import com.x8bit.bitwarden.data.vault.datasource.network.model.PolicyTypeJson
 import com.x8bit.bitwarden.data.vault.repository.VaultRepository
 import com.x8bit.bitwarden.data.vault.repository.model.DeleteSendResult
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -726,9 +727,14 @@ data class SearchState(
...
         @Parcelize
         data class Content(
-            val displayItems: List<DisplayItem>,
+            private val displayItems: List<DisplayItem>,
         ) : ViewState() {
             override val hasVaultFilter: Boolean get() = true
+
+            val displayItemsSorted: List<DisplayItem>
+                get() = displayItems.sortedWith { item1, item2 ->
+                    CompareStringSpecialCharWithPrecedence.compare(item1.title, item2.title)
```
david-livefront, line 736: These seems like an expensive call to be done in the UI many times, can we do this sorting before we construct the `Content` so that it only runs the one time.

dseverns-livefront: I may be misunderstanding what `runs the one time` means in this case since the sorting would need to take place each time the list updates, which is triggered via the search term. I can move the the call to the sort on the creation of the new lists in the `SearchTypeDataExtensions.kt`. Just as a note even with it now happing in the calls through the action channel it is still technically happing on `Dispacther.MAIN` I am not seeing any rise in recomposition either. either way happy to take logic call sight out of the Composable

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/util/SearchTypeDataExtensions.kt
```diff
@@ -334,7 +335,7 @@ private fun List<SendView>.toDisplayItemList(
             baseWebSendUrl = baseWebSendUrl,
             clock = clock,
         )
-    }
+    }.sortAlphabetically()
```
david-livefront, line 338: Can you move these all down a line:

```kotlin
)
.sortAlphabetically()
```

david-livefront (approved):

PR #3710, opened 2024-08-09
PM-8522: Fix vault tab nav bar title when logging in

shannon-livefront:
## 🎟️ Tracking

[PM-8522](https://bitwarden.atlassian.net/browse/PM-8522)

## 📔 Objective

Fix the nav bar vault title of an account just logging in with organizations.

## 📸 Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/9a0992b7-a4ad-4e7c-add9-3bd0ea69bbcb" /> | <video src="https://github.com/user-attachments/assets/02fdccee-6157-4d5b-9ca4-c1eb78c02bc8" /> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-8522]: https://bitwarden.atlassian.net/browse/PM-8522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
@@ -1,12 +1,17 @@
 package com.x8bit.bitwarden.ui.platform.feature.vaultunlockednavbar
 
 import androidx.annotation.StringRes
+import androidx.lifecycle.viewModelScope
 import com.x8bit.bitwarden.R
 import com.x8bit.bitwarden.data.auth.repository.AuthRepository
+import com.x8bit.bitwarden.data.auth.repository.model.UserState
 import com.x8bit.bitwarden.data.platform.manager.SpecialCircumstanceManager
 import com.x8bit.bitwarden.data.platform.manager.model.SpecialCircumstance
 import com.x8bit.bitwarden.ui.platform.base.BaseViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.update
 import javax.inject.Inject
 
 /**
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -17,22 +22,19 @@ class VaultUnlockedNavBarViewModel @Inject constructor(
...
-            ?.organizations
-            ?.isNotEmpty()
-            ?: false
-        val vaultRes = if (hasOrganization) R.string.vaults else R.string.my_vault
-        VaultUnlockedNavBarState(
-            vaultNavBarLabelRes = vaultRes,
-            vaultNavBarContentDescriptionRes = vaultRes,
-        )
-    },
+    initialState = VaultUnlockedNavBarState(
```
david-livefront, line 25: Hmmm, so this is not guaranteed to be correct on the first go but the event we listen too below will update it?

shannon-livefront: Correct. I'm not sure why it's not correct on init - I guess it just takes a moment to load the active account and determine if it has organizations? In this update version, I noticed that the tab bar title and the title of the screen itself update at the exact same time, which makes sense if that's the case

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -17,22 +22,19 @@ class VaultUnlockedNavBarViewModel @Inject constructor(
     private val authRepository: AuthRepository,
```
dseverns-livefront, line 22: ⛏️ this no longer needs to be a `val` since we are only using in `init`.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
dseverns-livefront: So just to add a bit to this, it looks like there may be a "race" with the organizations list of the active user after login(?). Going through the previous code in the debugger the orgs list is empty, then if I lock the same account and unlock the count will be 1 (for lukus+policy). Which is why I think this one works cause we are observing the state change in the user state value when the combine in `AuthRepostitoryImpl` emits the org list.

dseverns-livefront (approved):

PR #3733, opened 2024-08-14
[PM-10884] Catch ProviderException when generating a secure key

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10884

## 📔 Objective

Prevent the application from crashing when secret keys cannot be created due to an exception within the device KeyStore.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -58,6 +59,10 @@ class BiometricsEncryptionManagerImpl(
             // user removed all biometrics from the device
             settingsDiskSource.systemBiometricIntegritySource = null
             return null
+        } catch (e: ProviderException) {
+            // key generation failed because user cannot be authenticated by the OS.
+            settingsDiskSource.systemBiometricIntegritySource = null
+            return null
```
SaintPatrck, line 65: 🗒️ Unable to reproduce the reported crash. However, handling the exception should prevent it from occurring if the scenario arises again in production. Should it manifest, the app will exhibit the same behavior as the preceding `catch` block.

dseverns-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -89,7 +94,10 @@ class BiometricsEncryptionManagerImpl(
 
     /**
      * Generates a [SecretKey] from which the [Cipher] will be generated.
+     *
+     * @throws ProviderException if [KeyGenerator] is unable to generate a key.
      */
+    @Throws(ProviderException::class)
```
david-livefront, line 100: We generally don't do this, it's kinda a Java thing. How strongly do you feel about it?

dseverns-livefront (approved):

david-livefront (dismissed):

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
shannon-livefront: Do we also want to catch generic errors?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -89,6 +94,8 @@ class BiometricsEncryptionManagerImpl(
 
     /**
      * Generates a [SecretKey] from which the [Cipher] will be generated.
+     *
+     * @throws ProviderException if [KeyGenerator] is unable to generate a key.
```
brian-livefront, line 98: Is there a reason we don't do all the exception handling inside this `generateKey` function itself? We really want to avoid having any functions that we know throw exceptions. This could be, like `generateKeyOrNull()`

cc: @david-livefront

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
brian-livefront: @shannon-livefront Ideally we'd only catch the specific errors we know can occur so that we don't mask new and different problems.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
SaintPatrck: Thanks for making me dig into this a little deeper. Originally, I wanted to prevent a cascade of other changes needed to handle a nullable return value from this function. Main in `createCipher`.

Now that I'm taking a second look, I'm seeing that some call sites of `createCipher` may not be handled safely. That's making me think the changes I was trying to avoid are needed regardless. Otherwise `createCipher` is going to cause the same crash. 🤔

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/accountsetup/SetupUnlockViewModel.kt
```diff
@@ -70,10 +71,9 @@ class SetupUnlockViewModel @Inject constructor(
     }
 
     private fun handleEnableBiometricsClick() {
-        sendEvent(
-            SetupUnlockEvent.ShowBiometricsPrompt(
-                // Generate a new key in case the previous one was invalidated
-                cipher = biometricsEncryptionManager.createCipher(userId = state.userId),
```
SaintPatrck, line 76: 🗒️ Previously, we assumed `createCipher` was always successful and never threw an exception. Now that we're updating it to return null when exceptions are encountered, we're going to end up showing an error dialog and forcing the user to try again.

This also applies to [these](https://github.com/bitwarden/android/pull/3733/files#diff-5633ac2c8007036010438f407924e640e1f910c5fc7ffc3053dd13915404191aL161-L166) changes in `AccountSecurityViewModel`.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -88,16 +92,21 @@ class BiometricsEncryptionManagerImpl(
...
         val keyGen = KeyGenerator.getInstance(
             KeyProperties.KEY_ALGORITHM_AES,
             ENCRYPTION_KEYSTORE_NAME,
         )
         keyGen.init(keyGenParameterSpec)
-        keyGen.generateKey()
-        return requireNotNull(getSecretKey())
+        try {
+            keyGen.generateKey()
+        } catch (e: ProviderException) {
```
brian-livefront, line 106: I'm thinking we also need to catch `InvalidAlgorithmParameterException` in here now right, rather than in `getOrCreateCipher`? I'm assuming that's coming from both `generateKey` or `getSecretKey` right?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -88,16 +92,21 @@ class BiometricsEncryptionManagerImpl(
...
+            return null
+        }
+        return getSecretKey()
```
brian-livefront, line 109: Can this throw as well?

SaintPatrck: Yup. It's masking several exceptions that could be thrown by `KeyGenerator` and `keystore` calls. Am I correct in thinking we need to set `settingsDiskSource.systemBiometricIntegritySource = null` for all of these exceptions?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
SaintPatrck: Indeed!

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
david-livefront: I think what Brian wants is for any function that can throw an exception to be updated to handle the exception and return null.

Like:
```kotlin
    private fun getSecretKeyOrNull(): SecretKey? {
        keystore.load(null)
        return try {
            keystore.getKey(ENCRYPTION_KEY_NAME, null) as? SecretKey
        } catch (blah) {
            null
        }
    }
```

And then the consumers if those APIs would set things to null based on cirumstance.

```kotlin
    override fun getOrCreateCipher(userId: String): Cipher? {
        val secretKey = getSecretKeyOrNull()
            ?: generateKeyOrNull()
            ?: run {
                // user removed all biometrics from the device
                settingsDiskSource.systemBiometricIntegritySource = null
                return null
            }
```

brian-livefront: Yeah exactly...that's what I'm thinking. Any function call that can throw gets wrapped and then the places that call those functions can decide what to do in the failure case. Another option is to use `runCatching` in order to return a `Result` and then use `fold` or any of the other tools we need.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -88,24 +101,70 @@ class BiometricsEncryptionManagerImpl(
...
-        keyGen.generateKey()
-        return requireNotNull(getSecretKey())
+    @Suppress("TooGenericExceptionCaught")
+    private fun generateKeyOrNull(): SecretKey? {
+        val keyGen = try {
+            KeyGenerator.getInstance(
+                KeyProperties.KEY_ALGORITHM_AES,
+                ENCRYPTION_KEYSTORE_NAME,
+            )
+        } catch (e: NullPointerException) {
```
brian-livefront, line 114: Is this one something that can actually happen _inside_ the `KeyGenerator.getInstance` call?

SaintPatrck: It is declared as an explicit exception thrown within `getInstance`. In practice it shouldn't ever happen because we're passing `KeyProperties.KEY_ALGORITH_AES`. 

image

david-livefront: That is only if the algorithm is null, and we always pass in `KeyProperties.KEY_ALGORITHM_AES`, so it shouldn't happen, right?

brian-livefront: Right

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -52,13 +63,15 @@ class BiometricsEncryptionManagerImpl(
...
     override fun getOrCreateCipher(userId: String): Cipher? {
-        val secretKey = try {
-            getSecretKey() ?: generateKey()
-        } catch (e: InvalidAlgorithmParameterException) {
-            // user removed all biometrics from the device
-            settingsDiskSource.systemBiometricIntegritySource = null
-            return null
-        }
+        val secretKey =
+            getSecretKey()
```
david-livefront, line 67: Can we move this up a line

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -88,24 +101,67 @@ class BiometricsEncryptionManagerImpl(
...
+            return null
+        }
+
+        return getSecretKey()
     }
 
     /**
      * Returns the [SecretKey] stored in the keystore, or null if there isn't one.
      */
     private fun getSecretKey(): SecretKey? {
```
brian-livefront, line 136: Can we just rename this to `getSecretKeyOrNull` at this point too to match?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/accountsetup/SetupUnlockViewModel.kt
david-livefront: This make sense but we don't need an internal event for that since this is not suspending.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/accountsetup/SetupUnlockViewModel.kt
```diff
@@ -152,6 +162,32 @@ class SetupUnlockViewModel @Inject constructor(
             }
         }
     }
+
+    private fun handleReceiveCreateCipherResult(
+        action: SetupUnlockAction.Internal.ReceiveCreateCipherResult,
+    ) {
+        when (val cipher = action.cipher) {
```
david-livefront, line 169: Could this just be a `let`:
```kotlin
action
    .cipher
    ?.let {
        sendEvent(
            SetupUnlockEvent.ShowBiometricsPrompt(
                // Generate a new key in case the previous one was invalidated
                cipher = cipher,
            ),
        )
    }
    ?: run {
        mutableStateFlow.update {
            it.copy(
                dialogState = SetupUnlockState.DialogState.Error(
                    title = R.string.an_error_has_occurred.asText(),
                    message = R.string.generic_error_message.asText(),
                ),
            )
        }
    }
```

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/BiometricsEncryptionManagerImpl.kt
```diff
@@ -166,7 +221,7 @@ class BiometricsEncryptionManagerImpl(
         )
 
         try {
-            createCipher(userId)
+            createCipherOrNull(userId)
         } catch (e: Exception) {
```
brian-livefront, line 225: @SaintPatrck @david-livefront Would we be comfortable with removing this `try/catch` now? I feel like if we are not ready to remove it, that means we aren't catching enough in `createCipherOrNull`.

SaintPatrck: Removed it and added exception handling where needed.

brian-livefront (dismissed):

david-livefront (approved):

PR #3831, opened 2024-08-26
PM-11310 handle email registration special circumstance after successful login

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-11310

## 📔 Objective
Currently the special circumstance for the complete registration needs to be cleared since it interferes with AuthState related state navigation. The old way we were doing this via `onCleared` in the `CompleteRegistrationViewModel` was causing a conflict with the nav stack emitting a state based nav call after making the explicit call to go to the login screen.

- Categorize a `PreLogin` type of `SpecialCircumstance`
- Add function to manager to handle clearing a special circumstance after successful login.
- Observe user state change to update.

## 📸 Screenshots

https://github.com/user-attachments/assets/9bd6333b-c424-4984-8ddf-3438e1420504

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
@@ -98,7 +98,9 @@ class MainViewModel @Inject constructor(
             .drop(count = 1)
             // Trigger an action whenever the current user changes or we go into/out of a pending
             // account state (which acts like switching to a temporary user).
-            .map { it?.activeUserId to it?.hasPendingAccountAddition }
+            .map {
+                it?.activeAccount to it?.hasPendingAccountAddition
+            }
             .distinctUntilChanged()
             .onEach {
                 // Switching between account states often involves some kind of animation (ex:
@@ -106,7 +108,10 @@ class MainViewModel @Inject constructor(
                 // a refresh.
                 @Suppress("MagicNumber")
                 delay(500)
-                trySendAction(MainAction.Internal.CurrentUserStateChange)
+                val isLoggedIn = it.first?.isLoggedIn == true
+                trySendAction(
+                    MainAction.Internal.CurrentUserStateChange(isLoggedIn = isLoggedIn),
+                )
             }
             .launchIn(viewModelScope)
 
@@ -134,7 +139,7 @@ class MainViewModel @Inject constructor(
                 handleAutofillSelectionReceive(action)
             }
 
-            is MainAction.Internal.CurrentUserStateChange -> handleCurrentUserStateChange()
+            is MainAction.Internal.CurrentUserStateChange -> handleCurrentUserStateChange(action)
             is MainAction.Internal.ScreenCaptureUpdate -> handleScreenCaptureUpdate(action)
             is MainAction.Internal.ThemeUpdate -> handleAppThemeUpdated(action)
             is MainAction.Internal.VaultUnlockStateChange -> handleVaultUnlockStateChange()
@@ -149,8 +154,11 @@ class MainViewModel @Inject constructor(
         sendEvent(MainEvent.CompleteAutofill(cipherView = action.cipherView))
     }
 
-    private fun handleCurrentUserStateChange() {
+    private fun handleCurrentUserStateChange(action: MainAction.Internal.CurrentUserStateChange) {
         recreateUiAndGarbageCollect()
+        if (action.isLoggedIn) {
+            specialCircumstanceManager.clearSpecialCircumstanceAfterLogin()
+        }
     }
 
     private fun handleScreenCaptureUpdate(action: MainAction.Internal.ScreenCaptureUpdate) {
@@ -210,7 +218,7 @@ class MainViewModel @Inject constructor(
                     authRepository.hasPendingAccountAddition = true
                 }
                 specialCircumstanceManager.specialCircumstance =
-                    SpecialCircumstance.CompleteRegistration(
+                    SpecialCircumstance.PreLogin.CompleteRegistration(
                         completeRegistrationData = completeRegistrationData,
                         timestamp = clock.millis(),
                     )
@@ -329,7 +337,7 @@ sealed class MainAction {
         /**
          * Indicates a relevant change in the current user state.
          */
-        data object CurrentUserStateChange : Internal()
+        data class CurrentUserStateChange(val isLoggedIn: Boolean) : Internal()
 
         /**
          * Indicates that the screen capture state has changed.
--- a/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
@@ -98,9 +98,7 @@ class MainViewModel @Inject constructor(
             .drop(count = 1)
             // Trigger an action whenever the current user changes or we go into/out of a pending
             // account state (which acts like switching to a temporary user).
-            .map {
-                it?.activeAccount to it?.hasPendingAccountAddition
-            }
+            .map { it?.activeUserId to it?.hasPendingAccountAddition }
             .distinctUntilChanged()
             .onEach {
                 // Switching between account states often involves some kind of animation (ex:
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -98,15 +98,20 @@ class MainViewModel @Inject constructor(
             .drop(count = 1)
             // Trigger an action whenever the current user changes or we go into/out of a pending
             // account state (which acts like switching to a temporary user).
-            .map { it?.activeUserId to it?.hasPendingAccountAddition }
+            .map {
+                it?.activeAccount to it?.hasPendingAccountAddition
+            }
```
david-livefront, line 103: Can we leave this out?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/model/SpecialCircumstance.kt
```diff
@@ -97,4 +89,23 @@ sealed class SpecialCircumstance : Parcelable {
...
     data object VaultShortcut : SpecialCircumstance()
+
+    /**
+     * A subset of [SpecialCircumstance] that are only relevant in a pre-login state and should be
+     * cleared after a successful login.
+     *
+     * @see [SpecialCircumstanceManager.clearSpecialCircumstanceAfterLogin]
+     */
+    @Parcelize
+    sealed class PreLogin : SpecialCircumstance() {
```
david-livefront, line 100: So this totally works, I am curious what you think of this method vs the abstract boolean approach?

Just adding `abstract val shouldClearOnLogin: Boolean` to the class.

app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -149,8 +154,11 @@ class MainViewModel @Inject constructor(
         sendEvent(MainEvent.CompleteAutofill(cipherView = action.cipherView))
     }
 
-    private fun handleCurrentUserStateChange() {
+    private fun handleCurrentUserStateChange(action: MainAction.Internal.CurrentUserStateChange) {
         recreateUiAndGarbageCollect()
+        if (action.isLoggedIn) {
+            specialCircumstanceManager.clearSpecialCircumstanceAfterLogin()
```
david-livefront, line 160: Would it make more sense for the `SpecailCircumstanceManager to handle this itself:

```kotlin
class SpecialCircumstanceManagerImpl 
   ...
   
    init {
        authDiskSource
            .userAccountTokensFlow
            .filter { userAccountTokensList ->
                userAccountTokensList.any {
                    it.userId == authDiskSource.userState?.activeUserId && it.isLoggedIn
                }
            }
            .onEach {
                if (blah) {
                    specialCircumstance = null
                }
            }
            .launchIn()
    }
```

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationScreen.kt
```diff
@@ -180,6 +181,11 @@ fun CompleteRegistrationScreen(
             Spacer(modifier = Modifier.navigationBarsPadding())
         }
     }
+
+    // route OS back actions through the VM to clear the special circumstance
+    BackHandler {
+        handler.onBackClick()
```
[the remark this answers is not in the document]
dseverns-livefront: > You can also shorthand it to BackHandler(onBack = handler.onBackClick)

Style-wise I do think thats the way to go, but I find that way to be the _longhand_ way haha

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/model/SpecialCircumstance.kt
dseverns-livefront: 6 in one for the current use case, I do like the idea of having these types of in their own set/category, but that just works better for how I like to organize. I can appreciate the preference to avoid reliance on inheritance/typing stuff so I won't be hurt if you feel strongly going that route. LMK 👍

app/src/main/java/com/x8bit/bitwarden/MainViewModel.kt
dseverns-livefront: I really like this!

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/model/SpecialCircumstance.kt
dseverns-livefront: So this causes an issue with the Parcelable impl on the data objects, they would need be made data classes to be able to parcelize the `shouldClearOnLogin: Boolean`.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginViewModel.kt
```diff
@@ -35,10 +35,10 @@ private const val KEY_STATE = "state"
  */
 @HiltViewModel
 class LoginViewModel @Inject constructor(
-    private val authRepository: AuthRepository,
```
dseverns-livefront, line 38: just re-organized this to put the args above the vals

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/model/SpecialCircumstance.kt
```diff
@@ -97,4 +89,23 @@ sealed class SpecialCircumstance : Parcelable {
...
+        /**
+         * The app was launched via AppLink in order to allow the user complete an ongoing
+         * registration.
+         */
+        @Parcelize
+        data class CompleteRegistration(
```
caleb-livefront, line 106: Left this question on Slack but will also put it here for future reference: 

Can you explain the necessity of moving `CompleteRegistration` into the `PreLogin` subclass? Is it just for clarity? I am confused why `CompleteRegistration` wouldn't just be renamed to reflect the pre-login condition since it seems like it's only ever used for that, and there are no other permutations of `PreLogin` (*yet?)

caleb-livefront: Also, none of the others are contained within subclasses so all else being equal it feels like we'd prefer it to be at that same top level.

dseverns-livefront: Totally valid question see the [thread](https://github.com/bitwarden/android/pull/3831#discussion_r1731716428) above, also this is the code where I am checking for that status, https://github.com/bitwarden/android/pull/3831/files#diff-4bf84d954e5067e6026bf040a2c54f6c805d1dca3fa36da8f721168027b6e1aaR27-R32

There will be at least on other such case that will be like this. Besides the issue with parcellization and changing the signature of the existing `data object` events I opted this way as it only changes the relevant objects. We could just go with a like an `interface PreLoginEvent` and slap that on any that are applicable but I was kind of following the pattern of the `Internal` sub-sealed classes in some of the VM.
 
Happy to chat on slack/huddle to give more explanation. Also open to other ideas!

caleb-livefront: Well if there is definitely going to be at least one other case that is enough for me. Saw that thread but didn't realize its exact implications so that sounds good!

caleb-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/model/SpecialCircumstance.kt
david-livefront: While I am fine with this implementation, you do not need to do that. You just have to make it a getter.

```kotlin
object class Blah : SpecialCircumstance() {
    override val shouldClearOnLogin: Boolean get() = true
}
```

PR #3841, opened 2024-08-28
[PM-10902] Base64 encode sensitive 2FA nav args

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10902

## 📔 Objective

Base 64 encode password when passing it to the 2FA screen to prevent corruption when it includes characters that should be escaped for URL safety.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt
@@ -94,7 +94,7 @@ fun LoginScreen(
             }
 
             is LoginEvent.NavigateToTwoFactorLogin -> {
-                onNavigateToTwoFactorLogin(event.emailAddress, event.password)
+                onNavigateToTwoFactorLogin(event.emailAddress, event.base64EncodedPassword)
             }
 
             is LoginEvent.ShowToast -> {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt
```diff
@@ -94,7 +94,7 @@ fun LoginScreen(
             }
 
             is LoginEvent.NavigateToTwoFactorLogin -> {
-                onNavigateToTwoFactorLogin(event.emailAddress, event.password)
+                onNavigateToTwoFactorLogin(event.emailAddress, event.base64EncodedPassword)
```
dseverns-livefront, line 97: 💭 realize this is grandfathered in as not tested, but should we add a test now where in the lambda we can call something like 
`LoginScreen( onNavigateToTwoFactorLogin = { email, encodedString -> 
           lambdaCalled = true
           assertTrue(encodedString.isEncoded)
  },
)
```
probably overkill since I think we would need to write that check. but just a thought

**Review by dseverns-livefront — APPROVED** [review]
Just one thought, but I suppose it shouldn't be blocking!

File: app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt (hunk 1 above)
Comment (SaintPatrck) [reply app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt:97]: Yea, I didn't realize this was untested. I'll circle back around and make sure these changes are covered.

Comment (dseverns-livefront) [reply app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/login/LoginScreen.kt:97]:
```
this could work for the check

fun isBase64Encoded(str: String): Boolean {
    return try {
        Base64.getDecoder().decode(str)
        true
    } catch (e: IllegalArgumentException) {
        false
    }
}

dseverns-livefront (approved):
Siiiiick, thanks for adding the tests!

PR #3842, opened 2024-08-28
PM-11387 on new create account with email verification, attempt login…

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-11387

## 📔 Objective
- with the new create account through email verification we want to attempt logging in after the account is successfully created.
- If the login attempt is successful state based navigation takes over, currently takes user to the vault, in the future new onboarding flow.
- If the account is created, but the login attempt fails the user will land on the login with the MP page.
- Shows a toast once the account is created to provide feedback to the user.
- Went with approach of going through the ViewModel, advantage is that its already behind the feature flag and keeps the existing logic unchanged in the repository as well as allows for separate handling of the Register and Login results.

## 📸 Screenshots
| Login Succeeds | Login Fails |
|--------|--------|
| https://github.com/user-attachments/assets/cb60a8bf-5e10-4152-b33a-419b81064ed8 | https://github.com/user-attachments/assets/f177e5fa-a793-4859-8047-51288f482298 | 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
@@ -129,11 +129,11 @@ class CompleteRegistrationViewModel @Inject constructor(
 
     private fun handleInternalAction(action: Internal) {
         when (action) {
-            is Internal.AttemptLogin -> handleLoginAttempt(action)
             is Internal.GeneratedPasswordResult -> handleGeneratedPasswordResult(action)
             is ReceivePasswordStrengthResult -> handlePasswordStrengthResult(action)
             is Internal.ReceiveRegisterResult -> handleReceiveRegisterAccountResult(action)
             is Internal.UpdateOnboardingFeatureState -> handleUpdateOnboardingFeatureState(action)
+            is Internal.ReceiveLoginResult -> handleLoginResult(action)
         }
     }
 
@@ -218,11 +218,19 @@ class CompleteRegistrationViewModel @Inject constructor(
             }
 
             is RegisterResult.Success -> {
-                trySendAction(
-                    action = Internal.AttemptLogin(
+                viewModelScope.launch {
+                    val loginResult = authRepository.login(
+                        email = state.userEmail,
+                        password = state.passwordInput,
                         captchaToken = registerAccountResult.captchaToken,
-                    ),
-                )
+                    )
+                    sendAction(
+                        Internal.ReceiveLoginResult(
+                            loginResult = loginResult,
+                            captchaToken = registerAccountResult.captchaToken,
+                        ),
+                    )
+                }
             }
 
             RegisterResult.DataBreachFound -> {
@@ -260,29 +268,22 @@ class CompleteRegistrationViewModel @Inject constructor(
         }
     }
 
-    private fun handleLoginAttempt(action: Internal.AttemptLogin) {
-        viewModelScope.launch {
-            val result = authRepository.login(
-                email = state.userEmail,
-                password = state.passwordInput,
-                captchaToken = action.captchaToken,
-            )
-            handleDialogDismiss()
+    private fun handleLoginResult(action: Internal.ReceiveLoginResult) {
+        handleDialogDismiss()
+        sendEvent(
+            CompleteRegistrationEvent.ShowToast(
+                message = R.string.account_created_success.asText(),
+            ),
+        )
+        // If the login result is Success the state based navigation will take care of it.
+        // otherwise we need to navigate to the login screen.
+        if (action.loginResult !is LoginResult.Success) {
             sendEvent(
-                CompleteRegistrationEvent.ShowToast(
-                    message = R.string.account_created_success.asText(),
+                CompleteRegistrationEvent.NavigateToLogin(
+                    email = state.userEmail,
+                    captchaToken = action.captchaToken,
                 ),
             )
-            // If the login result is Success the state based navigation will take care of it.
-            // otherwise we need to navigate to the login screen.
-            if (result !is LoginResult.Success) {
-                sendEvent(
-                    CompleteRegistrationEvent.NavigateToLogin(
-                        email = state.userEmail,
-                        captchaToken = action.captchaToken,
-                    ),
-                )
-            }
         }
     }
 
@@ -630,6 +631,9 @@ sealed class CompleteRegistrationAction {
          *
          * @see [AuthRepository.login]
          */
-        data class AttemptLogin(val captchaToken: String?) : Internal()
+        data class ReceiveLoginResult(
+            val loginResult: LoginResult,
+            val captchaToken: String?,
+        ) : Internal()
     }
 }
--- a/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
@@ -269,7 +269,7 @@ class CompleteRegistrationViewModel @Inject constructor(
     }
 
     private fun handleLoginResult(action: Internal.ReceiveLoginResult) {
-        handleDialogDismiss()
+        clearDialogState()
         sendEvent(
             CompleteRegistrationEvent.ShowToast(
                 message = R.string.account_created_success.asText(),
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
```diff
@@ -259,6 +257,36 @@ class CompleteRegistrationViewModel @Inject constructor(
...
     }
 
+    private fun handleLoginAttempt(action: Internal.AttemptLogin) {
+        viewModelScope.launch {
+            val result = authRepository.login(
+                email = state.userEmail,
+                password = state.passwordInput,
+                captchaToken = action.captchaToken,
+            )
+            sendAction(Internal.ReceiveLoginResult(result, action.captchaToken))
```
caleb-livefront, line 267: Idk how much you guys are into naming params on this project, and these are pretty self-explanatory, but option to do so if you're into it

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
```diff
@@ -595,5 +623,18 @@ sealed class CompleteRegistrationAction {
          * Indicates a generated password has been received.
          */
         data class GeneratedPasswordResult(val generatedPassword: String) : Internal()
+
+        /**
+         * Indicates registration was successful and will now attempt to login and unlock the vault.
+         */
+        data class AttemptLogin(val captchaToken: String?) : Internal()
```
caleb-livefront, line 630: Do you think it's worth documenting here and/or in `ReceiveLoginResult` what happens when the `captchaToken` is null? It is not intuitive to me as someone who didn't work in the auth flow.

app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModelTest.kt
```diff
@@ -58,6 +59,11 @@ class CompleteRegistrationViewModelTest : BaseViewModelTest() {
     private val mutableUserStateFlow = MutableStateFlow<UserState?>(null)
     private val mockAuthRepository = mockk<AuthRepository>() {
         every { userStateFlow } returns mutableUserStateFlow
+        coEvery { login(any(), any(), any()) } returns LoginResult.Success
+
+        coEvery {
+            register(any(), any(), any(), any(), any(), any(), any())
```
caleb-livefront, line 65: 😵‍💫

app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModelTest.kt
```diff
@@ -58,6 +59,11 @@ class CompleteRegistrationViewModelTest : BaseViewModelTest() {
     private val mutableUserStateFlow = MutableStateFlow<UserState?>(null)
     private val mockAuthRepository = mockk<AuthRepository>() {
         every { userStateFlow } returns mutableUserStateFlow
+        coEvery { login(any(), any(), any()) } returns LoginResult.Success
```
caleb-livefront, line 62: I guess this is cool as long as the `verify` calls have specific params. I am surprised to not see a `confirmVerified` in here though, which would make me a little more comfortable with the any frenzy. Do you have a general philosophy on when it's cool to `any` all the things? All else being equal the specificity feels more robust to me but that would probably be a nightmare / not at all worth it for this class.

caleb-livefront: Some projects are also far more particular about verifying calls, and I seem to recall Bitwarden not being one of them, so I'm probably thinking too hard about this.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
```diff
@@ -215,13 +219,7 @@ class CompleteRegistrationViewModel @Inject constructor(
...
 
             is RegisterResult.Success -> {
-                mutableStateFlow.update { it.copy(dialog = null) }
-                sendEvent(
-                    CompleteRegistrationEvent.NavigateToLogin(
-                        email = state.userEmail,
-                        captchaToken = registerAccountResult.captchaToken,
-                    ),
-                )
+                trySendAction(Internal.AttemptLogin(registerAccountResult.captchaToken))
```
brian-livefront, line 222: Why exactly do we need an internal action triggered here? Why not just run the logic that happens in `handleLoginAttempt` here?

app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModelTest.kt
dseverns-livefront: so personally I feel like being generic up here is fine since I am setting them globally and then being specific where I am actually testing/calling verify. I will agree the amount of args here are scary looking. 🧟

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
dseverns-livefront: happy to un-Matryoshka doll this, I was trying to follow the pattern of the other "Receive*Result"

caleb-livefront (dismissed):
Looks great!

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
```diff
@@ -215,10 +218,8 @@ class CompleteRegistrationViewModel @Inject constructor(
             }
 
             is RegisterResult.Success -> {
-                mutableStateFlow.update { it.copy(dialog = null) }
-                sendEvent(
-                    CompleteRegistrationEvent.NavigateToLogin(
-                        email = state.userEmail,
+                trySendAction(
+                    action = Internal.AttemptLogin(
```
david-livefront, line 222: Why doe we need this event?

Can we just launch the scope here and then send the result event?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
dseverns-livefront: actually I guess this would maybe be a reason: https://github.com/bitwarden/android/pull/3842#discussion_r1736454964

cc: @david-livefront can you take a look at the previous commit and lmk if that would be better

david-livefront: I agree with Brian here, the pattern is usually to launch a scope to do the work and send an internal action to process the result (which usually updates the state).

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
dseverns-livefront: ok so basically opposite of the change I made from the first commit, same amount of chaining just different order, attemp the login here, send result through action channel so state is changes on that scope and bob's your uncle?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
dseverns-livefront: wow I need to work on my reading comprehension, good show lads, ty for the patience.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/completeregistration/CompleteRegistrationViewModel.kt
```diff
@@ -259,6 +268,25 @@ class CompleteRegistrationViewModel @Inject constructor(
         }
     }
 
+    private fun handleLoginResult(action: Internal.ReceiveLoginResult) {
+        handleDialogDismiss()
```
david-livefront, line 272: Could we make a helper function that both `handleLoginResult` and `handleDialogDismiss` call.

Ideally, we would not have a `handle...` function called from anywhere but the `handleAction` or `handleInternalAction` methods.

david-livefront (approved):

PR #3843, opened 2024-08-29
PM-11214 change navoptions to support single top without retaining VM

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-11214

## 📔 Objective
- If the tab the user is already on is clicked again, popBack to the top of the graph, this unblocks interaction with the sub screens as mentioned in the defect. 

**NOTE** Key difference is on the settings tab where tapping settings again will take you back to the start of the graph**
I do have another fix locked and loaded where the VM keeps hold of a `SelectedBottomTab` enum and no interaction with the navController takes place if its the currently selected one, this matches what is seen today in the settings screen where the user would just stay on the submenu view.

## 📸 Screenshots

https://github.com/user-attachments/assets/cecfe155-b13c-4f32-bf88-bbe938c9d74c

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
@@ -31,6 +31,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavController
 import androidx.navigation.NavDestination.Companion.hierarchy
 import androidx.navigation.NavGraph
@@ -267,12 +268,11 @@ private fun VaultBottomAppBar(
             VaultUnlockedNavBarTab.Generator,
             VaultUnlockedNavBarTab.Settings,
         )
+        // Collecting the back stack entry here as state is crucial to ensuring that the items
+        // below recompose when the navigation state changes to update the selected tab.
         val navBackStackEntry by navController.currentBackStackEntryAsState()
-        val currentDestination = navBackStackEntry?.destination
         destinations.forEach { destination ->
-            val isSelected = currentDestination?.hierarchy?.any {
-                it.route == destination.route
-            } == true
+            val isSelected = navBackStackEntry.isCurrentTab(destination)
 
             NavigationBarItem(
                 icon = {
@@ -327,7 +327,7 @@ private fun VaultBottomAppBar(
 private fun NavController.vaultUnlockedNavBarScreenNavOptions(
     tabToNavigateTo: VaultUnlockedNavBarTab,
 ): NavOptions {
-    val returnToCurrentSubRoot = isCurrentTab(tabToNavigateTo)
+    val returnToCurrentSubRoot = currentBackStackEntry.isCurrentTab(tabToNavigateTo)
     val startDestination = graph.findStartDestination().id
     val currentSubRootGraph = currentDestination?.parent?.id
     val popUpToDestination = graph
@@ -346,8 +346,8 @@ private fun NavController.vaultUnlockedNavBarScreenNavOptions(
 /**
  * Determine if the current destination is the same as the given tab.
  */
-private fun NavController.isCurrentTab(tab: VaultUnlockedNavBarTab): Boolean =
-    currentDestination?.hierarchy?.any {
+private fun NavBackStackEntry?.isCurrentTab(tab: VaultUnlockedNavBarTab): Boolean =
+    this?.destination?.hierarchy?.any {
         it.route == tab.route
     } == true
 
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
@@ -327,18 +327,25 @@ private fun NavController.vaultUnlockedNavBarScreenNavOptions(
     tabToNavigateTo: VaultUnlockedNavBarTab,
 ): NavOptions {
     val returnToCurrentSubRoot = currentBackStackEntry.isCurrentTab(tabToNavigateTo)
-    val startDestination = graph.findStartDestination().id
     val currentSubRootGraph = currentDestination?.parent?.id
+    // determine the destination to navigate to, if we are navigating to the same sub-root for the
+    // selected tab we want to find the start destination of the sub-root and pop up to it, which
+    // will maintain its state (i.e. scroll position). If we are navigating to a different sub-root,
+    // we can safely pop up to the start of the graph, the "home" tab destination.
     val popUpToDestination = graph
         .getSubgraphStartDestinationOrNull(currentSubRootGraph)
         .takeIf { returnToCurrentSubRoot }
-        ?: startDestination
+        ?: graph.findStartDestination().id
+    // If we are popping up the start of the whole nav graph we want to maintain the state of the
+    // the popped destinations in the other sub-roots. If we are navigating to the same sub-root,
+    // we want to pop off the nested destinations without maintaining their state.
+    val maintainStateOfPoppedDestinations = !returnToCurrentSubRoot
     return navOptions {
         popUpTo(popUpToDestination) {
-            saveState = !returnToCurrentSubRoot
+            saveState = maintainStateOfPoppedDestinations
         }
         launchSingleTop = true
-        restoreState = !returnToCurrentSubRoot
+        restoreState = maintainStateOfPoppedDestinations
     }
 }
 
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
@@ -1,6 +1,5 @@
 package com.x8bit.bitwarden.ui.platform.feature.vaultunlockednavbar
 
-import android.os.Parcelable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.consumeWindowInsets
@@ -42,7 +41,6 @@ import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.navOptions
-import com.x8bit.bitwarden.R
 import com.x8bit.bitwarden.ui.platform.base.util.EventsEffect
 import com.x8bit.bitwarden.ui.platform.base.util.max
 import com.x8bit.bitwarden.ui.platform.base.util.toDp
@@ -50,21 +48,18 @@ import com.x8bit.bitwarden.ui.platform.components.scaffold.BitwardenScaffold
 import com.x8bit.bitwarden.ui.platform.components.scrim.BitwardenAnimatedScrim
 import com.x8bit.bitwarden.ui.platform.components.util.rememberVectorPainter
 import com.x8bit.bitwarden.ui.platform.feature.search.model.SearchType
-import com.x8bit.bitwarden.ui.platform.feature.settings.SETTINGS_GRAPH_ROUTE
 import com.x8bit.bitwarden.ui.platform.feature.settings.navigateToSettingsGraph
 import com.x8bit.bitwarden.ui.platform.feature.settings.settingsGraph
+import com.x8bit.bitwarden.ui.platform.feature.vaultunlockednavbar.model.VaultUnlockedNavBarTab
 import com.x8bit.bitwarden.ui.platform.theme.RootTransitionProviders
-import com.x8bit.bitwarden.ui.tools.feature.generator.GENERATOR_GRAPH_ROUTE
 import com.x8bit.bitwarden.ui.tools.feature.generator.generatorGraph
 import com.x8bit.bitwarden.ui.tools.feature.generator.navigateToGeneratorGraph
-import com.x8bit.bitwarden.ui.tools.feature.send.SEND_GRAPH_ROUTE
 import com.x8bit.bitwarden.ui.tools.feature.send.navigateToSendGraph
 import com.x8bit.bitwarden.ui.tools.feature.send.sendGraph
 import com.x8bit.bitwarden.ui.vault.feature.vault.VAULT_GRAPH_ROUTE
 import com.x8bit.bitwarden.ui.vault.feature.vault.navigateToVaultGraph
 import com.x8bit.bitwarden.ui.vault.feature.vault.vaultGraph
 import com.x8bit.bitwarden.ui.vault.model.VaultItemCipherType
-import kotlinx.parcelize.Parcelize
 
 /**
  * Top level composable for the Vault Unlocked Screen.
@@ -91,7 +86,8 @@ fun VaultUnlockedNavBarScreen(
 
     EventsEffect(viewModel = viewModel) { event ->
         navController.apply {
-            val navOptions = vaultUnlockedNavBarScreenNavOptions(event.returnToSubgraphRoot)
+            val navOptions =
+                vaultUnlockedNavBarScreenNavOptions(tabToNavigateTo = event.tab)
             when (event) {
                 is VaultUnlockedNavBarEvent.NavigateToVaultScreen -> {
                     navigateToVaultGraph(navOptions)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -424,9 +424,6 @@ private sealed class VaultUnlockedNavBarTab : Parcelable {
  */
 private fun NavController.vaultUnlockedNavBarScreenNavOptions(): NavOptions =
     navOptions {
-        popUpTo(graph.findStartDestination().id) {
-            saveState = true
-        }
+        popUpTo(graph.findStartDestination().id)
         launchSingleTop = true
-        restoreState = true
```
brian-livefront, line 431: So why do we need to update this stuff? My worry is that this won't maintain state across tabs, which I'm prettty sure is something we want.

app/src/test/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreenTest.kt
```diff
@@ -27,10 +27,8 @@ class VaultUnlockedNavBarScreenTest : BaseComposeTest() {
         // When changing root navigation state, pop everything else off the back stack:
         popUpTo(fakeNavHostController.graphId) {
             inclusive = false
-            saveState = true
         }
         launchSingleTop = true
-        restoreState = true
```
brian-livefront, line 33: This concerns me a bit as well. I don't think we want to make this change.

If the problem is with clicking the same bottom nav icon multiple times, can't we just block _that_ from happening or make the nav options different for just those cases?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
david-livefront: I took a quick stab at this and while I do find it odd that this problem exists at all, I did find a way around it without modifying the nav args. We can make the navigation state based:

```kotlin
data class VaultUnlockedNavBarState(
    @StringRes val vaultNavBarLabelRes: Int,
    @StringRes val vaultNavBarContentDescriptionRes: Int,
    val tabState: TabState,
) {
    sealed class TabState {
        data object Vault : TabState()

        data object Send : TabState()

        data object Generator : TabState()

        data object Settings : TabState()
    }
}
```

```kotlin
    LaunchedEffect(state.tabState) {
        navController.run {
            val navOptions = vaultUnlockedNavBarScreenNavOptions()
            when (state.tabState) {
                VaultUnlockedNavBarState.TabState.Vault -> navigateToVaultGraph(navOptions)
                VaultUnlockedNavBarState.TabState.Send -> navigateToSendGraph(navOptions)
                VaultUnlockedNavBarState.TabState.Generator -> navigateToGeneratorGraph(navOptions)
                VaultUnlockedNavBarState.TabState.Settings -> navigateToSettingsGraph(navOptions)
            }
        }
    }
```

This seemed to work really well.

brian-livefront: If we go a state-based route we have to be very careful and make sure there isn't a mismatch in the state when restoring after process death.

dseverns-livefront: > I do have another fix locked and loaded where the VM keeps hold of a SelectedBottomTab enum and no interaction with the navController takes place if its the currently selected one, this matches what is seen today in the settings screen where the user would just stay on the submenu view. 

Would adding the VM's state to SavedStateHandle take care of that @brian-livefront

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -117,15 +130,36 @@ class VaultUnlockedNavBarViewModel @Inject constructor(
...
+    private fun sendNavigationEvent(tab: SelectedBottomTab) {
+        // If the tab is already selected, do nothing.
+        if (stateFlow.value.currentTab == tab) return
+        val navEvent = when (tab) {
+            SelectedBottomTab.Vault -> VaultUnlockedNavBarEvent.NavigateToVaultScreen
+            SelectedBottomTab.Send -> VaultUnlockedNavBarEvent.NavigateToSendScreen
+            SelectedBottomTab.Generator -> VaultUnlockedNavBarEvent.NavigateToGeneratorScreen
+            SelectedBottomTab.Settings -> VaultUnlockedNavBarEvent.NavigateToSettingsScreen
+        }
+        sendEvent(navEvent)
```
david-livefront, line 146: We do not need these events anymore.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -72,28 +85,28 @@ class VaultUnlockedNavBarViewModel @Inject constructor(
...
-        sendEvent(VaultUnlockedNavBarEvent.NavigateToVaultScreen)
+        sendNavigationEvent(SelectedBottomTab.Vault)
     }
 
     /**
      * Attempts to send [VaultUnlockedNavBarEvent.NavigateToSettingsScreen] event
      */
     private fun handleSettingsTabClicked() {
-        sendEvent(VaultUnlockedNavBarEvent.NavigateToSettingsScreen)
+        sendNavigationEvent(SelectedBottomTab.Settings)
```
david-livefront, line 109: Can this just be one line to update the state:

```kotlin
  mutableStateFlow.update { it.copy(currentTab = SelectedBottomTab.Settings) }
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -12,20 +14,26 @@ import dagger.hilt.android.lifecycle.HiltViewModel
...
 ) : BaseViewModel<VaultUnlockedNavBarState, VaultUnlockedNavBarEvent, VaultUnlockedNavBarAction>(
-    initialState = VaultUnlockedNavBarState(
-        vaultNavBarLabelRes = R.string.my_vault,
-        vaultNavBarContentDescriptionRes = R.string.my_vault,
-    ),
+    initialState = savedStateHandle[KEY_STATE]
+        ?: VaultUnlockedNavBarState(
+            vaultNavBarLabelRes = R.string.my_vault,
+            vaultNavBarContentDescriptionRes = R.string.my_vault,
+            currentTab = SelectedBottomTab.Vault,
```
david-livefront, line 35: We can simplify this from the `init` block:

```kotlin
tabState = when (specialCircumstancesManager.specialCircumstance) {
    SpecialCircumstance.GeneratorShortcut -> {
        specialCircumstancesManager.specialCircumstance = null
        SelectedBottomTab.Generator
    }

    SpecialCircumstance.VaultShortcut -> {
        specialCircumstancesManager.specialCircumstance = null
        SelectedBottomTab.Vault
    }

    else -> SelectedBottomTab.Vault
},
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -188,3 +222,14 @@ sealed class VaultUnlockedNavBarEvent {
...
+
+/**
+ * Enum to represent the current bottom tab that is selected.
+ */
+@Parcelize
+enum class SelectedBottomTab : Parcelable {
+    Vault,
+    Send,
+    Generator,
+    Settings,
```
david-livefront, line 234: Shouldn't these be all caps?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
david-livefront: I don't think that is what he is talking about, I think he is concerned about the saved state in the `NavController`.

He wrote [the code](https://github.com/bitwarden/android/blob/2a057bb1fb2f368dd43042bc7f2179668893d9d6/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavScreen.kt#L120) in the `RootNavScreen` to handle something similar.

david-livefront: Brian did mention another way to we could manage this in the Screen, where we check to see if the newly selected destination is already the current graph and ignore the event in that case. That would require us to Expose route information though.

david-livefront: Edit: I may have misunderstood him, I think he just wants to add this check in the `onClick` of the Navigation Item:


```kotlin
                onClick = {
                    if (isSelected) return@onClick <--- Here
                    when (destination) {
                        is VaultUnlockedNavBarTab.Vault -> vaultTabClickedAction()
                        VaultUnlockedNavBarTab.Send -> sendTabClickedAction()
                        VaultUnlockedNavBarTab.Generator -> generatorTabClickedAction()
                        VaultUnlockedNavBarTab.Settings -> settingsTabClickedAction()
                    }
                },
```

david-livefront: I did some additional testing and found that the root cause of the problem seems to be that the Lifecycle state of the screen is `DESTROYED` when the tab gets clicked multiple times in a row. I don't entirely understand why that is, it may be a bug in Android Lifecycle. Adding the `isSelected` check may be the easiest way to work around it, though state based navigation would work as well but incurs a lot more effort.

dseverns-livefront:
>Brian did mention another way to we could manage this in the Screen, where we check to see if the newly selected destination is already the current graph and ignore the event in that case. That would require us to Expose route information though.

So I think the only other thing is maybe need a special case for when graph is settings but, destination is not the main menu for giving this behavior noted in the ticket as expected:

>Normally, I would expect tapping Settings to return the user to the Settings main menu. Additionally, I would expect that the back arrow is always responsive.

david-livefront:
> Normally, I would expect tapping Settings to return the user to the Settings main menu. Additionally, I would expect that the back arrow is always responsive.

That is not how a BottomNavBar works, do you think we can follow up on this?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -168,23 +210,42 @@ sealed class VaultUnlockedNavBarAction {
...
+
+/**
+ * Enum to represent the current bottom tab that is selected.
+ */
+@Parcelize
+enum class BottomNavDestination : Parcelable {
+    VAULT,
+    SEND,
+    GENERATOR,
+    SETTINGS,
```
brian-livefront, line 250: Do we still need this? I thought this would all be covered by just the boolean being passed around.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -168,23 +210,42 @@ sealed class VaultUnlockedNavBarAction {
  * Models events for the bottom tab of the vault unlocked portion of the app.
  */
 sealed class VaultUnlockedNavBarEvent {
+    /**
+     * Whether to return to the root of the subgraph or not.
+     */
+    abstract val returnToSubgraphRoot: Boolean
```
brian-livefront, line 216: So I feel like this applies to all the navigation events but wouldn't technically apply if we ever had to add a different kind of event. This is probably fine for now but it's something to be aware of.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -117,15 +130,44 @@ class VaultUnlockedNavBarViewModel @Inject constructor(
         }
     }
     // #endregion BottomTabViewModel Action Handlers
+
+    /**
+     * Send the navigation event for the selected tab. If the tab is already selected, do nothing.
+     */
+    private fun updateTabStateAndSendNavEvent(tab: BottomNavDestination) {
```
brian-livefront, line 137: Do we still need this?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -421,12 +422,38 @@ private sealed class VaultUnlockedNavBarTab : Parcelable {
...
+        restoreState = !returnToCurrentSubRoot
     }
+}
+
+/**
+ * Helper function to determine the start destination of a subgraph.
+ *
+ * @param subgraphId the id of the subgraph to find the start destination of.
```
brian-livefront, line 450: `id` should generally be `ID`

app/src/test/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreenTest.kt
```diff
@@ -65,8 +65,11 @@ class VaultUnlockedNavBarScreenTest : BaseComposeTest() {
 
     @Test
     fun `NavigateToVaultScreen should navigate to VaultScreen`() {
-        composeTestRule.runOnIdle { fakeNavHostController.assertCurrentRoute("vault_graph") }
-        mutableEventFlow.tryEmit(VaultUnlockedNavBarEvent.NavigateToVaultScreen)
+        mutableEventFlow.tryEmit(VaultUnlockedNavBarEvent.NavigateToSendScreen)
+        composeTestRule.runOnIdle { fakeNavHostController.assertCurrentRoute("send_graph") }
+        mutableEventFlow.tryEmit(
+            VaultUnlockedNavBarEvent.NavigateToVaultScreen,
```
brian-livefront, line 71: We can undo all these formatting changes now right?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -90,21 +86,22 @@ fun VaultUnlockedNavBarScreen(
 
     EventsEffect(viewModel = viewModel) { event ->
         navController.apply {
-            val navOptions = vaultUnlockedNavBarScreenNavOptions()
+            val navOptions =
+                vaultUnlockedNavBarScreenNavOptions(tabToNavigateTo = event.tab)
```
david-livefront, line 90: Can we put this up a line

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -90,21 +86,22 @@ fun VaultUnlockedNavBarScreen(
...
             when (event) {
-                VaultUnlockedNavBarEvent.NavigateToVaultScreen -> {
+                is VaultUnlockedNavBarEvent.NavigateToVaultScreen -> {
                     navigateToVaultGraph(navOptions)
                 }
 
-                VaultUnlockedNavBarEvent.NavigateToSendScreen -> {
+                is VaultUnlockedNavBarEvent.NavigateToSendScreen -> {
```
david-livefront, line 96: I don't think these `is`s are needed

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -168,23 +169,37 @@ sealed class VaultUnlockedNavBarAction {
  * Models events for the bottom tab of the vault unlocked portion of the app.
  */
 sealed class VaultUnlockedNavBarEvent {
+
+    abstract val tab: VaultUnlockedNavBarTab
```
david-livefront, line 173: Wanna toss a kdoc on here

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarViewModel.kt
```diff
@@ -168,23 +169,37 @@ sealed class VaultUnlockedNavBarAction {
...
+    }
 
     /**
      * Navigate to the Vault screen.
      */
-    data object NavigateToVaultScreen : VaultUnlockedNavBarEvent()
+    data object NavigateToVaultScreen : VaultUnlockedNavBarEvent() {
+        override val tab: VaultUnlockedNavBarTab = VaultUnlockedNavBarTab.Vault(
+            labelRes = R.string.my_vault,
+            contentDescriptionRes = R.string.my_vault,
```
david-livefront, line 195: So these particular values are never actually used, we just care about the type, right?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/model/VaultUnlockedNavBarTab.kt
```diff
@@ -0,0 +1,107 @@
...
+ * Represents the different tabs available in the navigation bar
+ * for the unlocked portion of the vault.
+ *
+ * Each tab is modeled with properties that provide information on:
+ * - Regular icon resource
+ * - Icon resource when selected
+ * and other essential UI and navigational data.
+ *
+ * @property iconRes The resource ID for the regular (unselected) icon representing the tab.
+ * @property iconResSelected The resource ID for the icon representing the tab when it's selected.
```
david-livefront, line 21: You don't need the doc for the property is it is on the property below

dseverns-livefront: I ctrl v'd this whole file want me to change it?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -270,12 +267,11 @@ private fun VaultBottomAppBar(
             VaultUnlockedNavBarTab.Generator,
             VaultUnlockedNavBarTab.Settings,
         )
+        // Collecting the back stack entry here as state is crucial to ensuring that the items
+        // below recompose when the navigation state changes to update the selected tab.
```
brian-livefront, line 271: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -270,12 +267,11 @@ private fun VaultBottomAppBar(
...
         val navBackStackEntry by navController.currentBackStackEntryAsState()
-        val currentDestination = navBackStackEntry?.destination
         destinations.forEach { destination ->
-            val isSelected = currentDestination?.hierarchy?.any {
-                it.route == destination.route
-            } == true
+            val isSelected = navBackStackEntry.isCurrentTab(destination)
```
brian-livefront, line 274: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/VaultUnlockedNavBarScreen.kt
```diff
@@ -322,111 +318,48 @@ private fun VaultBottomAppBar(
...
+    tabToNavigateTo: VaultUnlockedNavBarTab,
+): NavOptions {
+    val returnToCurrentSubRoot = currentBackStackEntry.isCurrentTab(tabToNavigateTo)
+    val startDestination = graph.findStartDestination().id
+    val currentSubRootGraph = currentDestination?.parent?.id
+    val popUpToDestination = graph
+        .getSubgraphStartDestinationOrNull(currentSubRootGraph)
+        .takeIf { returnToCurrentSubRoot }
+        ?: startDestination
+    return navOptions {
```
brian-livefront, line 336: Is there more documentation we could add somewhere in here to explain what we're doing? Basically, to save the state of the destination we are popping and restore a previous state, _unless_ we are navigating to the same place.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/vaultunlockednavbar/model/VaultUnlockedNavBarTab.kt
david-livefront: Naw, if it's copy paste, we can leave it alone.

david-livefront (approved):
🚀

PR #3888, opened 2024-09-09
BITAU-98 Add EncryptionUtils helper functions to the bridge SDK

ahaisting-livefront:
## 🎟️ Tracking

https://livefront.atlassian.net/browse/BITAU-98

## 📔 Objective

The goal of this PR is to set up basic encryption/decryption logic in the Bridge SDK. This will be called by the PMA in future tickets when generating symmetric key, checking symmetric key, and ultimately implementing `BridgeService`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+private fun ByteArray.toSecretKey(): SecretKey =
+    SecretKeySpec(this, 0, this.size, "AES")
+
+/**
+ * Helper function for generating a [Cipher] that can be used for encrypting/decrypting using
+ * [SymmetricEncryptionKeyData].
+ */
+private fun generateCipher(): Cipher =
+    Cipher.getInstance("AES/CBC/PKCS5PADDING")
```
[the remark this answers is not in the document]
SaintPatrck: Or maybe have `toEncryptedSharedAccountData(...)` return `Result` and wrap it all in a `runCatching` block. I see a few other calls that could throw in there as well.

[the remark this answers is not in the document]
ahaisting-livefront: @SaintPatrck that's totally the plan 👍

bridge/src/main/java/com/bitwarden/bridge/model/AddTotpLoginItemDataJson.kt
```diff
@@ -10,7 +10,7 @@ import kotlinx.serialization.Serializable
  * @param totpUri A TOTP code URI to be added to the Bitwarden app.
  */
 @Serializable
-data class AddTotpLoginItemDataJson(
+internal data class AddTotpLoginItemDataJson(
```
ahaisting-livefront, line 13: I realized when working on this PR that these are just an implementation detail of the encryption. 

I love being able to use `internal` 😎

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,201 @@
+package com.bitwarden.bridge.util
```
ahaisting-livefront, line 1: S/O to @brian-livefront for originally writing all this code in his POC. I'm standing on the shoulders of a giant!

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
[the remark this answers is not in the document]
ahaisting-livefront: All the publicly available functions here are now wrapped in `Result` 👍

bridge/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 // For more info on versioning, see the README.
-val version = "0.1.0"
+val version = "0.2.0"
```
SaintPatrck, line 4: Does there need to be an entry added to `CHANGELOG` for this version change? Is artifact generation happening automatically or do we need to run assemble manually for each PR with a version change?

ahaisting-livefront: Great callout here.

1. Yep, a CHANGELOG update is required here. Thanks for the callout. This builds good SDK release habits! Added: a14b78c08dd8e3683122230354ab9e9011df77a9
2. Currently its all manual. My plan is to have a separate PR that includes the updated `.aar` in`/app` along with corresponding `/app` changes.

SaintPatrck (dismissed):

bridge/build.gradle.kts
brian-livefront: Do we need to update the version actually right now? Could we keep just using "snapshot" builds until we're ready to "release 0.1.0? I don't think we should bump the version until we know where we are releasing them.

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,201 @@
...
+    messageDigest.update(this.symmetricEncryptionKey.byteArray)
+    return@runCatching messageDigest.digest()
+}
+
+/**
+ * Encrypt [SharedAccountData].
+ *
+ * @param symmetricEncryptionKeyData Symmetric key used for encryption.
+ *
+ * This is intended to be used by the main Bitwarden app during a [IBridgeService.syncAccounts] call.
```
brian-livefront, line 52: I would think these lines might need to come before all the  @param in these

bridge/build.gradle.kts
ahaisting-livefront: Brian and I talked about this offline, and we decided to leave the version at 0.1.0 for now and keep calling every build a `SNAPSHOT` build. 

Then once we are in a stable place, we can bump the version to 1.0.0 and update the CHANGELOG accordingly.

ahaisting-livefront: 58d6f3e4716d47460e0aad390f1dc0a2d5a2bde0

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Generate a symmetric [SecretKey] that will used for encrypting IPC traffic.
+ *
+ * This is intended to be used for implementing [IBridgeService.getSymmetricEncryptionKeyData].
+ */
+fun generateSecretKey(): Result<SecretKey> = runCatching {
+    val keygen = KeyGenerator.getInstance("AES")
+    keygen.init(256, SecureRandom())
+    return@runCatching keygen.generateKey()
```
brian-livefront, line 30: Do we need these return calls here? I think we omit these for VaultSdkSourceImpl etc that also use runCatching

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
ahaisting-livefront: Ah yep, who knows why my brain used that order 😆 

4fd92d70551cfaeb135df6cf8354453eccb691b8

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
ahaisting-livefront: Yeah good question. 

I added these in 100% for my own readability sake. Looks like other usages of `runCatching` don't use them. Removed:

4a8f906f808c00074bd53db54351aa637a2346b7

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.jupiter.api.Test
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+import java.time.Instant
+import javax.crypto.Cipher
+import javax.crypto.KeyGenerator
+
+class EncryptionUtilTest {
+
+    private val TEST_SHARED_ACCOUNT_DATA = SharedAccountData(
```
brian-livefront, line 22: Do we typically append TEST to the name of sample data instances? Also, I think we tend to place these at the bottom of a test file.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    @Test
+    fun `generateSecretKey should return success`() {
+        val secretKey = generateSecretKey()
+        assert(secretKey.isSuccess)
+        assertNotNull(secretKey.getOrNull())
+    }
+
+    @Test
+    fun `when KeyGenerator getInstance throws, generateSecretKey should return failure`() {
```
brian-livefront, line 56: I believe all of these test names should first start with the name of the function being tested.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
```diff
@@ -0,0 +1,178 @@
...
+        mockkStatic(KeyGenerator::class)
+        every { KeyGenerator.getInstance("AES") } throws NoSuchAlgorithmException()
+        val secretKey = generateSecretKey()
+        assert(secretKey.isFailure)
+        unmockkStatic(KeyGenerator::class)
+    }
+
+    @Test
+    fun `toFingerprint should return success`() {
```
brian-livefront, line 65: Seems like there should be a "when" here, like "when encryption succeeds" or something

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
ahaisting-livefront: I swear I saw `TEST_` prefix elsewhere, but I must be seeing ghosts. Agreed on file location too. Updated:

0d465a81ae09aba1ae5216932b976624bfaefcbd

This also explains why the IDE was complaining to me about the name formatting of these... it is no longer complaining!

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
```diff
@@ -0,0 +1,178 @@
...
+        assert(secretKey.isSuccess)
+        assertNotNull(secretKey.getOrNull())
+    }
+
+    @Test
+    fun `when KeyGenerator getInstance throws, generateSecretKey should return failure`() {
+        mockkStatic(KeyGenerator::class)
+        every { KeyGenerator.getInstance("AES") } throws NoSuchAlgorithmException()
+        val secretKey = generateSecretKey()
+        assert(secretKey.isFailure)
```
david-livefront, line 60: Can you use the `assertTrue`

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
ahaisting-livefront: Sure thing!

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
ahaisting-livefront: True! Most were already correct, updated a few.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
ahaisting-livefront: 🫡

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,201 @@
...
+    val messageDigest = MessageDigest.getInstance("SHA-256")
+    messageDigest.reset()
+    messageDigest.update(this.symmetricEncryptionKey.byteArray)
+    messageDigest.digest()
+}
+
+/**
+ * Encrypt [SharedAccountData].
+ *
+ * This is intended to be used by the main Bitwarden app during a [IBridgeService.syncAccounts] call.
```
brian-livefront, line 50: I'm a bit torn...do we need to refer to `Bitwarden app` and `Bitwarden Authenticator app` or would referring to the `IBridgeServiceCallback` be enough? This is probably especially relevant to think through because once we add the "coroutines wrapper layer" it will actually be the _SDK_ code that calls these and not the Authenticator app directly.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
brian-livefront: I see the change in the commit you posted but in the full diff this has not changed....maybe that commit was reverted in a later one?

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
ahaisting-livefront: I like leaving this breadcrumb to try and help out any future reader/editor with more context. 

Re: the coroutines wrapping layer... I was imagining that layer only helping out with the Authenticator side of things (connecting to the service, registering callbacks). Here's the coroutines [wrapper ticket](https://livefront.atlassian.net/jira/software/c/projects/BITAU/boards/100?selectedIssue=BITAU-97), nothing there about helping out with the PMA side of things. To me, this makes sense, because in order to perform `syncAccounts`, we need the entire data layer of the main Bitwarden app, so it makes sense this logic would be in the BItwarden app not the SDK. 

Maybe I'm missing something though.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
ahaisting-livefront: I must have made an error while rebasing... thank you for catching this!!

Here's an update:

af046f9ff733328ac565300f1dca6eeae0054b41

ahaisting-livefront: Yeah I messed something up rebasing this morning for sure... here's the missing commit:

af046f9ff733328ac565300f1dca6eeae0054b41

bridge/src/main/java/com/bitwarden/bridge/model/SharedAccountDataJson.kt
```diff
@@ -13,7 +13,7 @@ import java.time.Instant
  * @param accounts The list of shared accounts.
  */
 @Serializable
-data class SharedAccountDataJson(
+internal class SharedAccountDataJson(
```
david-livefront, line 16: Can this be an internal data class?

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
brian-livefront: We spoke offline. We'll leave references to "the Bitwarden app" but remove those to "the Bitwarden Authenticator app", mainly since we're thinking we can make the `decrypt` call `internal`.

bridge/src/main/java/com/bitwarden/bridge/model/SharedAccountDataJson.kt
ahaisting-livefront: Totally :) Nice catch:

5643141517755254b869d585aa819c0fbbf29e8b

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
ahaisting-livefront: Thanks, Brian!

147c632f9269d7c0be4ac1bb62882d972beb9c57

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,195 @@
...
+import javax.crypto.spec.IvParameterSpec
+import javax.crypto.spec.SecretKeySpec
+
+/**
+ * Generate a symmetric [SecretKey] that will used for encrypting IPC traffic.
+ *
+ * This is intended to be used for implementing [IBridgeService.getSymmetricEncryptionKeyData].
+ */
+fun generateSecretKey(): Result<SecretKey> = runCatching {
+    val keygen = KeyGenerator.getInstance("AES")
```
brian-livefront, line 28: Can we look into using `KeyProperties.KEY_ALGORITHM_AES` etc. for these? When creating the Cipher we can probably use something like this from the biometrics code:

```kotlin
private const val CIPHER_TRANSFORMATION =
    KeyProperties.KEY_ALGORITHM_AES + "/" +
        KeyProperties.BLOCK_MODE_CBC + "/" +
        KeyProperties.ENCRYPTION_PADDING_PKCS7
```

ahaisting-livefront: So I was able to remove a few of the raw strings here. 

I still hit a sticky wicket with `KeyProperties.ENCRYPTION_PADDING_PKCS7` blowing up test because that algo is not available in our test environment. 

I failed to mock out that constant. To my knowledge and reading the [mockk docs](https://mockk.io/), I don't think mockk supports mocking a static constant. It can mock top level constants and extension functions, but I was't able to make it work for this static class constant.

brian-livefront: OK sounds good. Thanks for that context!

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
```diff
@@ -0,0 +1,207 @@
...
+    val decryptedModel = JSON.decodeFromString<SharedAccountDataJson>(
+        cipher.doFinal(this.encryptedAccountsJson.byteArray).decodeToString()
+    )
+    decryptedModel.toDomainModel()
+}
+
+/**
+ * Encrypt [AddTotpLoginItemData].
+ *
+ * This is intended to be used by the Bitwarden Authenticator app before requesting a new TOTP
```
brian-livefront, line 94: Still a reference to `Bitwarden Authenticator app` here.

bridge/src/test/java/com/bitwarden/bridge/util/EncryptionUtilTest.kt
```diff
@@ -0,0 +1,193 @@
...
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+import java.time.Instant
+import javax.crypto.Cipher
+import javax.crypto.KeyGenerator
+
+class EncryptionUtilTest {
+
+    @Test
+    fun `generateSecretKey should return success when there are no internal exceptions`() {
```
brian-livefront, line 25: 👍

bridge/src/main/java/com/bitwarden/bridge/util/EncryptionUtils.kt
ahaisting-livefront: 👍 missed this `encrypt`, good call:

b33e6d1ead8911babe05a4411b1f91a82f19b890

brian-livefront (approved):

ahaisting-livefront:
Thanks, @brian-livefront, @SaintPatrck, and @david-livefront !

PR #3958, opened 2024-09-24
PM-11535 Only add password to 2 factor login path if present

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-11535

## 📔 Objective
- Prevent crypto error for a SSO user with a MP due to wrongly encoded password being passed to login attempt.
- The way the route was being set  w/o specifying nullable args in the destination: was causing "null" to be decoded and instead of passing `null` to the login function it was attempting to send the resulting string which was incorrect and also was "\<?><?>e" encoded characters.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/twofactorlogin/TwoFactorLoginNavigation.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/twofactorlogin/TwoFactorLoginNavigation.kt
@@ -4,6 +4,8 @@ import androidx.lifecycle.SavedStateHandle
 import androidx.navigation.NavController
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavOptions
+import androidx.navigation.NavType
+import androidx.navigation.navArgument
 import com.x8bit.bitwarden.data.platform.annotation.OmitFromCoverage
 import com.x8bit.bitwarden.data.platform.datasource.network.util.base64UrlDecodeOrNull
 import com.x8bit.bitwarden.data.platform.datasource.network.util.base64UrlEncode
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/twofactorlogin/TwoFactorLoginNavigation.kt
```diff
@@ -34,8 +34,11 @@ fun NavController.navigateToTwoFactorLogin(
     password: String?,
     navOptions: NavOptions? = null,
 ) {
+    val requiredRoute = "$TWO_FACTOR_LOGIN_PREFIX/$emailAddress"
+    // only add the password if it is not null.
+    val routeOptions = password?.let { "?$PASSWORD=${it.base64UrlEncode()}" } ?: ""
```
david-livefront, line 39: Is there something inherently wrong with `?$PASSWORD=null`?

dseverns-livefront: the arg would then be `args[PASSWORD] == "null"` only in this case "null" gets encoded, then we are passing that string value to the "login( ..., ..., password = theMessIDescribeAbove)"

david-livefront: I wonder if we are just missing this?

```kotlin
        arguments = listOf(
            navArgument(EMAIL_ADDRESS) { type = NavType.StringType },
            navArgument(PASSWORD) {
                type = NavType.StringType
                nullable = true
            },
        ),
```

dseverns-livefront: you're 100% correct

dseverns-livefront: I am going to remove the tests, because A: we are just using the compose navigation here so no need to test it and B: it breaks the null case : {

david-livefront (approved):

PR #3987, opened 2024-09-27
BITAU-97 Add `AuthenticatorBridgeManager` 

ahaisting-livefront:
## 🎟️ Tracking

https://livefront.atlassian.net/browse/BITAU-97

## 📔 Objective

The goal here is to write a wrapper layer to make it easy for consuming applications (the Bitwarden Authenticator app)  to connect to AuthenticatorBridgeService.

## 📸 Screenshots

N/A

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/provider/AuthenticatorBridgeCallbackProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Get a [IAuthenticatorBridgeServiceCallback] that will call delegate [onAccountsSync] call
+     * to the given lambda.
+     *
+     * @param onAccountsSync Lambda that will be invoked when [onAccountsSync] calls back.
+     */
+    fun getCallback(
+        onAccountsSync: (EncryptedSharedAccountData) -> Unit,
+    ): IAuthenticatorBridgeServiceCallback
+}
```
ahaisting-livefront, line 21: TODO: newline.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/provider/SymmetricKeyStorageProvider.kt
```diff
@@ -0,0 +1,11 @@
+package com.bitwarden.authenticatorbridge.provider
+
+import com.bitwarden.authenticatorbridge.model.SymmetricEncryptionKeyData
+
+/**
+ * TODO: document me
```
david-livefront, line 6: Wanna document this

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/provider/SymmetricKeyStorageProvider.kt
```diff
@@ -0,0 +1,11 @@
...
+ */
+interface SymmetricKeyStorageProvider {
+
+    var symmetricKey: SymmetricEncryptionKeyData?
```
david-livefront, line 10: Wanna add a doc here too

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+/**
+ * Helper class for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+private class AuthenticatorBridgeServiceSafetyWrapper(
```
brian-livefront, line 186: Do we need a class for this? Can't we just use an extension function like:

```kotlin
    fun <T> IAuthenticatorBridgeService.safeCall(action: IAuthenticatorBridgeService.() -> T): T? =
        try {
            this?.let { action.invoke(it) }
        } catch (e: Exception) {
            null
        }
```

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/provider/SymmetricKeyStorageProvider.kt
ahaisting-livefront: Yes. ty :) 067fda133c5d96370a4cec544f168986e4cf765f

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/provider/SymmetricKeyStorageProvider.kt
ahaisting-livefront: 067fda133c5d96370a4cec544f168986e4cf765f

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+        mutableSharedAccountsStateFlow.value = decryptedAccounts
+            ?.let { AccountSyncState.Success(it.accounts) }
+            ?: AccountSyncState.Error
+    }
+
+    private fun unBindService() {
+        if (!isBound) return
+        isBound = false
+        bridgeService.safeCall { unregisterBridgeServiceCallback(authenticatorBridgeCallback) }
+        bridgeService = AuthenticatorBridgeServiceSafetyWrapper(service = null)
```
brian-livefront, line 175: I'd really prefer to just null this whole thing out and ditch the wrapper class. (See comment below.)

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+            val decryptedAccounts = data.decrypt(it).getOrNull()?.accounts ?: return@let null
+            AccountSyncState.Success(decryptedAccounts)
+        }
+        mutableSharedAccountsStateFlow.value = decryptedAccounts
+            ?.let { AccountSyncState.Success(it.accounts) }
+            ?: AccountSyncState.Error
+    }
+
+    private fun unBindService() {
+        if (!isBound) return
```
brian-livefront, line 172: Any reason to check this? Shouldn't we just all all this stuff no matter what?

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+        if (!isBound) {
+            mutableSharedAccountsStateFlow.value = AccountSyncState.Error
+        }
+    }
+
+    private fun onAccountsSync(data: EncryptedSharedAccountData) {
+        // Received account sync update. Decrypt with local symmetric key and update StateFlow:
+        val decryptedAccounts = symmetricKeyStorageProvider.symmetricKey?.let {
+            val decryptedAccounts = data.decrypt(it).getOrNull()?.accounts ?: return@let null
+            AccountSyncState.Success(decryptedAccounts)
```
brian-livefront, line 164: Not sure why you are wrapping and unwrapping this.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+
+        if (!isBound) {
+            mutableSharedAccountsStateFlow.value = AccountSyncState.Error
+        }
+    }
+
+    private fun onAccountsSync(data: EncryptedSharedAccountData) {
+        // Received account sync update. Decrypt with local symmetric key and update StateFlow:
+        val decryptedAccounts = symmetricKeyStorageProvider.symmetricKey?.let {
+            val decryptedAccounts = data.decrypt(it).getOrNull()?.accounts ?: return@let null
```
brian-livefront, line 163: I think I'd prefer to have a separate check to get the key and then a single final chain:

```kotlin
val symmetricKey = symmetricKeyStorageProvider.symmetricKey ?: AccountSyncState.Error
mutableSharedAccountsStateFlow.value = data
            .decrypt(it)
            .getOrNull()
            ?.accounts 
            ?.let { AccountSyncState.Success(it.accounts) }
            ?: AccountSyncState.Error
```

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+            )
+        }
+
+        isBound = try {
+            applicationContext.bindService(
+                intent,
+                bridgeServiceConnection,
+                Context.BIND_AUTO_CREATE,
+            )
+        } catch (e: SecurityException) {
```
brian-livefront, line 151: You should make sure this is the _only_ exception that can be thrown here. I don't know if there are others. So you'll want to either investigate the docs/source or play around with it.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+                override fun onStart(owner: LifecycleOwner) {
+                    bindService()
+                }
+
+                override fun onStop(owner: LifecycleOwner) {
+                    unBindService()
+                }
+            },
+        )
+    }
```
brian-livefront, line 135: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: mmmm, yes, I like that. My original intention with the class was to make it _impossible_ to call the service without the wrapper (by giving no access to the service directly), but I think we can trust future editors of this file to use this function and never call the service without `safeCall`.

Updated: c9fc5fb03ecc328e680d4dd05ae30e729248dc3f

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Callback registered with AuthenticatorBridgeService.
+     */
+    private val authenticatorBridgeCallback = callbackProvider.getCallback(::onAccountsSync)
+
+    /**
+     * Service connection that listens for connected and disconnected service events.
+     */
+    private val bridgeServiceConnection = object : ServiceConnection {
```
brian-livefront, line 75: I might go either way on this, but I do wonder a little if we should define a function or class that implements all this logic just so we don't have to have the bulk of our logic before we even get to our `init` block. Just an organizational thing...curious to hear what you think.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+                symmetricKeyStorageProvider.symmetricKey =
+                    bridgeService.safeCall { symmetricEncryptionKeyData }
+            }
+
+            if (symmetricKeyStorageProvider.symmetricKey == null) {
+                // This means bridgeService returned a null key, which means we can make
+                // no valid operations. We should disconnect form the service and expose to the
+                // calling application that authenticator sync is not enabled.
+                mutableSharedAccountsStateFlow.value = AccountSyncState.SyncNotEnabled
+                unBindService()
```
brian-livefront, line 102: Are you missing a `return` here?

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Calling `unbindService` if the service is not bound will throw an exception. Thus, the check.

We always call `unbindService` when we move to onStop. But there are other times we unbind, like if there are no accounts that have sync enabled, for example. So we have to be cautious here.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Service connection that listens for connected and disconnected service events.
+     */
+    private val bridgeServiceConnection = object : ServiceConnection {
+        override fun onServiceConnected(name: ComponentName, service: IBinder) {
+            bridgeService = IAuthenticatorBridgeService.Stub.asInterface(service)
+
+            // TODO: Add check for version mismatch between client and server SDKs:
+            // TODO: https://livefront.atlassian.net/browse/BITAU-72
```
brian-livefront, line 80: Honestly based on our designs for errors, it's possible we don't even need to worry about this right now. Just something to keep in mind.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Main AuthenticatorBridgeService access point.
+     */
+    private var bridgeService: IAuthenticatorBridgeService? = null
+
+    /**
+     * Internal state of [accountSyncStateFlow].
+     */
+    private
+    val mutableSharedAccountsStateFlow =
```
brian-livefront, line 64: Something weird going on with formatting here.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+    private val applicationContext: Context,
+    private val connectionType: AuthenticatorBridgeConnectionType,
+    private val symmetricKeyStorageProvider: SymmetricKeyStorageProvider,
+    callbackProvider: AuthenticatorBridgeCallbackProvider = StubAuthenticatorBridgeCallbackProvider(),
+    processLifecycleOwner: LifecycleOwner = ProcessLifecycleOwner.get(),
+) : AuthenticatorBridgeManager, DefaultLifecycleObserver {
+
+    /**
+     * Keep track of whether the service is bound. Useful for making sure we don't call
+     * [Context.unbindService] on a service that isn't bound.
```
brian-livefront, line 51: Would that be a problem?

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/model/AuthenticatorBridgeConnectionType.kt
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * Convert a [AuthenticatorBridgeConnectionType] to raw package name for connection.
+ */
+internal fun AuthenticatorBridgeConnectionType.toPackageName() =
+    when (this) {
+        AuthenticatorBridgeConnectionType.RELEASE -> "com.x8bit.bitwarden"
+        AuthenticatorBridgeConnectionType.DEV -> "com.x8bit.bitwarden.dev"
+    }
```
brian-livefront, line 28: 👍

brian-livefront: I'm curious what @david-livefront thinks but we don't usually mix classes and extension functions for that class in a single file, instead preferring to have a `util` package to place these in with a `....Extensions` file.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: It would throw an exception.

brian-livefront: What exception?

brian-livefront: To give more context, I'm not sure how much I trust that boolean; it makes me worry about threading issues and it getting out of sync somehow and that if we need to protect ourselves then we just protect ourselves more directly.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Yeah I think I just adjusted this too many times that it ended up a frankenstien monster 😆 

How's this: 8cedcad269fea09c211be12f3483a354920663bd

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Here's the exception that is thrown if you try to unbind a service that isn't bound: `java.lang.IllegalArgumentException: Service not registered`.

The idea for keeping track of `isBound` boolean came from the Android service docs. 

I see your point, though. There is some way this could get out of sync that I don't understand or am not thinking of. We could wrap our `Context.unbindService` call in `try`/`catch`, as a different way of protecting against this case.  

Do you have a preference either way?

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/model/AuthenticatorBridgeConnectionType.kt
ahaisting-livefront: Sure thing: a013a30d5c6c7b38c87d7d35ed3861264415f829

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Cleaned this up here: 8cedcad269fea09c211be12f3483a354920663bd

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
brian-livefront: Well I suppose the one thing to check is if `isBound` value is always getting set on the same thread (likely the main thread). If it is, then when you have is probably fine. Otherwise I think we may need to ditch it and do the try/catch. Again, for this particular feature we have to be super careful about anything that could throw an exception.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
brian-livefront: Oh yeah, even better 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
brian-livefront: OK. Yeah I guess this just goes into the discussion about what is safe and not safe to call here. So we can tie this up into that.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/model/AuthenticatorBridgeConnectionType.kt
david-livefront: Yea, an `AuthenticatorBridgeConnectionTypeExtensions` file sounds good, Looks like it's already done!

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+/**
+ * Helper function for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+fun <T> IAuthenticatorBridgeService?.safeCall(action: IAuthenticatorBridgeService.() -> T): T? =
```
brian-livefront, line 181: Can we make this return `Result<T>` instead? I feel it would be best to be able to distinguish between `null` and an error for some of the calls.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Yeah I see your point about the organization of the file being mildly annoying. I pulled the bulk of the lines into a private function here:

234e8e52db42f4817a1c32baeafa6a68b8f3bc7c

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Yeah I've been noodling on this a bit.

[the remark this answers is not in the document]
ahaisting-livefront: f9b1f1f20335521f3124cc4e2f9adc704da1e1f7

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,193 @@
...
+ * @param processLifecycleOwner Lifecycle owner that is used to listen for start/stop
+ * lifecycle events.
+ */
+class AuthenticatorBridgeManagerImpl(
+    private val applicationContext: Context,
+    private val connectionType: AuthenticatorBridgeConnectionType,
+    private val symmetricKeyStorageProvider: SymmetricKeyStorageProvider,
+    callbackProvider: AuthenticatorBridgeCallbackProvider = StubAuthenticatorBridgeCallbackProvider(),
+    processLifecycleOwner: LifecycleOwner = ProcessLifecycleOwner.get(),
+) : AuthenticatorBridgeManager, DefaultLifecycleObserver {
```
brian-livefront, line 47: We can remove `DefaultLifecycleObserver` here right?

ahaisting-livefront: Oh yeah: 9d789e6a4fbe204d334da38e06aa57a1baa911d1

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Ah, yep!

This was working before because `unBindService` prevents anything else from happening. But the early return makes it more explicit. 9f8aed67773b46ed30400943194f9f3cab0a9c81

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,196 @@
...
+ * application context.
+ * @param connectionType Specifies which build variant to connect to.
+ * @param symmetricKeyStorageProvider Provides access to local storage of the symmetric encryption
+ * key.
+ * @param callbackProvider Provides a way to construct a service callback that can be mocked in
+ * tests.
+ * @param processLifecycleOwner Lifecycle owner that is used to listen for start/stop
+ * lifecycle events.
+ */
+class AuthenticatorBridgeManagerImpl(
```
brian-livefront, line 41: Let's make this internal and create a `AuthenticatorBridgeFactory` class that we use to create instances of this with a `createAuthenticatorBridgeManager` manager.

ahaisting-livefront: bbff1a65de6823e5be4558d779173693064f58b7

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: We chatted offline! Based on the docs and looking at few other examples of service binding code, we decided this is sufficient. 

Also, I tested putting "banana" in for package name and the function did not throw.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,193 @@
...
+            .fold(
+                onSuccess = { it },
+                onFailure = { false },
+            )
+            ?: false
+
+        if (!haveCorrectKey) {
+            // If we don't have the correct key, query for key:
+            symmetricKeyStorageProvider.symmetricKey =
+                bridgeService.safeCall { symmetricEncryptionKeyData }.getOrNull()
```
brian-livefront, line 150: I would think we'd want the failure state here to end up emitting `AccountSyncState.Error` or whatever it's called. Same thing above. I think this is a case where `null` is different from an error.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+/**
+ * Helper function for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+fun <T> IAuthenticatorBridgeService?.safeCall(action: IAuthenticatorBridgeService.() -> T): Result<T?> =
```
brian-livefront, line 190: This can be private right? Or do you need this for testing? If you need it for testing, we should move it to a different package and name it `internal`

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,193 @@
...
+ * application context.
+ * @param connectionType Specifies which build variant to connect to.
+ * @param symmetricKeyStorageProvider Provides access to local storage of the symmetric encryption
+ * key.
+ * @param callbackProvider Provides a way to construct a service callback that can be mocked in
+ * tests.
+ * @param processLifecycleOwner Lifecycle owner that is used to listen for start/stop
+ * lifecycle events.
+ */
+internal class AuthenticatorBridgeManagerImpl(
```
brian-livefront, line 41: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+    private val applicationContext: Context,
```
david-livefront, line 42: Would it make sense to do a check on the context here:

```kotlin
internal class AuthenticatorBridgeManagerImpl(
    context: Context,
    ...
) {
    private val applicationContext: Context = context.applicationContext

```

ahaisting-livefront: I've never thought to do this! I like it. Thanks, Diz!

0a3003439171ebd549fdd9de2a77751759b6b1b1

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/factory/AuthenticatorBridgeFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.bitwarden.authenticatorbridge.manager.AuthenticatorBridgeManager
+import com.bitwarden.authenticatorbridge.manager.AuthenticatorBridgeManagerImpl
+import com.bitwarden.authenticatorbridge.manager.model.AuthenticatorBridgeConnectionType
+import com.bitwarden.authenticatorbridge.provider.SymmetricKeyStorageProvider
+
+/**
+ * Factory for supplying implementation instances of Authenticator Bridge SDK interfaces.
+ */
+class AuthenticatorBridgeFactory(
+    private val applicationContext: Context,
```
brian-livefront, line 13: I think we should do what you've done below and ensure that the context is only an application context:

```kotlin
class AuthenticatorBridgeFactory(
    context: Context,
) {
    private val applicationContext = context.applicationContext

```

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/factory/AuthenticatorBridgeFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+     * encryption key.
+     */
+    fun getAuthenticatorBridgeManager(
+        connectionType: AuthenticatorBridgeConnectionType,
+        symmetricKeyStorageProvider: SymmetricKeyStorageProvider,
+    ): AuthenticatorBridgeManager = AuthenticatorBridgeManagerImpl(
+        context = applicationContext,
+        connectionType = connectionType,
+        symmetricKeyStorageProvider = symmetricKeyStorageProvider,
+    )
```
brian-livefront, line 30: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+private const val AUTHENTICATOR_BRIDGE_SERVICE_CLASS =
+    "com.x8bit.bitwarden.data.platform.service.AuthenticatorBridgeService"
+
+/**
+ * Default implementation of [AuthenticatorBridgeManager].
+ *
+ * @param applicationContext The Context that will be used to bind to AuthenticatorBridgeService.
```
brian-livefront, line 31: You need to update these docs now.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+                bridgeServiceConnection,
+                Context.BIND_AUTO_CREATE,
+            )
+        } catch (e: SecurityException) {
+            unbindService()
+            false
+        }
+
+        if (!isBound) {
+            mutableSharedAccountsStateFlow.value = AccountSyncState.Error
```
brian-livefront, line 117: We'll need to think about what this particular error would mean for the UI....if you don't have the other app installed, there's no reason to expect this to work. It's almost like we need to do a check to see if the application exists before we even try to bind. If not, we should show something like `AccountSyncState.NoApplication` or something.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+/**
+ * Helper function for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+internal fun <T> IAuthenticatorBridgeService?.safeCall(action: IAuthenticatorBridgeService.() -> T): Result<T?> =
```
brian-livefront, line 192: I think we said this should either be private or internal and placed in a separate file.

ahaisting-livefront: I was confused by your original comment. 

Are private extensions allowed in the same file?

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Done.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/factory/AuthenticatorBridgeFactory.kt
ahaisting-livefront: Done.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Main AuthenticatorBridgeService access point.
+     */
+    private var bridgeService: IAuthenticatorBridgeService? = null
+
+    /**
+     * Internal state of [accountSyncStateFlow].
+     */
+    private val mutableSharedAccountsStateFlow =
+        MutableStateFlow<AccountSyncState>(AccountSyncState.Loading)
```
brian-livefront, line 59: So we can determine in the `init` block if this should be `Loading` because the BW app is installed and we have a shot at binding to it, or something more like `ApplicationNotFound` if the BW app doesn't exist.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
brian-livefront: Yes, private extensions are allowed in the same file. Public / internal ones generally are not.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: 285c4bafac0195b5ed129df0b3435e87cbd844a5

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Gotcha. Yeah this will only be used in this file so private is fine.

2e415b60af0d1972194d263ba3e49ac3f6a0b295

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -193,3 +193,13 @@ internal class AuthenticatorBridgeManagerImpl(
...
+/**
+ * Helper function for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+private fun <T> IAuthenticatorBridgeService?.safeCall(action: IAuthenticatorBridgeService.() -> T): Result<T?> =
+    runCatching {
+        this?.let { action.invoke(it) }
+    }
```
brian-livefront, line 204: Add a newline here.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -147,8 +148,18 @@ internal class AuthenticatorBridgeManagerImpl(
...
-            symmetricKeyStorageProvider.symmetricKey =
-                bridgeService.safeCall { symmetricEncryptionKeyData }.getOrNull()
+            bridgeService.safeCall { symmetricEncryptionKeyData }
+                .fold(
+                    onSuccess = {
+                        symmetricKeyStorageProvider.symmetricKey = it
+                    },
+                    onFailure = {
+                        mutableSharedAccountsStateFlow.value = AccountSyncState.Error
+                        unbindService()
```
brian-livefront, line 157: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: Done.

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
ahaisting-livefront: I added a check during the `bindService` call as well. Basically, if the app doesn't exist, don't bind. 

1df5c36e3738da75adcfea5df5bc5263aab90b80

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -191,7 +202,17 @@ internal class AuthenticatorBridgeManagerImpl(
...
     }
+
+    private fun isBitwardenAppInstalled(): Boolean =
+        // Check to see if correct Bitwarden app is installed:
+        try {
+            applicationContext.packageManager.getPackageInfo(connectionType.toPackageName(), 0)
+            true
+        } catch (e: NameNotFoundException) {
+            false
+        }
```
brian-livefront, line 214: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -95,6 +102,10 @@ internal class AuthenticatorBridgeManagerImpl(
     }
 
     private fun bindService() {
+        if (!isBitwardenAppInstalled()) {
+            mutableSharedAccountsStateFlow.value = AccountSyncState.AppNotInstalled
+            return
+        }
```
brian-livefront, line 108: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ * Helper function for wrapping all calls to [IAuthenticatorBridgeService] around try catch.
+ *
+ * This is important because all calls to [IAuthenticatorBridgeService] can throw
+ * DeadObjectExceptions as well as RemoteExceptions.
+ */
+private fun <T> IAuthenticatorBridgeService?.safeCall(action: IAuthenticatorBridgeService.() -> T): Result<T?> =
```
brian-livefront, line 223: Just noticing that this returns `Result<T?>` instead of `Result<T>`. Can we make this so the call is on `IAuthenticatorBridgeService` instead of `IAuthenticatorBridgeService?`? We should do a local check to get a non-null value for it, rather than baking that into the helper I would think.

ahaisting-livefront: Gotcha, I think this solves that: d0fbe8a743afaca40de1d88d426ee04b28786ab2

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,230 @@
...
+            ?.let { data.decrypt(it) }
+            ?.getOrNull()
+            ?.let { AccountSyncState.Success(it.accounts) }
+            ?: AccountSyncState.Error
+    }
+
+    private fun onServiceConnected(binder: IBinder) {
+        val service = IAuthenticatorBridgeService.Stub
+            .asInterface(binder)
+            .also { bridgeService = it }
```
brian-livefront, line 144: 👍

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerImpl.kt
```diff
@@ -0,0 +1,230 @@
...
+            symmetricKeyStorageProvider.symmetricKey?.toFingerprint()?.getOrNull()
+
+        // Query bridge service to see if we have a matching symmetric key:
+        val haveCorrectKey = service
+            .safeCall { checkSymmetricEncryptionKeyFingerprint(localKeyFingerprint) }
+            .fold(
+                onSuccess = { it },
+                onFailure = { false },
+            )
+            ?: false
```
brian-livefront, line 159: Do we need this now?

ahaisting-livefront: Nope! 2c1cbb60986ae5597ea50c90dbc286572040aa1e

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+
+    @AfterEach
+    fun teardown() {
+        unmockkConstructor(Intent::class)
+        unmockkStatic(IAuthenticatorBridgeService.Stub::class)
+        unmockkStatic(EncryptedSharedAccountData::decrypt)
+    }
+
+    @Test
+    fun `initial state should be Loading`() {
```
brian-livefront, line 71:
```suggestion
    fun `initial state should be Loading when Bitwarden app is present`() {
```

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+        assertEquals(AccountSyncState.Loading, manager.accountSyncStateFlow.value)
+    }
+
+    @Test
+    fun `initial state should be AppNotInstalled when package manager can't find app`() {
```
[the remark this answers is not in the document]
brian-livefront: Just for symmetry with the first one.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+        every { context.bindService(any(), any(), Context.BIND_AUTO_CREATE) } returns false
+
+        fakeLifecycleOwner.lifecycle.dispatchOnStart()
+
+        assertEquals(AccountSyncState.Error, manager.accountSyncStateFlow.value)
+        verify { context.bindService(any(), any(), Context.BIND_AUTO_CREATE) }
+    }
+
+    @Test
+    fun `onStart when package manager doesn't find bitwarden app should set state to AppNotInstalled`() {
```
brian-livefront, line 106:
```suggestion
    fun `onStart when Bitwarden app is not present should set state to AppNotInstalled`() {
```

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+        } throws SecurityException()
+
+        fakeLifecycleOwner.lifecycle.dispatchOnStart()
+
+        assertEquals(AccountSyncState.Error, manager.accountSyncStateFlow.value)
+        verify { context.bindService(any(), any(), Context.BIND_AUTO_CREATE) }
+    }
+
+    @Test
+    fun `onStart when bindService succeeds state should be Loading before service calls back`() {
```
brian-livefront, line 142:
```suggestion
    fun `onStart when Bitwarden app is present and bindService succeeds should be Loading before service calls back`() {
```

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+
+        assertEquals(AccountSyncState.Loading, manager.accountSyncStateFlow.value)
+        assertEquals(fakeSymmetricKeyStorageProvider.symmetricKey, SYMMETRIC_KEY)
+        verify { mockBridgeService.symmetricEncryptionKeyData }
+        verify { context.bindService(any(), any(), Context.BIND_AUTO_CREATE) }
+    }
+
+    @Test
+    @Suppress("MaxLineLength")
+    fun `onServiceConnected when symmetric key does match not query for symmetric key`() {
```
brian-livefront, line 216: Seems like something is missing in this one.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/util/FakeSymmetricKeyStorageProvider.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.bitwarden.authenticatorbridge.model.SymmetricEncryptionKeyData
+import com.bitwarden.authenticatorbridge.provider.SymmetricKeyStorageProvider
+
+/**
+ * A fake implementation of [SymmetricKeyStorageProvider] for testing purposes.
+ */
+class FakeSymmetricKeyStorageProvider : SymmetricKeyStorageProvider {
+
+    override var symmetricKey: SymmetricEncryptionKeyData? = null
+
```
brian-livefront, line 12: Can we remove the extra newlines on this to match `FakeLifecycleOwner`?

ahaisting-livefront: 955a1eb1b43d4cb1f6feb5c187845b70fad9e997

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
ahaisting-livefront: Yep, the word "should": 2680fd8f95f194121723ab94bfedc389de4468b6

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
brian-livefront: And actually...initial state of what? I think `accountSyncStateFlow` should probably be in the names of most or all of these functions.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
ahaisting-livefront: I would argue that the app being present is part of `bindService` succeeding. But more detail is fine too:

b3dc685f283950ff2aa3026ba972a973f8ba4d14

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
ahaisting-livefront: 12e7c4962e5ce2ead8820f59c3a073519533599c

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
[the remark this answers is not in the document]
ahaisting-livefront: 8ebc24aeec0b004f560b17bdb081dd161beec027

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
ahaisting-livefront: f65bac9e2439507d8e74afdb640e6d280e588ac7

authenticatorbridge/src/main/java/com/bitwarden/authenticatorbridge/factory/AuthenticatorBridgeFactory.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.content.Context
+import com.bitwarden.authenticatorbridge.manager.AuthenticatorBridgeManager
+import com.bitwarden.authenticatorbridge.manager.AuthenticatorBridgeManagerImpl
+import com.bitwarden.authenticatorbridge.manager.model.AuthenticatorBridgeConnectionType
+import com.bitwarden.authenticatorbridge.provider.SymmetricKeyStorageProvider
+
+/**
+ * Factory for supplying implementation instances of Authenticator Bridge SDK interfaces.
+ */
+class AuthenticatorBridgeFactory(
```
brian-livefront, line 12: We should probably add a `@OmitFromCoverage` so we don't need to add unnecessary tests for this.

ahaisting-livefront: Dizzy mentioned this as well. The tricky thing is that we don't have that annotation in this module... _yet_.

brian-livefront: Gotcha. Yeah the duplication of some things across modules (like some of the fakes too) is something we'll have to address as some point.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+        every {
+            context.bindService(
+                any(),
+                capture(serviceConnection),
+                Context.BIND_AUTO_CREATE
+            )
+        } returns true
+
+        fakeLifecycleOwner.lifecycle.dispatchOnStart()
+        serviceConnection.captured.onServiceConnected(mockk(), mockk())
```
brian-livefront, line 362: Can we handle this just by doing an `answers` where we trigger this?

```kotlin
 every { mockBridgeService.registerBridgeServiceCallback(capture(callback)) } answers {
    (it.invocation.args.first() as IAuthenticatorBridgeServiceCallback).onAccountsSync(encryptedAccounts)
}
```

Just keeps it in one place.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+        assertEquals(AccountSyncState.Loading, manager.accountSyncStateFlow.value)
+
+        callback.captured.onAccountsSync(encryptedAccounts)
+        assertEquals(AccountSyncState.Error, manager.accountSyncStateFlow.value)
+
+        assertEquals(fakeSymmetricKeyStorageProvider.symmetricKey, SYMMETRIC_KEY)
```
brian-livefront, line 411: I think the order of this assertion is backwards right? `SYMMETRIC_KEY` is the expected value, so it should be first.

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+                capture(serviceConnection),
+                Context.BIND_AUTO_CREATE
+            )
+        } returns true
+
+        fakeLifecycleOwner.lifecycle.dispatchOnStart()
+        serviceConnection.captured.onServiceConnected(mockk(), mockk())
+
+        assertEquals(AccountSyncState.Error, manager.accountSyncStateFlow.value)
+        assertEquals(fakeSymmetricKeyStorageProvider.symmetricKey, SYMMETRIC_KEY)
```
brian-livefront, line 280:
```suggestion
        assertEquals(SYMMETRIC_KEY, fakeSymmetricKeyStorageProvider.symmetricKey)
```

ahaisting-livefront: Ah, yep, those are backwards: 08264b27d598a86435238a14587e34d90d9895a6

authenticatorbridge/src/test/java/com/bitwarden/authenticatorbridge/manager/AuthenticatorBridgeManagerTest.kt
ahaisting-livefront: Ah, yep, those are backwards: https://github.com/bitwarden/android/commit/08264b27d598a86435238a14587e34d90d9895a6

brian-livefront: There are still some of these that are mixed up.

brian-livefront (approved):
Looks good!

david-livefront (approved):

PR #4006, opened 2024-10-01
[PM-9363] Disable cipher key encryption for older server versions

mpbw2:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-9363

## 📔 Objective

Disable cipher key encryption in SDK if server version < 2024.2.0

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
@@ -3,6 +3,7 @@ package com.x8bit.bitwarden.data.platform.manager
 import com.x8bit.bitwarden.data.platform.datasource.disk.model.ServerConfig
 import com.x8bit.bitwarden.data.platform.manager.model.FlagKey
 import com.x8bit.bitwarden.data.platform.repository.ServerConfigRepository
+import com.x8bit.bitwarden.data.platform.util.isServerVersionAtLeast
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
@@ -16,7 +17,8 @@ class FeatureFlagManagerImpl(
 ) : FeatureFlagManager {
 
     override val sdkFeatureFlags: Map<String, Boolean>
-        get() = mapOf(CIPHER_KEY_ENCRYPTION_KEY to true)
+        get() = mapOf(CIPHER_KEY_ENCRYPTION_KEY to
+            isServerVersionAtLeast(serverConfigRepository, "2024.2.0"))
 
     override fun <T : Any> getFeatureFlagFlow(key: FlagKey<T>): Flow<T> =
         serverConfigRepository
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
 private const val CIPHER_KEY_ENCRYPTION_KEY = "enableCipherKeyEncryption"
+private const val CIPHER_KEY_ENC_MIN_SERVER_VERSION = "2024.2.0"
 
 /**
  * Primary implementation of [FeatureFlagManager].
@@ -17,8 +18,13 @@ class FeatureFlagManagerImpl(
 ) : FeatureFlagManager {
 
     override val sdkFeatureFlags: Map<String, Boolean>
-        get() = mapOf(CIPHER_KEY_ENCRYPTION_KEY to
-            isServerVersionAtLeast(serverConfigRepository, "2024.2.0"))
+        get() = mapOf(
+            CIPHER_KEY_ENCRYPTION_KEY to
+                isServerVersionAtLeast(
+                    serverConfigRepository.getLocalServerConfig(),
+                    CIPHER_KEY_ENC_MIN_SERVER_VERSION,
+                ),
+        )
 
     override fun <T : Any> getFeatureFlagFlow(key: FlagKey<T>): Flow<T> =
         serverConfigRepository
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlin.text.split
+import kotlin.text.toIntOrNull
+
+private const val VERSION_SEPARATOR = "."
+private const val SUFFIX_SEPARATOR = "-"
+
+/**
+ * Checks if the server version is greater than another provided version, returns true if it is.
+ */
+ fun isServerVersionAtLeast(serverConfig: ServerConfigRepository, version: String): Boolean {
```
SaintPatrck, line 13: 🎨 Can we change `serverConfig` argument to type `ServerConfig?` or possibly even `serverVersion: String?` instead of passing the repository?

mpbw2: I ended up using `ServerConfig?` instead of another string since I kept second guessing myself on which string was which regardless of the names

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
```diff
@@ -16,7 +18,13 @@ class FeatureFlagManagerImpl(
 ) : FeatureFlagManager {
 
     override val sdkFeatureFlags: Map<String, Boolean>
-        get() = mapOf(CIPHER_KEY_ENCRYPTION_KEY to true)
+        get() = mapOf(
+            CIPHER_KEY_ENCRYPTION_KEY to
+                isServerVersionAtLeast(
+                    serverConfigRepository.getLocalServerConfig(),
```
david-livefront, line 24: Can you just access it via the stateFlow directly?

```kotlin
serverConfigRepository.serverConfigStateFlow.value
```

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+    if (serverVersionParts == null || otherVersionParts == null) {
+        return false
+    }
+
+    for (i in serverVersionParts.indices) {
+        val serverPart = serverVersionParts.getOrNull(i)?.toIntOrNull() ?: 0
+        val otherPart = otherVersionParts.getOrNull(i)?.toIntOrNull() ?: 0
+
+        if (serverPart > otherPart) {
+            return true
```
david-livefront, line 30: Can we lift the return

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+ fun isServerVersionAtLeast(serverConfig: ServerConfig?, version: String): Boolean {
+    val serverVersion = serverConfig
+        ?.serverData
+        ?.version
+
+    val serverVersionParts = getVersionComponents(serverVersion)
+    val otherVersionParts = getVersionComponents(version)
+
+    if (serverVersionParts == null || otherVersionParts == null) {
+        return false
```
david-livefront, line 22: Can we add a test

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+        val otherPart = otherVersionParts.getOrNull(i)?.toIntOrNull() ?: 0
+
+        if (serverPart > otherPart) {
+            return true
+        } else if (serverPart < otherPart) {
+            return false
+        }
+    }
+
+    return true // Versions are equal
```
david-livefront, line 36: Comment should be above the return not on the same line.

Can we also add a test for this case

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
mpbw2: We have to walk through the entire loop to know if the version parts are the same. Added a comment for clarity.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt
```diff
@@ -16,7 +18,13 @@ class FeatureFlagManagerImpl(
 ) : FeatureFlagManager {
 
     override val sdkFeatureFlags: Map<String, Boolean>
-        get() = mapOf(CIPHER_KEY_ENCRYPTION_KEY to true)
+        get() = mapOf(
+            CIPHER_KEY_ENCRYPTION_KEY to
+                isServerVersionAtLeast(
+                    serverConfigRepository.serverConfigStateFlow.value,
```
david-livefront, line 24: 👍

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlin.text.split
+import kotlin.text.toIntOrNull
+
+private const val VERSION_SEPARATOR = "."
+private const val SUFFIX_SEPARATOR = "-"
+
+/**
+ * Checks if the server version is greater than another provided version, returns true if it is.
+ */
+fun isServerVersionAtLeast(serverConfig: ServerConfig?, version: String?): Boolean {
```
david-livefront, line 13: Last question, why would the version ever be null?

mpbw2: It's shouldn't be, but I felt weird skipping the null check further down just to satisfy the IDE so I made it nullable. Thoughts?

david-livefront: I think we can make `version` nonnull and just do the empty check

app/src/main/java/com/x8bit/bitwarden/data/platform/util/ServerVersionUtils.kt
david-livefront: Ahh, missed that

david-livefront (approved):

PR #4091, opened 2024-10-14
PM-12296: Only match port when present on both uris

david-livefront:
## 🎟️ Tracking

[PM-12296](https://bitwarden.atlassian.net/browse/PM-12296)

## 📔 Objective

This PR updates the `HOST` matching logic to only match on port when both the source and the target have a port present.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-12296]: https://bitwarden.atlassian.net/browse/PM-12296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/ciphermatching/CipherMatchingManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/manager/ciphermatching/CipherMatchingManagerImpl.kt
@@ -7,8 +7,10 @@ import com.x8bit.bitwarden.data.platform.manager.ResourceCacheManager
 import com.x8bit.bitwarden.data.platform.repository.SettingsRepository
 import com.x8bit.bitwarden.data.platform.util.firstWithTimeoutOrNull
 import com.x8bit.bitwarden.data.platform.util.getDomainOrNull
+import com.x8bit.bitwarden.data.platform.util.getHostOrNull
 import com.x8bit.bitwarden.data.platform.util.getHostWithPortOrNull
 import com.x8bit.bitwarden.data.platform.util.getWebHostFromAndroidUriOrNull
+import com.x8bit.bitwarden.data.platform.util.hasPort
 import com.x8bit.bitwarden.data.platform.util.isAndroidApp
 import com.x8bit.bitwarden.data.platform.util.regexOrNull
 import com.x8bit.bitwarden.data.vault.repository.VaultRepository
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/manager/ciphermatching/CipherMatchingManagerImpl.kt
```diff
@@ -210,9 +212,15 @@ private fun LoginUriView.checkForMatch(
             UriMatchType.EXACT -> exactIfTrue(loginViewUri == matchUri)
 
             UriMatchType.HOST -> {
-                val loginUriHost = loginViewUri.getHostWithPortOrNull()
-                val matchUriHost = matchUri.getHostWithPortOrNull()
-                exactIfTrue(matchUriHost != null && loginUriHost == matchUriHost)
+                if (loginViewUri.hasPort() && matchUri.hasPort()) {
```
david-livefront, line 215: I think we could technically just check if the `matchUri` has a port since it is the value that might not have it, whereas the `loginViewUri` is the value stored to disk and _should_ be accurate.

That could make the functionality more confusing but it is an option.

vvolkgang: That's what I had initially too in my snippet. Assuming this may change under our feet at some point in the future, I thought it could start failing to match cases where users just set a domain / IP with the knowledge that it didn't matter back then, and it would be difficult for us to troubleshoot why. 🤔

david-livefront: I would hope that it would not change under our feet since the AutofillService provides us the `webDomain` and `webScheme` as 2 separate components and we have to put it together ourselves. Hopefully the port would be added as a new property that we would have to manage ourselves. Having said that, it could happen. 

Do you think I should change this to only check for the `matchUri.hasPort()`

vvolkgang: I would leave it as is.

vvolkgang (approved):
👏

david-livefront:
Thanks @vvolkgang & @dseverns-livefront

PR #4203, opened 2024-10-30
PM-13848 Handle URIs with ports and host matching

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-13948

## 📔 Objective
- If a port was specified with no scheme, the `URI` was not parsing the string correctly and was using the `host` as the `scheme`

## 📸 Screenshots
before: 

https://github.com/user-attachments/assets/a0248aeb-e46f-49ac-8ae1-54b5306a7828

After: 

https://github.com/user-attachments/assets/927c4711-0a73-4d4f-beee-35331533a65a

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
@@ -63,7 +63,10 @@ fun String.getDomainOrNull(resourceCacheManager: ResourceCacheManager): String? 
  */
 @OmitFromCoverage
 fun String.hasPort(): Boolean {
-    val uri = this.toUriOrNull() ?: return false
+    val uri = this
+        .toUriOrNull()
+        ?.let { addSchemeToUriIfNecessary(this, uri = it) }
+        ?: return false
     return uri.port != -1
 }
 
@@ -73,15 +76,7 @@ fun String.hasPort(): Boolean {
 @OmitFromCoverage
 fun String.getHostOrNull(): String? = this.toUriOrNull()
     ?.let { uri ->
-        if (
-            uri.host == null &&
-            this.contains(".") &&
-            !this.hasHttpProtocol()
-        ) {
-            URI("https://$this")
-        } else {
-            uri
-        }
+        addSchemeToUriIfNecessary(uriString = this, uri = uri)
     }?.host
 
 /**
@@ -89,7 +84,10 @@ fun String.getHostOrNull(): String? = this.toUriOrNull()
  */
 @OmitFromCoverage
 fun String.getHostWithPortOrNull(): String? {
-    val uri = this.toUriOrNull() ?: return null
+    val uri = this
+        .toUriOrNull()
+        ?.let { addSchemeToUriIfNecessary(uriString = this, uri = it) }
+        ?: return null
     return uri.host?.let { host ->
         val port = uri.port
         if (port != -1) {
@@ -112,5 +110,27 @@ fun String.findLastSubstringIndicesOrNull(substring: String): IntRange? {
         IntRange(lastIndex, endIndex)
     } else {
         null
+    }
+}
+
+/**
+ * Private helper method which takes a string and a [URI] derived from that string and adds
+ * an HTTPS scheme to the URI if it is necessary. (i.e. foo.com:9090 -> https://foo.com:9090)
+ * This would be necessary if you have a source string like "foo.com:9090" and you want to want
+ * to extract the [URI.getHost] or [URI.getPort] from it which would fail.
+ */
+private fun addSchemeToUriIfNecessary(uriString: String, uri: URI): URI? {
+    if (uriString.toUriOrNull()?.equals(uri) != true) return null
+    return if (
+    // see if the string contains a host pattern
+        uriString.contains(".") &&
+        // if it does but the URI's host is null, add an https scheme
+        uri.host == null &&
+        // provided that scheme does not exist already.
+        !uriString.hasHttpProtocol()
+    ) {
+        URI("https://$uriString")
+    } else {
+        uri
     }
 }
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
@@ -15,16 +15,7 @@ private const val ANDROID_APP_PROTOCOL: String = "androidapp://"
  */
 fun String.toUriOrNull(): URI? =
     try {
-        val uri = URI(this)
-        if (
-            uri.host == null &&
-            this.contains(".") &&
-            !this.hasHttpProtocol()
-        ) {
-            URI("https://$this")
-        } else {
-            uri
-        }
+        URI(this)
     } catch (e: URISyntaxException) {
         null
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
```diff
@@ -15,7 +15,16 @@ private const val ANDROID_APP_PROTOCOL: String = "androidapp://"
  */
 fun String.toUriOrNull(): URI? =
     try {
-        URI(this)
+        val uri = URI(this)
+        if (
+            uri.host == null &&
+            this.contains(".") &&
+            !this.hasHttpProtocol()
```
david-livefront, line 22: What if it has a different protocol, like `androidapp`?

dseverns-livefront: in that case if it has a valid scheme then host would be present, the specific case something `foo.com:1000` when put in the `URI("foo.com:1000")` -> uri.host == null and uri.scheme == "foo.com", assuming its parsing for the first `:` to get the scheme.

dseverns-livefront: so it it has a valid scheme and a host we would be presuming that `uri.host != null` and the `URI()` will parse it correctly.

app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
```diff
@@ -71,7 +71,18 @@ fun String.hasPort(): Boolean {
  * Extract the host from this [String] if possible, otherwise return null.
  */
 @OmitFromCoverage
-fun String.getHostOrNull(): String? = this.toUriOrNull()?.host
+fun String.getHostOrNull(): String? = this.toUriOrNull()
+    ?.let { uri ->
+        if (
+            uri.host == null &&
+            this.contains(".") &&
+            !this.hasHttpProtocol()
```
david-livefront, line 79: Can we add a little comment or doc just explaining why these extra steps are needed

app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
dseverns-livefront: did a little changing of where we do this check to only be when grabbing the host, also added tests

david-livefront (dismissed):

app/src/main/java/com/x8bit/bitwarden/data/platform/util/URIExtensions.kt
```diff
@@ -45,6 +45,27 @@ fun URI.parseDomainNameOrNull(resourceCacheManager: ResourceCacheManager): Domai
             )
         }
 
+/**
+ * Adds and HTTPS scheme to a valid URI if that URI has a valid host in the raw string but
+ * the standard parsing of `URI("foo.com")` is not able to determine a valid host.
+ * (i.e. val uri = URI("foo.bar:1090") -> uri.host == null)
+ */
+fun URI.addSchemeToUriIfNecessary(): URI {
```
vvolkgang, line 53: Great find! 

🤔 Assuming https means this will fail to match http hosts right? It's still common for lan services / local domains. If that's the case, could we instead change the match detection to only compare schema if it's present in our cipher? IIRC that's already the case for the port.

vvolkgang (changes requested):

app/src/test/java/com/x8bit/bitwarden/data/util/UriExtensionsTest.kt
```diff
@@ -321,4 +322,38 @@ class UriExtensionsTest {
...
+    }
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `addSchemeToUriIfNecessary should add https when https scheme is missing and a port is present`() {
+        val uriWithPort = URI("example.com:8080")
+        assertEquals(URI("https://example.com:8080"), uriWithPort.addSchemeToUriIfNecessary())
+    }
+
+    @Test
```
dseverns-livefront, line 339: Test for http:// schemes

app/src/main/java/com/x8bit/bitwarden/data/platform/util/URIExtensions.kt
dseverns-livefront: @vvolkgang if the URL explicitly has an `http` scheme it will still match correctly, where this is being called it is specifically used to ensure the resulting URI is parsed correctly when attempting to grab out specific pieces: i.e. `uri.host` and `uri.port` specifically. Without a scheme this is not possible, so we are only adding the scheme if it does not already exist. If the user adds `http://foo.bar` or just `foo.bar` for an domain that is use `http` it will still match correctly. It will only add the `https` scheme if there is no other scheme present and it has a "host" or domain pattern based on the `.`. I am not seeing in the CipherMatching where we are comparing the schemes for the matching.

vvolkgang: I meant when the cipher is setup for a local service without a schema (e.g. `rpi.local`) and our app receives a request for `http://rpi.local`. If I understood correctly we would try to match `https://rpi.local` (from our cipher) to `http://rpi.local` (from the browser), would it fail in that case?

dseverns-livefront: @vvolkgang sorry for delay, been trying to track down one of the prod crashes, I tested this with an `http` site, with host matching, it is in fact working (video below). With the default domain matching it did not, but not because of the scheme. The host/domain of `testphp.vulnweb.com` being shortened to `vulnweb.com`, I believe this exposed another area where the URI is not being parsed as expected, adding this extension to our `String.getDomainOrNul()` call fixes this too.

https://github.com/user-attachments/assets/b59796d8-397c-46a2-ba5b-57ee530ead06

vvolkgang: Awesome, thanks for checking and improving that too!

vvolkgang (approved):
@dseverns-livefront confirmed this will work for local http services that are setup without the schema in our cipher. @david-livefront given that you did the initial review would appreciate you approving the recent changes!

qiaoyun680:
> ## 🎟️ Tracking
> https://bitwarden.atlassian.net/browse/PM-13948
> 
> ## 📔 Objective
> * If a port was specified with no scheme, the `URI` was not parsing the string correctly and was using the `host` as the `scheme`
> 
> ## 📸 Screenshots
> before:
> 
>  beforepm13xxx.mp4 
> After:
> 
>  pm13xxxafter.mp4 
> ## ⏰ Reminders before review
> * Contributor guidelines followed
> * All formatters and local linters executed and passed
> * Written new unit and / or integration tests where applicable
> * Used internationalization (i18n) for all UI strings
> * CI builds passed
> * Communicated to DevOps any deployment requirements
> * Updated any necessary documentation or informed the documentation team
> 
> ## 🦮 Reviewer guidelines
> * 👍 (`:+1:`) or similar for great changes
> * 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
> * ❓ (`:question:`) for questions
> * 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
>   issue and could potentially benefit from discussion
> * 🎨 (`:art:`) for suggestions / improvements
> * ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
> * 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
> * ⛏ (`:pick:`) for minor or nitpick changes
This issue has not been resolved. If the URL contains a port number, the matching process will match all port entries for all hosts. It should match based on both the host and the port number.

PR #4227, opened 2024-11-04
PM-14433 update flow type to nullable so we can handle when an empty …

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-14433

## 📔 Objective
- Avoid call on a null object when a query is made against an empty table returning a Flow. 
- According to the [documentation](https://developer.android.com/reference/androidx/room/Query), this should have thrown a NPE, but it seems that there is an open [issue](https://issuetracker.google.com/issues/213175894) where it actually is returning Flow<T?> and emitting `null`. Updating the contract to return a `Flow<DomainEntity>` so it matches the result we get from Room, then we filter out the null emissions where we are mapping back to the JSON.
- FWIW: I checked other areas in the codebase where we are returning `Flow`s from Room and they are all `Flow<List<T>>` so they will return an empty list when the table is empty.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceImpl.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.filterNotNull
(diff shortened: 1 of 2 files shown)


dseverns-livefront:
Added `hold` label as I have been unable to reproduce the issue where the table is read from before the data is sync'd. I believe that there was particular data/ migration scenario that exposed this issue because this would have been happening a lot more frequently if it was something that happened each time the app sync'd for the first time

dseverns-livefront:
Adding back `hold` as this was merged to hotfix-v2024.11.0

PR #4263, opened 2024-11-08
PM-14433 update flow type to nullable so we can handle gracefully and avoid crash

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14433
Cherry-picked from #4231 (5a4b8d64) on branch `release/hotfix-v2024.11.0`.

## 📔 Objective
- Avoid call on a null object when a query is made against an empty table returning a Flow. 
- According to the [documentation](https://developer.android.com/reference/androidx/room/Query), this should have thrown a NPE, but it seems that there is an open [issue](https://issuetracker.google.com/issues/213175894) where it actually is returning Flow<T?> and emitting `null`. Updating the contract to return a `Flow<DomainEntity>` so it matches the result we get from Room, then we filter out the null emissions where we are mapping back to the JSON.
- FWIW: I checked other areas in the codebase where we are returning `Flow`s from Room and they are all `Flow<List<T>>` so they will return an empty list when the table is empty.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceImpl.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceImpl.kt
```diff
@@ -124,6 +125,7 @@ class VaultDiskSourceImpl(
     override fun getDomains(userId: String): Flow<SyncResponseJson.Domains> =
         domainsDao
             .getDomains(userId)
+            .filterNotNull()
```
david-livefront, line 128: So this will be removed when the cherry-pick from my PR goes through, right?

SaintPatrck: Correct

dseverns-livefront (approved):

PR #4305, opened 2024-11-14
PM-14429 Set the min and max range of the slider to match the restrictions not update the min with the computed min.

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-14429

## 📔 Objective
- When decreasing the  min numbers or special characters the value continues to change because the length of the pw does not change and the minimum range is changing.

## 📸 Screenshots
Before: See story
After: 

https://github.com/user-attachments/assets/56807212-d0f7-4ae4-a9dc-a5d8afc1cb1d

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
@@ -455,8 +455,12 @@ private fun ColumnScope.PasswordTypeContent(
 
     BitwardenSlider(
         value = passwordTypeState.length,
-        onValueChange = passwordHandlers.onPasswordSliderLengthChange,
-        range = passwordTypeState.computedMinimumLength..passwordTypeState.maxLength,
+        onValueChange = { newValue, isUserInteracting ->
+            if (newValue >= passwordTypeState.computedMinimumLength) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -455,8 +455,12 @@ private fun ColumnScope.PasswordTypeContent(
 
     BitwardenSlider(
         value = passwordTypeState.length,
-        onValueChange = passwordHandlers.onPasswordSliderLengthChange,
-        range = passwordTypeState.computedMinimumLength..passwordTypeState.maxLength,
+        onValueChange = { newValue, isUserInteracting ->
+            if (newValue >= passwordTypeState.computedMinimumLength) {
```
david-livefront, line 459: Do we need this `if` check?

dseverns-livefront: It is preventing this behavior where the if the user pulls the slider back it generates a bunch of pw. The same thing happens now but not as fast. More of an edge case but that's why its there.
https://github.com/user-attachments/assets/60117016-cf84-4a0f-a804-352708dc37e2

PR #4359, opened 2024-11-21
PM-15109 only accept numeric values for account pin lock value

dseverns-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-15109

## 📔 Objective
- prevent users from entering or pasting non numeric values in for the text field.

## 📸 Screenshots

https://github.com/user-attachments/assets/2c168289-c382-407c-a668-e159f6b134bc

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/PinInputDialog.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/PinInputDialog.kt
@@ -107,7 +107,11 @@ fun PinInputDialog(
                     label = stringResource(id = R.string.pin),
                     value = pin,
                     autoFocus = true,
-                    onValueChange = { pin = it },
+                    onValueChange = { newValue ->
+                        if (newValue.all { it.isDigit() }) {
+                            pin = newValue
+                        }
+                    },
                     keyboardType = KeyboardType.Number,
                     modifier = Modifier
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/PinInputDialog.kt
```diff
@@ -107,7 +107,11 @@ fun PinInputDialog(
                     label = stringResource(id = R.string.pin),
                     value = pin,
                     autoFocus = true,
-                    onValueChange = { pin = it },
+                    onValueChange = { newValue ->
+                        if (newValue.all { it.isDigit() }) {
+                            pin = newValue
```
david-livefront, line 112: What do you think of this:

```kotlin
onValueChange = { pin = it.filter { it.isDigit() } },
```

dseverns-livefront: I guess the pros: are 86 one level of `{}`.
cons: we are technically reassigning pin each time, which wouldn't really cause recomps as the state's value wouldn't change.
either way we are iterating through whole string, so for the sack of pseudo-efficiency and laziness I would feel fine to leave it, but if the nesting is something we want to avoid this change would also be fine with me.
Your call?

PR #4400, opened 2024-12-02
[PM-8217] New device notice email access UI

andrebispo5:
## 🎟️ Tracking

[PM-8217 Add notice to users with no 2FA](https://bitwarden.atlassian.net/browse/PM-8217)
## 📔 Objective

Add first screen for the notice to users with no 2FA

## 📸 Screenshots
image

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+import com.x8bit.bitwarden.ui.platform.components.scaffold.BitwardenScaffold
+import com.x8bit.bitwarden.ui.platform.components.toggle.BitwardenSwitch
+import com.x8bit.bitwarden.ui.platform.components.util.rememberVectorPainter
+import com.x8bit.bitwarden.ui.platform.theme.BitwardenTheme
+
+/**
+ * The top level composable for the new device notice email access screen.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Suppress("LongMethod")
```
david-livefront, line 51: This does not look like the method is too long

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+        when (event) {
+            NavigateToTwoFactorOptions -> onNavigateToTwoFactorOptions()
+        }
+    }
+
+    val scrollBehavior = TopAppBarDefaults.pinnedScrollBehavior(rememberTopAppBarState())
+    BitwardenScaffold(
+        modifier = Modifier
+            .fillMaxSize()
+            .nestedScroll(scrollBehavior.nestedScrollConnection),
```
david-livefront, line 68: There is no toolbar, do we need this at all?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/**
+ * The main content of the screen.
+ */
+@Composable
+private fun MainContent(
+    email: String,
+    isEmailAccessEnabled: Boolean = false,
+    onEmailAccessToggleChanged: (Boolean) -> Unit = {},
```
david-livefront, line 156: Why do these have default values?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+) {
+    Spacer(modifier = Modifier.size(24.dp))
```
david-livefront, line 158: This can just be height, not size

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+            textAlign = TextAlign.Center,
+        )
+    }
+}
+
+/**
+ * The main content of the screen.
+ */
+@Composable
+private fun MainContent(
```
david-livefront, line 153: Since this does not have a wrapping contain with a modifier, I believe the pattern requires us to scope the function to provide that extra context.

```kotlin
private fun ColumnScope.MainContent(
```

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
andrebispo5: You are right

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
andrebispo5: Removing it

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+    isEmailAccessEnabled: Boolean = false,
+    onEmailAccessToggleChanged: (Boolean) -> Unit = {},
+) {
+    Spacer(modifier = Modifier.size(24.dp))
+    Column(
+        modifier = Modifier
+            .padding(horizontal = 22.dp)
+            .fillMaxWidth()
+            .clip(RoundedCornerShape(size = 4.dp))
+            .background(BitwardenTheme.colorScheme.background.secondary),
```
david-livefront, line 164: We should not be using custom stuff like this, we already have `BitwardenContentCard` to handle stuff like this.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
andrebispo5: Just left them after a layout rework. Removing them.

david-livefront:
@andrebispo5 Could you link the Figma files for this?

This does not seem to match current design patterns used in the app.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessViewModel.kt
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    private fun handleContinueClick() {
+        sendEvent(NavigateToTwoFactorOptions)
+    }
+
+    private fun handleEmailAccessToggle(action: EmailAccessToggle) {
+        mutableStateFlow.update {
+            it.copy(isEmailAccessEnabled = action.newState)
```
david-livefront, line 47: What is the intended purpose of the switch?

Do we not send the code if they do not have access to their email?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -118,15 +110,17 @@ private fun NewDeviceNoticeEmailAccessContent(
...
 @Composable
-private fun HeaderContent() {
+private fun HeaderContent(
+    modifier: Modifier = Modifier,
+) {
     Image(
         painter = rememberVectorPainter(id = R.drawable.warning),
         contentDescription = null,
-        modifier = Modifier.size(120.dp),
+        modifier = modifier.size(120.dp),
```
david-livefront, line 119: The `Modifier` should only ever be applied to the top level component, if you wanted to pass one in like this, you would want to wrap it in a column.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessViewModel.kt
andrebispo5: If the user has access to the email we can dismiss this flow.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
andrebispo5: The `BitwardenContentCard` didn't seem to adapt to what I wanted.

david-livefront: That seems to indicate that the UI is non-standard, I don't think we should be doing that.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,194 @@
...
+}
+
+@Composable
+private fun NewDeviceNoticeEmailAccessContent(
+    email: String,
+    isEmailAccessEnabled: Boolean,
+    onEmailAccessToggleChanged: (Boolean) -> Unit = {},
+    onContinueClick: () -> Unit = {},
+) {
+    Column(
```
david-livefront, line 82: Since you went the route of adding a top-level column, be sure to pass in a modifier.

We also do not need default params for the `onEmailAccessToggleChanged` and `onContinueClick`

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+@Suppress("MaxLineLength")
+@Composable
+private fun HeaderContent(
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally,
+        modifier = modifier.padding(horizontal = 24.dp),
```
david-livefront, line 119: Why is this using a non-standard amount of padding?

Can we discuss this with design because this screen is going to be difficult to maintain since it uses so many non-standard design patterns.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/newdevicenotice/NewDeviceNoticeEmailAccessScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Header content containing the warning icon and title.
+ */
+
+@Suppress("MaxLineLength")
+@Composable
+private fun HeaderContent(
+    modifier: Modifier = Modifier,
+) {
```
SaintPatrck, line 115: 🎨 If you scope this it will inherit the `modifier` and other context from its parent Column scope. That would mean you don't need a wrapping `Column` within `HeaderContent` to set the `modifier` and `horizontalAlignment`. It would look like...

```kotlin
fun NewDeviceNoticeEmailAccessContent(...) {
    Column(
        horizontalAlignment = Alignment.CenterHorizontally,
        modifier = modifier
            .standardHorizontalMargin()
            .fillMaxSize()
            .verticalScroll(state = rememberScrollState()),
    ) {
        Spacer(...)
        HeaderContent()
        // other components
    }
}

private fun ColumnScope.HeaderContent() {
    Image(...)
    Spacer(...)
    // other components
} 
```

This works because `HeaderContent` is using the same `modifier` and `horizontalAlignment` from the parent Column scope. It doesn't work quite as cleanly for `MainContent` because you're not using the same horizontal alignment as its parent Column scope.

Consider this more of a FYI than an actual change request since there's nothing technically wrong with the way you've done it.

andrebispo5: Oh I got it and thanks for the explanation, it makes sense specially the part that it's the same as the parent. I'll change it since it clears a lot of unnecessary code lines 👍

SaintPatrck (approved):

PR #4456, opened 2024-12-12
[PM-10515] CI build info on version copy

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10515

## 📔 Objective

Add more info output into our About view when an user presses the version copy button. This extra information will display more device information, like brand, model OS version and OS sdk. This should also contain build info like the branch and commit from the build as the build source URL.

Example:
```
© Bitwarden Inc. 2015-2024

Version: 2024.9.0 (1)
📱 google sdk_gphone64_arm64 🤖 14@34 📦 dev
🧱 commit: bitwarden/android/main@abc123
💻 build source: bitwarden/android/actions/runs/123/attempts/1
```

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/about/AboutViewModel.kt
```diff
@@ -92,7 +94,15 @@ class AboutViewModel @Inject constructor(
 
     private fun handleVersionClick() {
         clipboardManager.setText(
-            text = state.copyrightInfo.concat("\n\n".asText()).concat(state.version),
+            text = state.copyrightInfo.concat("\n\n".asText()).concat(state.version)
```
david-livefront, line 97: Can we update this to do full chaining.

```kotlin
text = state
    .copyrightInfo
    .concat("\n\n".asText())
    .concat(state.version)
    ...
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/about/utils/CIBuildInfo.kt
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * **WARNING:** This object is automatically generated by the CI scripts.
+ * Do not modify this file directly unless you are also updating the
+ * corresponding CI scripts that generate it.
+ */
+object CIBuildInfo {
+    /**
+     * Contains build information generated by the CI process.
+     */
+    val info: List<Pair<String, String>> = emptyList()
```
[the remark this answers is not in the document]
SaintPatrck: I second that sentiment. Unless there's data we can't access from the gradle script we should be putting build tasks there instead of standalone scripts.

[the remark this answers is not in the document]
vvolkgang: Yeah that's the purpose of it, passing data from the workflow. iOS was done first and this followed a similar approach which helps maintaining them but, I don't oppose a different route as long as we're able to easily change the content without app code changes, having a map / keyvalue pairs list instead of individual variables.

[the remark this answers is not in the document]
SaintPatrck: Even if we format the CI info in another location, there is more formatting taking place [here](https://github.com/bitwarden/android/pull/4456/files#diff-317dd91445d7e22f8ced693b2c1a3cd5dc6322fe7cd8ff2d28bb8dd1b878d1bcR126-R145), so it's likely app changes would be required regardless. IMO, it would be easier to maintain if all of the formatting for the copy was in one place. Some of the copy information is not available at compiletime, so formatting within the viewmodel at runtime seems like the most appropriate place.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/about/utils/CIBuildInfo.kt
SaintPatrck, line 1: What do you think about moving this to a `*.generated` package? `about/model/generated/CIBuildInfo.kt`, maybe?

Alternatively, if we write these values to a gradle properties file, like `ci.properties`, we can access them from all the projects in the repo and we can assign the formatted string to a `buildConfigField("String", "CI_INFO", "\"${getFormattedBuildInfo()}\")`. That approach would allow us to reference `BuildConfig.CI_INFO` and not have to worry about generating source files ourself.

If we want to get even more fancy and introduce type safety, we can define `CiBuildInfo` as a data class and declare it as the `buildConfigField` type. 

Example with both suggestions:
```kotlin

val ciProperties = Properties().apply { load(File(rootDir, "ci.properties") }

// Pre-formatted string
buildConfigField(
    type = "String",
    name = "CI_INFO",
    value = "\"${ciProperties.getOrDefault("ci.info", "local")}\"",
)

// Alternate with type safety
buildConfigField(
    type = "com.x8bit.bitwarden.ui.platform.feature.settings.about.utils.CIBuildInfo", 
    name = "CI_INFO", 
    value = "new com.x8bit.bitwarden.ui.platform.feature.settings.about.utils.CIBuildInfo(
        ${ciProperties.getOrDefault("ci.branch.name", "local")}, 
        ${ciProperties.getOrDefault("ci.foo.bar", "local")}",
    )
)
```

scripts/update_app_ci_build_info.sh
SaintPatrck, line 1: @vvolkgang and I were talking about moving this bash script into Fastlane's realm of responsibility. If we did the same with the`release-notes.sh` all of our build related operations would be under Fastlane's scope. That would help maintenance going forward. Thoughts?

vvolkgang: We'll migrate these to fastlane actions in a future pr!

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/about/utils/CIBuildInfo.kt
vvolkgang: We're going with the `ci.properties` approach @SaintPatrck suggested, build config field was added as part of https://github.com/bitwarden/android/pull/4471 and this PR will be updated to use that instead.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/about/AboutViewModel.kt
```diff
@@ -91,8 +92,35 @@ class AboutViewModel @Inject constructor(
...
+                .concat(state.version)
+                .concat("\n\n".asText())
+                .concat("$deviceBrandModel $osInfo $buildInfo".asText())
+                .let { text ->
+                    if (ciBuildInfoString.isEmpty()) {
+                        text
+                    } else {
+                        text.concat("\n".asText()).concat(ciBuildInfoString.asText())
+                    }
+                },
```
SaintPatrck, line 123: ⛏️ 🎨 We can use `takeUnless` here. Reads a little better, imo. Up to you though 😄 
```suggestion
                .concat( 
                    "\n$ciBuildInfoString"
                    .takeUnless { ciBuildInfoString.isEmpty() }
                    .orEmpty()
                    .asText()
                ),
```

ci.properties
SaintPatrck, line 1: I don't think we should track this file. "local" will be printed if it doesn't exist, thanks to [this](https://github.com/bitwarden/android/pull/4456/files#diff-8cff73265af19c059547b76aca8882cbaa3209291406f52df1dafbbc78e80c46L43-L45) check and [this](https://github.com/bitwarden/android/pull/4456/files#diff-8cff73265af19c059547b76aca8882cbaa3209291406f52df1dafbbc78e80c46R71) default value.

app/build.gradle.kts
```diff
@@ -68,7 +68,7 @@ android {
         buildConfigField(
             type = "String",
             name = "CI_INFO",
-            value = "\"${ciProperties.getOrDefault("ci.info", "local")}\""
+            value = "${ciProperties.getOrDefault("ci.info", "local")}"
```
SaintPatrck, line 71: The property value is being written with surrounding double quotes but we still need to make sure this default value has them.

```suggestion
            value = "${ciProperties.getOrDefault("ci.info", \""local\"")}"
```

SaintPatrck (changes requested):

SaintPatrck (approved):

PR #4460, opened 2024-12-12
PM-15976 - App crashes when non-english language user tries to create account

phil-livefront:
## 🎟️ Tracking

[PM-15976](https://bitwarden.atlassian.net/browse/PM-15976)

## 📔 Objective

- This PR addresses a crash caused by an IndexOutOfBoundsException in the `createClickableAnnotatedString` function. The crash occurred when attempting to create links for text highlights that were not present in the main string, which could happen in scenarios where translations were incomplete or mismatched.

We did have translations for `strTerms` and `strPrivacy` but not for `strTermsAndPrivacy` which was breaking out logic since `startIndex` came back as `-1` which caused the `java.lang.IndexOutOfBoundsException: setSpan (-1 ... 0) starts before 0`
```
val strTerms = stringResource(id = R.string.terms_of_service)
val strPrivacy = stringResource(id = R.string.privacy_policy)
val strTermsAndPrivacy = stringResource(
    id = R.string.by_continuing_you_agree_to_the_terms_of_service_and_privacy_policy,
)
```

## 📸 Screenshots

https://github.com/user-attachments/assets/b0c185bb-5a82-4c1a-8e47-d1fdf8ccb4de

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/Text.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/Text.kt
@@ -199,6 +199,12 @@ fun createClickableAnnotatedString(
                     mainString.lastIndexOf(text, ignoreCase = true)
                 }
             }
+
+            // Skip adding the link if the text to highlight is not found in the main string.
+            // This can happen if the highlighted text is correctly translated, but the main string
+            // is not yet translated, causing the startIndex to be -1.
+            if (startIndex < 0) { continue }
+
             val endIndex = startIndex + highlight.textToHighlight.length
             val link = LinkAnnotation.Clickable(
                 tag = highlight.textToHighlight,


app/src/main/java/com/x8bit/bitwarden/ui/platform/base/util/Text.kt
```diff
@@ -199,6 +199,12 @@ fun createClickableAnnotatedString(
                     mainString.lastIndexOf(text, ignoreCase = true)
                 }
             }
+
+            // Skip adding the link if the text to highlight is not found in the main string.
+            // This can happen if the highlighted text is correctly translated, but the main string
+            // is not yet translated, causing the startIndex to be -1.
+            if (startIndex < 0) { continue }
```
david-livefront, line 206: You don't need the curly braces

david-livefront (approved):

dseverns-livefront (changes requested):
So this will prevent the crash, but it breaks the functionality, I think what we may need to do is construct the final string from the string resources like `<string name = foo>Foo bar %1$s, baz cuh bruh, ong %2$s</string>` and in code do like :
```
val highlightOne = stringResource(R.string.boul)
val highlightTwo = stringResource(R.string.jawn)
val mainString = stringResource(R.string.foo, highlightOne, highlightTwo)
val annotated = createClickableAnnotatedString(
    mainString = mainString,
    highlights = listOf(...),
)
```

david-livefront:
@dseverns-livefront That would fix the issue and maintain the functionality, but mistakes can still occur and this function should not crash in those scenarios.

I think we should move forward with this solution and apply your solution in all the appropriate spots.

dseverns-livefront (approved):

PR #4483, opened 2024-12-17
[PM-12391] Respect PIN unlock setting during FIDO user verification

SaintPatrck:
## 🎟️ Tracking

PM-12391

## 📔 Objective

Check the `isUnlockWithPinEnabled` setting instead of checking the `vaultUnlockType` of the active account to determine if a PIN prompt should be shown.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -10,7 +10,6 @@ import com.x8bit.bitwarden.data.auth.repository.model.BreachCountResult
 import com.x8bit.bitwarden.data.auth.repository.model.UserState
 import com.x8bit.bitwarden.data.auth.repository.model.ValidatePasswordResult
 import com.x8bit.bitwarden.data.auth.repository.model.ValidatePinResult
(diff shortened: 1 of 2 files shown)


dseverns-livefront (approved):
This looks good for this fix and seems to be inline with how we are checking this elsewhere. I am wondering if we should update this in `AuthRepositoryImpl`
```kotlin
// from 
private fun getVaultUnlockType(
        userId: String,
    ): VaultUnlockType =
        when {
            authDiskSource.getPinProtectedUserKey(userId = userId) != null -> {
                VaultUnlockType.PIN
            }

            else -> {
                VaultUnlockType.MASTER_PASSWORD
            }
        }

// to 
private fun getVaultUnlockType(
        userId: String,
    ): VaultUnlockType =
        when {
            authDiskSource.getEncryptedPin(userId = userId) != null -> {
                VaultUnlockType.PIN
            }


            else -> {
                VaultUnlockType.MASTER_PASSWORD
            }
        }
``` 
since that seems to be the delta on why that value was not working when getting from the user state vs the settings repository. I am not sure if there is a specific reason for this difference though so may be safer to not touch it for this change.

SaintPatrck:
I like that suggestion. Like you say, we'll need to make sure it doesn't have any negative downstream effects. I'll create a ticket to revisit this for a broader improvement later and we can go forward with this fix in the interim.

PR #4512, opened 2025-01-01
[PM-15969] Users with `Can Edit` access cannot assign collections

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15969

## 📔 Objective

Fix bug where users with Can Edit permission in one collection could not use "assign to collection".

## 📸 Screenshots

https://github.com/user-attachments/assets/9c9b98de-f5f3-4fdc-a294-f16d27de27ec

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/util/CollectionViewExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/util/CollectionViewExtensions.kt
@@ -110,13 +110,8 @@ fun List<CollectionView>?.hasDeletePermissionInAtLeastOneCollection(
 /**
  * Checks if the user has permission to assign an item to a collection.
  *
- * Assigning to a collection is not allowed when the item is in a collection that the user does not
- * have "manage" permission for and is also in a collection they cannot view the passwords in.
- *
- * E.g., If an item is in A collection with "view except passwords" or "edit except passwords"
- * permission and in another with "manage" permission, the user **cannot** assign the item to other
- * collections. Conversely, if an item is in a collection with "manage" permission and another with
- * "view" or "edit" permission, the user **can** assign the item to other collections.
+ * Assigning to a collection is only allowed when the item is in a collection that the user does
+ * have "manage" or "edit" permission.
  */
 fun List<CollectionView>?.canAssignToCollections(currentCollectionIds: List<String>?): Boolean {
     if (this.isNullOrEmpty()) return true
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/util/CollectionViewExtensions.kt
```diff
@@ -126,17 +121,9 @@ fun List<CollectionView>?.canAssignToCollections(currentCollectionIds: List<Stri
...
-        // permission in any collection the item is not in.
-        this
-            .none {
-                currentCollectionIds.contains(it.id) &&
-                    (it.permission == CollectionPermission.EDIT_EXCEPT_PASSWORD ||
-                        it.permission == CollectionPermission.VIEW_EXCEPT_PASSWORDS)
-            }
+                (it.permission == CollectionPermission.MANAGE ||
+                    it.permission == CollectionPermission.EDIT)
+        }
```
fedemkr, line 126: ⛏️ I can't review over [line 120](https://github.com/bitwarden/android/pull/4512/files#diff-846fdecb131a468c99d7593706091841f57a3e82056281272a5bd72308038a66R120) but shouldn't that comment be updated to include `EDIT` permission as well?

fedemkr (dismissed):
LGTM, just an old comment improvement ⛏️

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/util/CollectionViewExtensions.kt
andrebispo5: You are right! While I was overviewing the PR changes, I didn't saw the comment on line 120 as well, thanks!

dseverns-livefront: do we still need the `none` check? I have been trying to puzzle this out for a bit, what we are saying there is that that if the `CollectionView` exists in any of the collections where it is restricted by those hidden password values then we don't want to allow that item to be assigned to another collection by the user. I guess I am looking for clarity on the business rule since the `CollectionView.permission` is something that was added specifically to capture these cases. I can't really figure out the edge case where this would happen, but would we potentially allow a user to supercede a restriction on an item put in place by an organization?

andrebispo5: @dseverns-livefront the change was made based on new requirements where higher permissions override lower ones. And if you think about it, it doesn't make sense. If you can manage an item in one collection you have access, doesn't make sense that in the other collection you don't.

dseverns-livefront: > change was made based on new requirements where higher permissions override lower  

That about solves it for me then, if the requirements changed. And yeah that would definitely be a super (strange) edge case. Thanks for the added context @andrebispo5 !

PR #4574, opened 2025-01-16
[PM-13626] Remember last opened view for 5 minutes

aj-rosado:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-13626

## 📔 Objective

After lock, for the next 5 minutes, the user should be redirected to the previous screen if they were:

- Send
- Verification Code
- Generator
- Search

To achieve this, a new manager was added to control the behaviour and saving the last screen on `SettingsDiskSource` (should this be on a different disk source?). The navigation is provided by deeplink using the existing SpecialCircumstance.

The current screen is always saved `onResume` and cleared `onStop` (to ensure we don't navigate to previous screens)
Before clearing, checking if the user is not locked, does the trick to not clear in case of lock when the app was on background

A different task will add more screens

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
```diff
@@ -502,6 +504,17 @@ class AuthDiskSourceImpl(
         )
     }
 
+    override fun getLastLockTimestamp(userId: String): Long {
+        return getLong(key = LAST_LOCK_TIMESTAMP.appendIdentifier(userId)) ?: Long.MIN_VALUE
```
david-livefront, line 508: This should just return a nullable value, the disk source should not have any idea what the default values are

app/src/main/java/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
```diff
@@ -502,6 +504,17 @@ class AuthDiskSourceImpl(
...
+    }
+
+    override fun storeLastLockTimestamp(userId: String, lastLockTimestamp: Instant) {
```
david-livefront, line 511: We should allow a nullable value so the data can be cleared.

david-livefront: We should also be clearing this value when `clearData` is called.

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
```diff
@@ -482,6 +485,13 @@ class SettingsDiskSourceImpl(
         )
     }
 
+    override fun storeAppResumeScreen(userId: String, screenData: AppResumeScreenData?) {
+        putString(RESUME_SCREEN.appendIdentifier(userId), json.encodeToString(screenData))
```
david-livefront, line 489: This needs to be able to handle null better:

```kotlin
putString(
    key = RESUME_SCREEN.appendIdentifier(userId),
    value = screenData?.let { json.encodeToString(it) },
)
```

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    override fun getResumeScreen(): AppResumeScreenData? {
+        return authRepository.activeUserId?.let { userId ->
+            settingsDiskSource.getAppResumeScreen(userId)
+        }
+    }
+
+    override fun getResumeSpecialCircumstance(): SpecialCircumstance? {
+        val timeNowMinus5Min = Instant.now().minusSeconds(UNLOCK_NAVIGATION_TIME)
```
david-livefront, line 41: Please use the clock to retrieve the time.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+        val lastLockTimestamp = Instant.ofEpochMilli(
+            authDiskSource.getLastLockTimestamp(
+                userId = authRepository.activeUserId ?: "",
```
david-livefront, line 44: Shouldn't we be checking if the user ID is null before this?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+        }
+        return when (val resumeScreenData = getResumeScreen()) {
+            AppResumeScreenData.GeneratorScreen -> SpecialCircumstance.GeneratorShortcut
+            AppResumeScreenData.SendScreen -> SpecialCircumstance.SendShortcut
+            is AppResumeScreenData.SearchScreen -> SpecialCircumstance.SearchShortcut(
+                resumeScreenData.searchTerm,
+            )
+
+            AppResumeScreenData.VerificationCodeScreen ->
+                SpecialCircumstance.VerificationCodeShortcut
```
david-livefront, line 58: This should have brackets around it

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+
+            else -> null
+        }
+    }
+
+    @Suppress("MagicNumber")
```
david-livefront, line 64: Where is the magic number?

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+    override fun clearResumeScreen() {
+        if (vaultLockManager.isVaultUnlocked(userId = authRepository.activeUserId ?: "")) {
```
david-livefront, line 66: Do not just pass in an empty string, check if the id is null first.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/di/PlatformManagerModule.kt
```diff
@@ -329,4 +332,20 @@ object PlatformManagerModule {
...
+    @Singleton
+    fun provideAppResumeManager(
+        settingsDiskSource: SettingsDiskSource,
+        authDiskSource: AuthDiskSource,
+        authRepository: AuthRepository,
+        vaultLockManager: VaultLockManager,
+    ): AppResumeManager {
+        return AppResumeManagerImpl(
+            settingsDiskSource,
+            authDiskSource,
```
david-livefront, line 346: Can we add some names here

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -268,6 +269,10 @@ class VaultLockManagerImpl(
         )
         if (!wasVaultLocked) {
             mutableVaultStateEventSharedFlow.tryEmit(VaultStateEvent.Locked(userId = userId))
+            authDiskSource.storeLastLockTimestamp(
+                userId = userId,
+                lastLockTimestamp = Instant.now(),
```
david-livefront, line 274: Please use the clock

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -80,16 +83,23 @@ class SearchViewModel @Inject constructor(
...
     specialCircumstanceManager: SpecialCircumstanceManager,
+    private val appResumeManager: AppResumeManager,
 ) : BaseViewModel<SearchState, SearchEvent, SearchAction>(
     // We load the state from the savedStateHandle for testing purposes.
     initialState = savedStateHandle[KEY_STATE]
         ?: run {
             val searchType = SearchArgs(savedStateHandle).type
             val userState = requireNotNull(authRepo.userStateFlow.value)
             val specialCircumstance = specialCircumstanceManager.specialCircumstance
+            var searchTerm: String = ""
```
david-livefront, line 94: Can we make this a val

```kotlin
val searchTerm = if (specialCircumstance is SpecialCircumstance.SearchShortcut) {
    specialCircumstance
        .searchTerm
        .also { specialCircumstanceManager.specialCircumstance = null }
} else {
    ""
}
```

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -718,6 +730,16 @@ class SearchViewModel @Inject constructor(
             .data
             ?.cipherViewList
             ?.firstOrNull { it.id == cipherId }
+
+    private fun handleOnResumed() {
+        appResumeManager.setResumeScreen(
+            AppResumeScreenData.SearchScreen(searchTerm = state.searchTerm),
```
david-livefront, line 736: Why do we do this?

Can't we just let the viewModel store the searchTerm?

david-livefront: Nevermind, I see what is going ho here

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchScreen.kt
```diff
@@ -143,6 +147,23 @@ fun SearchScreen(
...
+@Composable
+private fun LivecycleEventEffect(viewModel: SearchViewModel) {
+    LivecycleEventEffect { _, event ->
+        when (event) {
+            Lifecycle.Event.ON_RESUME -> {
+                viewModel.trySendAction(SearchAction.LifecycleResume)
+            }
+
+            Lifecycle.Event.ON_STOP -> {
+                viewModel.trySendAction(SearchAction.LifecyclePause)
```
david-livefront, line 159: I think is going to be confusing since there is a paused lifecycle state, can we can this event `LifecycleStop`.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -718,6 +730,16 @@ class SearchViewModel @Inject constructor(
...
+        )
+    }
+
+    private fun handleOnPaused() {
+        appResumeManager.clearResumeScreen()
```
david-livefront, line 741: How does this work?

When the screen stops, this will be cleared and there will nothing to restore?

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/send/SendScreen.kt
```diff
@@ -64,6 +66,17 @@ fun SendScreen(
             { viewModel.trySendAction(SendAction.RefreshPull) }
         },
     )
+    LivecycleEventEffect { _, event ->
+        when (event) {
+            Lifecycle.Event.ON_RESUME -> {
+                viewModel.trySendAction(SendAction.LifecycleResume)
+            }
+            Lifecycle.Event.ON_STOP -> {
```
david-livefront, line 74: Instead of listening to these lifecycle events, can we just do this in the VM?

Use `init` instead of `resume` and use `onCleared` instead of `stop`.

app/src/test/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceTest.kt
```diff
@@ -1239,4 +1242,33 @@ class SettingsDiskSourceTest {
...
+    @Test
+    fun `storeAppResumeScreen should update SharedPreferences`() {
+        val mockUserId = "mockUserId"
+        val resumeScreenKey = "bwPreferencesStorage:resumeScreen_$mockUserId"
+        val expectedData = AppResumeScreenData.GeneratorScreen
+        settingsDiskSource.storeAppResumeScreen(mockUserId, expectedData)
+        assertEquals(
+            expectedData,
+            fakeSharedPreferences.getString(resumeScreenKey, "")?.let
+            {
```
david-livefront, line 1269: This curly brace should be up a line

app/src/test/java/com/x8bit/bitwarden/data/platform/datasource/disk/util/FakeSettingsDiskSource.kt
```diff
@@ -380,6 +384,14 @@ class FakeSettingsDiskSource : SettingsDiskSource {
         createSendActionCount = count
     }
 
+    override fun storeAppResumeScreen(userId: String, screenData: AppResumeScreenData?) {
+        storedAppResumeScreenData[userId] = Json.encodeToString(screenData)
```
david-livefront, line 388: Make sure to handle null:

```kotlin
storedAppResumeScreenData[userId] = screenData?.let { Json.encodeToString(it) }
```

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
aj-rosado: There is no magic number anymore, this should have been removed

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
aj-rosado: When the app is backgrounded this is called but even if the vault is locked, it will run the onResume of current screen before moving to the vault locked screen

Then the manager deals with the logic to know if this should be cleared or not.

`if (vaultLockManager.isVaultUnlocked(userId = userId)) ` if vault is locked, it won't clear the `AppResumeScreen`

It surprised me to work this way and at first had a different implementation to deal with it (the MagicNumber existed for that reason).

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/send/SendScreen.kt
aj-rosado: Both `init` and `OnCleared` are not always called when moving into and out the screen, navigating from `Send` to Vault screen does not call this and we need to clear it. After the first time we go into the `Send`, the init is not called

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
david-livefront: I see, so the `isVaultUnlocked` is what makes this all work 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -80,16 +83,24 @@ class SearchViewModel @Inject constructor(
...
     specialCircumstanceManager: SpecialCircumstanceManager,
+    private val appResumeManager: AppResumeManager,
 ) : BaseViewModel<SearchState, SearchEvent, SearchAction>(
     // We load the state from the savedStateHandle for testing purposes.
     initialState = savedStateHandle[KEY_STATE]
         ?: run {
             val searchType = SearchArgs(savedStateHandle).type
             val userState = requireNotNull(authRepo.userStateFlow.value)
             val specialCircumstance = specialCircumstanceManager.specialCircumstance
+            val searchTerm = if (specialCircumstance is SpecialCircumstance.SearchShortcut) {
```
dseverns-livefront, line 94: this could also be: 
``` kotlin
(specialCircumstance as? SpecialCircumstance.SearchShortcut)
                ?.searchTerm
                ?.also {
                    specialCircumstanceManager.specialCircumstance = null
                }
                .orEmpty()

Comment (dseverns-livefront) [inline app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt:94]: this could also be: 
``` kotlin
(specialCircumstance as? SpecialCircumstance.SearchShortcut)
                ?.searchTerm
                ?.also {
                    specialCircumstanceManager.specialCircumstance = null
                }
                .orEmpty()

File: app/src/main/java/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSource.kt [hunk 19]
```diff
@@ -352,4 +353,14 @@ interface AuthDiskSource {
      * Stores the new device notice state for the given [userId].
      */
     fun storeNewDeviceNoticeState(userId: String, newState: NewDeviceNoticeState?)
+
+    /**
+     * Gets the last lock timestamp for the given [userId].
+     */
+    fun getLastLockTimestamp(userId: String): Long?
```
david-livefront, line 360: These should return the same types, either both longs or both Instants. I kinda prefer `Instant` but I'll leave it up to you

app/src/main/java/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
```diff
@@ -502,6 +504,17 @@ class AuthDiskSourceImpl(
         )
     }
 
+    override fun getLastLockTimestamp(userId: String): Long? {
+        return getLong(key = LAST_LOCK_TIMESTAMP.appendIdentifier(userId))
+    }
+
+    override fun storeLastLockTimestamp(userId: String, lastLockTimestamp: Instant?) {
```
david-livefront, line 511: This should be cleared in `clearData`

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,78 @@
...
+class AppResumeManagerImpl(
+    private val settingsDiskSource: SettingsDiskSource,
+    private val authDiskSource: AuthDiskSource,
+    private val authRepository: AuthRepository,
+    private val vaultLockManager: VaultLockManager,
+    private val clock: Clock,
+) : AppResumeManager {
+    private companion object {
+        // 5 minutes
+        private const val UNLOCK_NAVIGATION_TIME: Long = 5 * 60
```
david-livefront, line 24: Can we add `_MS` to the end of this to indicate units.

Also, this can be a top level property of the file.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    override fun getResumeSpecialCircumstance(): SpecialCircumstance? {
+        val userId = authRepository.activeUserId ?: return null
+        val timeNowMinus5Min = clock.instant().minusSeconds(UNLOCK_NAVIGATION_TIME)
+        val lastLockTimestamp = Instant.ofEpochMilli(
+            authDiskSource.getLastLockTimestamp(
+                userId = userId,
+            ) ?: Long.MIN_VALUE,
+        )
```
david-livefront, line 49: We don't need to handle a null timestamp at all:
```kotlin
val lastLockTimestamp = authDiskSource
    .getLastLockTimestamp(userId = userId)
    ?.let { Instant.ofEpochMilli(it) }
    ?: return
```

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    override fun getResumeSpecialCircumstance(): SpecialCircumstance? {
+        val timeNowMinus5Min = Instant.now().minusSeconds(UNLOCK_NAVIGATION_TIME)
+        val lastLockTimestamp = Instant.ofEpochMilli(
+            authDiskSource.getLastLockTimestamp(
+                userId = authRepository.activeUserId ?: "",
+            ),
+        )
+        if (timeNowMinus5Min.isAfter(lastLockTimestamp)) {
+            return null
```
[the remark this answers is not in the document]
david-livefront: I know I said to do this but I think we need to actually call `storeAppResumeScreen` directly:

```kotlin
 settingsDiskSource.storeAppResumeScreen(userId = userId, screenData = null)
```

This way we can be sure the value is nulled out even if the vault is locked.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    override fun clearResumeScreen() {
+        val userId = authRepository.activeUserId ?: return
+        if (vaultLockManager.isVaultUnlocked(userId = userId)) {
+                settingsDiskSource.storeAppResumeScreen(
+                    userId = userId,
+                    screenData = null,
+                )
```
david-livefront, line 75: Please format this file

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -268,6 +270,10 @@ class VaultLockManagerImpl(
         )
         if (!wasVaultLocked) {
             mutableVaultStateEventSharedFlow.tryEmit(VaultStateEvent.Locked(userId = userId))
+            authDiskSource.storeLastLockTimestamp(
+                userId = userId,
+                lastLockTimestamp = clock.instant(),
```
david-livefront, line 275: 👍

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -142,6 +153,8 @@ class SearchViewModel @Inject constructor(
             is SearchAction.SearchTermChange -> handleSearchTermChange(action)
             is SearchAction.VaultFilterSelect -> handleVaultFilterSelect(action)
             is SearchAction.OverflowOptionClick -> handleOverflowItemClick(action)
+            is SearchAction.LifecycleResume -> handleOnResumed()
+            is SearchAction.LifecycleStop -> handleOnPaused()
```
david-livefront, line 157: The function is still called `handleOnPaused` even though you are stopping. Can we rename this

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -1064,6 +1087,16 @@ sealed class SearchAction {
         val overflowAction: ListingItemOverflowAction,
     ) : SearchAction()
 
+    /**
+     * Indicates the UI has been entered a resumed lifecycle state.
+     */
+    data object LifecycleResume : SearchAction()
+
+    /**
+     * Indicates the UI has been entered a paused lifecycle state.
```
david-livefront, line 1096: This doc refers to being paused instead of stopped

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -110,7 +110,9 @@ fun GeneratorScreen(
             Lifecycle.Event.ON_RESUME -> {
                 viewModel.trySendAction(GeneratorAction.LifecycleResume)
             }
-
+            Lifecycle.Event.ON_STOP -> {
+                viewModel.trySendAction(GeneratorAction.LifecycleStop)
+            }
```
david-livefront, line 115: Wanna run the formatter here, there should be some spaces.

I would recommend adding a macro for this. There are instructions [here](https://github.com/bitwarden/android?tab=readme-ov-file#setup).

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorViewModel.kt
```diff
@@ -2139,6 +2150,11 @@ sealed class GeneratorAction {
      */
     data object LifecycleResume : GeneratorAction()
 
+    /**
+     * Indicates the UI has been entered a resumed lifecycle state.
```
david-livefront, line 2154: This refers to resume instead of stop

app/src/test/java/com/x8bit/bitwarden/data/auth/datasource/disk/util/FakeAuthDiskSource.kt
```diff
@@ -314,13 +317,21 @@ class FakeAuthDiskSource : AuthDiskSource {
         return storedNewDeviceNoticeState[userId] ?: NewDeviceNoticeState(
             displayStatus = NewDeviceNoticeDisplayStatus.HAS_NOT_SEEN,
             lastSeenDate = null,
-            )
+        )
```
david-livefront, line 320: 👍

app/src/test/java/com/x8bit/bitwarden/ui/platform/feature/rootnav/RootNavViewModelTest.kt
```diff
@@ -53,6 +53,31 @@ class RootNavViewModelTest : BaseViewModelTest() {
             dispatcherManager = FakeDispatcherManager(),
         )
 
+    private val mockVaultUnlockedUserState = UserState(
```
david-livefront, line 56: Can we just make this a constant at the bottom of the file.

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
aj-rosado: Formatter is not making any changes on this file

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -7,7 +7,9 @@ import com.x8bit.bitwarden.data.platform.manager.model.AppResumeScreenData
 import com.x8bit.bitwarden.data.platform.manager.model.SpecialCircumstance
 import com.x8bit.bitwarden.data.vault.manager.VaultLockManager
 import java.time.Clock
-import java.time.Instant
+
+// 5 minutes
+private const val UNLOCK_NAVIGATION_TIME_MS: Long = 5 * 60
```
dseverns-livefront, line 12: ⛏️ I appreciate the comment here, but the `_MS` immediately makes me think `millis`, `UNLOCK_NAVIGATION_TIME_MINUTES` can never go wrong being UBER explicit.

david-livefront: I think that is my bad, I told him to ad `_MS` because I thought it was milliseconds.

It should be `UNLOCK_NAVIGATION_TIME_SECONDS` since the value is in seconds

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
david-livefront: Hmm, are you sure you have the correct formatter configured. Because this when I run the formatter it does this:

```kotlin
    LivecycleEventEffect { _, event ->
        when (event) {
            Lifecycle.Event.ON_RESUME -> {
                viewModel.trySendAction(GeneratorAction.LifecycleResume)
            }

            Lifecycle.Event.ON_STOP -> {
                viewModel.trySendAction(GeneratorAction.LifecycleStop)
            }

            else -> Unit
        }
    }
```

aj-rosado:
Changed quite a bit the implementation in order to reduce the duplicated code and simplify the implementation. 
Thank you @dseverns-livefront for the suggestions and help 👍

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorViewModel.kt
```diff
@@ -77,6 +78,7 @@ class GeneratorViewModel @Inject constructor(
     private val authRepository: AuthRepository,
     private val policyManager: PolicyManager,
     private val reviewPromptManager: ReviewPromptManager,
+    private val appResumeManager: AppResumeManager,
```
dseverns-livefront, line 81: we can remove this now!

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+
+import com.x8bit.bitwarden.data.auth.datasource.disk.AuthDiskSource
+import com.x8bit.bitwarden.data.auth.repository.AuthRepository
+import com.x8bit.bitwarden.data.platform.datasource.disk.SettingsDiskSource
+import com.x8bit.bitwarden.data.platform.manager.model.AppResumeScreenData
+import com.x8bit.bitwarden.data.platform.manager.model.SpecialCircumstance
+import com.x8bit.bitwarden.data.vault.manager.VaultLockManager
+import java.time.Clock
+
+// 5 minutes
```
david-livefront, line 11: This comment should not be needed. Or you could you make it a javadoc.

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+            settingsDiskSource.getAppResumeScreen(userId)
+        }
+    }
+
+    override fun getResumeSpecialCircumstance(): SpecialCircumstance? {
+        val userId = authRepository.activeUserId ?: return null
+        val timeNowMinus5Min = clock.instant().minusSeconds(UNLOCK_NAVIGATION_TIME_SECONDS)
+        val lastLockTimestamp = authDiskSource.getLastLockTimestamp(
+            userId = userId,
+        ) ?: return null
```
david-livefront, line 45: formatting:

```kotlin
       val lastLockTimestamp = authDiskSource
           .getLastLockTimestamp(userId = userId)
           ?: return null
```

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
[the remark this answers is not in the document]
david-livefront: Thanks for that!

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -208,6 +211,21 @@ class VaultViewModel @Inject constructor(
     }
 
     private fun handleLifecycleResumed() {
+        when (specialCircumstanceManager.specialCircumstance) {
+            is SpecialCircumstance.SearchShortcut -> {
+                sendEvent(VaultEvent.NavigateToVaultSearchScreen)
```
david-livefront, line 216: Should this one also clear the Special Circumstance?

app/src/test/java/com/x8bit/bitwarden/data/auth/datasource/disk/util/FakeAuthDiskSource.kt
```diff
@@ -471,6 +482,13 @@ class FakeAuthDiskSource : AuthDiskSource {
         assertEquals(policies, storedPolicies[userId])
     }
 
+    /**
+     * Assert that the [lastLockTimestamp] was stored successfully using the [userId].
+     */
+    fun assertNotNullLastLockTimestamp(userId: String) {
```
david-livefront, line 488: Shouldn't we pass in a value to compare against?

app/src/test/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceTest.kt
```diff
@@ -1253,4 +1256,52 @@ class SettingsDiskSourceTest {
...
+            },
+        )
+    }
+
+    @Test
+    fun `storeAppResumeScreen should save null when passed`() {
+        val mockUserId = "mockUserId"
+        val resumeScreenKey = "bwPreferencesStorage:resumeScreen_$mockUserId"
+        val expectedData = AppResumeScreenData.GeneratorScreen
+            settingsDiskSource.storeAppResumeScreen(mockUserId, expectedData)
```
david-livefront, line 1293: Wanna run the formatter here

app/src/test/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerTest.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `getResumeScreen should return null when there is no app resume screen saved`() =
+        runTest {
+            val actualValue = appResumeManager.getResumeScreen()
+            assertNull(actualValue)
+        }
+
+    @Suppress("MaxLineLength")
```
david-livefront, line 65: Can we remove these suppressions on the tests that do not need tham

app/src/test/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerTest.kt
```diff
@@ -0,0 +1,172 @@
...
+    private val vaultLockManager: VaultLockManager = mockk {
+        every { isVaultUnlocked(USER_ID) } returns true
+    }
+
+    private val fixedClock: Clock = Clock.fixed(
+        Instant.parse("2023-10-27T12:00:00Z"),
+        ZoneOffset.UTC,
+    )
+
+    private val authDiskSource: AuthDiskSource = mockk {
```
david-livefront, line 34: Can we use the fake instead

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
aj-rosado:
```
val lastLockTimestamp = authDiskSource
            .getLastLockTimestamp(
                userId = userId
            ) ?: return null
```

            
If the ")" is on the bottom line, formatter will format it like above

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
aj-rosado: SpecialCircumstance keeps the `searchTerm` value and will be cleared on the `SearchViewModel`

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/AppResumeManagerImpl.kt
david-livefront: The `?:` should be on its own line for proper chaining.

This would also be fine:

```kotlin
val lastLockTimestamp = authDiskSource
    .getLastLockTimestamp(
        userId = userId
    )
    ?: return null
```

app/src/test/java/com/x8bit/bitwarden/data/auth/datasource/disk/util/FakeAuthDiskSource.kt
```diff
@@ -471,6 +481,13 @@ class FakeAuthDiskSource : AuthDiskSource {
         assertEquals(policies, storedPolicies[userId])
     }
 
+    /**
+     * Assert that the [lastLockTimestamp] was stored successfully using the [userId].
+     */
+    fun assertNotNullLastLockTimestamp(userId: String, expectedValue: Instant?) {
```
david-livefront, line 487: This name no longer makes sense, can we go with `assertLastLockTimestamp`

app/src/test/java/com/x8bit/bitwarden/ui/tools/feature/send/SendScreenTest.kt
```diff
@@ -60,19 +63,22 @@ class SendScreenTest : BaseComposeTest() {
...
-            SendScreen(
-                viewModel = viewModel,
-                onNavigateToAddSend = { onNavigateToNewSendCalled = true },
-                onNavigateToEditSend = { onNavigateToEditSendId = it },
-                onNavigateToSendFilesList = { onNavigateToSendFilesListCalled = true },
-                onNavigateToSendTextList = { onNavigateToSendTextListCalled = true },
-                onNavigateToSearchSend = { onNavigateToSendSearchCalled = true },
-                intentManager = intentManager,
-            )
+            CompositionLocalProvider(LocalAppResumeStateManager provides appResumeStateManager) {
```
david-livefront, line 71: This should be injected into `SendScreen`

aj-rosado: `AppResumeStateManager` is present on the `LocalManagerProvider` that is injected on the `MainActivity`, and is used by the `RegisterScreenDataOnLifecycleEffect` that is injected on the `SendScreen`

When injecting `RegisterScreenDataOnLifecycleEffect` on the test class, `AppResumeStateManager` was returning null. That's why `CompositionLocalProvider` is called on the test class.
Should it still be called on the `SendScreen` instead?

dseverns-livefront: Hmm, so for the other composition locals we are passing them in the screen composables as parameter which makes the testing easier since we can just pass the mock in directly. With the Effect here its baked in already. One thing we could do that I would be in favor of is adding this to the base compose test like 
```
protected fun ComposeContentTestRule.setContentWithLocalManagers(
        localAppResumeStateManager: AppResumeStateManager = mockk<AppReviewStateManager>(relaxed = true),
        /** etc. **/
        testContent: @Composable () -> Unit
    ) {
        setContent { 
            CompositionLocalProvider(
                LocalAppResumeStateManager provides localAppResumeStateManager
                /** etc. **/
            ) { 
                testContent()
            }
        }
    }

Comment (david-livefront) [reply app/src/test/java/com/x8bit/bitwarden/ui/tools/feature/send/SendScreenTest.kt:71]: The preferred way is to inject it into the screen. We already do this in multiple places.

```kotlin
fun SendScreen(
    onNavigateToAddSend: () -> Unit,
    onNavigateToEditSend: (sendItemId: String) -> Unit,
    onNavigateToSendFilesList: () -> Unit,
    onNavigateToSendTextList: () -> Unit,
    onNavigateToSearchSend: (searchType: SearchType.Sends) -> Unit,
    viewModel: SendViewModel = hiltViewModel(),
    intentManager: IntentManager = LocalIntentManager.current,
    appResumeStateManager: AppResumeStateManager = LocalAppResumeStateManager.current, <-- Here
) {

    ...

    RegisterScreenDataOnLifecycleEffect(appResumeStateManager = appResumeStateManager) {
        AppResumeScreenData.SendScreen
    }
```


```kotlin
@Composable
fun RegisterScreenDataOnLifecycleEffect(
    appResumeStateManager: AppResumeStateManager = LocalAppResumeStateManager.current, <-- Here
    appResumeStateProvider: () -> AppResumeScreenData,
) {
    val appResumeStateManager = LocalAppResumeStateManager.current
    LivecycleEventEffect { _, event ->
        when (event) {
            Lifecycle.Event.ON_RESUME -> {
                appResumeStateManager.updateScreenData(data = appResumeStateProvider())
            }

            Lifecycle.Event.ON_STOP -> {
                appResumeStateManager.updateScreenData(data = null)
            }

            else -> Unit
        }
    }
}
```

dseverns-livefront: @aj-rosado lets go with David's suggestion since that follows the patterns already established.

dseverns-livefront (approved):
Thanks for making the changes and keeping up with it!

david-livefront (approved):
Thanks for making those updates!

PR #4605, opened 2025-01-21
PM-17404: Set app delegate on theme change

david-livefront:
## 🎟️ Tracking

[PM-17404](https://bitwarden.atlassian.net/browse/PM-17404)

## 📔 Objective

This PR sets the theme on the `AppCompatDelegate` whenever the user changes the theme in app. This ensure that all image resources are updated properly.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-17404]: https://bitwarden.atlassian.net/browse/PM-17404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/MainActivity.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/MainActivity.kt
@@ -66,13 +66,14 @@ class MainActivity : AppCompatActivity() {
             )
         }
 
-        // Within the app the language will change dynamically and will be managed
-        // by the OS, but we need to ensure we properly set the language when
-        // upgrading from older versions that handle this differently.
+        // Within the app the language and theme will change dynamically and will be managed by the
+        // OS, but we need to ensure we properly set the values when upgrading from older versions
+        // that handle this differently or when the activity restarts.
         settingsRepository.appLanguage.localeName?.let { localeName ->
             val localeList = LocaleListCompat.forLanguageTags(localeName)
             AppCompatDelegate.setApplicationLocales(localeList)
         }
+        AppCompatDelegate.setDefaultNightMode(settingsRepository.appTheme.osValue)
         setContent {
             val state by mainViewModel.stateFlow.collectAsStateWithLifecycle()
             val navController = rememberNavController()
@@ -93,6 +94,16 @@ class MainActivity : AppCompatActivity() {
                                 Toast.LENGTH_SHORT,
                             )
                             .show()
+                    }
+
+                    is MainEvent.UpdateAppLocale -> {
+                        AppCompatDelegate.setApplicationLocales(
+                            LocaleListCompat.forLanguageTags(event.localeName),
+                        )
+                    }
+
+                    is MainEvent.UpdateAppTheme -> {
+                        AppCompatDelegate.setDefaultNightMode(event.osTheme)
                     }
                 }
             }
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
@@ -10,7 +10,6 @@ import com.x8bit.bitwarden.ui.platform.feature.settings.appearance.model.AppThem
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.onSubscription
-import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import java.time.Instant
 
@@ -50,6 +49,7 @@ class SettingsDiskSourceImpl(
     private val json: Json,
 ) : BaseDiskSource(sharedPreferences = sharedPreferences),
     SettingsDiskSource {
+    private val mutableAppLanguageFlow = bufferedMutableSharedFlow<AppLanguage?>(replay = 1)
     private val mutableAppThemeFlow = bufferedMutableSharedFlow<AppTheme>(replay = 1)
 
     private val mutableLastSyncFlowMap = mutableMapOf<String, MutableSharedFlow<Instant?>>()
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -73,6 +73,7 @@ class MainActivity : AppCompatActivity() {
             val localeList = LocaleListCompat.forLanguageTags(localeName)
             AppCompatDelegate.setApplicationLocales(localeList)
         }
+        AppCompatDelegate.setDefaultNightMode(settingsRepository.appTheme.osValue)
```
phil-livefront, line 76: 🎨 mainly just a suggestion and not going to hold up the PR but should we put this logic within the `SettingsRepository` and then call a function there to change the `AppCompatDelegate` instead of putting it directly in the `MainActivity`?

phil-livefront (dismissed):

app/src/main/java/com/x8bit/bitwarden/MainActivity.kt
dseverns-livefront: Typically we want to avoid adding this type of thing as a dependency to something that may outlive it. This is a weird one since its a static piece but ultimately it seems to be tied to the activity anyway.

>An Activity can only be linked with one AppCompatDelegate instance, therefore the instance returned from create(Activity, AppCompatCallback) should be retained until the Activity is destroyed.

phil-livefront: yeah makes total sense to me.  My only response back is the `SettingsRepository` is a singleton and the `MainActivity` won't outlive it.   I only brought this up since the `appTheme` lives within the settings repo so containing all changes to that within one place allows a single source of control vs having changes be made in two separate places.  

```
   override val appThemeStateFlow: StateFlow<AppTheme>
        get() = settingsDiskSource
            .appThemeFlow
            .stateIn(
                scope = unconfinedScope,
                started = SharingStarted.Eagerly,
                initialValue = settingsDiskSource.appTheme,
            )
```

dseverns-livefront: > MainActivity won't outlive it and vice versa due to how the architecture is set up
Exactly why we don't want to have any reference in the singleton scoped, since that would be a potential leak. Unique situation here is when the AutoFill/Accessibility service(s) are running that same instance of the @Singleton scoped classes will remain with the process, where as the activity could be destroyed and recreated several times. Honestly in terms of real risk of that here, I may be being overly cautious since this is a static call and we are not actually passing in a reference to the activity, but my personal feeling is keep anything relating to any Android Lifecycle as isolated to that space as possible.

phil-livefront: It looks like we already observing the app theme changes in the `MainViewModel` and we are using this to update the `BitwardenTheme` via compose state.  Could we also add an Event to also handle these changes within the `MainActivity` and then have a single spot where these changes are occurring?

dseverns-livefront: We could do this for the `locales` as well.

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/appearance/AppearanceScreen.kt
```diff
@@ -57,6 +59,15 @@ fun AppearanceScreen(
...
         when (event) {
             AppearanceEvent.NavigateBack -> onNavigateBack.invoke()
+            is AppearanceEvent.UpdateAppLocales -> {
+                AppCompatDelegate.setApplicationLocales(
+                    LocaleListCompat.forLanguageTags(event.localeName),
+                )
+            }
+
+            is AppearanceEvent.UpdateAppTheme -> {
+                AppCompatDelegate.setDefaultNightMode(event.osTheme)
```
phil-livefront, line 69: couldn't this be done in the `MainActivity` with the current flow that's set up in `MainViewModel` but an Event there?  

```
settingsRepository
            .appThemeStateFlow
            .onEach { trySendAction(MainAction.Internal.ThemeUpdate(it)) }
            .launchIn(viewModelScope)
```

phil-livefront: Then we could remove the call `AppCompatDelegate.setDefaultNightMode(settingsRepository.appTheme.osValue)` that happens on every `onCreate` and only call it when things need to be change but it'll still be respected on every launch.

app/src/main/java/com/x8bit/bitwarden/MainActivity.kt
david-livefront: @dseverns-livefront & @phil-livefront 
I went ahead and added the the events to the `AppearanceViewModel` I did not migrate the logic into the `MainActivity` thought. What does everyone think about it?

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/appearance/AppearanceScreen.kt
david-livefront: Hmm, I do believe you are correct

david-livefront: We still need it to happen in `onCreate` to ensure that if the activity is recreated, the delegate is updated.

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
```diff
@@ -10,7 +10,6 @@ import com.x8bit.bitwarden.ui.platform.feature.settings.appearance.model.AppThem
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.onSubscription
-import kotlinx.serialization.encodeToString
```
dseverns-livefront, line 13: LOL there are like 10 PRs up trying to get this line removed rn.

phil-livefront (approved):

david-livefront:
Thanks @dseverns-livefront & @phil-livefront

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
david-livefront: I think I may beat them all

PR #4607, opened 2025-01-21
PM-16621 add explore generator card to be able to trigger coach marks

dseverns-livefront:
## 🎟️ Tracking
[PM-16621](https://bitwarden.atlassian.net/browse/PM-16621)

## 📔 Objective
- Show the action card when necessary
- Once card has been interacted with store that value and don't show the card anymore.

## 📸 Screenshots

https://github.com/user-attachments/assets/a141736f-83a0-4a02-b600-9b438cd7f8fb

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-16621]: https://bitwarden.atlassian.net/browse/PM-16621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSource.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSource.kt
@@ -356,4 +356,19 @@ interface SettingsDiskSource {
      * Stores the given [count] completed create send actions for the device.
      */
     fun storeCreateSendActionCount(count: Int?)
+
+    /**
+     * Gets the Boolean value of if the Generator CoachMark tour has been interacted with.
+     */
+    fun getHasSeenGeneratorCoachMark(): Boolean?
+
+    /**
+     * Stores a value for if the Generator CoachMark tour has been interacted with
+     */
+    fun storeHasSeenGeneratorCoachMark(hasSeen: Boolean?)
+
+    /**
+     * Returns an [Flow] to observe updates to the "HasSeenGeneratorCoachMark" value.
+     */
+    fun getHasSeenGeneratorCoachMarkFlow(): Flow<Boolean?>
 }
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
@@ -10,7 +10,6 @@ import com.x8bit.bitwarden.ui.platform.feature.settings.appearance.model.AppThem
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.onSubscription
-import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import java.time.Instant
 
@@ -40,6 +39,7 @@ private const val IS_VAULT_REGISTERED_FOR_EXPORT = "isVaultRegisteredForExport"
 private const val ADD_ACTION_COUNT = "addActionCount"
 private const val COPY_ACTION_COUNT = "copyActionCount"
 private const val CREATE_ACTION_COUNT = "createActionCount"
+private const val HAS_SEEN_GENERATOR_COACH_MARK = "hasSeenGeneratorCoachMark"
 
 /**
  * Primary implementation of [SettingsDiskSource].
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FirstTimeActionManagerImpl.kt
```diff
@@ -154,6 +154,17 @@ class FirstTimeActionManagerImpl @Inject constructor(
             }
             .distinctUntilChanged()
 
+    override val hasSeenGeneratorCoachMarkFlow: Flow<Boolean>
+        get() = settingsDiskSource
+            .getHasSeenGeneratorCoachMarkFlow()
+            .map { it ?: false }
```
phil-livefront, line 160: 🎨 Just curious why the Boolean is nullable for `fun getHasSeenGeneratorCoachMark(): Boolean?` when we are mapping it to `false`.  Could we just set it as false by default and avoid the nullable case?

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorViewModel.kt
```diff
@@ -746,6 +774,14 @@ class GeneratorViewModel @Inject constructor(
         }
     }
 
+    private fun handleHasSeenCoachMarkValueChange(
+        action: GeneratorAction.Internal.HasSeenGeneratorCoachMarkValueChangeReceive,
+    ) {
+        mutableStateFlow.update {
+            it.copy(hasSeenExploreGeneratorCard = action.hasSeenCoachMark)
```
phil-livefront, line 781: 🎨 Instead of passing a Boolean value to directly what if we send an indication to toggle its state downstream? The way is set up now seems to allow us to potentially send `false` again and re-trigger the tour for the user. If we protect the toggling it would ensure that the value is only toggled once at the disk level, preventing unintended multiple updates.

phil-livefront: This should have been pinned on line 226 on `FirstTimeActionManagerImpl` 

```
override fun hasSeenGeneratorCoachMarkTour() {
        settingsDiskSource.storeHasSeenGeneratorCoachMark(hasSeen = true)
    }
```

dseverns-livefront: @david-livefront How would you feel about moving this into the disk source itself to really ensure that once it is set, it can't be changed? Also moving the defaulting of the value to the disksource  `null ?: false` since we aren't reseting this value with the clear data.

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
```diff
@@ -482,6 +485,22 @@ class SettingsDiskSourceImpl(
         )
     }
 
+    override fun getHasSeenGeneratorCoachMark(): Boolean? =
```
david-livefront, line 488: This could be a `val`

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FirstTimeActionManagerImpl.kt
```diff
@@ -154,6 +154,17 @@ class FirstTimeActionManagerImpl @Inject constructor(
...
+            .combine(
+                featureFlagManager.getFeatureFlagFlow(FlagKey.OnboardingFlow),
```
david-livefront, line 162: Nice

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSource.kt
```diff
@@ -356,4 +356,19 @@ interface SettingsDiskSource {
      * Stores the given [count] completed create send actions for the device.
      */
     fun storeCreateSendActionCount(count: Int?)
+
+    /**
+     * Gets the Boolean value of if the Generator CoachMark tour has been interacted with.
+     */
+    fun getHasSeenGeneratorCoachMark(): Boolean?
```
phil-livefront, line 363: 🎨 Since we are using this logic on other screens, should we make an enum for the screen and have one function in the disk source that accepts the enum type instead of making funs for each individual screen?

phil-livefront: might save some time for other screens if they get added but also I can see both sides so its just an idea vs a change request kinda thing.  Go birds

app/src/main/java/com/x8bit/bitwarden/data/platform/manager/FirstTimeActionManagerImpl.kt
phil-livefront: Is that an architecture design we made or is this android specific?

PR #4486, opened 2024-12-17
#4416 PM-15037 mTLS

rohm1:
## 🎟️ Tracking

https://github.com/bitwarden/android/issues/4416

## 📔 Objective

This PR adds a "Client certificate" menu to the custom environment menu. By clicking on a "Choose" button, the native certificate selection menu will open. User can choose a certificate that will be used for connections to a self-hosted instance.

Other apps implement some dynamic logic: if the sever requires mTLS, the user is prompted to select a certificate. While this behavior is more flexible for the user, it is more complicated to implement. With the chosen approach, the app settings must be cleared and a new server must be set up. As users generally do not enable/disable mTLS on a regular basis, I think the tradeoff is acceptable. Additionally the only information that is saved is the certificate name (key alias). If a user update/renew a certificate, as long as the certificate name does not change, the new certificate will be used and there is no need to reset the app. This issue could also be worked around by adding a certificate selection menu to the settings.

## 📸 Screenshots

Screenshot (Dec 17, 2024 21_10_48)

## Manual tests I did

- Upgrade from main does not crash and old settings are still saved/available/used
- It is still possible to connect to server without mTLS
- A certificate can be selected and saved in the settings. Certificate is used for connections to the server
- Close the app and restart: Certificate setting has been persisted and connection is still using mTLS

Tested on Android 15 (Pixel 6) and Android 14 (Samsung Galaxy Note 20 Ultra 5G)

## 🦮 Reviewer guidelines

- 📝 I am not an Android dev and this is my first time with Kotlin. While I have tried to follow the architecture, I'm sure I missed some things and and others are not in the right place. I apologize in advance and look forward how to improve the code
- ❓ The Android APIs use the _key alias_ wording [(KeyChain.choosePrivateKeyAlias),](https://developer.android.com/reference/android/security/KeyChain#choosePrivateKeyAlias(android.app.Activity,%20android.security.KeyChainAliasCallback,%20java.lang.String[],%20java.security.Principal[],%20java.lang.String,%20int,%20java.lang.String)) and this is also the wording I chose in the code. For the UI I chose to go with _Client certificate_. While this is generally bad to have two wordings for the same thing, _key alias_ cannot be used in the UI because confusing for the user, and _user certificate_ in code is kind of weird because that would be something else that the Android wording. Open for suggestions
- 🌱 Following the current approach, we can add a Certificate selection screen to the settings. This way it is not necessary anymore to reset the app to choose a new certificate.

PR #4608, opened 2025-01-21
[PM-17424] Implement KeyManager for handling private keys

SaintPatrck:
## 🎟️ Tracking

PM-17424

Relates to #4606, #4486

## 📔 Objective

Implement a new manager, `KeyManager`, for managing key-related data.
This includes:
- Importing a mutual TLS certificate into the AndroidKeyStore.
- Removing a mutual TLS certificate from the AndroidKeyStore.
- Retrieving a mutual TLS certificate from the AndroidKeyStore or system KeyChain.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/ImportPrivateKeyResult.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/ImportPrivateKeyResult.kt
@@ -1,0 +1,45 @@
+package com.x8bit.bitwarden.data.platform.datasource.disk.model
+
+/**
+ * Models the result of importing a private key.
+ */
+sealed class ImportPrivateKeyResult {
+
+    /**
+     * Represents a successful result of importing a private key.
+     *
+     * @property alias The alias assigned to the imported private key.
+     */
+    data class Success(val alias: String) : ImportPrivateKeyResult()
+
+    /**
+     * Represents a generic error during the import process.
+     */
+    sealed class Error : ImportPrivateKeyResult() {
+
+        /**
+         * Indicates that the provided key is unrecoverable or the password is incorrect.
+         */
+        data object UnrecoverableKey : Error()
+
+        /**
+         * Indicates that the certificate chain associated with the key is invalid.
+         */
+        data object InvalidCertificateChain : Error()
+
+        /**
+         * Indicates that the specified alias is already in use.
+         */
+        data object DuplicateAlias : Error()
+
+        /**
+         * Indicates that an error occurred during the key store operation.
+         */
+        data object KeyStoreOperationFailed : Error()
+
+        /**
+         * Indicates the provided key is not supported.
+         */
+        data object UnsupportedKey : Error()
+    }
+}
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/MutualTlsCertificate.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/MutualTlsCertificate.kt
@@ -1,0 +1,36 @@
+package com.x8bit.bitwarden.data.platform.datasource.disk.model
+
+import java.security.PrivateKey
+import java.security.cert.X509Certificate
+
+/**
+ * Represents a mutual TLS certificate.
+ */
+data class MutualTlsCertificate(
+    val alias: String,
+    val privateKey: PrivateKey,
+    val certificateChain: List<X509Certificate>,
+) {
+    /**
+     * Leaf certificate of the chain.
+     */
+    val leafCertificate: X509Certificate?
+        get() = certificateChain.lastOrNull()
+
+    /**
+     * Root certificate of the chain.
+     */
+    val rootCertificate: X509Certificate?
+        get() = certificateChain.firstOrNull()
+
+    override fun toString(): String = leafCertificate
+        ?.let {
+            buildString {
+                appendLine("Subject: ${it.subjectDN}")
+                appendLine("Issuer: ${it.issuerDN}")
+                appendLine("Valid From: ${it.notBefore}")
+                appendLine("Valid Until: ${it.notAfter}")
+            }
+        }
+        ?: ""
+}
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/MutualTlsKeyHost.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/model/MutualTlsKeyHost.kt
@@ -1,0 +1,16 @@
+package com.x8bit.bitwarden.data.platform.datasource.disk.model
+
+/**
+ * Location of the key data.
+ */
+enum class MutualTlsKeyHost {
+    /**
+     * Key is stored in the system key chain.
+     */
+    KEY_CHAIN,
+
+    /**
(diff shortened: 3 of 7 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
```diff
@@ -0,0 +1,37 @@
...
+    sharedPreferences: SharedPreferences,
+) : BaseEncryptedDiskSource(
+    sharedPreferences = sharedPreferences,
+    encryptedSharedPreferences = encryptedPreferences,
+), KeyDiskSource {
+
+    override fun storeMutualTlsKey(
+        userId: String,
+        alias: String?,
+        host: MutualTlsKeyHost?,
```
david-livefront, line 23: We don't actually use the host here?

SaintPatrck: It will be used by the KeyRepo (coming soon) to determine where the key is stored; system KeyChain or AndroidKeyStore.

Alternatively, we could make `KeyDiskSource` responsible for delegating between KeyChain and KeyStore. That would mean passing `Context` down to a `DiskSource`, which I don't see happening anywhere. We don't really pass `Context` to repositories either. 🤔 

I plan on invoking `KeyChain.choosePrivateKeyAlias(activity)` from a `KeyChainManager` available to Composables. Another option is to move the other `KeyChain` calls that only require `Context` to this `KeyChainManager`. Downside to that approach is the number of Actions and Events that will have to be wired up.

david-livefront: I don't see an inherent issue with letting the disk source determine whether it needs to got to the key chain or shared preferences, since they are bot `disk` (sorta). Similar to how the other disk sources determine between encrypted and unencrypted preferences.

What I was mainly pointing out was that we do not actually store the `host` value

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+        return ImportPrivateKeyResult.Success(alias)
+    }
+
+    override fun removeMutualTlsKey(
+        alias: String,
+        host: MutualTlsKeyHost,
+    ): RemoveMutualTlsKeyResult =
+        when (host) {
+            MutualTlsKeyHost.ANDROID_KEY_STORE -> removeKeyFromAndroidKeyStore(alias)
+            MutualTlsKeyHost.KEY_CHAIN -> RemoveMutualTlsKeyResult.Success
```
dseverns-livefront, line 108: ❓ I am assuming that this would be for a cert manually added to the devices through the security settings, which we would not be able to remove. If that is correct I am guessing there is some handling up stream like maybe redirecting them to the settings or something? I think that this makes sense to have `MutualTlsKeyHost` if there are new ones we need to support in the future. I think my question is what are we using this result for and would it make sense to have a like a `NotMyProblem` result type?

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+    ): ImportPrivateKeyResult {
+        // Step 1: Load PKCS12 bytes into a KeyStore.
+        val pkcs12KeyStore: KeyStore = key
+            .inputStream()
+            .use { stream ->
+                try {
+                    KeyStore.getInstance(KEYSTORE_TYPE_PKCS12)
+                        .also { it.load(stream, password.toCharArray()) }
+                } catch (e: KeyStoreException) {
+                    Timber.e(e, "Failed to load PKCS12 bytes")
```
david-livefront, line 42: I assume we are OK logging these failures?

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+import java.security.PrivateKey
+import java.security.UnrecoverableKeyException
+import java.security.cert.Certificate
+import java.security.cert.CertificateException
+import java.security.cert.X509Certificate
+
+/**
+ * Default implementation of [KeyDiskSource].
+ */
+class KeyDiskSourceImpl(
```
david-livefront, line 24: I have no real issue with the logic but the name of this class uses the DiskSource moniker, and it feels like this is doing a lot of logic outside the scope of that. Should this be considered a manager or repo?

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
SaintPatrck: Correct. `KEY_CHAIN` represents keys installed via system settings and we cannot remove them from within the app.

Now that you ask, I don't think we need this result. At least not right now. At the moment, this will only ever be called when the user logs out, so we wouldn't want to stop that process if removal of the key fails. It may be better to ignore any errors that occur. 🤔 

If/when we add proper key management UI within the app, linking to the system settings for system certificates may be something we do, and it would also create a need for the result.

I'll remove the result for now. We can add it back later if it's needed.

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
SaintPatrck: I was originally going to make KeyRepository responsible for interacting with KeyStore and KeyChain and limiting KeyDiskSource responsibility to interacting with SharedPrefs. After taking another look, I realized we don't need to store anything in shared prefs, so we're only interacting with KeyStore and KeyChain. I moved those interactions here and left this a `DiskSource` because both KeyChain and KeyStore are (_sorta_) forms of disk persistence like you stated earlier. I'm not strongly attached to `DiskSource` if you feel like `Manager` is more appropriate. I honestly see valid arguments for both. 

`Manager` may be more appropriate simply because it doesn't implement BaseDiskSource and is not dependent on SharedPrefs, while all the other `DiskSource`'s are. 🤔

david-livefront: I would be fine with `manager`. Though I will say that the `VaultDiskSource` does not implement the `BaseDiskSource` since it is interacting with `Room`.

app/src/main/java/com/x8bit/bitwarden/data/platform/datasource/disk/KeyDiskSourceImpl.kt
david-livefront: No reservation just double checking

david-livefront (approved):

SaintPatrck:
Thanks @david-livefront and @dseverns-livefront

PR #4610, opened 2025-01-22
PM-15804, PM-17130: Add logic to monitor when the screen on state to ensure the vault locks properly

david-livefront:
## 🎟️ Tracking

[PM-17130](https://bitwarden.atlassian.net/browse/PM-17130)
[PM-15804](https://bitwarden.atlassian.net/browse/PM-15804)

## 📔 Objective

This PR adds a broadcast receiver to the `VaultLockManager` in order to ensure that we lock the vault at the soonest possible moment.

There seems to be an issue in the way `delay` works inside a coroutine scope when the screen is off. The `delay` will take significantly longer to complete and cause the vault to not lock when it is supposed to. This new solution restarts the delay with an appropriate remaining time when the screen is turned back on to ensure we lock the vault correctly.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-17130]: https://bitwarden.atlassian.net/browse/PM-17130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-15804]: https://bitwarden.atlassian.net/browse/PM-15804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
@@ -1,5 +1,9 @@
 package com.x8bit.bitwarden.data.vault.manager
 
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.content.IntentFilter
 import com.bitwarden.core.InitOrgCryptoRequest
 import com.bitwarden.core.InitUserCryptoMethod
 import com.bitwarden.core.InitUserCryptoRequest
@@ -50,6 +54,8 @@ import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import java.time.Clock
+import java.util.concurrent.ConcurrentHashMap
 import kotlin.time.Duration.Companion.minutes
 
 /**
@@ -62,6 +68,7 @@ private const val MAXIMUM_INVALID_UNLOCK_ATTEMPTS = 5
  */
 @Suppress("TooManyFunctions", "LongParameterList")
 class VaultLockManagerImpl(
+    private val clock: Clock,
     private val authDiskSource: AuthDiskSource,
     private val authSdkSource: AuthSdkSource,
     private val vaultSdkSource: VaultSdkSource,
@@ -70,13 +77,15 @@ class VaultLockManagerImpl(
     private val userLogoutManager: UserLogoutManager,
     private val trustedDeviceManager: TrustedDeviceManager,
     dispatcherManager: DispatcherManager,
+    context: Context,
 ) : VaultLockManager {
     private val unconfinedScope = CoroutineScope(dispatcherManager.unconfined)
 
     /**
-     * This [Map] tracks all active timeout [Job]s that are running using the user ID as the key.
+     * This [Map] tracks all active timeout [Job]s that are running and their associated data using
+     * the user ID as the key.
      */
-    private val userIdTimerJobMap = mutableMapOf<String, Job>()
+    private val userIdTimerJobMap: MutableMap<String, TimeoutJobData> = ConcurrentHashMap()
 
     private val activeUserId: String? get() = authDiskSource.userState?.activeUserId
 
@@ -96,6 +105,10 @@ class VaultLockManagerImpl(
         observeUserSwitchingChanges()
         observeVaultTimeoutChanges()
         observeUserLogoutResults()
+        context.registerReceiver(
+            ScreenStateBroadcastReceiver(),
+            IntentFilter(Intent.ACTION_SCREEN_ON),
+        )
     }
 
     override fun isVaultUnlocked(userId: String): Boolean =
@@ -363,7 +376,7 @@ class VaultLockManagerImpl(
 
     private fun handleOnForeground() {
         val userId = activeUserId ?: return
-        userIdTimerJobMap[userId]?.cancel()
+        userIdTimerJobMap.remove(key = userId)?.job?.cancel()
     }
 
     private fun observeUserSwitchingChanges() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -601,6 +620,36 @@ class VaultLockManagerImpl(
...
+     * when the screen is off. We do not cancel the job when the screen is off, this allows the
+     * job to complete as-soon-as possible if the screen is powered off for an extended period.
+     */
+    private inner class ScreenStateBroadcastReceiver : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent) {
+            userIdTimerJobMap.map { (userId, data) ->
+                handleTimeoutActionWithDelay(
+                    userId = userId,
+                    vaultTimeoutAction = data.vaultTimeoutAction,
+                    delayInMs = data.durationMs - (clock.millis() - data.startTimeMs),
```
david-livefront, line 637: This will start a new timer with an appropriate amount of time remaining. If the timer should have expired already, then this number will be negative and the delay should return immediately.

dseverns-livefront: or we could add `(clock.millis() - data.startTimeMs).coerceAtLeast(0L)` ?

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -363,7 +376,7 @@ class VaultLockManagerImpl(
 
     private fun handleOnForeground() {
         val userId = activeUserId ?: return
-        userIdTimerJobMap[userId]?.cancel()
+        userIdTimerJobMap.remove(key = userId)?.job?.cancel()
```
dseverns-livefront, line 379: This is what would prevent the user from being locked out while using the app if they, lock the screen then come back in an unlocked state correct? I am just wondering if there is a potential race with this or response to the broadcast receiver.

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
david-livefront: Testing a Broadcast Receiver is always a little wonky. I can give it a whirl though.

As for the `coerceAtLeast(0L)`, is the concern that it could have a _longer_ duration if the start time is after the current time?

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -601,6 +620,36 @@ class VaultLockManagerImpl(
...
+    /**
+     * A custom [BroadcastReceiver] that listens for when the screen is powered on and restarts the
+     * vault timeout jobs to ensure they complete at the correct time.
+     *
+     * This is necessary because the [delay] function in a coroutine will no keep accurate time
+     * when the screen is off. We do not cancel the job when the screen is off, this allows the
+     * job to complete as-soon-as possible if the screen is powered off for an extended period.
+     */
+    private inner class ScreenStateBroadcastReceiver : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent) {
```
brian-livefront, line 632: Doesn't really matter here but I _think_ technically this `context` can be non-null.

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -601,6 +620,36 @@ class VaultLockManagerImpl(
...
+            userIdTimerJobMap.map { (userId, data) ->
+                handleTimeoutActionWithDelay(
```
brian-livefront, line 634: 👍

app/src/test/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerTest.kt
```diff
@@ -98,6 +108,13 @@ class VaultLockManagerTest {
         dispatcherManager = fakeDispatcherManager,
     )
 
+    @AfterEach
+    fun tearDown() {
+        verify(exactly = 1) {
+            context.registerReceiver(any(), any())
+        }
```
brian-livefront, line 115: Do we have other tests that do verifications in a teardown method?

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
david-livefront: The screen should be turned on before the app is foregrounded, I am unsure if there is a scenario where this can happen. 🤔

app/src/test/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerTest.kt
david-livefront: Not sure but I can move this into it's own test.

phil-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
dseverns-livefront: yeah, which I feel is unlikely, and I guess we still could test for negative case for the function overall.

dseverns-livefront (approved):

david-livefront:
Thanks @phil-livefront & @dseverns-livefront

PR #4613, opened 2025-01-23
PM-16622 PM-16623 and PM-16624 Add the first three coach marks to the generator tour

dseverns-livefront:
## 🎟️ Tracking
[PM-16622](https://bitwarden.atlassian.net/browse/PM-16622)
[PM-16623](https://bitwarden.atlassian.net/browse/PM-16623)
[PM-16624](https://bitwarden.atlassian.net/browse/PM-16624)

## 📔 Objective
- Show the first three coach marks on the segmented controls on the generator screen.
- NOTE: I did resolve with Emily that it's fine if the "highlight" effect for these is a per option implementation.

## 📸 Screenshots

https://github.com/user-attachments/assets/0e309bbb-e4b9-47de-b9d2-a2879576d8b5

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-16622]: https://bitwarden.atlassian.net/browse/PM-16622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-16623]: https://bitwarden.atlassian.net/browse/PM-16623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-16624]: https://bitwarden.atlassian.net/browse/PM-16624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -443,7 +513,79 @@ private fun MainStateOptionsItem(
         modifier = modifier
             .fillMaxWidth()
             .testTag(tag = "GeneratorTypePicker"),
-    )
+    ) { option ->
+        when (option.testTag) {
+            "password_option" -> {
+                CoachMarkHighlight(
+                    key = ExploreGeneratorCoachMark.PASSWORD_MODE,
+                    title = stringResource(R.string.coachmark_1_of_6),
```
david-livefront, line 521: It is probably still to long but can we open this up a bit:

```kotlin
description = stringResource(
    id = R.string.use_the_generator_to_create_secure_passwords_passphrases_and_usernames,
),
```

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -443,7 +513,79 @@ private fun MainStateOptionsItem(
...
+                ) {
+                    SegmentedButtonOptionContent(option = option)
+                }
+            }
+
+            "passphrase_option" -> {
+                CoachMarkHighlight(
+                    key = ExploreGeneratorCoachMark.PASSPHRASE_MODE,
+                    title = stringResource(R.string.coachmark_2_of_6),
+                    description = stringResource(R.string.passphrases_are_strong_passwords_that_are_often_easier_to_remember_and_type_than_random_passwords),
```
david-livefront, line 540: Ditto

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -443,7 +513,79 @@ private fun MainStateOptionsItem(
...
+                ) {
+                    SegmentedButtonOptionContent(option = option)
+                }
+            }
+
+            "username_option" -> {
+                CoachMarkHighlight(
+                    key = ExploreGeneratorCoachMark.USERNAME_MODE,
+                    title = stringResource(R.string.coachmark_3_of_6),
+                    description = stringResource(R.string.unique_usernames_add_an_extra_layer_of_security_and_can_help_prevent_hackers_from_finding_your_accounts),
```
david-livefront, line 564: ditto ditto

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -443,7 +513,79 @@ private fun MainStateOptionsItem(
         modifier = modifier
             .fillMaxWidth()
             .testTag(tag = "GeneratorTypePicker"),
-    )
+    ) { option ->
+        when (option.testTag) {
```
david-livefront, line 517: Could we do this by index or something. We should not rely on test tags for this.

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
```diff
@@ -101,6 +116,8 @@ fun GeneratorScreen(
     viewModel: GeneratorViewModel = hiltViewModel(),
     onNavigateToPasswordHistory: () -> Unit,
     onNavigateBack: () -> Unit,
+    onDimNavBarRequest: (Boolean) -> Unit,
+    scrimClickCount: State<Int>,
```
david-livefront, line 120: Can you tell me a little more about this `scrimClickCount`?

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
dseverns-livefront: yeah, we will still need the suppression but can def do this!

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
dseverns-livefront: Within the `CoachMarkContainer` we handle any clicks outside the tooltips and then essential re-show the tooltip for the current coachmark (default behaviour dismissed this via google locking it down). Since the bottom bar is outside the container we need to capture if the user clicks in that space otherwise it will dismiss the tool tip, I am using this value in the launched effect to call the function if this state updates.

app/src/main/java/com/x8bit/bitwarden/ui/platform/components/coachmark/model/CoachMarkHighlightShape.kt
```diff
@@ -1,16 +1,23 @@
...
+sealed class CoachMarkHighlightShape {
     /**
-     * A square-shaped highlight.
+     * A rounded rectangle shape which has a radius to round the corners by.
+     *
+     * @property radius the radius to use to round the corners of the rectangle shape.
+     * Defaults to [ROUNDED_RECT_DEFAULT_RADIUS]
      */
-    SQUARE,
+    data class RoundedRectangle(
```
dseverns-livefront, line 15: image

Better support items that use `CircleShape` which result in the "pill" look.

app/src/main/java/com/x8bit/bitwarden/ui/platform/components/segment/BitwardenSegmentedButton.kt
```diff
@@ -58,29 +68,39 @@ fun BitwardenSegmentedButton(
...
-                            text = option.text,
-                            style = BitwardenTheme.typography.labelLarge,
-                        )
-                    },
-                    icon = {
-                        // No icon required
-                    },
-                    modifier = Modifier.semantics { option.testTag?.let { testTag = it } },
-                )
+                optionContent(index, option)
```
phil-livefront, line 71: clean.

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
phil-livefront: I may be just missing something but whats the benefit of keeping a count tho?

dseverns-livefront: @phil-livefront I just chose an incrementing count cause that seemed like a simple thing to keep modifying which is easy to observer changes to, If you look at the launched effect, it will run each time that value changes, again incrementing a number just made sense here. The reason I did this vs passing a lambda back to execute is that could leak a reference if the generator screen leaves the composition and but the bottom nav is still there. I am open to something different.

app/src/main/java/com/x8bit/bitwarden/ui/platform/components/segment/BitwardenSegmentedButton.kt
```diff
@@ -32,13 +33,22 @@ import kotlinx.collections.immutable.ImmutableList
  * @param modifier Modifier.
  * @param windowInsets The insets to be applied to this composable.
  */
+@Suppress("MaxLineLength")
```
david-livefront, line 36: Is this still needed?

app/src/main/java/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreen.kt
phil-livefront: Ah i see! make sense! One idea comes to mind, feel free to ignore but could we use a `Flow` here instead?

```
In the VM
private val _eventFlow = MutableSharedFlow<Unit>()
val eventFlow = _eventFlow.asSharedFlow()

in the Screen
val eventFlow = viewModel.eventFlow

LaunchedEffect(Unit) {
    eventFlow.collectLatest {
       ...
    }
}
```

david-livefront: Since this scrim is being handled outside, couldn't we just make it a no-op? Do we need to force a re-show here or would it still dismiss?

dseverns-livefront: it would still dismiss, would not add this level of ick if not needed :)

david-livefront: Why does that cause a dismiss?

app/src/main/res/values/strings.xml
```diff
@@ -1117,10 +1117,10 @@ Do you want to switch to this account?</string>
   <string name="login_credentials">Login Credentials</string>
   <string name="autofill_options">Autofill Options</string>
   <string name="use_this_button_to_generate_a_new_unique_password">Use this button to generate a new unique password.</string>
-  <string name="coachmark_1_of_3">1 of 3</string>
-  <string name="coachmark_2_of_3">2 of 3</string>
+  <string name="coachmark_1_of_3">1 OF 3</string>
```
phil-livefront, line 1120: 🔨 could this be dependent on an `int` state vs hardcoded?  Like we pass in the current step and then pass in the total steps?  Feel free to ignore.

phil-livefront (approved):

app/src/main/res/values/strings.xml
dseverns-livefront: I think if the UI was built in a way where it was tied to the list of something and we were going dynamic then fo sho. With this I think the juice ain't worth the squeeze.

phil-livefront: makes sense!  I was more thinking if product wants to add another coach mark to an existing flow, instead of having to add any more strings (also more translations) for a specific step we can use one string key and it'll be dynamic but again, just be spit bawllllin' here

PR #4721, opened 2025-02-13
[PM-18261] Added default value for linkedIdType to get my vault loading

cseufert:
## 🎟️ Tracking

This address helpdesk issue #587728

## 📔 Objective

This patch allows me to load my vault in the >= 2024.12 versions of the Android Mobile App.

This is a POC/Workaround to solve my issue of the vault not loading, however I have no idea how my vault has over 100 records with linkedType = 0.

## 📸 Screenshots
image

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


SaintPatrck:
Hi @cseufert
Thank you for investigating and submitting this change. Unfortunately we won't be able to blindly fallback to `LOGIN_PASSWORD` as it could lead to other issues if the linked field is not part of a LOGIN cipher or was not originally meant to be linked to the PASSWORD field.

We're brainstorming approaches to resolve and remediate the issue you're encountering. In the meantime, it would be helpful if you could provide more details so that we can track down the root cause.

1. Are you able to identify what client these erroneous records were created with?
2. What method of creation was used for these records? I.e., import, auto-fill creation, manual creation, etc..
3. If you create a new item, or add a new custom Linked field to an existing item, from various clients do you notice if those items are assigned a valid value.
4. Were you expecting those records to have custom Linked fields in the first place?
5. Are all of the instances of invalid linked field types in your vault supposed to be LOGIN_PASSWORD?
6. Have you tried linking other fields in different item types? Are they assigned valid values?

cseufert:
I agree this is not the correct way to fix this, however without this my vault or anyone I have shared logins with can load the mobile app since December 2024.

I do not know where these records came from but outside of using the mobile app and browser extension I have used the following:
- We did import our database from LastPass around 3 years ago
- I used the CLI tooling to generate a bunch of logins at least 2 years ago

I suspect using the cli tool is the culprit, however i no longer have the script used to do that.

I have been unable drag a link between the data returned under ciphers in the sync api, and the items in my vault.

How would I decrypt one of the linked fields to see an example of the values, is there a tool / cli command to do this?

SaintPatrck:
@cseufert now that your vault is loading you will likely see custom Linked fields in some items that were not visible in the previous release. Any items with these Linked custom fields that you did not explicitly create are the ones causing issues because the identifier for the field they are linked to does not exist. If you go through your vault and remove the linked custom fields it will also remove the unexpected data that is the root cause of the crash.

You may want to consider exporting your vault to a json file and reviewing the custom field's linked field types. Only values defined in `LinkedIdTypeJson` should be used, and they should correspond with the cipher type (e.g., Login, SecureNote, Card, Identity, etc..).

We have a section in our help docs on [using linked custom fields](https://bitwarden.com/help/auto-fill-custom-fields/#using-linked-custom-fields) if you're not familiar with them already.

If you don't mind, I'd like to push a few commits to this PR that will introduce a proper default fallback value. We can use that to gracefully handle unexpected values without making assumptions.

cseufert:
@SaintPatrck this seems like a great solution, I have tried exporting the vault to JSON via the web interface (vault.bitwarden.com) however none of the LinkedId: 0 values are in that export, I wonder if the JS SDK ignores these silently. I cant seem to get the cli tool to install, some problem installing NPM packages that I have not had time to get to the bottom of yet to see if that has the same issue.

app/src/test/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModelTest.kt
```diff
@@ -1626,6 +1626,48 @@ class VaultAddEditViewModelTest : BaseViewModelTest() {
...
+        runTest {
+            val stateWithName = createVaultAddItemState(
+                vaultAddEditType = VaultAddEditType.AddItem(VaultItemCipherType.LOGIN),
+                commonContentViewState = createCommonContentViewState(
+                    name = "mockName-1",
+                    customFieldData = listOf(
+                        VaultAddEditState.Custom.LinkedField(
+                            "mockItemId",
+                            "mockCustomFieldName",
+                            VaultLinkedFieldType.UNKNOWN,
```
david-livefront, line 1640: Can we add some names to these params

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
```diff
@@ -363,6 +363,21 @@ class VaultAddEditViewModel @Inject constructor(
...
+        // proceeding.
+        content.common.customFieldData
+            .filterIsInstance<VaultAddEditState.Custom.LinkedField>()
+            .firstOrNull { it.vaultLinkedFieldType == VaultLinkedFieldType.UNKNOWN }
+            ?.let {
+                showErrorDialog(
+                    title = R.string.invalid_custom_field_link.asText(),
+                    message = R.string.custom_field_x_has_an_invalid_link.asText(it.name),
+                )
+                return@onContent
```
david-livefront, line 378: Do we know how this UI reacts to an unknown type in the list, does the user have a way to alter or remove these unknown items at this time?

SaintPatrck: Yes. I added screenshots to the PR description that show how it looks. The `SELECT_TEXT` is shown in the drop-down when viewing and editing if the current value is unknown. In edit more, when the field is tapped the user is presented with valid link options for that cipher type. If they attempt to save without correcting the issue this dialog is shown.

image

image

app/src/main/java/com/x8bit/bitwarden/ui/vault/model/VaultLinkedFieldType.kt
```diff
@@ -43,6 +44,7 @@ enum class VaultLinkedFieldType(
     FIRST_NAME(id = 416.toUInt(), label = R.string.first_name.asText()),
     LAST_NAME(id = 417.toUInt(), label = R.string.last_name.asText()),
     FULL_NAME(id = 418.toUInt(), label = R.string.full_name.asText()),
+    UNKNOWN(id = 0.toUInt(), label = SELECT_TEXT),
```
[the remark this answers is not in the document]
SaintPatrck: It's not a resource, but I can add one specifically for this.

SaintPatrck:
Hold for Product review.

SaintPatrck:
@cseufert Unfortunately we will not be moving forward with this change. Please reach out to our CS team and reference this ticket. They will work with you to clean up the erroneous data in your vault.

PR #4845, opened 2025-03-11
PM-11356 prevent extra soft-keyboard showing.

dseverns-livefront:
## 🎟️ Tracking
[PM-11356](https://bitwarden.atlassian.net/browse/PM-11356)

## 📔 Objective
- When the user account and unlocked status change we trigger the garbage collection which causes the screen to completely compose twice, which causes the focus request on the text field to happen twice which is the cause of the "extra" keyboard showing. 
- Adding a small delay means the first attempt to draw the screen that will ultimately get garbage collected will not show the keyboard as the text field will not autofocus.

## 📸 Screenshots
# BEFORE: 

https://github.com/user-attachments/assets/0f099687-3232-46e0-98ee-bf5afc6970cc

# AFTER

https://github.com/user-attachments/assets/e109d8f6-d9e9-42ec-9a99-fdcf3f1043e5

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11356]: https://bitwarden.atlassian.net/browse/PM-11356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreen.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreen.kt
@@ -67,7 +67,10 @@ import com.x8bit.bitwarden.ui.platform.manager.biometrics.BiometricsManager
 import com.x8bit.bitwarden.ui.platform.theme.BitwardenTheme
 import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.toImmutableList
+import kotlinx.coroutines.delay
 import javax.crypto.Cipher
+
+private const val AUTO_FOCUS_DELAY = 415L
 
 /**
  * The top level composable for the Vault Unlock screen.
@@ -251,6 +254,17 @@ fun VaultUnlockScreen(
         ) {
             Spacer(modifier = Modifier.height(12.dp))
             if (!state.hideInput) {
+                // When switching from an unlocked account to a locked account, garbage collection
+                // is triggered which causing this screen to be composed twice. Adding this delay
+                // prevents the MP or Pin field from auto focusing on the first composition which
+                // creates a visual jank where the keyboard shows, disappears, and then shows again.
+                var autoFocusDelayCompleted by rememberSaveable {
+                    mutableStateOf(false)
+                }
+                LaunchedEffect(Unit) {
+                    delay(AUTO_FOCUS_DELAY)
+                    autoFocusDelayCompleted = true
+                }
                 BitwardenPasswordField(
                     label = state.vaultUnlockType.unlockScreenInputLabel(),
                     value = state.input,
@@ -261,7 +275,7 @@ fun VaultUnlockScreen(
                     showPasswordTestTag = state
                         .vaultUnlockType
                         .inputFieldVisibilityToggleTestTag,
-                    autoFocus = state.showKeyboard,
+                    autoFocus = state.showKeyboard && autoFocusDelayCompleted,
                     imeAction = ImeAction.Done,
                     keyboardActions = KeyboardActions(
                         onDone = remember(viewModel) {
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreenTest.kt
```diff
@@ -474,8 +476,9 @@ class VaultUnlockScreenTest : BaseComposeTest() {
     }
 
     @Test
-    fun `state with input and without biometrics should request focus on input field`() {
+    fun `state with input and without biometrics should request focus on input field`() = runTest {
         mutableStateFlow.update { it.copy(hideInput = false, isBiometricEnabled = false) }
+        dispatcher.advanceTimeByAndRunCurrent(500L)
```
dseverns-livefront, line 481: Rounded up to 500 here, just....cause :)

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreen.kt
```diff
@@ -251,6 +254,17 @@ fun VaultUnlockScreen(
         ) {
             Spacer(modifier = Modifier.height(12.dp))
             if (!state.hideInput) {
+                // When switching from an unlocked account to a locked account, garbage collection
+                // is triggered which causing this screen to be composed twice. Adding this delay
```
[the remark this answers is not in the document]
dseverns-livefront: maybe the language should be updated in the comment since it is the activity recreation causing this... referenced the gc since the way this was linked in the above function, but seems like what you are saying is we are recreating the UI cause of the account/unlock change and doing the gc on the side.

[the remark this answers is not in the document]
david-livefront: Can you validate if it is specifically one or the other and make sure the comment is accurate?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreen.kt
```diff
@@ -251,6 +254,18 @@ fun VaultUnlockScreen(
...
+                // When switching from an unlocked account to a locked account, the
+                // current activity is recreated and therefore the composition takes place
+                // twice. Adding this delay prevents the MP or Pin field
+                // from auto focusing on the first composition which creates a visual jank where
+                // the keyboard shows, disappears, and then shows again.
+                var autoFocusDelayCompleted by rememberSaveable {
+                    mutableStateOf(false)
+                }
+                LaunchedEffect(Unit) {
+                    delay(AUTO_FOCUS_DELAY)
```
phil-livefront, line 266: ❓ should this logic go into the VM or is it specific enough to the view that wouldn't make sense?  Also noticed, which isn't part of your PR code changes, that we are only hitting this `if` condition if `!state.hideInput` but we are also checking that within the `state.showKeyboard` property.  

```
    /**
     * Indicates if we want force focus on Master Password \ PIN input field and show keyboard.
     */
    val showKeyboard: Boolean get() = !showBiometricLogin && !hideInput
```

phil-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockScreen.kt
david-livefront: I think the `showKeyboard` function was defined independent of the UI, so it considers all the variables. I don't think it hurts anything.

I also think this logic is fine living in the screen.

PR #4846, opened 2025-03-11
[PM-13257] Checking if user is navigating from vault before showing prompt for biometrics

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13257

## 📔 Objective

User should not be automatically prompted for biometrics when naviting from vault to lock screen. On all other scenarios user should be automatically prompted.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
@@ -99,6 +99,8 @@ class VaultLockManagerImpl(
     override val vaultStateEventFlow: Flow<VaultStateEvent>
         get() = mutableVaultStateEventSharedFlow.asSharedFlow()
 
+    override var isFromLockFlow: Boolean = false
+
     init {
         observeAppCreationChanges()
         observeAppForegroundChanges()
@@ -118,6 +120,7 @@ class VaultLockManagerImpl(
         mutableVaultUnlockDataStateFlow.value.statusFor(userId) == VaultUnlockData.Status.UNLOCKING
 
     override fun lockVault(userId: String) {
+        isFromLockFlow = true
         setVaultToLocked(userId = userId)
     }
 
--- a/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
@@ -119,14 +119,17 @@ class VaultLockManagerImpl(
     override fun isVaultUnlocking(userId: String): Boolean =
         mutableVaultUnlockDataStateFlow.value.statusFor(userId) == VaultUnlockData.Status.UNLOCKING
 
-    override fun lockVault(userId: String) {
-        isFromLockFlow = true
+    override fun lockVault(userId: String, isUserInitiated: Boolean) {
+        isFromLockFlow = isUserInitiated
         setVaultToLocked(userId = userId)
     }
 
     override fun lockVaultForCurrentUser() {
         activeUserId?.let {
-            lockVault(it)
+            lockVault(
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -118,6 +120,7 @@ class VaultLockManagerImpl(
         mutableVaultUnlockDataStateFlow.value.statusFor(userId) == VaultUnlockData.Status.UNLOCKING
 
     override fun lockVault(userId: String) {
+        isFromLockFlow = true
```
david-livefront, line 123: This needs to be scoped to the user that locked the vault, for instance a user with an `immediate` timeout action could cause a race condition and suppress the biometric prompt for the user when they switch to another account.

There is also a concern here that we don't always want to set the value to true here, since we do not want to do this if the vault is locked automatically.

Something like this:
```kotlin
    override fun lockVault(userId: String, isUserInitiated: Boolean) {
        isFromLockFlow = isUserInitiated
        setVaultToLocked(userId = userId)
    }
```

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockViewModel.kt
```diff
@@ -422,7 +428,7 @@ class VaultUnlockViewModel @Inject constructor(
 
     private fun promptForBiometricsIfAvailable() {
         val cipher = biometricsEncryptionManager.getOrCreateCipher(state.userId)
-        if (state.showBiometricLogin && cipher != null) {
+        if (state.showBiometricLogin && cipher != null && !vaultLockManager.isFromLockFlow) {
```
david-livefront, line 431: Maybe set this on the state during initialization

app/src/main/java/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -119,14 +119,17 @@ class VaultLockManagerImpl(
...
+        isFromLockFlow = isUserInitiated
         setVaultToLocked(userId = userId)
     }
 
     override fun lockVaultForCurrentUser() {
         activeUserId?.let {
-            lockVault(it)
+            lockVault(
+                userId = it,
+                isUserInitiated = true,
```
david-livefront, line 131: So this is technically true at this time but we probably should not assume that all callers of this method will be user initiated.

Can we update this to be:
```kotlin
override fun lockVaultForCurrentUser(isUserInitiated: Boolean) {
    activeUserId?.let {
        lockVault(
            userId = it,
            isUserInitiated = isUserInitiated,
        )
    }
}
```

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/vaultunlock/VaultUnlockViewModel.kt
```diff
@@ -120,6 +123,10 @@ class VaultUnlockViewModel @Inject constructor(
             .launchIn(viewModelScope)
 
         promptForBiometricsIfAvailable()
+
+        // only when navigating from vault to lock we should not display biometrics
+        // subsequent views of the lock screen should display biometrics if available
+        vaultLockManager.isFromLockFlow = false
```
david-livefront, line 129: Shouldn't you also clear the the VM state too?

aj-rosado: Not sure if it is necessary (from testing it is not). But it makes sense to me to clear it here so added the change.

david-livefront (approved):

PR #4916, opened 2025-03-25
PM-19131 - Custom hidden fields not working properly if TOTP also configured

phil-livefront:
## 🎟️ Tracking

[PM-19131](https://bitwarden.atlassian.net/browse/PM-19131)

## 📔 Objective

- Previously, the isVisible property on custom hidden fields was being reset every time the TOTP clock updated (once per second). This caused a UX issue where a visible hidden field would incorrectly revert to a hidden state on each tick.

- The initial fix attempted to preserve visibility state by comparing against the `previousState`. However, `FieldView` via the SDK does not have a unique id property, and matching by name proved unreliable since multiple custom fields can share the same name.

- To address this, the visibility state is now managed locally in the view using rememberSaveable, allowing the UI to persist visibility independently of the model.

- This pattern is likely a better fit for view-driven state, such as `isVisible` and could also be extended to other field types (like password) to avoid similar issues and reduce model-driven UI state churn.  Happy to have a convo about that.  

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19131]: https://bitwarden.atlassian.net/browse/PM-19131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/component/VaultItemCustomField.kt
```diff
@@ -47,8 +53,11 @@ fun CustomField(
                 BitwardenPasswordField(
                     label = customField.name,
                     value = customField.value,
-                    showPasswordChange = { onShowHiddenFieldClick(customField, it) },
-                    showPassword = customField.isVisible,
+                    showPasswordChange = {
+                        onShowHiddenFieldClick(customField, it)
+                        isVisible = it
```
david-livefront, line 58: This is bypassing the the prompt for password check. We cannot do this here, it has to be managed by the ViewModel.

phil-livefront: I am probably just misunderstanding but the same value is being passed along with `onShowHiddenFieldClick` so im not following how this is bypassing any existing behavior.

david-livefront: Unless I am mistaken, you are making the value visible before the password prompt which bypasses that security feature. We cannot do that.

phil-livefront: which prompt are you talking about?  just want to make sure we're on the same page here.

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemViewModel.kt
```diff
@@ -1653,7 +1659,6 @@ data class VaultItemState(
                  * @property uris The URI associated with the login item.
                  * @property passwordRevisionDate An optional string indicating the last time the
                  * password was changed.
-                 * @property totpCodeItemData The optional data related the TOTP code.
```
david-livefront: Can we leave this here

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemViewModel.kt
```diff
@@ -1662,7 +1667,7 @@ data class VaultItemState(
                  *
                  * **NOTE** [canViewTotpCode] currently supports a deprecated edge case where an
                  * organization supports TOTP but not through the current premium model.
-                 * This additional field is added to allow for [isPremiumUser] to be an independent
+                 * This additional field is added to allow for [] to be an independent
```
david-livefront: Can we put this back?

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemViewModel.kt
```diff
@@ -1688,6 +1694,7 @@ data class VaultItemState(
                             fido2CredentialCreationDateText != null ||
                             totpCodeItemData != null
 
+
```
david-livefront, line 1697: Wanna drop this

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/util/CipherViewExtensions.kt
```diff
@@ -57,7 +57,14 @@ fun CipherView.toViewState(
             name = name,
             requiresReprompt = (reprompt == CipherRepromptType.PASSWORD && hasMasterPassword) &&
                 previousState?.common?.requiresReprompt != false,
-            customFields = fields.orEmpty().map { it.toCustomField() },
+            customFields = fields.orEmpty().map { fieldView ->
+                fieldView.toCustomField(
+                    previousState = previousState
+                        ?.common
+                        ?.customFields
+                        ?.find { it.id == fieldView.hashCode().toString() },
```
david-livefront, line 65: So you did go with the hashcode?

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemViewModel.kt
```diff
@@ -1687,7 +1693,7 @@ data class VaultItemState(
                             passwordData != null ||
                             fido2CredentialCreationDateText != null ||
                             totpCodeItemData != null
-
+
```
david-livefront, line 1696: One more. lol

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/util/CipherViewExtensions.kt
```diff
@@ -195,27 +202,40 @@ fun CipherView.toViewState(
...
         FieldType.HIDDEN -> VaultItemState.ViewState.Content.Common.Custom.HiddenField(
+            id = this.hashCode().toString(),
             name = name.orEmpty(),
             value = value.orZeroWidthSpace(),
             isCopyable = !value.isNullOrBlank(),
-            isVisible = false,
+            isVisible = (previousState as?
+                VaultItemState.ViewState.Content.Common.Custom.HiddenField)
+                ?.isVisible
+                ?: false,
```
david-livefront, line 228: 👍

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/util/CipherViewExtensions.kt
```diff
@@ -195,27 +202,40 @@ fun CipherView.toViewState(
         },
     )
 
-private fun FieldView.toCustomField(): VaultItemState.ViewState.Content.Common.Custom =
+
+/**
+ * Transforms [FieldView] into [VaultItemState.ViewState.Content.Common.Custom].
+ */
+fun FieldView.toCustomField(
```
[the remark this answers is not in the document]
phil-livefront: so i could use the function in the test which allowed me to use `fieldView.hashCode().toString()`.  I didnt see a reason this couldn't be public, did i miss something?
`

app/src/test/java/com/x8bit/bitwarden/ui/vault/feature/item/util/VaultItemTestUtil.kt
```diff
@@ -188,29 +187,36 @@ fun createCommonContent(
...
-                VaultItemState.ViewState.Content.Common.Custom.TextField(
+                FieldView(
                     name = "text",
                     value = "value",
-                    isCopyable = true,
-                ),
-                VaultItemState.ViewState.Content.Common.Custom.HiddenField(
+                    type = FieldType.TEXT,
+                    linkedId = null,
+                ).toCustomField(null),
```
david-livefront, line 195: Instead of doing this, Can we just leave it as it was and hardcode it?

```
VaultItemState.ViewState.Content.Common.Custom.BooleanField(
    id = "12345",
    value = true,
    type = FieldType.BOOLEAN,
```

phil-livefront: We can't because then the hash ids dont match since its based on the FieldView

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/util/CipherViewExtensions.kt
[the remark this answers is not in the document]
david-livefront: We generally do not make any function public purely for testing purposes. ideally, this would not be needed.

app/src/test/java/com/x8bit/bitwarden/ui/vault/feature/item/util/VaultItemTestUtil.kt
david-livefront: I'm a little confused, this is all test data is it not?

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/util/CipherViewExtensions.kt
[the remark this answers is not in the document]
phil-livefront: see comment below on the comment about the mock data changes.  I know this isnt the ideal situation but imo this solution isn't the most ideal either, if we had a 1:1 matching `id` from the sdk to the domain model it would make way more sense.

app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemViewModel.kt
```diff
@@ -1597,11 +1597,14 @@ data class VaultItemState(
                  * Represents a custom field, TextField, HiddenField, BooleanField, or LinkedField.
                  */
                 sealed class Custom : Parcelable {
+                    abstract val id: String
```
david-livefront, line 1600: Can we add doc for this... Nothing crazy just like: The unique ID of the custom field.

app/src/test/java/com/x8bit/bitwarden/ui/vault/feature/item/util/VaultItemTestUtil.kt
phil-livefront: yeah... its because its being generated from the FieldView itself and comparing it to our domain models `VaultItemState.ViewState.Content.Common.Custom.HiddenField` so the ids will never match.

david-livefront (approved):

PR #5012, opened 2025-04-09
[PM-20026] RTL text direction

### LINKED ISSUE #4753: RTL text direction
### Steps To Reproduce

**To reproduce the bug with a vault item:**
1. Change app language to a language with an RTL script such as Hebrew or Arabic, or enable "Force RTL layout direction" in Android's Developer Settings
2. Open Bitwarden (and unlock vault)
3. Create a new login vault item with the following details and save it:
* Name: not applicable (just type something)
* Username: `_user`
* Password: `pass*`
* URL: `https://bitwarden.com/` _(don't forget the ending slash)_
* Custom field (text): `$custom-text@#`
4. Open the newly created vault item and examine its username, password, URL and custom field

**To reproduce the bug with the password generator:**
1. Open the password generator
2. Generate a password which ends or starts with a number or a special character
3. Copy the generated password and compare it with the displayed password

### Expected Result

The username, password, URL, and custom field of a vault item should appear exactly as they were typed.

The generated password that is displayed in the Password Generator should be identical to the copied password.

### Actual Result

The details of the vault item appear as follows:
* Username: `user_`
* Password: `*pass`
* URL: `/https://bitwarden.com`
* Custom field (text): `#@custom-text$`

Notice that the special characters at the start and end of each field have been reversed (same thing happens with the Password Generator).

### Additional Context

The bug most likely happens because the direction of text fields automatically follows the direction of the layout (RTL).

This behavior is unwanted because it causes confusion and error when copying a displayed field visually.

Most websites do not allow passwords or usernames to contain RTL scripts, and when such characters appear in a URL, they must be (percent) encoded. So there is no benefit to apply an RTL direction to such fields.

Note that the RTL text direction is desired in the notes field of a login vault item as well as in a secure note and in the various fields of an identity vault item, because they are more likely to be written from right-to-left. So the current behavior should remain unchanged for them.

Lastly, it should be noted that the password and text fields of a Send are also affected.

**UPDATE:** It seems that this issue is present in the master password input field as well!

### Build Version

2025.1.2

### Environment Details

- OS Version: Android 14

### Issue Tracking Info

codokie:
## 🎟️ Tracking

#4753

## 📔 Objective

Fixing text direction of LTR text in RTL layout, and vice versa.

## 📸 Screenshots

Screenshot

(see "Steps To Reproduce" in the tracked issue)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


S-Kakar:
Thank you for your contribution! We've added this to our internal Community PR board for review.
ID: PM-20026

codokie:
Note that there is still a problem with bi-directional text in the password history view.

Historic custom hidden passwords are not formatted properly. See the following screenshot:

bug

(Password is 1234 5678 in both cases, only field label has been changed. Current behavior is the one to the left.)

Upon some digging I've found [this](https://github.com/bitwarden/android/blob/8767f2dd24229d8d5c79f22e0205f0316abe577d/app/src/main/java/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultAddItemStateExtensions.kt#L196) to be the root cause.

**Possible solution:**

`PasswordHistoryView.password` should not include the label of the hidden field. Instead, in `PasswordHistoryListItem`, add a separate `Text` composable for the label.

I could not figure out how to add a `label` property to `PasswordHistoryView` data class, so I may not be able to implement the fix for this. It also seems that old fields are stored in the cipher, so there would be a need to migrate them.

codokie:
@david-livefront @brian-livefront

Could you please review the proposed changes?

Thank you

app/src/main/java/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/pendingrequests/PendingRequestsScreen.kt
```diff
@@ -313,14 +313,17 @@ private fun PendingRequestItem(
...
             Text(
                 text = timestamp,
                 style = BitwardenTheme.typography.labelSmall,
                 color = BitwardenTheme.colorScheme.text.secondary,
-                textAlign = TextAlign.End,
+                textAlign = if (LocalLayoutDirection.current == LayoutDirection.Rtl) {
+                    TextAlign.Left
+                } else {
+                    TextAlign.Right
+                },
```
SaintPatrck: We should avoid manually setting the text alignment in screens like this.

We also do not want to use `Left` and `Right` alignments. We should be using `Start` and `End` so they respect the selected language's layout direction.

app/src/main/java/com/x8bit/bitwarden/ui/platform/theme/type/BitwardenTypography.kt
```diff
@@ -28,3 +29,31 @@ data class BitwardenTypography(
...
+        bodyMediumEmphasis = bodyMediumEmphasis.copy(textAlign = newTextAlign),
+        bodySmall = bodySmall.copy(textAlign = newTextAlign),
+        labelLarge = labelLarge.copy(textAlign = newTextAlign),
+        labelMedium = labelMedium.copy(textAlign = newTextAlign),
+        labelSmall = labelSmall.copy(textAlign = newTextAlign),
+        sensitiveInfoSmall = sensitiveInfoSmall.copy(textAlign = newTextAlign),
+        sensitiveInfoMedium = sensitiveInfoMedium.copy(textAlign = newTextAlign),
+        eyebrowMedium = eyebrowMedium.copy(textAlign = newTextAlign),
+    )
+}
```
SaintPatrck: All files must end with a newline. 

The [Setup](https://github.com/bitwarden/android/blob/main/README.md#setup) section of README has instructions on how to create a macro to enforce this and other [Code style guidelines](https://github.com/bitwarden/android/blob/main/docs/STYLE_AND_BEST_PRACTICES.md).

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/enterprisesignon/EnterpriseSignOnScreen.kt
```diff
@@ -133,7 +132,6 @@ private fun EnterpriseSignOnScreenContent(
         Spacer(modifier = Modifier.height(height = 12.dp))
         Text(
             text = stringResource(id = R.string.log_in_sso_summary),
-            textAlign = TextAlign.Start,
```
SaintPatrck: We should not be removing alignment assignments in the screens. `Start` and `End` will respect the layout direction of the selected language.

app/src/main/java/com/x8bit/bitwarden/ui/platform/theme/type/Typography.kt
```diff
@@ -228,6 +245,7 @@ val bitwardenTypography: BitwardenTypography = BitwardenTypography(
             trim = LineHeightStyle.Trim.None,
         ),
         platformStyle = PlatformTextStyle(includeFontPadding = false),
+        textDirection = TextDirection.Ltr,
```
SaintPatrck: Should this be using `TextDirection.Content`?

app/src/main/java/com/x8bit/bitwarden/ui/platform/theme/type/Typography.kt
```diff
@@ -240,6 +258,7 @@ val bitwardenTypography: BitwardenTypography = BitwardenTypography(
...
```
SaintPatrck: Should this be using `TextDirection.Content`?

SaintPatrck (commented):
🤔 I'm wondering if having a `BidiTextManager` and `ProvidableCopositionLocal<BidiTextManager>` would be a better way to handle formatting. Something along the lines of

```kotlin
interface BidiTextManager {
    fun unicodeWrap(text: String): String
}

val LocalBidiTextManager: ProvidableCompositionLocal<BidiTextManager> = compositionLocalOf {
    error("CompositionLocal BidiTextManager not present")
}
```

Then we can apply the formatting in our common Composables.

An example from `BitwardenTextField` would look like...

```kotlin
fun BitwardenTextField(...) {
    var textFieldValueState by remember { mutableStateOf(TextFieldValue(text = formattedText)) }
    val textFieldValue = textFieldVauleState.copy(
        text = LocalBidiTextManager.current.unicodeWrap(value),
    )
    // Remaining logic omitted...
}
```

With that approach I was able to achieve the following results

Image

SaintPatrck:
Can you update your fork & branches? There are several conflicts since we've moved common logic and UI components to their own module.

codokie:
Thanks a lot @SaintPatrck for the review.

Your idea of using `BidiTextManager` seems much cleaner than my approach, although I haven’t been able to test it yet. Does the `unicodeWrap` function need to be implemented?

I will try to update my fork and resolve the conflicts later.

SaintPatrck:
@codokie I opened #5825, which implements the `BidiTextManager`. I'm not sure if the verification codes are being handled correctly. Could you take a look at the screenshots and confirm that they're appearing correctly with RTL languages?

Regarding the Hidden Field history; that will have to be addressed separately due to how it's being constructed.

codokie:
Closing in favor of #6309

PR #5101, opened 2025-04-25
Use Google's Digital Asset Links API to verify digital asset links

SaintPatrck:
## 🎟️ Tracking

TBD

## 📔 Objective

Refactor DigitalAssetLinkApi and DigitalAssetLinkService implementations to use Google's digital asset link REST API.

By using Google's REST API, we no longer have to derive the host URL ourselves, and move the responsibility of validating asset link files to Google.

Additionally, relyingPartyId is no longer needed to check asset links and has been removed from corresponding argument lists.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/java/com/x8bit/bitwarden/data/platform/util/CallingAppInfoExtensions.kt
```diff
@@ -7,14 +7,15 @@ import java.security.MessageDigest
 
 /**
  * Returns the application's signing certificate hash formatted as a hex string if it has a single
- * signing certificate. Otherwise `null` is returned.
+ * signing certificate. Otherwise an empty String is returned.
  */
 @OptIn(ExperimentalStdlibApi::class)
-fun CallingAppInfo.getSignatureFingerprintAsHexString(): String? {
+fun CallingAppInfo.getSignatureFingerprintAsHexString(): String {
```
SaintPatrck, line 13: This change causes the DigitalAssetLink API to fail if the signature is empty or invalid when it is required, so we don't have to pre-validate it ourselves.

david-livefront: Hmmm, I think I would prefer returning null here instead of an empty string. Then adding the `orEmpty()` where it is needed.

Returning null indicates that there it was unable to get the value, returning null requires the caller to know that it can be empty and handle it. We lose that information by returning an empty string here.

david-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/environment/EnvironmentScreen.kt
```diff
@@ -153,6 +153,30 @@ fun EnvironmentScreen(
...
+                title = dialog.title(),
+                message = dialog.message(),
+                confirmButtonText = stringResource(R.string.replace_certificate),
+                dismissButtonText = stringResource(R.string.cancel),
+                onConfirmClick = remember(viewModel) {
+                    {
+                        viewModel.trySendAction(
+                            EnvironmentAction.ConfirmOverwriteCertificateClick(
+                                triggeringAction = dialog.triggeringAction,
+                            ),
```
david-livefront, line 167: Can we get some tests added?

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/environment/EnvironmentViewModel.kt
```diff
@@ -393,6 +413,27 @@ class EnvironmentViewModel @Inject constructor(
...
+                certificateManager.importMutualTlsCertificate(
+                    key = bytes,
+                    alias = alias,
+                    password = password,
+                )
+            }
+            .map { result ->
+                sendAction(
+                    EnvironmentAction.Internal.ImportKeyResultReceive(result),
+                )
```
david-livefront, line 434: It's a bit odd to do this in a `map`, can we use an `also` instead?

app/src/main/res/values/strings.xml
```diff
@@ -1259,4 +1259,7 @@ Do you want to switch to this account?</string>
   <string name="all_logs_deleted">All logs deleted</string>
   <string name="log_deleted">Log deleted</string>
   <string name="app_settings">App settings</string>
+  <string name="replace_existing_certificate">Replace existing certificate?</string>
+  <string name="a_certificate_with_the_alias_x_already_exists_do_you_want_to_replace_it">A certificate with this alias \"%s\" already exists. Do you want to replace it?\nReplace the certificate may impact your connection to any environments currently using it.</string>
+    <string name="replace_certificate">Replace certificate</string>
```
david-livefront, line 1264: I think this one is overly indented.

david-livefront: Do you know if it's possible to get Crowdin to respect our preferred indentation so this isn't a problem in the future?

SaintPatrck: Crowdin doesn't control this. I'm not sure when or how, but at some point the wrong indention was applied and we've just not fixed it.

https://github.com/bitwarden/android/pull/5109 should fix it.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/environment/EnvironmentScreen.kt
SaintPatrck: Oops. Missed this. Thanks!

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/environment/EnvironmentViewModel.kt
SaintPatrck: 👍 We only really need the first `map`. ⚡

david-livefront: Can we still use the `also` instead of the `map` since we are not using the result `map` seems a bit odd here

SaintPatrck: That `map` is unwrapping the `Result<T>`. I actually need to swap it for a `.onFailure {}.onSuccess {}` chain so we notify the user if `fileManager.uriToByteArray(uri)` fails.

app/src/main/java/com/x8bit/bitwarden/ui/auth/feature/environment/EnvironmentViewModel.kt
```diff
@@ -384,15 +391,60 @@ class EnvironmentViewModel @Inject constructor(
...
+
+    private suspend fun saveCertificate(
+        uri: Uri,
+        alias: String,
+        password: String,
+    ) {
+        fileManager
+            .uriToByteArray(uri)
+            .onFailure {
+                showErrorDialog(
```
david-livefront, line 430: This is a pattern break, you should never update the state from inside a suspend function.

Can we do this instead:
```kotlin
val result = fileManager.uriToByteArray(uri)
sendAction(EnvironmentAction.Internal.NewAction(result))


private fun handNewAction(action: EnvironmentAction.Internal.NewAction) {
    action.result.fold(
        onFailure = {
            showErrorDialog(
                message = R.string.generic_error_message.asText(),
                throwable = it,
            )
        },
        onSuccess = {
            // Do the other thing
        },
    )
}
```

david-livefront: You could also just add the new action to avoid messing with the old one:
```kotlin
        fileManager
            .uriToByteArray(uri)
            .fold(
                onFailure = {
                    sendAction(
                        EnvironmentAction.Internal.ShowDialog(
                            ...
                        ),
                    )
                },
                onSuccess = {
                    sendAction(
                        EnvironmentAction.Internal.ImportKeyResultReceive(
                            certificateManager.importMutualTlsCertificate(
                                key = bytes,
                                alias = alias,
                                password = password,
                            ),
                        ),
                    )
                },
            )
```

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/manager/Fido2CredentialManagerImpl.kt
```diff
@@ -113,9 +113,7 @@ class Fido2CredentialManagerImpl(
                     packageName = callingAppInfo.packageName,
                     sha256CertFingerprint = callingAppInfo
                         .getSignatureFingerprintAsHexString()
-                        ?: return Fido2CredentialAssertionResult
-                            .Error
-                            .InvalidAppSignature,
```
david-livefront, line 118: Can we ditch this error state?

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/model/Fido2CredentialAssertionResult.kt
```diff
@@ -24,21 +24,6 @@ sealed class Fido2CredentialAssertionResult {
          */
         data object MissingHostUrl : Error()
 
-        /**
-         * Indicates the calling application signature was invalid.
-         */
-        data object InvalidAppSignature : Error()
```
david-livefront, line 30: 👍

david-livefront (approved):

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/manager/Fido2OriginManagerImpl.kt
```diff
@@ -11,61 +10,50 @@ import timber.log.Timber
 
 private const val GOOGLE_ALLOW_LIST_FILE_NAME = "fido2_privileged_google.json"
 private const val COMMUNITY_ALLOW_LIST_FILE_NAME = "fido2_privileged_community.json"
+private const val DELEGATE_PERMISSION_HANDLE_ALL_URLS = "delegate_permission/common.handle_all_urls"
```
Nailik, line 13: i think get_login_creds should be used

SaintPatrck: Hi @Nailik,

I see no issue with adjusting our logic to check for _either_. I'm hesitant to replace it entirely or revert back to requiring both because many relying parties have not updated their links to include `get_login_creds`, and fewer use it exclusively. 

In order to provide the expected behavior to most users (and to align with Google Password Manager behavior which often drives user expectation) we elected to only require `handle_all_urls` as that's what the majority of relying party implementations still have declared.

network/src/main/kotlin/com/bitwarden/network/service/DigitalAssetLinkServiceImpl.kt
```diff
@@ -11,13 +11,17 @@ class DigitalAssetLinkServiceImpl(
...
-            .getDigitalAssetLinks(
-                url = "$scheme$relyingParty/.well-known/assetlinks.json",
-            )
-            .toResult()
+    override suspend fun checkDigitalAssetLinksRelations(
+        packageName: String,
+        certificateFingerprint: String,
+        relation: String,
+    ): Result<DigitalAssetLinkCheckResponseJson> = digitalAssetLinkApi
+        .checkDigitalAssetLinksRelations(
```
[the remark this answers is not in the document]
SaintPatrck: Good question. 🤔 This may be an implementation bug on our end. I'll have to double check which one, but I believe `source` or `target` info need to be from the Bitwarden package.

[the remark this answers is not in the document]
Nailik: hm i think it should only be used if a website should be associated with an app.

because then you can upload an assetlink json file to your website containing the delegations (get_login_creds and handle_all_urls) with your app package name and fingerprint to prove that your app should get access to credentials that were stored for this site.

app/src/main/java/com/x8bit/bitwarden/data/autofill/fido2/manager/Fido2OriginManagerImpl.kt
Nailik: according to the android docs it should be an "and" and check for both https://developer.android.com/identity/sign-in/credential-manager?hl=de#add-support-dal
I remember having this issue where the google password manager also checks for both.

SaintPatrck: I agree that the RP implementation reads as though Providers should expect both, and it is somewhat misleading. I had to get clarification from Google about this because we had issues with RP's such as Paypal that do no declare `get_all_creds` but expect passkeys to work with all providers. That issue was compounded due to the fact that it works with Google Password Manager, despite both relations not being declared in their asset links. The RP docs say "_and_", but Google recommends **providers** check for "_either_".

From Google...
> The documentation you linked above is guidance to RPs, not to providers. As explained in the issue in your second link, we instruct RPs to include both relations because, for Google Password Manager, there is an ongoing migration from one to the other. The documentation does not imply requirements on providers... [^1]

> ...we strongly recommend [validating assetlinks] by checking for the presence of either relation. [^1]

They fail to mention in the RP docs that "handle_all_urls" should be included because Google Password Manager _may_ migrate to `get_login_creds`. 🤦 

[^1]: https://issuetracker.google.com/issues/397181253

Issue #5084, opened 2025-04-22
Host Match Detection broken for URLs with ports


NicholasFlamy:
### Steps To Reproduce

https://github.com/bitwarden/mobile/issues/2970
1. Save an autofill item for a URL with a port.
2. See how it behaves on sites.

### Expected Result

Works correctly. Or at least works like the old app.

### Actual Result

Doesn't work.

### Build Version

2025.3.0

### What server are you connecting to?

Self-host

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

S-Kakar:
Thank you for your report! We've added this to our internal board for review.
ID: PM-20513

NicholasFlamy:
Note: #4531 is focused on Host detection not working WHICH IS A KNOWN AND REPRODUCIBLE ISSUE CAUSED BY CHROMIUM. This issue is focused on the fact that the new app behaves different from the old one.

NicholasFlamy:
Big difference #1 is that leaving out the URL scheme, `http://` or `https://`, makes the autofill not work if Match Detection is set to Host.

Big difference #2 is that Accessibility Autofill no longer handles ports. On the old app, enabling both accessibility and inline autofill meant the inline showed all options for that base domain which the accessibility button to open the Bitwarden app would open the app to just the properly mathing URLs.

NicholasFlamy:
Small, unrelated thing: Great job on the native app. It LITERALLY takes less than a 1/4 second to open to the main screen when opening it from the Android home screen.
Seriously, great job! (I would like a compact mode appearance option but overall great.)

SaintPatrck:
Hi @NicholasFlamy,

There is a limitation with the Autofill framework that prevents us from matching based on port, path, and query string parameters. They are completely omitted from the request we receive. I left more details about this limitation on https://github.com/bitwarden/mobile/issues/2124#issuecomment-2839759140.

The Accessibility behavior you're describing should be resolved with #5118.

I'm not able to replicate the issue with inline autofill and accessibility being enabled at the same time. Could you elaborate on the behavior you're experiencing when both are enabled?

NicholasFlamy:
> The Accessibility behavior you're describing should be resolved with [#5118](https://github.com/bitwarden/android/pull/5118).

I hope it does, I read the description and it seems to address both parts of the issue.


[...]

PR #5118, opened 2025-05-01
[PM-14846] Improve IP Address and Port Handling in StringExtensions

SaintPatrck:
## 🎟️ Tracking

PM-14846
PM-20513
Relates to #5084 

## 📔 Objective

Correct parsing of IP addresses with and without scheme and port. `URI` cannot parse an IP address with a port when it does not also include the scheme (http, https), so autofill matching triggered from the Quick Tile was not working properly.

-   **New Function:** Added `isIpAddress()` to check if a string is a valid IPv4 address, optionally with a port and scheme.
-   **Regex Update:** Added the `IP_ADDRESS_WITH_OPTIONAL_PORT` regex. It validates IPv4 addresses with or without ports (0-65535 range), with an optional `http://` or `https://` prefix.
-   **URI Parsing:** Modified `toUriOrNull()` to handle IP addresses without schemes by adding a default "https://" scheme before parsing.
- **Tests update**: Added new tests for `isIpAddress` function.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
+++ b/app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
@@ -11,12 +11,54 @@ import java.net.URISyntaxException
 private const val ANDROID_APP_PROTOCOL: String = "androidapp://"
 
 /**
+ * Regex to match an IP address (IPv4) with an optional port within the valid range (0-65535).
+ *
+ * This regex matches the following:
+ * - An optional "http://" or "https://" prefix.
+ * - A standard IPv4 address (four groups of digits separated by dots).
+ * - An optional colon (:) followed by a port number (digits).
+ * - The port number, if present, must be within the range of 0-65535.
+ *
+ * Example valid strings:
+ * - 192.168.1.1
+ * - 10.0.0.5:8080
+ * - 255.255.255.0:9000
+ * - 0.0.0.0
+ * - 1.1.1.1:0
+ * - 1.1.1.1:65535
+ * - https://192.168.1.1
+ * - https://10.0.0.5:8080
+ * - http://255.255.255.0:9000
+ * - http://10.0.0.5:8080
+ *
+ * Example invalid strings:
+ * - 256.1.1.1 (invalid IP component)
+ * - 10.1.1 (missing IP components)
+ * - 10.1.1.1:abc (non-numeric port)
+ * - 10.1.1.1:-1 (invalid port)
+ * - 10.1.1.1: (missing port)
+ * - 1.1.1.1:65536 (invalid port)
+ * - 1.1.1.1:99999 (invalid port)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/x8bit/bitwarden/data/platform/util/StringExtensions.kt
```diff
@@ -10,13 +10,58 @@ import java.net.URISyntaxException
...
-    } catch (e: URISyntaxException) {
+        // URI cannot parse IP addresses without a scheme, so add one if necessary.
+        if (
+            this.isIpAddress() &&
+            !this.hasHttpProtocol()
+        ) {
+            URI("https://$this")
+        } else {
+            URI(this)
+        }
```
SaintPatrck, line 63: I considered replacing `URI` with `android.net.Uri`, but that introduces different problems and changes the behavior of this function. 

- `Uri.parse()` is too lenient. It accepts URI's we consider invalid; e.g., `not a uri`.
- `Uri` does not return the correct port when the URI consists of `{IP}:{PORT}`, without the scheme.
- `Uri` is an Android class, so we cannot use it in unit tests. We would rely on the `Roboelectric` implementation of `Uri`, which may or may not accurately reflect the actual implementation from Google.

app/src/test/java/com/x8bit/bitwarden/data/util/StringExtensionsTest.kt
```diff
@@ -223,4 +230,49 @@ class StringExtensionsTest {
...
     }
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `getHostWithPortOrNull should return host with port when URI is IP address with port and scheme`() {
+        val uriString = "https://192.168.1.1:8080"
+        assertEquals("192.168.1.1:8080", uriString.getHostWithPortOrNull())
+    }
+
+    @Suppress("MaxLineLength")
```
david-livefront, line 241: Can we remove this suppression?

david-livefront (approved):

app/src/test/java/com/x8bit/bitwarden/data/util/StringExtensionsTest.kt
```diff
@@ -223,4 +230,48 @@ class StringExtensionsTest {
...
     }
+
+    @Suppress("MaxLineLength")
+    @Test
+    fun `getHostWithPortOrNull should return host with port when URI is IP address with port and scheme`() {
+        val uriString = "https://192.168.1.1:8080"
+        assertEquals("192.168.1.1:8080", uriString.getHostWithPortOrNull())
+    }
+
+    @Test
```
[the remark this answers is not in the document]
SaintPatrck: First time I've seen `ParameterizedTest`. Looks interesting. I'll keep it in mind for future use. Thanks for pointing it out!

PR #5219, opened 2025-05-19
[BRE-769] Use Fastlane to keep github releases in sync with mobile deploy versions

michalchecinski:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BRE-769

## 📔 Objective

Add a workflow to check for new Android versions submitted to the store. If a new version exists, publish the GitHub release as the latest and non-pre-release.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
...
+        if: steps.check_version.outputs.version_released == 'true'
+        id: make_release
+        run: |
+          latest_prerelease_version=${{ steps.get_latest_prerelease.outputs.latest_prerelease_version }}
+          echo "Making release $latest_prerelease_version latest and non-pre-release"
+          curl -X PATCH \
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/bitwarden/android/releases/$latest_prerelease_id \
+            -d '{"prerelease": false, "draft": false, "make_latest": true}'
```
SaintPatrck, line 92:
```suggestion
            -d '{"prerelease": false, "draft": false, "make_latest": true}'
            
```

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Get latest pre-release name
+        id: get_latest_prerelease
+        run: |
+          latest_release=$(jq -r 'first' <<< $(curl --silent https://api.github.com/repos/bitwarden/android/releases))
```
vvolkgang, line 25: We've been using gh cli which helps simplify these.

* Most recent GitHub Release, including drafts and pre-releases: `gh release list --json name,tagName,isDraft,isPrerelease -L 1`
* Last GitHub Release marked as Released: `gh release view --json name,isPrerelease`

Sending the returned json to `GITHUB_OUTPUT` should enable you to directly access fields in other steps without parsing fields individually. 🤔

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
...
+              echo "version_released=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Make GitHub release latest and non-pre-release
+        if: steps.check_version.outputs.version_released == 'true'
+        id: make_release
+        run: |
+          latest_prerelease_version=${{ steps.get_latest_prerelease.outputs.latest_prerelease_version }}
+          echo "Making release $latest_prerelease_version latest and non-pre-release"
+          curl -X PATCH \
```
vvolkgang, line 88: Followup from the GH CLI comment, the following should work but confirm in a test repo first:

`gh release edit <tag> --latest`

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
...
+
+          az storage blob download --account-name $ACCOUNT_NAME --container-name $CONTAINER_NAME \
+          --name authenticator_play_store-creds.json --file ${{ github.workspace }}/secrets/authenticator_play_store-creds.json --output none
+
+      - name: Check released version
+        if: steps.get_latest_prerelease.outputs.is_latest_prerelease == 'true'
+        id: appstore_version
+        run: |
+          appstore_version=$(bundle exec fastlane run getLatestVersion \
+          json_key:${{ github.workspace }}/secrets/authenticator_play_store-creds.json }})
```
vvolkgang, line 62: Did you test this? Seems to be mixing things, it's passing a `json_key` argument to the lane and the lane is accessing `options[:serviceCredentialsFile]` argument which is a firebase thing. It's also accessing the Authenticator app secret while running the Password Manager lane. 

Looking at `publishProdToPlayStore` lane you probably don't need to set anything for Password Manager at least. 

While at it, would help some ongoing related work if your lanes receive the package ID as an argument.

Untested sample code:
```
  desc "Get latest published version"
  lane :getLatestVersion do
      version_codes = google_play_track_version_codes(
        package_name: ptions[:package_id],
        track: "production",
      )
      puts version_codes.max
  end
```

```
appstore_version=$(bundle exec fastlane run getLatestVersion package_id:com.x8bit.bitwarden)
```

vvolkgang (changes requested):

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
+name: Publish GitHub Release as newest
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 3 * * 1-5'
```
vvolkgang, line 6: Given the scheduled trigger, if for some reason we have to change a release back to pre-release or draft, any way we can stop this workflow from re-processing the same release again?

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,92 @@
...
+
+jobs:
+  publish-release:
+    name: Publish GitHub Release
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      actions: read
```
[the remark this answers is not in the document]
michalchecinski: We need the permissions clause either on workflow or job level, by the new workflow linter rule.

.github/workflows/publish-github-release.yml
michalchecinski: Thanks. Changed it to gh cli. 

> Sending the returned json to GITHUB_OUTPUT should enable you to directly access fields in other steps without parsing fields individually. 🤔

The data is returned as a list, so I don't think it's possible.

.github/workflows/publish-github-release.yml
michalchecinski: Fixed.

.github/workflows/publish-github-release.yml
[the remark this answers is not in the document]
vvolkgang: I was referring to `actions: read` specifically, which I'm assuming came from `github-release.yml`. I'm removing it from that other workflow, just confirmed in [this run](https://github.com/bitwarden/android/actions/runs/15148111568/job/42588593310) that we don't need it, you probably don't need it here too.

.github/workflows/publish-github-release.yml
vvolkgang: 🤔 We could follow the [_version.yml ](https://github.com/bitwarden/ios/blob/main/.github/workflows/_version.yml#L109-L126) approach. The gist:
1. When this workflow changes a release, upload a json file with state info - e.g., release tag, release title, initial state (drift / pre-release) and changed to state (release)
2. Before changing a release, get the artifact from the last run of this workflow and confirm if we're about to change the same release.

.github/workflows/publish-github-release.yml
vvolkgang: You can use `--jq '....'` to retrieve the object:

`gh release list --json isDraft,name,tagName -L 1 --jq '.[]'` something like that should work.

fastlane/Fastfile
```diff
@@ -384,4 +394,14 @@ platform :android do
...
   end
+
+  desc "Get latest Authenticator published version"
+  lane :getLatestAuthenticatorVersion do |options|
+      version_codes = google_play_track_version_codes(
+        package_name: "com.bitwarden.authenticator",
+        track: "production",
+        json_key: options[:serviceCredentialsFile]
+      )
+      puts version_codes.max
```
vvolkgang, line 405: Just hit me that this should be retrieving the version code (e.g. `20100` in the [last release](https://github.com/bitwarden/android/releases/tag/v2025.4.0)) and the workflow is comparing it to the release tag (e.g. `v2025.4.0`).

.github/workflows/publish-github-release.yml
vvolkgang: @michalchecinski Was working on something unrelated that required downloading artifacts from the last run of a workflow, remembered this may be useful for you too:

`gh run download $(gh run list --workflow=build.yml --event=schedule --limit 1 --json databaseId --jq '.[0].databaseId')`

* Replacing the workflow filename with `$(basename $GITHUB_WORKFLOW)` may work but I didn't test it. 
* Filtering event name would skip `workflow_dispatch` events we may use to test the workflow.

.github/workflows/publish-github-release.yml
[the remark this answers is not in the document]
michalchecinski: Removed

.github/workflows/publish-github-release.yml
michalchecinski: I've implemented the whole setup to do this. But when I was testing it (it works as expected) I realized that the release cannot be converted into a draft once released. So this is unnecessary, it seems. I should have checked earlier 😅

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,215 @@
...
+        with:
+          name: release-info
+          run-id: ${{ steps.get_previous_run.outputs.previous_run_id }}
+
+      - name: Parse previous state
+        id: parse_previous_state
+        if: steps.get_previous_run.outputs.has_previous_run == 'true'
+        run: |
+          if [ -f "release-info.json" ]; then
+            previous_release_tag=$(jq -r '.release_tag // empty' release-info.json)
```
vvolkgang, line 74: ⛏️ You can probably use `fromJSON(...) >> GITHUB_OUTPUT`  and skip manually parsing each field.

vvolkgang (changes requested):
Some minor nitpicks but I'm still not sure how this works, what am I missing? Seeing a test run would help!

1. fastlane `version_codes.max` should be the build number (e.g. `20100`)
2. The previous value will be compared to `latest_draft_version` which is currently set to `tagName` (e.g. `v2025.4.0`)
3. Tangentially, drafts can sometimes have tags like `untagged-123912901`.

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,215 @@
...
+              exit 0
+          fi
+
+          latest_draft_version=$(jq -r '.tagName' <<< $latest_release)
+          echo "latest_draft_version=$latest_draft_version" >> $GITHUB_OUTPUT
+
+          latest_draft_name=$(jq -r '.name' <<< $latest_release)
+          echo "latest_draft_name=$latest_draft_name" >> $GITHUB_OUTPUT
+
+      - name: Get previous run ID
```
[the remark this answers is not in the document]
michalchecinski: Great idea. Done.

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,215 @@
...
+        id: get_previous_run
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          workflow_name="$(basename "$GITHUB_WORKFLOW")"
+
+          # Get the last successful scheduled run ID
+          previous_run_id=$(gh run list --workflow=$workflow_name --event=schedule --status=success --limit 1 --json databaseId --jq '.[0].databaseId // empty')
```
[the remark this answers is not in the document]
michalchecinski: Hmm, I don't get it

.github/workflows/publish-github-release.yml
michalchecinski: Can't do that. I could use the `fromJSON()` function here, but differently than your suggestion.

Use `fromJSON` instead of `jq`:
```
- name: Read JSON configuration
  id: json-reader
  run: |
    content=$(cat release-info.json)
    echo "release-info-json=$content" >> $GITHUB_OUTPUT
    
- name: Process JSON data
  run: |
    echo "Release tag: ${{ fromJSON(steps.json-reader.outputs.release-info-json).release_tag }}"
    echo "Initial state: ${{ fromJSON(steps.json-reader.outputs.release-info-json).initial_state }}"
    echo "Changed to: ${{ fromJSON(steps.json-reader.outputs.release-info-json).changed_to_state }}"   
```

I'd rather have the fields parsed as outputs to use them directly, without needing to do anything with them in the steps that use values.

michalchecinski:
@vvolkgang I've extracted the action to gh-actions repo to reuse it both here and in the iOS repo: https://github.com/bitwarden/gh-actions/pull/401

fastlane/Fastfile
```diff
@@ -384,4 +398,17 @@ platform :android do
       mapping: "authenticator/build/outputs/mapping/release/mapping.txt",
     )
   end
+
+  desc "Get latest Authenticator published version"
+  lane :getLatestAuthenticatorVersion do |options|
+      release_name = google_play_track_release_names(
+        package_name: "com.bitwarden.authenticator",
+        track: "production",
+        json_key: options[:serviceCredentialsFile]
```
vvolkgang: [From a previous comment,](https://github.com/bitwarden/android/pull/5219#discussion_r2097961272), we don't need `json_key` / `serviceCredentialsFile` here.

fastlane/Fastfile
```diff
@@ -265,6 +265,18 @@ platform :android do
     releaseNotes
   end
 
+  desc "Get latest published version for a given package name and Play Store track"
+  lane :getLatestPlayStoreVersion do |options|
```
vvolkgang, line 269: @michalchecinski fyi I updated this to consolidate lanes and remove the use of `json_key` / `serviceCredentialsFile`.

fastlane/Fastfile
```diff
@@ -265,6 +265,18 @@ platform :android do
...
+      release_name = google_play_track_release_names(
```
vvolkgang: For Google Play Store, we should use `version_code` as you previously had, parsing them from the GitHub Release title. Multiple builds may have the same name but they can't have the same `version_code`.

.github/workflows/publish-github-release-passoword-manager.yml
```diff
@@ -0,0 +1,22 @@
+name: Publish Password Manager GitHub Release as newest
```
vvolkgang: 🤔 We don't really need multiple files and thinking ahead of extending this to other repos, it won't scale. We can just pivot the main workflow to "handle all draft releases" and log an array of published releases, assuming that if the array is empty then we didn't publish anything.

.github/workflows/_publish-github-release.yml
```diff
@@ -0,0 +1,255 @@
...
+      actions: read
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Get latest draft name
+        id: get_latest_draft
```
vvolkgang: 1. 🎨 Other steps are running even when this is empty. Instead of adding `if: ...` lines back to every single step, we could move this to it's own job?
2.  [Followup](https://github.com/bitwarden/android/pull/5219/files#r2122013536), we can remove the need for multiple workflows by handling a list of draft releases instead, parsing the App from the Release Title.

fastlane/Fastfile
michalchecinski: @vvolkgang And what about the iOS?

fastlane/Fastfile
michalchecinski: Hmm, so how will the lane authenticate to the store? Sorry, I probably don't fully get how the Fastfile works 😅

.github/workflows/publish-github-release-passoword-manager.yml
michalchecinski: I'll try to do that. Another solution could be to publish one artifact per release type, instead of the array.

Also, this reusable workflow won't work for other repo types, as it uses Fastlane to get the data. We'll probably need to make the workflow more robust (maybe even a custom python script) to handle other type of projects.

fastlane/Fastfile
vvolkgang: iOS needs the combination version name + version number to be unique, which we get from the release title too. A given version number (e.g. `1000`) can exist in multiple version names (e.g. `2025.5.0`, `2025.5.1`). That said, I don't know what info you can retrieve from AppStoreConnect.

.github/workflows/publish-github-release-passoword-manager.yml
vvolkgang: Yep, good idea!

Fair. I thought your `check_release_command` would deal with that but we can cross that bridge when we get there.

fastlane/Fastfile
michalchecinski: @vvolkgang Please check if now it's okay. I want to check both release name and version for both systems to have the consolidated approach.

.github/workflows/publish-github-release.yml
```diff
@@ -1,12 +1,100 @@
...
+  #         echo "version_number: $version_number"
+
+  #         OUTPUT=$(bundle exec fastlane getLatestPlayStoreVersion package_name:com.bitwarden.authenticator track:production)
+  #         version_name=$(echo "$OUTPUT" | grep 'version_name: ' | sed 's/.*version_name: //')
+  #         version_number=$(echo "$OUTPUT" | grep 'version_number: ' | cut -d' ' -f3)
+
+  #         echo "Retrieved BWA version info:"
+  #         echo
+  #         echo "version_name: $version_name"
+  #         echo "version_number: $version_number"
```
brian-livefront: Can this commented-out code be removed?

michalchecinski: Yep, this sould be removed. Thanks for catchin it!

.github/workflows/publish-github-release.yml
```diff
@@ -1,12 +1,35 @@
-name: Publish GitHub Release as newest
+name: Publish Password Manager GitHub Release as newest
```
david-livefront: Should we mention the Authenticator here too?

fastlane/Fastfile
```diff
@@ -265,6 +265,40 @@ platform :android do
...
+    track = options[:track]
+    # Hardcoding paths for now, it will simplify transitioning to the new build .env files
+    case package_name
+    when "com.x8bit.bitwarden", "com.x8bit.bitwarden.beta"
+      json_key = "secrets/play_creds.json"
+    when "com.bitwarden.authenticator"
+      json_key = "secrets/authenticator_play_store-creds.json"
+    else
+      UI.important "Unexpected package name: #{package_name}, using default play store json key"
+      json_key = "secrets/play_creds.json"
```
david-livefront, line 280: Should we just fail here instead?

.github/workflows/publish-github-release.yml
michalchecinski: Yep, fixed.

fastlane/Fastfile
michalchecinski: @vvolkgang what's your view?

vvolkgang: These defaults have been helpful dealing with silly typos while iterating given the slow turnaround of workflows. Not a hill I would die on tho, feel free to change.

david-livefront: That makes sense and I am also not dying on this hill either. It just felt weird to sign an unknown package with the Play Store key

vvolkgang (approved):

fastlane/Fastfile
vvolkgang: fair enough. No package signing here, it's the API key required to retrieve info from the store.

.github/workflows/publish-github-release.yml
```diff
@@ -1,14 +1,35 @@
...
 jobs:
-  stub:
-    runs-on: ubuntu-24.04
-    name: Stub
-    steps:
-      - name: Stub
-        run: echo "This is a stub job to trigger the workflow."
+  publish-release-password-manager:
+    name: Publish Password Manager Release
+    uses: bitwarden/gh-actions/.github/workflows/_publish-mobile-github-release.yml@BRE-769-Use-Fastlane-to-keep-github-releases-in-sync-with-mobile-deploy-versions
```
SaintPatrck: Should this be updated so it's not explicitly referencing the working branch? This branch will not exist after the PR merges.

michalchecinski: Great finding!

SaintPatrck (approved):

PR #5423, opened 2025-06-25
Bump authenticatorbridge version to 1.1.0

SaintPatrck:
## 🎟️ Tracking

N/A

## 📔 Objective

Updated `authenticatorbridge` AAR files in `app` and `authenticator` modules.

Updated CHANGELOG to include notes on changes included in the release.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


authenticatorbridge/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ v1.1.0 (pending)
...
+v1.1.0
+--------
+
+### API Changes
+
+- No API changes were made in this release.
+
+### Other changes
+
+- Replace internal Build version check functions with `isBuildVersionAtLeast` function in `core` module.
```
brian-livefront: Does this change any internal functionality or is it more of a refactor? I'm just curious if it makes sense to hold on releasing a new version until there are functional changes.

SaintPatrck: It's just a refactor. No functional changes. I was holding this until we knew if more changes are required for the other Authenticator tickets we discussed. Sounds like there will not be, so I'm for not releasing.

brian-livefront: OK sounds good. If that's the case, do you want to update this PR to add this changelog documentation to then pending list? Or do we even need to mention this internal change if there's no behavior difference?

authenticatorbridge/README.md
```diff
@@ -47,13 +47,22 @@ Some things to note:
 
 ## Dependencies
 
+> [!IMPORTANT]
+> The SDK does not come packaged with these dependencies, so consumers of the SDK must provide them.
+
+### Project Dependencies
```
brian-livefront: What's the difference between a "project dependency" and an "application dependency"?

authenticatorbridge/README.md
```diff
@@ -47,13 +47,22 @@ Some things to note:
 
 ## Dependencies
 
+> [!IMPORTANT]
+> The SDK does not come packaged with these dependencies, so consumers of the SDK must provide them.
```
brian-livefront, line 51: It might be worth looking into providing this as in a local maven dependency with a POM file at some point. That would solve this transitive dependency issue. This is something we did on another project recently that @david-livefront would have some background on.

authenticatorbridge/README.md
SaintPatrck: I was going back-and-forth on whether to list `core` in `Application Dependencies` or not. I landed on `Project` to distinguish it from the third-party maven dependencies.

`annotations` is not listed either. 

I'm open to suggestion if you think there's a better way to express that, or I can expand on the description.

authenticatorbridge/README.md
SaintPatrck: I agree. I've done the same in few other projects. With Authenticator finally getting some love we can justify setting that up.

The only potential impediment is the SDK is dependent on our other modules. If we make this SDK a local maven dependency we may have to do the same with the the other modules (or at least `core` and `annotations`). 

I'll get a ticket created to track/investigate this.

authenticatorbridge/README.md
brian-livefront: Gotcha ok. So this is listing the module dependencies. Yeah I guess if we wanted to call that out in a separate section makes sense. Maybe calling it `Module Dependencies` would be a bit more clear? If we do that here you'd probably want to do that with the other modules as well.

authenticatorbridge/README.md
brian-livefront: Sounds good, thanks Patrick! That's a really good point about the local module dependencies....worth investigating a little at least.

brian-livefront: I wonder if this new fused library support would help? https://developer.android.com/build/publish-library/fused-library

SaintPatrck: Nice find! That looks very promising. I wish it was available a few years ago. 😆

PR #5462, opened 2025-07-02
[PM-22640] Re-added isScreenCaptureAllowed to the MainViewModel state

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22640

## 📔 Objective

`isScreenCaptureAllowed` needs to be present on the `MainViewModel` state in order to when `MainActivity` is recreated  it correctly updates the flags.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
@@ -84,7 +84,7 @@ class MainActivity : AppCompatActivity() {
             val navController = rememberBitwardenNavController(name = "MainActivity")
             SetupEventsEffect(navController = navController)
             val state by mainViewModel.stateFlow.collectAsStateWithLifecycle()
-            handleScreenCaptureSettingChange(isScreenCaptureAllowed = state.isScreenCaptureAllowed)
+            updateScreenCapture(isScreenCaptureAllowed = state.isScreenCaptureAllowed)
             LocalManagerProvider(featureFlagsState = state.featureFlagsState) {
                 ObserveScreenDataEffect(
                     onDataUpdate = remember(mainViewModel) {
@@ -186,10 +186,6 @@ class MainActivity : AppCompatActivity() {
                 }
 
                 is MainEvent.UpdateAppTheme -> AppCompatDelegate.setDefaultNightMode(event.osTheme)
-
-                is MainEvent.ScreenCaptureSettingChange -> {
-                    handleScreenCaptureSettingChange(isScreenCaptureAllowed = event.isAllowed)
-                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -84,6 +84,7 @@ class MainActivity : AppCompatActivity() {
             val navController = rememberBitwardenNavController(name = "MainActivity")
             SetupEventsEffect(navController = navController)
             val state by mainViewModel.stateFlow.collectAsStateWithLifecycle()
+            handleScreenCaptureSettingChange(isScreenCaptureAllowed = state.isScreenCaptureAllowed)
```
david-livefront: Interesting, so why do we need the event if this works?

aj-rosado: The event is to ensure this is updated when the user changes the setting.
Without the event it would only apply when the activity was recreated.

david-livefront: The state should do that as well, are you saying that does not work?

aj-rosado: Without state, it would apply the flags the first time it runs but when the MainActivity is recreated, it would not add the flags again

david-livefront: I understand, but my question why do we need the event? When the state changes, it also updates the setting.

Is that occurring?

aj-rosado:
```
private fun handleScreenCaptureSettingChange(isScreenCaptureAllowed: Boolean) {
        if (isScreenCaptureAllowed) {
            window.clearFlags(WindowManager.LayoutParams.FLAG_SECURE)
        } else {
            window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
        }
    }
```

    
  Both setting and state are being updated but are not adding/clearing this flag. The event is making sure this is added/cleared from the `MainActivity`

david-livefront: When the state changes the entire `setContent` should be rerun and the window flags are updated value should be applied.

I just tested this which the event disabled and it works fine without the event. Can you explain the problem that the event solves?

aj-rosado: you are correct the event is not doing anything here.

The only necessary change is `bufferedMutableSharedFlow` on the SettingsDiskSource.

I am going to revert the changes.
Thank you!

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
```diff
@@ -93,7 +93,7 @@ class SettingsDiskSourceImpl(
 
     private val mutableHasSeenGeneratorCoachMarkFlow = bufferedMutableSharedFlow<Boolean?>()
 
-    private val mutableScreenCaptureAllowedFlow = bufferedMutableSharedFlow<Boolean?>(replay = 1)
+    private val mutableScreenCaptureAllowedFlow = bufferedMutableSharedFlow<Boolean?>()
```
david-livefront, line 96: 👍

aj-rosado:
Thanks @david-livefront

PR #5617, opened 2025-07-29
PM-15229: Accomidate system bars on specific Android 15 revisions

david-livefront:
## 🎟️ Tracking

PM-15229

## 📔 Objective

This PR updates the way we recreate the activity on Android 15 in order to avoid various UI issues.

* The windowBackground has been set for Android 15 specifically and matches the secondary background color of the app. This makes the color issues less jaring to users.
* The processes that trigger a recreate have been updated to trigger less frequently and avoid double triggers.
* The way we restart the activity on specific revisions of Android 15 has changed to avoid the `recreate` API completely (source of the issue).

Bug tracker for the `recreate` API issue: https://issuetracker.google.com/issues/370180732

## 📸 Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bf93287a-9809-42ea-8332-bfd34af06a60" width="300"/> | <video src="https://github.com/user-attachments/assets/251e34f9-d4b7-42eb-9380-62a29099aa24" width="300"/> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
@@ -15,6 +15,7 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
+import androidx.core.app.ActivityCompat
 import androidx.core.os.LocaleListCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
@@ -42,6 +43,8 @@ import com.x8bit.bitwarden.ui.platform.util.appLanguage
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.flow.map
 import javax.inject.Inject
+
+private const val ANDROID_15_BUG_MAX_REVISION: Int = 241007
 
 /**
  * Primary entry point for the application.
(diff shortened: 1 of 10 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -135,36 +139,23 @@ class MainViewModel @Inject constructor(
...
+                // pending account state (which acts like switching to a temporary user).
+                .map { it?.activeUserId to it?.hasPendingAccountAddition }
+                .distinctUntilChanged(),
+            vaultRepository
+                .vaultStateEventFlow
+                .filter { it is VaultStateEvent.Locked },
+        )
+            // This debounce ensure we do not emit multiple times rapidly and also acts as a short
+            // delay to give animations time to finish (ex: account switcher).
+            .debounce(timeoutMillis = ANIMATION_DEBOUNCE_DELAY_MS)
```
david-livefront, line 156: Merging these events and then debouncing them performs double duty, the delay allows the animation to finish but it also ensure the flash only occurs once, without this change, we would sometimes get double flashes.

app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -221,7 +224,35 @@ class MainActivity : AppCompatActivity() {
...
+            // This is done to avoid a bug in specific older revisions of Android 15. The bug has
+            // been fixed but certain phones that are no longer supported will never get the fix.
+            // The OS bug is tracked here: https://issuetracker.google.com/issues/370180732
+            startActivity(
+                Intent
+                    .makeMainActivity(componentName)
+                    .addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION),
+            )
+            finish()
+            overrideActivityTransition(OVERRIDE_TRANSITION_CLOSE, 0, 0)
```
david-livefront, line 252: With the other changes made in this PR, we could drop this change. This would remove the flicker but there would be statusbar and navigationbar issues when using dynamic colors, the edge-to-edge would not work properly, and the in-app theme selector would not change the statusbar and navigation bar colors.

vvolkgang: +1 on dropping all of this if possible, it'll be harder for us to cleanup later down the line. The theme defaults and debounce are sensible, I'm assuming the debounce will also be useful on slower devices / high load scenarios. 

Incredible detective work here! :godmode:

david-livefront: Correct, I don't think we will ever be able to remove this code until our minimum is Android 16 or we remove the recreate call. Both of which are going to be awhile 😭 .

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #5682, opened 2025-08-12
[PM-24577] Provision SDK with AccountKeys

SaintPatrck:
## 🎟️ Tracking

PM-24577

## 📔 Objective

This commit refactors the vault unlocking logic to utilize AEAD (Authenticated Encryption with
Associated Data) keys if they are available and the `EnrollAeadOnKeyRotation` feature flag is
enabled.

Specifically, the following changes were made:

- The `VaultLockManager.unlockVault` method and its implementations now accept optional `signingKey`
  and `securityState` parameters. These parameters are used to initialize the user's crypto with
  AEAD keys if provided.
- The `AuthRepositoryImpl` and `VaultRepositoryImpl` were updated to retrieve the `signingKey` and
  `securityState` from `AuthDiskSource` when the `EnrollAeadOnKeyRotation` feature flag is active.
  These keys are then passed to the `VaultLockManager` during the vault unlocking process.
- The `AuthenticatorBridgeRepositoryImpl` was also updated to use AEAD keys from `AuthDiskSource`
  when initializing crypto if the feature flag is enabled.
- Corresponding tests were updated to reflect these changes, including mocking the
  `FeatureFlagManager` and providing the new parameters in `unlockVault` calls.
- Dependency injection modules were updated to provide the `FeatureFlagManager` to the necessary
  repositories and managers.

## ⏰ Reminders before review

- CI builds passed
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -563,6 +566,7 @@ class AuthRepositoryImpl(
                             .map { keys }
                     }
                     .onSuccess { keys ->
+                        // TODO: Should this overwrite accountKeys private key?
```
SaintPatrck: ❓ Should the Org private key overwrite the accountKeys private key received from Sync?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -1197,6 +1208,7 @@ class AuthRepositoryImpl(
                     )
                     .onSuccess {
                         rsaKeys?.private?.let {
+                            // TODO: Should this overwrite the accountToken private key?
```
SaintPatrck: ❓ Should this RSA private key overwrite the accountKeys private key received from Sync?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -1805,6 +1817,7 @@ class AuthRepositoryImpl(
         loginResponse.privateKey?.let {
             // Only set the value if it's present, since we may have set it already
             // when we completed the key connector conversion.
+            // TODO: Should this overwrite accountKeys private key?
```
SaintPatrck: ❓ In the common login handler, should the login response private key overwrite the accountKeys private key received from Sync?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -1942,6 +1959,7 @@ class AuthRepositoryImpl(
                             userId = profile.userId,
                             userKey = keyConnectorResponse.encryptedUserKey,
                         )
+                        // TODO: Should this overwrite the accountKeys private key?
```
SaintPatrck: ❓ During KeyConnector login, should the KeyConnector private key overwrite the accountKeys private key received from Sync?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -1038,17 +1045,26 @@ class VaultRepositoryImpl(
     ): VaultUnlockResult {
         val account = authDiskSource.userState?.accounts?.get(userId)
             ?: return VaultUnlockResult.InvalidStateError(error = NoActiveUserException())
-        val privateKey = authDiskSource.getPrivateKey(userId = userId)
+        val accountKeys = authDiskSource.getAccountKeys(userId = userId)
+            ?.takeIf { featureFlagManager.getFeatureFlag(FlagKey.EnrollAeadOnKeyRotation) }
+        val privateKey = accountKeys
+            ?.publicKeyEncryptionKeyPair
+            ?.wrappedPrivateKey
+            ?: authDiskSource.getPrivateKey(userId = userId)
```
quexten: Please do not feature flag this code. The flag is meant specifically just to guard whether a migration happens on user-key rotation / if mobile wants the usage of account keys feature flagged, we should introduce a separate flag used to roll this out, before we enable the migrations via `EnrollAeadOnKeyRotation`.

If we used the flag as-is, this would mean that KM stopping the roll-out, by disabling the flag, would lead to already migrated users to only provide the private key + the new AEAD symmetric key, but not the signing key / security state. This partial v2 state is not allowed and would lead to the SDK failing to unlock.


```suggestion
        val accountKeys = authDiskSource.getAccountKeys(userId = userId)?;
        val privateKey = accountKeys
            ?.publicKeyEncryptionKeyPair
            ?.wrappedPrivateKey
            ?: authDiskSource.getPrivateKey(userId = userId)
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultLockManagerImpl.kt
```diff
@@ -643,16 +648,23 @@ class VaultLockManagerImpl(
     ): VaultUnlockResult {
         val account = authDiskSource.userState?.accounts?.get(userId)
             ?: return VaultUnlockResult.InvalidStateError(error = NoActiveUserException())
-        val privateKey = authDiskSource.getPrivateKey(userId = userId)
+        val accountKeys = authDiskSource.getAccountKeys(userId = userId)
```
quexten, line 650: Same comment as below; Please use a separate feature flag for the mobile feature or don't flag this.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/repository/AuthenticatorBridgeRepositoryImpl.kt
```diff
@@ -126,9 +129,19 @@ class AuthenticatorBridgeRepositoryImpl(
         account: AccountJson,
         decryptedUserKey: String,
     ): VaultUnlockResult {
-        val privateKey = authDiskSource
-            .getPrivateKey(userId = userId)
-            ?: return VaultUnlockResult.InvalidStateError(MissingPropertyException("Private key"))
+        val accountKeys = authDiskSource.getAccountKeys(userId = userId)
```
quexten, line 129: Same comment as below, please use a separate feature flag, or don't flag it.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: They are one and the same. However, all of:
- private key
- signing key
- security state
need to be provided together for a v2-crypto user. So putting null values here, would not work for v2 users, and would fail to initialize the SDK. I'll update the server PR shortly to include the signing key & security state in these token responses too.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: Generating new private keys here, and sending them to the server is not correct, and on the server is rejected: 

https://github.com/bitwarden/server/blob/87877aeb3da7c8b953f89abf0a6fbb7325e4f23b/src/Api/Auth/Models/Request/Accounts/SetPasswordRequestModel.cs#L37

https://github.com/bitwarden/server/blob/87877aeb3da7c8b953f89abf0a6fbb7325e4f23b/src/Core/Auth/Models/Api/Request/Accounts/KeysRequestModel.cs#L36

I'll get back in a bit here to check what the actual right approach should be.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: As far as I understand, the keys here are the *user public-key encryption keypair*, passed through from: `makeRegisterTdeKeysAndUnlockVault`. The initial creation of a tde user is still initialized with cryptov1, and we have not yet started the epic moving any registration flow over to v2, so in this case, we don't need changes here. These users are expected to only have a private key, and a symmetric account encryption key (not the new AEAD kind, but the old AES256-CBC-HMAC).

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: The keys are the same. However, the server should be returning signing keys and security state here too, and we should set "accountKeys" here.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: ^ LoginResponse should now include the accountKeys, as per: https://github.com/bitwarden/server/pull/5982#issuecomment-3187903972. We can follow the same strategy as for sync - if it is there use it, if it is not use the existing "private key".

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
SaintPatrck: I've removed the feature flag. Thanks for calling that out.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -1870,6 +1890,8 @@ class AuthRepositoryImpl(
                             masterKey = it.masterKey,
                             userKey = key,
                         ),
+                        securityState = null,
+                        signingKey = null,
```
david-livefront: Shouldn't we be fetching this data from the `loginResponse`?

david-livefront: Also, should we be using `privateKeyOrNull()` up above here?

SaintPatrck: Yup. Missed this one. Thank you.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
SaintPatrck: @quexten Any concern with approving the current changes and investigating this separately?

quexten: Yeah, that's fine. We just need to check this before releasing user crypto v2.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -1952,13 +1984,15 @@ class AuthRepositoryImpl(
...
         // Attempt to unlock the vault with auth request if possible.
         // These values will only be null during the Just-in-Time provisioning flow.
-        val privateKey = loginResponse.privateKey
+        val privateKey = loginResponse.privateKeyOrNull()
         val key = loginResponse.key
         if (privateKey != null && key != null) {
             deviceData?.let { model ->
                 return unlockVault(
                     accountProfile = profile,
                     privateKey = privateKey,
```
quexten, line 1993: I believe this should also use conditionally the `accountKeys?.publicKeyEncryptionKeyPair?.wrappedPrivateKey / private key since privateKey is now deprecated.

quexten: Nevermind, I see this function's implementation defined below.

david-livefront (approved):

PR #5871, opened 2025-09-11
Refactor Vault Sync Logic into VaultSyncManager

SaintPatrck:
## 🎟️ Tracking

TBD

## 📔 Objective

This change introduces `VaultSyncManager` to encapsulate the vault synchronization logic previously handled within `VaultRepositoryImpl`.

Key changes:
- Created `VaultSyncManager` interface and `VaultSyncManagerImpl` implementation.
- Moved sync-related logic, including security stamp checks, organization crypto initialization, and data storage, from `VaultRepositoryImpl` to `VaultSyncManagerImpl`.
- `VaultRepositoryImpl` now delegates sync operations to `VaultSyncManager`.
- Introduced `SecurityStampMismatchException` for clearer error handling.
- Moved `SyncVaultDataResult` to `com.x8bit.bitwarden.data.vault.manager.model`.
- Updated DI modules and tests to reflect these changes.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
@@ -1437,7 +1437,10 @@ class VaultRepositoryImpl(
     }
     //endregion Push Notification helpers
 
-    private suspend fun syncInternal(userId: String, forced: Boolean): SyncVaultDataResult =
+    private suspend fun syncInternal(
+        userId: String,
+        forced: Boolean,
+    ): SyncVaultDataResult =
         vaultSyncManager
             .sync(userId = userId, forced = forced)
             .also { result ->


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/error/SecurityStampMismatchException.kt
```diff
@@ -0,0 +1,6 @@
+package com.x8bit.bitwarden.data.platform.error
+
+/**
+ * An exception indicating that the security stamps for the current user do not match.
+ */
+class SecurityStampMismatchException : IllegalStateException("Security stamps do not match!")
```
brian-livefront, line 6: Do we use custom exceptions elsewhere? One of the reasons I'm wondering is because having a custom exception makes it tempting to try and use this for a control flow, rather than just modeling this as an expected state of sealed class result. It looks like maybe right now you're just using this for logging purposes mainly though?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManager.kt
```diff
@@ -0,0 +1,29 @@
...
+ * This interface provides a way to trigger a sync process, which updates the local
+ * database with the latest changes from the server.
+ */
+interface VaultSyncManager {
+    /**
+     * Initiates a synchronization process for the user's vault data.
+     *
+     * This function fetches the latest data from the remote server and updates the local
+     * vault cache. It can be a standard sync or a "forced" sync, which typically
+     *  bypasses local cache checks and fetches everything anew.
```
brian-livefront: I _think_ there's an extra space here:

```suggestion
     * bypasses local cache checks and fetches everything anew.
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManager.kt
```diff
@@ -0,0 +1,29 @@
...
+     * @param forced If true, performs a full, forced synchronization, ignoring any recent sync
+     * timestamps. If false, performs a standard incremental sync.
+     *
+     * @return A [SyncVaultDataResult] indicating the outcome of the synchronization, such as
+     * success or failure with details.
+     */
+    suspend fun sync(
+        userId: String,
+        forced: Boolean,
+    ): SyncVaultDataResult
```
brian-livefront, line 28: I'm curious if it makes sense to pull `syncIfNecessary` into this as well. Maybe not, but I'm curious to hear your thinking on keeping that inside `VaultRepository(Impl)`.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+     * vault cache. It can be a standard sync or a "forced" sync, which typically
+     *  bypasses local cache checks and fetches everything anew.
+     *
+     * @param userId The unique identifier of the user whose vault is to be synchronized.
+     * @param forced If true, performs a full, forced synchronization, ignoring any recent sync
+     * timestamps. If false, performs a standard incremental sync.
+     *
+     * @return A [SyncVaultDataResult] indicating the outcome of the synchronization, such as
+     * success or failure with details.
+     */
```
brian-livefront: We shouldn't need to duplicate the documentation that is already on the interface right?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/model/SyncVaultDataResult.kt
```diff
@@ -16,5 +16,5 @@ sealed class SyncVaultDataResult {
      *
      * @property throwable The exception that caused the failure, if any.
      */
-    data class Error(val throwable: Throwable?) : SyncVaultDataResult()
+    data class Error(val throwable: Throwable) : SyncVaultDataResult()
```
brian-livefront, line 19: If this isn't nullable we should change the docs above. However, I am curious why this part changed....or what it meant previously for the error to be `null` actually.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -364,7 +358,7 @@ class VaultRepositoryImpl(
         mutableFoldersStateFlow.updateToPendingOrLoading()
         mutableCollectionsStateFlow.updateToPendingOrLoading()
         mutableSendDataStateFlow.updateToPendingOrLoading()
-        syncJob = ioScope.launch { syncInternal(userId = userId, forced = forced) }
+        syncJob = ioScope.launch { sync(userId = userId, forced = forced) }
```
brian-livefront: 👍

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -378,14 +372,15 @@ class VaultRepositoryImpl(
         if (lastSyncInstant == null ||
             currentInstant.isAfter(lastSyncInstant.plus(30, ChronoUnit.MINUTES))
         ) {
-            sync()
+            sync(forced = false)
```
brian-livefront, line 375: 👍

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -1500,90 +1437,13 @@ class VaultRepositoryImpl(
...
-                            }
-                        },
-                        onFailure = {
-                            updateVaultStateFlowsToError(throwable = it)
-                            return SyncVaultDataResult.Error(throwable = it)
-                        },
-                    )
+    private suspend fun sync(userId: String, forced: Boolean): SyncVaultDataResult =
+        vaultSyncManager.sync(userId = userId, forced = forced)
+            .also { result ->
```
brian-livefront, line 1446: Minor, but I think formatting-wise we'd typically have

```kotlin
        vaultSyncManager
            .sync(userId = userId, forced = forced)
            .also { result ->
                // ...
            }
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -1500,90 +1437,13 @@ class VaultRepositoryImpl(
...
+                if (result is SyncVaultDataResult.Error) {
+                    updateVaultStateFlowsToError(throwable = result.throwable)
+                }
```
brian-livefront, line 1449: 👍

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -1500,90 +1437,13 @@ class VaultRepositoryImpl(
...
-                                    ?.isNotEmpty() == true
-                                return SyncVaultDataResult.Success(itemsAvailable = itemsAvailable)
-                            }
-                        },
-                        onFailure = {
-                            updateVaultStateFlowsToError(throwable = it)
-                            return SyncVaultDataResult.Error(throwable = it)
-                        },
-                    )
+    private suspend fun sync(userId: String, forced: Boolean): SyncVaultDataResult =
```
brian-livefront: Just curious...any reason not to just keep the name `syncInternal` and the signature as multi-line? Smaller diff that way.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
brian-livefront: We'll need to be very careful to make sure the folks call this function rather than `vaultSyncManager.sync` directly if they ever need to add sync calls in this class now.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/error/SecurityStampMismatchException.kt
SaintPatrck: We have a couple other exceptions defined; `MissingPropertyException` and `NoActiveUserException`. They're all being used for error reporting purposes. Now that we present a "Share error logs" action in error dialogs we bubble the exception up to the UI for easy sharing.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
david-livefront: It will also help with disambiguating between the `sync` call on the `VaultSyncManager`

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/error/SecurityStampMismatchException.kt
brian-livefront: OK sounds good!

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
brian-livefront: Right I was thinking that as well. When I first saw this being called elsewhere I thought that maybe the repository was implementing the new interface. I think clearly disambiguating would be a good idea.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManager.kt
SaintPatrck: I went back and forth on that internally. My thinking for leaving the `ifNecessary` logic in VaultRepo came down to; `SyncManager` is responsible for _performing_ sync while the `VaultRepo` is responsible for determining _if_ sync should be performed. I see valid argument for moving it into SyncManager though. Makes VaultRepo more lightweight, which I'm totally for too. If you or @david-livefront feel strongly one way or the other, we can take that approach.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/model/SyncVaultDataResult.kt
SaintPatrck: This was originally nullable because we were not propagating the timestamp mismatch for error reporting. Previously, passing `null` prevented the `Send error report` action from being presented. Now that we have an explicit exception for timestamp mismatch there are no calls to this without a Throwable.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
SaintPatrck: Indeed. It would be nice if we could make SyncManager package private in some manner to prevent accidental usage. I can update docs, if you think that would help prevent calling VaultSyncManager directly.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
SaintPatrck: Sounds good. Reverted back to `syncInternal`.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
SaintPatrck: You're right. Thanks!

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManager.kt
david-livefront: While I do like the idea of having _all_ the sync stuff in one place, that might get a little messy since we have to ensure all the flows get updated properly.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -378,12 +372,13 @@ class VaultRepositoryImpl(
...
         ) {
-            sync()
+            sync(forced = false)
         }
     }
 
     override suspend fun syncForResult(): SyncVaultDataResult {
-        val userId = activeUserId ?: return SyncVaultDataResult.Error(throwable = null)
+        val userId = activeUserId
+            ?: return SyncVaultDataResult.Error(throwable = NoActiveUserException())
```
david-livefront, line 381: 👍

SaintPatrck:
Thanks @david-livefront!

@brian-livefront any other comments/concerns/questions before this merges?

brian-livefront (approved):

PR #5982, opened 2025-10-07
Improve CXF message handling

SaintPatrck:
## 🎟️ Tracking

PM-26599

## 📔 Objective

Improve handling of the FIDO Credential Exchange (CXF) message formats.

It introduces the necessary data models for the CXF protocol, including `CredentialExchangeProtocolMessage`, `CredentialExchangeExportResponse`, and `CredentialExchangeVersion`.

The `CredentialExchangeCompletionManager` is updated to build and encode the export response according to the CXF specification, including the header and base64-encoded payload. The import logic in `CredentialExchangeImportManager` is also updated to parse this new format, validate versioning, and decode the payload before processing.

### Import to Bitwarden

https://github.com/user-attachments/assets/0efcfd50-f6f6-404b-bc37-9745be3fe99f

### Import from Bitwarden

https://github.com/user-attachments/assets/1bcbb31a-e970-40a2-b31e-3b7ef30bbf10

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
```diff
@@ -19,58 +34,105 @@ class CredentialExchangeImportManagerImpl(
...
-                            ImportCredentialsUnknownErrorException().asFailure()
-                        }
+    ): ImportCxfPayloadResult {
+        val credentialExchangeExportResult = json
+            .decodeFromStringOrNull<CredentialExchangeProtocolMessage>(payload)
+            ?: return ImportCxfPayloadResult.Error(
+                ImportCredentialsInvalidJsonException("Invalid CXP JSON."),
+            )
+
+        if (SUPPORTED_CXP_FORMAT_VERSIONS[credentialExchangeExportResult.version.major]
```
david-livefront, line 51: How does this work exactly, the major version contains a set of supported minor versions?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
```diff
@@ -19,58 +34,105 @@ class CredentialExchangeImportManagerImpl(
...
+                ImportCredentialsInvalidJsonException("Unsupported CXF version."),
+            )
+        }
 
-                        ImportCiphersResponseJson.Success -> {
-                            ImportCxfPayloadResult
-                                .Success(itemCount = cipherList.size)
-                                .asSuccess()
+        val accountsJson = try {
+            json.encodeToString(exportResponse.accounts.firstOrNull())
```
david-livefront, line 86: Is it OK to encode null here?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
SaintPatrck: Correct. This may change later, once we determine how much of the validation and payload mutation we can offload to the SDK. IMO, the SDK should be handling all of this, but it only accepts the `account` JSON at the moment. 😞

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
SaintPatrck: We could check it's not empty and exit early. This effectively achieves the same because we're catching `SerializationExceptions` and propagating it to the caller.

cxf/src/main/kotlin/com/bitwarden/cxf/model/CredentialExchangeExportResponse.kt
```diff
@@ -0,0 +1,55 @@
...
+     * [Item](https://fidoalliance.org/specs/cx/cxf-v1.0-rd-20250313.html#entity-item) objects
+     * associated with this account.
+     */
+    @Serializable
+    data class Account(
+        val id: String,
+        val username: String,
+        val email: String,
+        val collections: JsonArray,
+        val items: JsonArray,
```
david-livefront, line 53: Do we not care about modeling these?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
david-livefront: Gotcha

cxf/src/main/kotlin/com/bitwarden/cxf/model/CredentialExchangeExportResponse.kt
SaintPatrck: No. We have to extract a single "account" JSON object from the array because the SDK only consumes a single account (for now). We don't really care about the actual properties, just need the JSON. This seems like the better approach so that we don't have to continue updating the `collections` and `items` models as the Spec changes.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerImpl.kt
david-livefront: Parsing null will not throw, it will make the string `null`

SaintPatrck: Ah. True. I'll check if it's empty first, in that case.

david-livefront (approved):

PR #5825, opened 2025-09-03
[PM-25423] Add BidiTextManager for handling bidirectional text

SaintPatrck:
## 🎟️ Tracking

PM-25423

## 📔 Objective

This commit introduces `BidiTextManager` to manage and ensure correct display of text containing both LTR and RTL characters. It includes:
- `BidiTextManager` interface with a `unicodeWrap` method.
- `BidiTextManagerImpl` as the default implementation, using `androidx.core.text.BidiFormatter`.
- `LocalBidiTextManager` CompositionLocal for providing the manager in Compose UI.

## 📸 Screenshots

Coming soon!

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #6090, opened 2025-10-28
[PM-25423] Force LTR direction for password and sensitive fields

SaintPatrck:
## 🎟️ Tracking

PM-25423

## 📔 Objective

This commit introduces a `VisualTransformation` to enforce Left-To-Right (LTR) text direction in password fields and other sensitive inputs. This ensures consistent layout and prevents rendering issues, particularly in Right-To-Left (RTL) locales where characters like asterisks might otherwise appear incorrectly aligned.

A `CompoundVisualTransformation` utility has also been added to allow chaining multiple `VisualTransformation`s together.

Specific changes:
- Created `ForceLtrVisualTransformation` to wrap text with LTR Unicode control characters.
- Created `CompoundVisualTransformation` to combine multiple transformations, preserving correct offset mapping.
- Applied `forceLtrVisualTransformation` to `BitwardenPasswordField` and `BitwardenHiddenPasswordField`, combining it with the existing `PasswordVisualTransformation` or `nonLetterColorVisualTransformation`.
- Enforced LTR direction on generated passwords in the Password History screen.
- Applied LTR transformation to sensitive fields in the "Identity" and "Login" item views, such as TOTP, URI, SSN, and passport number.

## 📸 Screenshots

Coming soon!

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/passwordhistory/PasswordHistoryScreenTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/passwordhistory/PasswordHistoryScreenTest.kt
@@ -100,7 +100,7 @@ class PasswordHistoryScreenTest : BitwardenComposeTest() {
             )
         }
 
-        composeTestRule.onNodeWithText("${LRO}${password.password}$PDF").assertIsDisplayed()
+        composeTestRule.onNodeWithText("$LRO${password.password}$PDF").assertIsDisplayed()
         composeTestRule.onNodeWithContentDescription("Copy").performClick()
 
         verify {


ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenPasswordField.kt
```diff
@@ -140,6 +141,21 @@ fun BitwardenPasswordField(
         TextToolbarType.NONE -> BitwardenEmptyTextToolbar
     }
     var lastTextValue by remember(value) { mutableStateOf(value = value) }
+
+    val visualTransformation = when {
+        !showPassword -> compoundVisualTransformation(
+            PasswordVisualTransformation(),
+            forceLtrVisualTransformation(),
```
david-livefront: If we are obscuring the content, do we need to make this compound?

Can it just be the `PasswordVisualTransformation`?

SaintPatrck: Nope. Updated. Nice catch!

SaintPatrck:
@codokie We've pivoted to a different approach for handling bidi text. Would you be willing to check behavior with these changes? I'm particularly curious if the Password fields handle manual RTL input correctly.

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/passwordhistory/PasswordHistoryScreenTest.kt
```diff
@@ -164,6 +167,7 @@ class PasswordHistoryScreenTest : BitwardenComposeTest() {
             )
         }
 
-        composeTestRule.onNodeWithText("Password1").assertIsDisplayed()
+        @Suppress("StringTemplate")
```
david-livefront: What's the suppression all about here?

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemScreenTest.kt
```diff
@@ -62,6 +64,7 @@ import org.junit.Test
 import java.time.Instant
 
 @Suppress("LargeClass")
+
```
david-livefront: Can we drop the newline

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/passwordhistory/PasswordHistoryScreenTest.kt
SaintPatrck: I had both control characters wrapped in curly brackets because I felt it was easier to read, and `detekt` doesn't like that. I guess it "fixed" it automatically before I applied the suppression. 😢

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/passwordhistory/PasswordHistoryScreenTest.kt
```diff
@@ -98,7 +100,7 @@ class PasswordHistoryScreenTest : BitwardenComposeTest() {
             )
         }
 
-        composeTestRule.onNodeWithText(password.password).assertIsDisplayed()
+        composeTestRule.onNodeWithText("${LRO}${password.password}$PDF").assertIsDisplayed()
```
david-livefront: This one can just be `"$LRO${password.password}$PDF"`

ui/src/main/kotlin/com/bitwarden/ui/platform/components/util/ForceLtrVisualTransformation.kt
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * Remembers a [ForceLtrVisualTransformation] transformation.
+ *
+ * This is an optimization to avoid creating a new [ForceLtrVisualTransformation] on every
+ * recomposition.
+ */
+@Composable
+fun forceLtrVisualTransformation(): VisualTransformation = remember {
```
david-livefront, line 79: Does the Remember do anything here?

SaintPatrck: Only to expose ForceLtrTransformation since it's `internal` and follow the existing pattern(s).

codokie:
> @codokie We've pivoted to a different approach for handling bidi text. Would you be willing to check behavior with these changes? I'm particularly curious if the Password fields handle manual RTL input correctly.

The usage of unicode characters to force LTR direction on text does seem to fix the first and third regressions of the unicodeWrap() attempt (see [this](https://github.com/bitwarden/android/pull/5825#issuecomment-3252914352) comment). However, the 2nd regression (wrong cursor position) is still very much an issue.

As for manual RTL input in password fields, there seems to be a problem with that because LTR direction is being forced regardless of the directionality of the input. For example, the password `סיסמה1234*` is supposed to appear as `*סיסמה1234` (asterisk is shifted). Imitating the logic of [TextDirection.ContentOrLtr](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/style/TextDirection#ContentOrLtr()) could prove to be a solution. That said, it is not really a big deal because most services forbid passwords with RTL characters, and LTR input seems to work as expected in password fields (except for the cursor alignment issue).

Upon brief testing, I’ve noticed a few more quirks that occur when an RTL app language is selected:

* Editing a URI field does not force LTR text direction on the input (typing `https://bitwarden.com/` appears as `/https://bitwarden.com`).
* Editing the phone number field of an identity item does not force LTR text direction on the input (typing  `123 456 7890` appears as `7890 456 123`).
* In the footnote of the Settings -> About page, the copyright symbol is placed on the right (`Bitwarden Inc. 2015-2025 ©`) instead of on the left (`© Bitwarden Inc. 2015-2025`).
* LTR notes written in a language such as English are uncomfortable to read/write because RTL text direction is applied to them. If any non-strong character is used at the start or end of a line, then it is placed on the other end (e.g. `Hello world!` appears as `!Hello world`). Bi-directional notes (mixture of LTR with RTL text) are also not rendered properly. In order to fully support them, the user should be able to override (through some kind of text formatting UI) the direction of each line of text separately. This should obviously not be a priority at the moment because note formatting has not been implemented yet.

It still seems to me that #5012 is a better approach than utilizing Unicode, mainly because the native Android implementation to set text direction takes care to align the text cursor correctly. The only downside is that it also aligns the text according to its direction, but it is easy to override this behavior with basically no side effects.

Lastly, maintainability is also an important factor to consider. With the changes that this PR proposes, future PRs that add new types of vault items (for example) could be prone to omitting or misusing the new `forceLtr` property or the `forceLtrVisualTransformation()` function.

Thanks for dedicating time to find the best solution to this matter, issuing a fix could greatly improve the usability of the app for users of RTL languages. Please let me know if I could help with anything.

SaintPatrck:
Superseded by #6309

Issue #5951, opened 2025-09-27
[PM-26241] Clicking on text fields using TalkBack when editing an item does not bring up the keyboard to edit


Jonathans859:
### Steps To Reproduce

1. Open Bitwarden and edit an item of your choice (only tested with login items for now)
2. Start TalkBack and focus any edit field, be it username or password, for example.
3. Double tap the screen in order to open the keyboard and input in the field.

### Expected Result

The keyboard pops up to input/edit the text.

### Actual Result

The keyboard is not appearing, nothing happens. Disabling TalkBack and clicking on the item as (usual, without TalkBack) works.

### Build Version

version 2025.9.0  com.x8bit.bitwarden

### What server are you connecting to?

US

### Environment Details

Pixel 9 Pro, Android 16 security update from Sep 5th

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

rmcdowell-bitwarden:
Hi there,

Thank you for your report! 

I was able to reproduce this issue, and I have flagged this to our engineering team.

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.

Thanks once again!

PR #6129, opened 2025-11-06
[PM-27869] fix/[PM-26241] : draw out keyboard on talkback click

dev-sharma3624:
## 🎟️ Tracking

https://github.com/bitwarden/android/issues/5951

## 📔 Objective

* **Fixes Bug:** [PM-26241]
* **The Problem:** The on-screen keyboard failed to appear when TalkBack users selected a text field on the "edit item" screen.
* **The Solution:** This PR ensures that the text field correctly requests and receives focus when activated via TalkBack, which properly triggers the keyboard to display.
* **Impact:** Restores the ability for screen reader users to edit item details, ensuring the feature is accessible.

### Test Updates

* **Context:** The fix for the TalkBack focus behavior revealed that two unit tests were brittle. They were built on an incorrect assumption that text would always be inserted at the **start** of the field.

* **Solution:** This PR also includes a commit to update these tests, making them more robust and aligning them with the new, correct cursor behavior.

## 📸 Screenshots

[before.webm](https://github.com/user-attachments/assets/5ca3b6ba-4b4d-4285-9140-47b21c35ec30)

[after.webm](https://github.com/user-attachments/assets/dd7fd4f0-32fa-4b2c-a54e-b399cf10f4c7)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26241]: https://bitwarden.atlassian.net/browse/PM-26241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
+++ b/ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
@@ -45,6 +45,7 @@ import androidx.compose.ui.platform.TextToolbar
 import androidx.compose.ui.semantics.CustomAccessibilityAction
 import androidx.compose.ui.semantics.customActions
 import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.TextRange
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.text.input.TextFieldValue
@@ -278,14 +279,16 @@ fun BitwardenTextField(
         val isDropDownExpanded = filteredAutoCompleteList.isNotEmpty() && hasFocused
         ExposedDropdownMenuBox(
             expanded = isDropDownExpanded,
-            onExpandedChange = { hasFocused = false },
+            onExpandedChange = {
+                hasFocused = !hasFocused
+                focusRequester.requestFocus()
+            },
             modifier = modifier.defaultMinSize(minHeight = 60.dp),
         ) {
             Column(
                 modifier = Modifier
                     .onGloballyPositioned { widthPx = it.size.width }
                     .onFocusEvent { focusState -> hasFocused = focusState.hasFocus }
-                    .focusRequester(focusRequester)
                     .cardStyle(
                         cardStyle = cardStyle,
                         paddingTop = 6.dp,
@@ -377,8 +380,14 @@ fun BitwardenTextField(
                         .nullableTestTag(tag = textFieldTestTag)
                         .menuAnchor(type = ExposedDropdownMenuAnchorType.PrimaryEditable)
                         .fillMaxWidth()
+                        .focusRequester(focusRequester)
                         .onFocusChanged { focusState ->
                             focused = focusState.isFocused
+                            if (focused) {
+                                textFieldValueState = textFieldValueState.copy(
+                                    selection = TextRange(textFieldValueState.text.length),
+                                )
+                            }
                         },
                 )
                 supportingContent
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreenTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/generator/GeneratorScreenTest.kt
@@ -25,6 +25,7 @@ import androidx.compose.ui.test.onSiblings
 import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performScrollTo
 import androidx.compose.ui.test.performTextInput
+import androidx.compose.ui.test.performTextReplacement
 import androidx.compose.ui.test.performTouchInput
 import androidx.compose.ui.test.swipeRight
 import androidx.compose.ui.text.AnnotatedString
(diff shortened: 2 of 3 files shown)


dev-sharma3624:
@bitwarden-bot @david-livefront Is there any update on this PR?

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
```diff
@@ -377,8 +384,14 @@ fun BitwardenTextField(
                         .nullableTestTag(tag = textFieldTestTag)
                         .menuAnchor(type = ExposedDropdownMenuAnchorType.PrimaryEditable)
                         .fillMaxWidth()
+                        .focusRequester(textFieldFocusRequestor)
```
david-livefront: I think we can simplify this by putting the original `focusRequester` here and removing it from the `Column`. Then we can avoid adding the `textFieldFocusRequestor`.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
```diff
@@ -278,13 +280,18 @@ fun BitwardenTextField(
...
+                textFieldFocusRequestor.requestFocus()
+            },
             modifier = modifier.defaultMinSize(minHeight = 60.dp),
         ) {
             Column(
                 modifier = Modifier
                     .onGloballyPositioned { widthPx = it.size.width }
-                    .onFocusEvent { focusState -> hasFocused = focusState.hasFocus }
+                    .onFocusEvent { focusState ->
+                        hasFocused = focusState.hasFocus
```
david-livefront: Can we remove this change.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
```diff
@@ -377,8 +384,14 @@ fun BitwardenTextField(
                         .nullableTestTag(tag = textFieldTestTag)
                         .menuAnchor(type = ExposedDropdownMenuAnchorType.PrimaryEditable)
                         .fillMaxWidth()
+                        .focusRequester(textFieldFocusRequestor)
                         .onFocusChanged { focusState ->
                             focused = focusState.isFocused
+                            if (focused) {
+                                textFieldValueState = textFieldValueState.copy(
+                                    selection = TextRange(textFieldValueState.text.length)
```
david-livefront: I assume this is to ensure the selection goes to the end?

dev-sharma3624: Yes

david-livefront: Be sure to add a trailing comma here, otherwise the linter will get angry

dev-sharma3624: Thanks for the review. I'll make those adjustments shortly and see if we can actually make it work without a second focus requestor as you pointed out correctly.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
dev-sharma3624: It worked perfectly as you suggested. Thanks for the review! I’ve updated it to use the original focus requester.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
```diff
@@ -377,8 +381,14 @@ fun BitwardenTextField(
                         .nullableTestTag(tag = textFieldTestTag)
                         .menuAnchor(type = ExposedDropdownMenuAnchorType.PrimaryEditable)
                         .fillMaxWidth()
+                        .focusRequester(focusRequester)
```
david-livefront, line 383: Can we remove the `focusRequester` from the `Column` as well?

dev-sharma3624: Oh yes definitely, It should have been done that way. This is my mistake.

david-livefront (dismissed):

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenTextField.kt
```diff
@@ -377,8 +380,14 @@ fun BitwardenTextField(
...
                         .onFocusChanged { focusState ->
                             focused = focusState.isFocused
+                            if (focused) {
+                                textFieldValueState = textFieldValueState.copy(
+                                    selection = TextRange(textFieldValueState.text.length),
```
david-livefront, line 388: Looks like there are two tests failing because of the update to text selection here. 🤔 

Can you update those

dev-sharma3624: You're absolutely right, thanks for catching that.

The tests were written with an implicit assumption that new text would be inserted at the start of the field.

My fix for the focus behavior moves the cursor to the end, which caused that old test logic to fail. I've just pushed a commit to update the tests to align with this new, correct behavior. 👍

david-livefront (approved):

PR #6279, opened 2025-12-17
[PM-28468] Updated validation and navigation for MigrateToMyItems

aj-rosado:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28468

## 📔 Objective

Implement navigation to the vault migration flow that prompts users with a Personal Ownership policy to migrate their personal vault items to an organization. The `VaultMigrationManager` reactively monitors vault state, sync completion, and network connectivity to determine when migration is required, displaying the prompt only when all conditions are met (policy active, feature flag enabled, personal ciphers exist, and network available). 

The implementation includes an optimized database query using an indexed `organizationId` column to efficiently check for personal ciphers without loading full cipher JSON data, and ensures the migration prompt correctly appears or disappears as conditions change (e.g., going offline/online or after sync completes in multi-account scenarios).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerTest.kt
@@ -44,7 +44,8 @@ class VaultMigrationManagerTest {
         every { vaultUnlockDataStateFlow } returns mutableVaultUnlockDataStateFlow
     }
 
-    private val mutableLastSyncTimeFlow = MutableStateFlow<Instant?>(Instant.now())
+    private val mutableLastSyncTimeFlow =
+        MutableStateFlow<Instant?>(Instant.parse("2023-10-27T12:00:00Z"))
     private val settingsDiskSource: SettingsDiskSource = mockk {
         every { getLastSyncTimeFlow(any()) } returns mutableLastSyncTimeFlow
     }
@@ -503,7 +504,7 @@ class VaultMigrationManagerTest {
             expectNoEvents()
 
             // Sync completes - should now trigger migration check
-            mutableLastSyncTimeFlow.value = Instant.now()
+            mutableLastSyncTimeFlow.value = Instant.parse("2023-10-27T12:00:00Z")
 
             assertEquals(
                 VaultMigrationData.MigrationRequired(


app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -523,6 +546,34 @@ class VaultSyncManagerImpl(
...
+    ): Boolean {
+        return policyManager
+            .getActivePolicies(PolicyTypeJson.PERSONAL_OWNERSHIP)
+            .any() &&
+            featureFlagManager.getFeatureFlag(FlagKey.MigrateMyVaultToMyItems) &&
+            connectionManager.isNetworkConnected &&
+            cipherListView.any { it.organizationId == null }
+    }
+
+    // Updates [shouldMigratePersonalVaultFlow].
```
SaintPatrck: ⛏️ Don't really need this comment. The function name is descriptive enough.

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/migratetomyitems/MigrateToMyItemsViewModel.kt
```diff
@@ -23,13 +25,22 @@ private const val KEY_STATE = "state"
...
 ) : BaseViewModel<MigrateToMyItemsState, MigrateToMyItemsEvent, MigrateToMyItemsAction>(
     initialState = savedStateHandle[KEY_STATE] ?: run {
-        val args = savedStateHandle.toMigrateToMyItemsArgs()
+
+        val orgId = policyManager.getPersonalOwnershipPolicyOrganizationId()
+        val orgName = authRepository.userStateFlow.value
+            ?.activeAccount
+            ?.organizations
+            ?.firstOrNull { it.id == orgId }
+            ?.name
```
SaintPatrck: ❓ Passing args makes for a simpler ViewModel; specifically in regards to handling scenarios when either `orgId` or `orgName` are null. Is there a reason why this approach is taken instead?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -523,6 +546,33 @@ class VaultSyncManagerImpl(
...
+            cipherListView.any { it.organizationId == null }
+    }
+
+    private fun verifyAndUpdateIfUserShouldMigrateVaultToMyItems(cipherList: List<Cipher>) {
+        val shouldMigrate = policyManager
+            .getActivePolicies(PolicyTypeJson.PERSONAL_OWNERSHIP)
+            .any() &&
+            featureFlagManager.getFeatureFlag(FlagKey.MigrateMyVaultToMyItems) &&
+            connectionManager.isNetworkConnected &&
+            cipherList.any { it.organizationId == null }
```
SaintPatrck: ⛏️ 🎨 Not a mandatory change request, but some of this duplication could be eliminated if `userShouldMigrate()` is changed to accept `hasPersonalItems` as a lambda.

```kotlin
private fun userShouldMigrateVault(
    hasPersonalItems: () -> Boolean,
) : Boolean {
    return connectionManager.isNetworkConnected &&
        featureFlagManager.getFeatureFlag(...) &&
        policyManager.getActivePolicies(...).any() &&
        hasPersonalItems
}

private fun verifyAndUpdateIfUserShouldMigrateVaultToMyItems(
    cipherList: List<Cipher>,
) {
    mutableShouldMIgratePersonalVaultFlow.update {
        userShouldMigrateVault { 
            cipherList.any { it.organizationId == null } 
        }
    }
}

Comment (SaintPatrck) [reply app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt]: This would also be a performance improvement since it would short-circuit before attempting to iterate over the entire cipher collection when any of the preceding conditions are false.

File: app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/migratetomyitems/MigrateToMyItemsViewModel.kt (hunk 2 above)
Comment (aj-rosado) [reply app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/migratetomyitems/MigrateToMyItemsViewModel.kt]: I have changed this quite a bit from original idea. `RootNavViewModel` only verifies if it should navigate from the `shouldMigratePersonalVaultFlow`, not having any info regarding the org that should do the migration.

This was changed because if the vault was updated on a different client, the migration flow would still be displayed as it would not check with the most synced data. Now we are syncing before displaying the screen to make sure we are not displaying the screen unless it is needed.

This being said, I agree that if we could pass the parameters would simplify the VM, not sure about a good way to achieve it, only that I can think about is the flow having the Organization object?

File: app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManager.kt [hunk 4]
```diff
@@ -62,6 +63,12 @@ interface VaultSyncManager {
      */
     val sendDataStateFlow: StateFlow<DataState<SendData>>
 
+    /**
+     * Flow that emits when conditions are met for the user to migrate their personal vault.
+     * Updated after each sync to reflect current policy and vault state.
+     */
+    val shouldMigratePersonalVaultFlow: StateFlow<VaultMigrationData>
```
SaintPatrck: ⛏️ since this is no longer a simple Boolean value, can we rename it to something like `vaultMigrationDataStateFlow`? That would align with the naming of other flows defined above.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -409,6 +428,14 @@ class VaultSyncManagerImpl(
                     )
                     .fold(
                         onSuccess = { result ->
+                            // We need to be sure the data on device is updated
+                            // before sending the user to the migration screen
+                            if (userShouldMigrateVault {
+                                    result.successes.any { it.organizationId == null }
```
SaintPatrck: `it` is being shadowed by the outer `map { }` arg. 

```suggestion
                                    result.successes.any { cipher -> cipher.organizationId == null }
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -409,6 +428,14 @@ class VaultSyncManagerImpl(
...
+                                }
+                            ) {
+                                sync(forced = true)
```
david-livefront: IDK how I feel about having side-affects in this flow 🤔

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -523,6 +550,44 @@ class VaultSyncManagerImpl(
...
+        return policyManager
+            .getActivePolicies(PolicyTypeJson.PERSONAL_OWNERSHIP)
+            .any() &&
+            featureFlagManager.getFeatureFlag(FlagKey.MigrateMyVaultToMyItems) &&
+            connectionManager.isNetworkConnected &&
+            hasPersonalItems()
+    }
+
+    private fun verifyAndUpdateIfUserShouldMigrateVaultToMyItems(cipherList: List<Cipher>) {
+        val userId = activeUserId ?: return
```
david-livefront: Can we just pass in the `userId` to avoid any conflicts

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -523,6 +550,44 @@ class VaultSyncManagerImpl(
...
+                val orgId = policyManager.getPersonalOwnershipPolicyOrganizationId()
+                val orgName = authDiskSource
+                    .getOrganizations(userId = userId)
+                    ?.firstOrNull { it.id == orgId }
+                    ?.name
+
+                if (orgId != null && orgName != null) {
+                    VaultMigrationData.MigrationRequired(
+                        organizationId = orgId,
+                        organizationName = orgName,
```
david-livefront: The `VaultSyncManager` handling the migration feels a little odd to me. It make it responsible for more scope than it maybe should be responsible for.

What would you think of a `VaultMigrationManager` that consumed the cipher data to handle the migration?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
SaintPatrck: ❓ Is there a specific scenario or issue you're encountering that requires this additional `sync()` call before navigating? Asking because this feels out of place. I would expect the disk source data to already be up-to-date here. If not, the next sync (manual or automatic) should trigger migration if it's needed.

If the concern is navigating to migration screens with stale data, could we perform `sync` or `syncForResult` in the `MigateToMyItemsViewModel` immediately before triggering the migration?

aj-rosado: I do not feel totally comfortable with it as well. Any idea of a better approach?
Otherwise we will display the takeover to the user until a sync is forced even if the user does not need to do the migration anymore

aj-rosado: Yes the issues is displaying the Migrate screen when is not necessary.
So we navigate to the screen, sync and check again if the user still needs the migration and dismiss it if he does not? Looks good to me actually, this should be an edge case

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
```diff
@@ -402,6 +409,11 @@ class VaultSyncManagerImpl(
             .onStart { mutableDecryptCipherListResultFlow.updateToPendingOrLoading() }
             .map {
                 vaultLockManager.waitUntilUnlocked(userId = userId)
+                // Verify migration state after unlock with the cipher list from disk
+                vaultMigrationManager.verifyAndUpdateMigrationState(
+                    userId = userId,
+                    cipherList = it,
+                )
```
david-livefront: This still feels like an inappropriate place to have this side-affect.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerImpl.kt
```diff
@@ -0,0 +1,65 @@
...
+                } else {
+                    VaultMigrationData.NoMigrationRequired
+                }
+            } else {
+                VaultMigrationData.NoMigrationRequired
+            }
+        }
+    }
+
+    override fun shouldMigrateVault(hasPersonalItems: () -> Boolean): Boolean {
```
david-livefront: Can this just be a private helper function. It is never used outside this class.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerImpl.kt
SaintPatrck: Agreed. I think we can encapsulate all of this into `VaultMigrationManager` if we shuffle around where and how we're observing things. 🤔

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/migratetomyitems/MigrateToMyItemsViewModel.kt
```diff
@@ -42,6 +46,11 @@ class MigrateToMyItemsViewModel @Inject constructor(
         stateFlow
             .onEach { savedStateHandle[KEY_STATE] = it }
             .launchIn(viewModelScope)
+        viewModelScope.launch {
+            // We need to ensure we have the most recent data. No need to do anything
+            // RootNavViewModel will take care of navigating if needed
+            vaultSyncManager.syncForResult(forced = true)
```
david-livefront: Does this need to be for result if we aren't doing anything with the result

SaintPatrck: Nope. And we can clean up a few other things in here now, if you like the changes to `VaultMigrationManager`.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerImpl.kt
```diff
@@ -0,0 +1,124 @@
...
+            .launchIn(unconfinedScope)
+    }
+
+    /**
+     * Observes cipher data for the given user and updates migration state when changes occur.
+     * Only emits updates after the user has synced at least once to ensure data freshness.
+     */
+    private fun observeCipherDataAndUpdateMigrationState(userId: String) =
+        vaultDiskSource
+            .getCiphersFlow(userId = userId)
```
david-livefront: This is going to put a large amount of ciphers into memory.

What do you think of adding the `organizationId` to the `CipherEntity` and adding a new function to the `VaultDiskSource`, something like this:

```kotlin
fun hasCiphersWithoutOrganizationFlow(userId: String): Flow<Boolean>
```

That will optimize this quite a bit.

SaintPatrck: I like it!

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/entity/CipherEntity.kt
```diff
@@ -2,18 +2,25 @@ package com.x8bit.bitwarden.data.vault.datasource.disk.entity
...
+        Index(value = ["user_id", "organization_id"]),
+    ],
+)
 data class CipherEntity(
     @PrimaryKey(autoGenerate = false)
     @ColumnInfo(name = "id")
     val id: String,
 
-    @ColumnInfo(name = "user_id", index = true)
+    @ColumnInfo(name = "user_id")
```
david-livefront, line 23: 👍

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/entity/CipherEntity.kt
```diff
@@ -2,18 +2,25 @@ package com.x8bit.bitwarden.data.vault.datasource.disk.entity
...
 
 /**
  * Entity representing a cipher in the database.
  */
-@Entity(tableName = "ciphers")
+@Entity(
+    tableName = "ciphers",
+    indices = [
+        Index(value = ["user_id"]),
+        Index(value = ["user_id", "organization_id"]),
```
david-livefront, line 15: Is the basic `user_id` needed since you already added it to the `ColumnInfo`?


Do we need the compound Index? We do not query by a specific org id, so I'm unsure if that will gain us anything.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/database/VaultDatabase.kt
```diff
@@ -27,7 +27,7 @@ import com.x8bit.bitwarden.data.vault.datasource.disk.entity.SendEntity
         FolderEntity::class,
         SendEntity::class,
     ],
-    version = 8,
+    version = 9,
```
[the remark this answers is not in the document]
SaintPatrck: @claude we fallback to destructive migrations, which in turn trigger a full sync. Migrations are not required in most cases.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/entity/CipherEntity.kt
SaintPatrck: I _removed_ it from the `ColumnInfo`, so it is needed here.

We don't gain much from this particular query, but we'll be fetching all personal ciphers from the DB when we implement the actual migration API's. That's where the new index will really shine.

SaintPatrck: Technically we don't need the individual `user_id` index **and** the composite index. At least I don't think we do. 🤔 

@claude is the individual `user_id` column index needed if it's include in the composite index of `user_id` and `organization_id`?

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    private val mutableVaultMigrationDataStateFlow =
+        MutableStateFlow<VaultMigrationData>(value = VaultMigrationData.NoMigrationRequired)
+
+    override val vaultMigrationDataStateFlow: StateFlow<VaultMigrationData>
+        get() = mutableVaultMigrationDataStateFlow.asStateFlow()
+
+    init {
+        // Observe cipher data changes and automatically verify migration state
+        mutableVaultMigrationDataStateFlow
```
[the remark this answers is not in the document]
SaintPatrck: @claude you are misunderstanding how it works. `subscriptionCount` refers to external collectors of the exposed `vaultMigrationDataStateFlow`, not the internal observer operations.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * Uses optimized [VaultDiskSource.hasPersonalCiphersFlow] query that checks only the
+     * indexed organizationId column without loading full cipher JSON data.
+     */
+    private fun observeCipherDataAndUpdateMigrationState(userId: String) =
+        vaultDiskSource
+            .hasPersonalCiphersFlow(userId = userId)
+            .filter {
+                // Only process cipher data after sync has occurred at least once
+                settingsDiskSource.getLastSyncTime(userId = userId) != null
```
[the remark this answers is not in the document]
SaintPatrck: Great observation!

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param hasPersonalCiphers Boolean indicating if the user has any personal ciphers.
+     * @return true if migration conditions are met, false otherwise.
+     */
+    private fun shouldMigrateVault(hasPersonalCiphers: Boolean): Boolean =
+        policyManager
+            .getActivePolicies(PolicyTypeJson.PERSONAL_OWNERSHIP)
+            .any() &&
+            featureFlagManager.getFeatureFlag(FlagKey.MigrateMyVaultToMyItems) &&
+            connectionManager.isNetworkConnected &&
```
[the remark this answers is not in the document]
SaintPatrck: Another good catch!

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/vaultunlocked/VaultUnlockedNavigation.kt
```diff
@@ -263,6 +264,11 @@ fun NavGraphBuilder.vaultUnlockedGraph(
         importLoginsScreenDestination(
             onNavigateBack = { navController.popBackStack() },
         )
+
+        migrateToMyItemsDestination(
+            onNavigateToVault = { navController.popBackStack() },
+            onNavigateToLeaveOrganization = { },
```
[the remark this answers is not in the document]
SaintPatrck: @claude navigation to that screen is out of scope for this pull request. Navigation to `LeaveOrganization` screen will be completed in a subsequent PR.

app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerTest.kt
```diff
@@ -0,0 +1,656 @@
...
+            )
+
+            // Emit personal ciphers exist - should not trigger yet since no sync
+            mutableHasPersonalCiphersFlow.value = true
+
+            // Should remain NoMigrationRequired since sync hasn't happened
+            expectNoEvents()
+
+            // Sync completes - should now trigger migration check
+            mutableLastSyncTimeFlow.value = Instant.now()
```
david-livefront: Can we use `Instant.parse("2023-10-27T12:00:00Z")` instead.

Or event `Instant.MIN`

app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultMigrationManagerTest.kt
```diff
@@ -0,0 +1,656 @@
...
+        every { hasPersonalCiphersFlow(any()) } returns mutableHasPersonalCiphersFlow
+    }
+
+    private val mutableVaultUnlockDataStateFlow =
+        MutableStateFlow<List<VaultUnlockData>>(emptyList())
+    private val vaultLockManager: VaultLockManager = mockk {
+        every { vaultUnlockDataStateFlow } returns mutableVaultUnlockDataStateFlow
+    }
+
+    private val mutableLastSyncTimeFlow = MutableStateFlow<Instant?>(Instant.now())
```
david-livefront: Same thing here can we avoid the `now()`

aj-rosado:
Thank you @SaintPatrck and @david-livefront

aj-rosado:
Thank you @SaintPatrck and @david-livefront

PR #6298, opened 2025-12-27
[PM-30259] Add iodéOS browser to community FIDO2 privileged list

lucasmz-dev:
Obtained from https://fdroid.iode.tech/repo/


vincentvidal:
I'm the maintainer of iodéOS browser. Thanks @lucasmz-dev for this PR, this would be an interesting feature. I checked the sha256 fingerprint, it's ok.

PR #6330, opened 2026-01-07
[PM-30516] Remove extra comma on FIDO2 community list

lucasmz-dev:
Oops.


SaintPatrck:
We added support for trailing commas to our JSON instance (#6326) so this isn't necessary. Thanks for noticing and circling back to fix it though.

lucasmz-dev:
oh that's nice

PR #6384, opened 2026-01-20
[PM-30899] Store account keys upon SSO user creation

SaintPatrck:
## 🎟️ Tracking

[PM-30899](https://bitwarden.atlassian.net/browse/PM-30899)

## 📔 Objective

Update the `createAccountKeys` API call to handle and store the full response object, which includes the `accountKeys`.

When a new SSO user is created, the `accountKeys` received from the `createAccountKeys` endpoint are now stored in `authDiskSource`.

This fixes an issue where an SSO user was forced to re-authenticate after app process death, even when the device is trusted.

Specific changes include:
- Modified `AccountsService` and related API definitions to expect and return a `CreateAccountKeysResponseJson` object instead of `Unit`.
- Introduced the `CreateAccountKeysResponseJson` data class to model the API response.
- Updated `AuthRepositoryImpl` to store the `accountKeys` from the response after creating a new SSO user.
- Storing the new SSO user data in the `AuthRepositoryImpl` is done in a UserState to ensure state-based navigation does not occur too soon.
- Adjusted associated tests to reflect these changes, including mocking the new response structure and verifying that `accountKeys` are stored correctly.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-30899]: https://bitwarden.atlassian.net/browse/PM-30899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
@@ -2,6 +2,7 @@ package com.x8bit.bitwarden.data.auth.repository
 
 import com.bitwarden.core.AuthRequestMethod
 import com.bitwarden.core.InitUserCryptoMethod
+import com.bitwarden.core.RegisterTdeKeyResponse
 import com.bitwarden.core.WrappedAccountCryptographicState
 import com.bitwarden.core.data.manager.dispatcher.DispatcherManager
 import com.bitwarden.core.data.repository.error.MissingPropertyException
@@ -14,6 +15,7 @@ import com.bitwarden.crypto.Kdf
 import com.bitwarden.data.datasource.disk.ConfigDiskSource
 import com.bitwarden.data.repository.util.toEnvironmentUrls
 import com.bitwarden.data.repository.util.toEnvironmentUrlsOrDefault
+import com.bitwarden.network.model.CreateAccountKeysResponseJson
 import com.bitwarden.network.model.DeleteAccountResponseJson
 import com.bitwarden.network.model.GetTokenResponseJson
 import com.bitwarden.network.model.IdentityTokenAuthModel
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -465,6 +466,13 @@ class AuthRepositoryImpl(
                                 publicKey = keys.publicKey,
                                 encryptedPrivateKey = keys.privateKey,
                             )
+                            .onSuccess { response ->
+                                authDiskSource
+                                    .storeAccountKeys(
+                                        userId = userId,
+                                        accountKeys = response.accountKeys,
```
david-livefront: Should this be done in the `onSuccess` down below?

If `organizationResetPasswordEnroll` fails, we will have persisted these keys without completing the full flow.

david-livefront: There is also this comment down below:

> // TDE and SSO user creation still uses crypto-v1. These users are not
> // expected to have the AEAD keys so we only store the private key for now.
> // See https://github.com/bitwarden/android/pull/5682#discussion_r2273940332
> // for more details.

Do we know if this is still true?

SaintPatrck: > Should this be done in the onSuccess down below?

Good catch. Yes.

@quexten what's the status of TDE & SSO user creation? Is it safe to remove the crypto-v1 usages yet?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -465,19 +466,23 @@ class AuthRepositoryImpl(
                                 publicKey = keys.publicKey,
                                 encryptedPrivateKey = keys.privateKey,
                             )
-                            .map { keys }
+                            .map { response -> keys to response.accountKeys }
```
david-livefront: Does `response.accountKeys` contain the private key?

If so, we can just pass that along, right?

If not, this looks great! 😄

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: The "accountKeys" already contain the "privateKey". You should be able to drop the entire privateKey state from the codebase entirely. Clients has a corresponding PR doing so.

As to the status of TDE/SSO, KM has work in progress to port it over to V2 encryption. We have SDK functionality here, but it is not yet used on mobile. The android task is here: https://bitwarden.atlassian.net/browse/PM-27241?search_id=95804c4c-f91b-42b4-b512-d2bcf097f856

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
quexten: You should not need to pass the private key directly anywhere anymore. If you do find an API that needs it, please let me know. Otherwise you can just pass the account cryptographic state / "accountKeys" everywhere.

If a request model requires it, then just set the privateKey from the accountKeys, but only extract it at the last possible moment. (That's the strategy we've been trying on clients).

quexten:
> Additionally, during a soft logout, the user's accountKeys are preserved and restored along with other account data upon re-authentication.

Could you expand on why we are preserving them? A re-authentication should always just sync the latest copy of these from the server, so I'm not sure why we need to keep them during a soft logout?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
david-livefront: The methods associated with the private key (storePrivateKey, getPrivateKey, etc.) are all deprecated but have not yet been fully removed.
Removing those functions and any associated functionality will be handled separately from this change.

david-livefront:
Thanks @aj-rosado

PR #6472, opened 2026-02-03
[PM-30279] Extract credential provider handling to dedicated activity

SaintPatrck:
## 🎟️ Tracking

PM-30279

## 📔 Objective

This PR extracts credential provider handling into a dedicated `CredentialProviderActivity` to improve security and architecture. The new activity serves as a transparent trampoline that receives credential intents from the Android CredentialManager framework, securely relays them to `MainActivity` via `CredentialProviderRequestManager`, and forwards results back to the framework. This approach hardens against malicious intent attacks by ensuring credential data is never passed through intent extras to exported activities, while preserving the single-Activity architecture where all UI is hosted by MainActivity. 

## 📸 Screenshots

Coming soon!

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -85,15 +85,6 @@
                 <data android:pathPattern="/redirect-connector.*" />
             </intent-filter>
             <intent-filter>
-                <action android:name="com.x8bit.bitwarden.credentials.ACTION_CREATE_PASSKEY" />
-                <action android:name="com.x8bit.bitwarden.credentials.ACTION_GET_PASSKEY" />
-                <action android:name="com.x8bit.bitwarden.credentials.ACTION_CREATE_PASSWORD" />
-                <action android:name="com.x8bit.bitwarden.credentials.ACTION_GET_PASSWORD" />
-                <action android:name="com.x8bit.bitwarden.credentials.ACTION_UNLOCK_ACCOUNT" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-            <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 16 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/CredentialProviderRequestManagerImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+import javax.inject.Singleton
+
+/**
+ * Primary implementation of [CredentialProviderRequestManager].
+ *
+ * Uses an [AtomicReference] for thread-safe get-and-clear semantics, ensuring
+ * the pending request is only processed once.
+ */
+@Singleton
+data object CredentialProviderRequestManagerImpl : CredentialProviderRequestManager {
```
david-livefront: Can we make this a `class`?

That will ensure no one is bypassing the dagger graph and just accessing the object singleton directly.

app/src/main/kotlin/com/x8bit/bitwarden/MainViewModel.kt
```diff
@@ -292,6 +290,14 @@ class MainViewModel @Inject constructor(
         intent: Intent,
         isFirstIntent: Boolean,
     ) {
+        // Check for pending credential request from CredentialProviderActivity first.
+        // This is a one-shot read that clears the pending request, ensuring credential
+        // data is never passed through intent extras to this exported activity.
+        credentialProviderRequestManager.getPendingCredentialRequest()?.let { request ->
+            handleCredentialRequest(request)
+            return
```
david-livefront: Does this need to be treated as a special case?

Can it just be the highest priority case in the `when` down below?

SaintPatrck: Ah. I moved it to the bottom of the `when` but forgot to remove this. Thanks.

app/src/main/kotlin/com/x8bit/bitwarden/CredentialProviderActivity.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        intent = intent.validate()
+        super.onCreate(savedInstanceState)
+
+        if (savedInstanceState == null) {
+            // Process credential intent (sets pending request on CredentialProviderRequestManager)
+            viewModel.trySendAction(CredentialProviderAction.ReceiveFirstIntent(intent))
+            launchMainActivityForResult()
```
david-livefront, line 61: I am unsure if this will work but could we remove the `forResult` part here and just finish the activity?

If we switch the `CredentialProviderRequestManagerImpl` to use a channel (Channel<CredentialProviderRequest>(capacity = Int.MAX_VALUE)), the value should be waiting for the `MainActivity` when it launches.

SaintPatrck: `forResult` is required in order to forward the operation result back to the OS. We can't `finish()` the `CredentialActivity` until we have `setResult(..)` with the result of the credential request operation.

SaintPatrck: Confirmed that if we simply start `MainActivity` then `finish()` `CredentialActivity` before `setResult()` is called, the system treats the operation as cancelled. That results in a false-positive where we save the passkey but not saved by the relying party.

david-livefront: Ahh, that makes perfect sense.

app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/CredentialManagerPendingIntentManagerImpl.kt
```diff
@@ -83,6 +86,7 @@ class CredentialManagerPendingIntentManagerImpl(
     ): PendingIntent {
         val intent = Intent(CREATE_PASSWORD_ACTION)
             .setPackage(context.packageName)
+            .setClassName(context, CREDENTIAL_ACTIVITY_CLASS)
```
david-livefront: Can we make these type safe?

```kotlin
.setClass(context, CredentialProviderActivity::class.java)
```

app/src/main/AndroidManifest.xml
```diff
@@ -120,6 +111,35 @@
...
+             (Theme.NoDisplay requires finish() before onResume(), incompatible with ActivityResult).
+
+             Note: Unlike AuthCallbackActivity, this does NOT use noHistory="true" because it
+             must remain in the back stack to receive the ActivityResult callback from
+             MainActivity. -->
+        <activity
+            android:name=".CredentialProviderActivity"
+            android:exported="false"
+            android:launchMode="singleTop"
+            android:theme="@android:style/Theme.Translucent.NoTitleBar">
```
david-livefront, line 131: What's the difference between this theme and the `@android:style/Theme.NoDisplay"` we use elsewhere?

SaintPatrck: `Theme.NoDisplay` requires `finish()` be called before `onResume()` completes. This works in other places because the proxy Activities are not required to complete their corresponding operation. Since `CredentialActivity` must remain in the stack we have to use this theme.

It's an uncommon scenario, that's why it's mentioned in the comment above.

david-livefront: 🤦 I need to read better. Thanks for the explanation

PR #6519, opened 2026-02-11
[PM-31615] feat: Updated Send network models to support email verification

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31615

## 📔 Objective

Update models to support the email verification on sends

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/data/vault/datasource/disk/VaultDiskSourceTest.kt
@@ -584,7 +584,8 @@ private const val SEND_JSON = """
     "text": "mockText-1"
   },
   "key": "mockKey-1",
-  "expirationDate": "2023-10-27T12:00:00.000Z"
+  "expirationDate": "2023-10-27T12:00:00.000Z",
+  "authType": 1
 }
 """
 


network/src/main/kotlin/com/bitwarden/network/model/SendJsonRequest.kt
```diff
@@ -64,4 +69,7 @@ data class SendJsonRequest(
 
     @SerialName("hideEmail")
     val shouldHideEmail: Boolean?,
+
+    @SerialName("emails")
+    val emails: String?,
```
david-livefront, line 74: Is this supposed to be a single string to represent multiple emails?

network/src/main/kotlin/com/bitwarden/network/model/SyncResponseJson.kt
```diff
@@ -931,12 +932,18 @@ data class SyncResponseJson(
...
+        val authType: SendAuthTypeJson?,
+
         @SerialName("accessId")
         val accessId: String?,
 
         @SerialName("password")
         val password: String?,
 
+        @SerialName("emails")
+        val emails: String?,
```
david-livefront, line 948: Can we add a doc for this new property

network/src/main/kotlin/com/bitwarden/network/model/SendJsonRequest.kt
aj-rosado: yes, it is a comma separated list of strings

https://github.com/bitwarden/server/blob/6bfacec3e85bb521b69a3488db1b4b75edb05796/src/Api/Tools/Models/Request/SendRequestModel.cs#L106

PR #6520, opened 2026-02-11
PM-31925: Replace 'android' reference with logic in LibraryExtension

david-livefront:
## 🎟️ Tracking

[PM-31925](https://bitwarden.atlassian.net/browse/PM-31925)

## 📔 Objective

This PR removes the last reference to the deprecated `android` variable.

[PM-31925]: https://bitwarden.atlassian.net/browse/PM-31925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/build.gradle.kts
+++ b/ui/build.gradle.kts
@@ -50,6 +50,7 @@ configure<LibraryExtension> {
     testFixtures {
         enable = true
     }
+    sourceSets["main"].res.directories.add("src/test/res")
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


ui/build.gradle.kts
```diff
@@ -50,6 +50,7 @@ configure<LibraryExtension> {
     testFixtures {
         enable = true
     }
+    sourceSets["main"].res.directories.add("src/test/res")
```
david-livefront, line 53: So this should not be required at all because we have the `unitTests.isIncludeAndroidResources` flag enabled. So that is something we should monitor moving forward.

david-livefront:
Thanks @SaintPatrck

PR #6528, opened 2026-02-12
PM-31664: Add new SnackbarRelay type specific for the View Screen

david-livefront:
## 🎟️ Tracking

[PM-31664](https://bitwarden.atlassian.net/browse/PM-31664)

## 📔 Objective

This PR adds two new SnackbarRelay values for the archiving and unarchiving an item from the `VaultItemScreen`. This is required because when the normal relay event is sent, the `VaultItemViewModel` was consuming it right away and the subsequent screens we not getting the event. Providing it its own relay type resolves this issue.

[PM-31664]: https://bitwarden.atlassian.net/browse/PM-31664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
@@ -167,7 +167,9 @@ class SearchViewModel @Inject constructor(
         snackbarRelayManager
             .getSnackbarDataFlow(
                 SnackbarRelay.CIPHER_ARCHIVED,
+                SnackbarRelay.CIPHER_ARCHIVED_VIEW,
                 SnackbarRelay.CIPHER_UNARCHIVED,
+                SnackbarRelay.CIPHER_UNARCHIVED_VIEW,
                 SnackbarRelay.CIPHER_DELETED,
                 SnackbarRelay.CIPHER_DELETED_SOFT,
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/search/SearchViewModel.kt
```diff
@@ -167,7 +167,9 @@ class SearchViewModel @Inject constructor(
         snackbarRelayManager
             .getSnackbarDataFlow(
                 SnackbarRelay.CIPHER_ARCHIVED,
+                SnackbarRelay.CIPHER_ARCHIVED_VIEW,
                 SnackbarRelay.CIPHER_UNARCHIVED,
+                SnackbarRelay.CIPHER_UNARCHIVED_VIEW,
```
david-livefront, line 172: I am open to some better naming pattern here. Thoughts?

SaintPatrck: Hm. Nothing better immediately comes to mind. Maybe just add docs to explain the difference?

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6554, opened 2026-02-18
Replace ZonedDateTime with Instant

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR replaces the usage of `ZonedDateTime` with `Instant`. All the dates in the app are in UTC already and the Bitwarden SDK uses Instant as well. This should simplify logic within the app and remove the need to convert class types when interacting with the SDK.


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/entity/OrganizationEventEntity.kt
```diff
@@ -24,7 +24,7 @@ data class OrganizationEventEntity(
     val cipherId: String?,
 
     @ColumnInfo(name = "date")
-    val date: ZonedDateTime,
+    val date: Instant,
```
david-livefront, line 27: The schema does not need to be updated because we have a `TypeConvertor` that is transforming the value to a Long. So the stored datatype has not actually changed.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/util/VaultSdkCipherExtensions.kt
```diff
@@ -55,7 +53,7 @@ fun Cipher.toEncryptedNetworkCipher(
             ?.associate { requireNotNull(it.id) to it.toNetworkAttachmentRequest() },
         reprompt = reprompt.toNetworkRepromptType(),
         passwordHistory = passwordHistory?.toEncryptedNetworkPasswordHistoryList(),
-        lastKnownRevisionDate = ZonedDateTime.ofInstant(revisionDate, ZoneOffset.UTC),
+        lastKnownRevisionDate = revisionDate,
```
david-livefront, line 56: Little conversions like this are just not needed anymore.

core/src/main/kotlin/com/bitwarden/core/data/serializer/InstantSerializer.kt
```diff
@@ -6,28 +6,31 @@ import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
...
+        get() = PrimitiveSerialDescriptor(serialName = "Instant", kind = PrimitiveKind.STRING)
 
-    override fun deserialize(decoder: Decoder): ZonedDateTime =
-        decoder.decodeString().let { dateString ->
-            ZonedDateTime.parse(dateString, dateTimeFormatterDeserialization)
-        }
+    override fun deserialize(decoder: Decoder): Instant =
+        ZonedDateTime
+            .parse(decoder.decodeString(), dateTimeFormatterDeserialization)
+            .toInstant()
```
david-livefront, line 31: This is one of the few remaining places that still uses the `ZonedDateTime`.

I did test out using `Instant.parse(decoder.decodeString())` and did not see any negative ramification of the change but I am unsure of a good way to validate that it works for any and all variations the server could send back to us.

So for the moment, I have left it as a `ZonedDateTime`, which we know works, and then convert it to an `Instant`.

network/src/main/kotlin/com/bitwarden/network/api/UnauthenticatedIdentityApi.kt
```diff
@@ -29,6 +30,7 @@ internal interface UnauthenticatedIdentityApi {
     @Suppress("LongParameterList")
     @FormUrlEncoded
     suspend fun getToken(
+        @Url()
         @Field(value = "scope", encoded = true) scope: String,
```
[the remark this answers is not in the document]
david-livefront: @claude This is resolved.

core/src/main/kotlin/com/bitwarden/core/data/serializer/InstantSerializer.kt
david-livefront: Any concerns with migrating directly to `Instant` with this:

```kotlin
override fun deserialize(decoder: Decoder): Instant = Instant.parse(decoder.decodeString())
```

SaintPatrck: I like that `ZonedDateTime` is more permissive. I'm assuming the `DateTimeFormatter` pattern accepts `.` or `:` as a separator for a reason, and if `:` is received, `Instant.parse()` is going to fail. `ZonedDateTime` also handles various timezone offset formats more gracefully.

If we're not able to guarantee the server's format now and in the future, `ZonedDateTime` feels like the safer approach. The flexibility is worth the couple of extra lines, imo.

david-livefront: Fair enough, we will run with this for now.

SaintPatrck (approved):
LGTM!

david-livefront:
Thanks @SaintPatrck

PR #6558, opened 2026-02-20
[PM-30916] bug: Create passkey myitems

andrebispo5:
## 🎟️ Tracking

<!-- Paste the link to the Jira or GitHub issue or otherwise describe / point to where this change is coming from. -->
https://bitwarden.atlassian.net/browse/PM-30916

## 📔 Objective

<!-- Describe what the purpose of this PR is, for example what bug you're fixing or new feature you're adding. -->
Fixes a bug where users belonging to an organization with the PERSONAL_OWNERSHIP policy enabled could not create Passkeys into the MyItems (DEFAULT_USER_COLLECTION) collection.                                                               
                                                                                                                                                                                                                                                 
**Root Cause**                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                 
  Two issues were identified:                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                 
  1. toCipherView() in VaultAddItemStateExtensions.kt used originalCipher?.collectionIds.orEmpty() which returns an empty list for new items, ignoring the selected collection IDs from the UI state.                                            
  2. saveCredential() in Fido2CredentialStoreImpl.kt always called createCipher() for new passkeys, which doesn't support organization collections. When PERSONAL_OWNERSHIP is enabled, ciphers must be created using                            
  createCipherInOrganization() with the appropriate collection IDs.                                                                                                                                                                              
                                                                                                                                                                                                                                                 
  **Changes**                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                 
  - Updated toCipherView() to use selected collection IDs from the UI state when creating new items                                                                                                                                              
  - Updated saveCredential() to use createCipherInOrganization() when the cipher belongs to an organization and has collection IDs

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
@@ -87,11 +87,13 @@ class Fido2CredentialStoreImpl(
                     }
                     ?: decryptedCipherView.createCipher()
 
-                if (result is CreateCipherResult.Error) {
-                    throw result.error
-                        ?: IllegalStateException(
+                when (result) {
+                    CreateCipherResult.Success -> Unit
+                    is CreateCipherResult.Error -> {
+                        throw result.error ?: IllegalStateException(
                             result.errorMessage ?: "Failed to save credential",
                         )
+                    }
                 }
             }
             .onFailure { throw it }


app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
```diff
@@ -79,11 +79,23 @@ class Fido2CredentialStoreImpl(
             .map { decryptedCipherView ->
                 decryptedCipherView.id
                     ?.let { vaultRepository.updateCipher(it, decryptedCipherView) }
-                    ?: vaultRepository.createCipher(decryptedCipherView)
+                    ?: decryptedCipherView.createCipher()
             }
```
SaintPatrck: I think we need to check the results here, otherwise we're ignoring any errors from create/update. 

Right now, `onFailure` is only going to catch and throw the decryption failure because it's a `Result<>`. `updateCipher()` and `createCipher()` return custom results that won't trigger `onFailure`. 

If either `updateCipher` or `createCipher` return `Error` results we need to explicitly throw so that the SDK can propagate the error back to the app. `save` is called by the SDK when we call `vaultSdkSource.registerFido2Credential()` and `vaultSdkSource.authenticateFido2Credential()`.

Looks like this is a pre-existing bug, but since we're refactoring and fixing a somewhat related issue I think it makes sense to correct it in this PR.

SaintPatrck: This would be a good opportunity to update the implementation docs since we can't modify the interface docs.

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultAddItemStateExtensionsTest.kt
```diff
@@ -963,6 +964,94 @@ class VaultAddItemStateExtensionsTest {
...
+                username = "mockUsername-1",
+                password = "mockPassword-1",
+            ),
+        )
+
+        val result = viewState.toCipherView(clock = FIXED_CLOCK, isPremiumUser = true)
+
+        assertEquals(
+            listOf("collection-1", "collection-3"),
+            result.collectionIds,
```
david-livefront: Can we test against the full result

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultAddItemStateExtensionsTest.kt
```diff
@@ -963,6 +964,94 @@ class VaultAddItemStateExtensionsTest {
...
+                username = "mockUsername-1",
+                password = "mockPassword-1",
+            ),
+        )
+
+        val result = viewState.toCipherView(clock = FIXED_CLOCK, isPremiumUser = true)
+
+        assertEquals(
+            listOf("original-collection-1", "original-collection-2"),
+            result.collectionIds,
```
david-livefront: Same here

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
andrebispo5: Can you check the last commit to see if that is inline with your suggestion?

david-livefront:
LGTM, but we should let @SaintPatrck take another pass too

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
```diff
@@ -76,14 +78,43 @@ class Fido2CredentialStoreImpl(
...
+            .onSuccess { decryptedCipherView ->
+                val result = decryptedCipherView.id
+                    ?.let {
+                        vaultRepository
+                            .updateCipher(it, decryptedCipherView)
+                            .toCreateCipherResult()
+                    }
+                    ?: decryptedCipherView.createCipher()
+
+                if (result is CreateCipherResult.Error) {
```
david-livefront: Can we make this a `when`

```kotlin
                when (result) {
                    CreateCipherResult.Success -> Unit
                    is CreateCipherResult.Error -> {
                        throw result.error ?: IllegalStateException(
                            result.errorMessage ?: "Failed to save credential",
                        )
                    }
                }
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/model/Fido2CredentialStoreImpl.kt
SaintPatrck: Looks good! Let's convert it to a `when`, like David suggested and ship it. 🥳

PR #6563, opened 2026-02-23
[PM-28834] bug: Setting configuration for VR devices on MainActivity

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28834

## 📔 Objective

Set the configuration on MainActivity so that it appears properly on HorizonOS devices.


app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -212,6 +212,14 @@ class MainActivity : AppCompatActivity() {
         .takeIf { it }
         ?: super.dispatchKeyEvent(event)
 
+    override fun onPostCreate(savedInstanceState: Bundle?) {
+        super.onPostCreate(savedInstanceState)
+        // resize only one time at the start
+        if (!mainViewModel.stateFlow.value.hasResizeBeenRequested) {
```
david-livefront, line 219: So if the activity is recreated, we still don't want to make the request again?

app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -267,4 +275,31 @@ class MainActivity : AppCompatActivity() {
...
         }
     }
+
+    private fun isHorizonOSDevice(): Boolean {
+        return Build.MANUFACTURER.equals("Oculus", ignoreCase = true) ||
+            Build.MANUFACTURER.equals("Meta", ignoreCase = true)
+    }
+
+    @Suppress("MagicNumber", "TooGenericExceptionCaught")
+    private fun setHorizonOSAppLayout() {
```
david-livefront: I wonder if we can bury this in an extension?

```kotlin
fun Activity.setHorizonOSAppLayout(
    onResizeRequested: () -> Unit,
)
```

app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -267,4 +275,31 @@ class MainActivity : AppCompatActivity() {
...
+                val clazz = Class.forName("horizonos.view.WindowExt")
+                val method = clazz.getMethod(
+                    "requestWindowResize",
+                    android.view.Window::class.java,
+                    Int::class.javaPrimitiveType,
+                    Int::class.javaPrimitiveType,
+                )
+                method.invoke(null, window, 1024, 640)
+                mainViewModel.trySendAction(MainAction.Internal.ResizeHasBeenRequested)
+            } catch (t: Throwable) {
```
david-livefront: Which part of this code will throw?

Can we wrap only the logic that is capable of throwing

app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
aj-rosado: yes, that is the intended behaviour, resize request runs only once during startup, after the UI is created, and not in response to user-driven panel resizing.

david-livefront (approved):

PR #6581, opened 2026-02-25
Remove unnecessary opt-in annotations

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR removes som opt-in annotations that are no longer needed from the codebase.


david-livefront:
Thanks @SaintPatrck

PR #6582, opened 2026-02-25
Replace test workflow with sharded parallel CI execution

SaintPatrck:
## 🎟️ Tracking

No Jira ticket — CI infrastructure improvement.

## 📔 Objective

Replace the existing sequential `test.yml` workflow with a sharded version that runs tests across 7 parallel matrix jobs + 1 coverage notification job, targeting ~12-14 minute wall-clock time versus the current ~45-50 minute sequential run.

**New Fastlane lanes:** `checkLint`, `testAppShard`, `testLibraries`, `generateCoverageReport`

**Job structure:**

A single `test` job uses a matrix strategy with 7 entries:

| Matrix group | Fastlane lane | What it runs |
|-----|------|-------------|
| `app-data` | `testAppShard` | `:app` data tests |
| `app-ui-auth-tools` | `testAppShard` | `:app` UI auth, tools, autofill, credentials, and root-level tests |
| `app-ui-platform` | `testAppShard` | `:app` UI platform tests |
| `app-ui-vault` | `testAppShard` | `:app` UI vault tests |
| `authenticator` | `testLibraries` | `:authenticator` tests |
| `lib-core-network-bridge` | `testLibraries` | `:core :network :cxf :authenticatorbridge :testharness` tests |
| `lib-data-ui` | `testLibraries` | `:data :ui` tests |

Additional jobs:
- `static-analysis` — detekt + lint
- `coverage-notify` — Sends `send-notifications` to Codecov after all test jobs complete

The old `test.yml` is replaced in-place (no parallel workflow needed).

## ⚠️ Codecov behavior with per-shard uploads

Each shard uploads its own coverage report to Codecov with a unique flag (e.g., `app-data`, `lib-core-network-bridge`). Codecov processes these uploads asynchronously, so **the PR comment from the Codecov bot will update incrementally** as each shard's report is processed. During this intermediate state, you may see:

- A warning about "a different number of reports uploaded between BASE and HEAD"
- Artificially low project coverage (e.g., 27% instead of ~86%)

Intermediate Codecov comment showing partial coverage
*Example: Codecov PR comment after only 3 of 7 shard uploads have been processed.*

This is expected. Once all 7 shard uploads are processed and the `coverage-notify` job sends the `send-notifications` signal, Codecov merges all reports and the PR comment updates to reflect the full project coverage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
             fastlane_options: "--tests com.x8bit.bitwarden.data.*"
           - group: app-ui-auth-tools
             fastlane_method: testAppShard
-            fastlane_options: "--tests com.x8bit.bitwarden.ui.auth.* --tests com.x8bit.bitwarden.ui.tools.* --tests com.x8bit.bitwarden.ui.autofill.* --tests com.x8bit.bitwarden.ui.credentials.* --tests com.x8bit.bitwarden.MainViewModelTest --tests com.x8bit.bitwarden.CredentialProviderViewModelTest --tests com.x8bit.bitwarden.AuthCallbackViewModelTest --tests com.x8bit.bitwarden.AutofillCallbackViewModelTest"
+            fastlane_options: "--tests com.x8bit.bitwarden.ui.auth.* --tests com.x8bit.bitwarden.ui.tools.* --tests com.x8bit.bitwarden.ui.autofill.* --tests com.x8bit.bitwarden.ui.credentials.*"
           - group: app-ui-platform
             fastlane_method: testAppShard
             fastlane_options: "--tests com.x8bit.bitwarden.ui.platform.*"
@@ -68,6 +68,12 @@ jobs:
           _FASTLANE_METHOD: ${{ matrix.fastlane_method }}
           _FASTLANE_OPTIONS: ${{ matrix.fastlane_options }}
         run: |
+          if [ "$_GROUP" = "app-ui-auth-tools" ]; then
+            _TOP_LEVEL_TESTS=$(basename -a -s .kt app/src/test/kotlin/com/x8bit/bitwarden/*Test.kt \
+              | xargs -I{} printf ' --tests com.x8bit.bitwarden.{}')
+            _FASTLANE_OPTIONS="${_FASTLANE_OPTIONS} ${_TOP_LEVEL_TESTS}"
+          fi
+
           if [ "$_GROUP" = "static-analysis" ]; then
             bundle exec fastlane "$_FASTLANE_METHOD"
           else


.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,390 @@
...
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - shard_name: data
+            test_filter: '--tests "com.x8bit.bitwarden.data.*"'
+          # NOTE: Root-level tests (e.g., MainViewModelTest) must be explicitly
+          # listed here since they don't match any package glob pattern.
+          # If a new test is added to `com.x8bit.bitwarden` (not a sub-package),
```
vvolkgang: Too easy to miss 😔 . Would a class name wildcard work? e.g. `com.x8bit.bitwarden.*Test`. Or some test case group that's easier for devs to infer the pattern.

.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,390 @@
+name: Test (Sharded)
+
+on:
+  push:
+    branches:
+      - "main"
+      - "rc"
+      - "hotfix-rc"
```
vvolkgang: Just changed this in another PR for `test.yml`: 

```suggestion
      - main
      - release/**/*
```

.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,390 @@
...
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
+          persist-credentials: false
+
+      - name: Validate Gradle wrapper
+        uses: gradle/actions/wrapper-validation@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
+
+      - name: Cache Gradle files
```
vvolkgang: After https://github.com/bitwarden/android/pull/6583/ merges these can be replaced with the Setup Android Build action step.

.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,390 @@
...
+      - "rc"
+      - "hotfix-rc"
+  pull_request:
+    types: [ opened, synchronize ]
+  merge_group:
+    types: [ checks_requested ]
+  workflow_dispatch:
+
+env:
+  _JAVA_VERSION: 21
```
vvolkgang: After https://github.com/bitwarden/android/pull/6583/ merges this can be removed.

.github/workflows/test-sharded.yml
SaintPatrck: True. I wasn't a huge fan of this. I like the filename pattern. Great idea.

.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,244 @@
...
+            app/build/reports/detekt/
+
+  test-app:
+    name: "Test App (${{ matrix.shard_name }})"
+    runs-on: ubuntu-24.04
+    permissions:
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
```
vvolkgang: 💭 We can follow a similar approach to consolidate all jobs into a single one, as steps are pretty much the same between them. Example of inputs following a "TestGroup" analogy:

```
group: app-data
fastlane_method: testAppShard
fastlane_options: 'testFilter: --tests "com.x8bit.bitwarden.data.*"'
```

* `group` would be used for the Job name (`name: "Test ${{ matrix.group }}"`), codecov.io flag and "Upload test reports" filename.
* It'll probably be easier to have a single public Fastfile lane that receives `group` and `fastlane_options` (removing the need for `fastlane_method` in this matrix), internally calling the right tasks or routing to private lanes.
* For the "Upload test reports" path, we may be able to get away with a wildcard like `**/build/reports/tests/`, not sure. Listing all paths should also work too, there's a argument to control what happens when [files aren't found](https://github.com/actions/upload-artifact?tab=readme-ov-file#customization-if-no-files-are-found) `if-no-files-found:`, defaults to `warn`.

.github/workflows/test-sharded.yml
```diff
@@ -0,0 +1,244 @@
+name: Test (Sharded)
+
```
vvolkgang: 🎨 Something we should add to other workflows too, `run-name:` that makes it easier to identify what triggered it. E.g. this should format the title as "Test PR 123" for pull requests falling back to "Test branchX" (or tag or commit hash) for everything else.

`run-name: Test ${{ github.event.pull_request && format('PR {0}', github.event.pull_request.number) || github.event.ref_name }}`

.github/workflows/test.yml
```diff
@@ -31,49 +31,123 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        include:
+          # App shards
+          - group: app-data
+            fastlane_method: testAppShard
+            fastlane_options: "testFilter:'--tests \"com.x8bit.bitwarden.data.*\"'"
+          - group: app-ui-auth-tools
+            fastlane_method: testAppShard
+            fastlane_options: "testFilter:'--tests \"com.x8bit.bitwarden.ui.auth.*\" --tests \"com.x8bit.bitwarden.ui.tools.*\" --tests \"com.x8bit.bitwarden.ui.autofill.*\" --tests \"com.x8bit.bitwarden.ui.credentials.*\" --tests \"com.x8bit.bitwarden.*Test\"'"
```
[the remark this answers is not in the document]
SaintPatrck: @vvolkgang We rarely add files to this root package, so I think it's acceptable to list them explicitly. We can add documentation somewhere in hopes of reducing the chance of this getting missed. I recently added `TROUBLESHOOTING` so maybe that's the right place?

SaintPatrck:
@vvolkgang `Test` is still showing as required. Guess we can't avoid updating the required checks. 😞 

Can you do that?

.github/workflows/test.yml
```diff
@@ -15,12 +16,11 @@ env:
   _GITHUB_ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}/attempts/${{ github.run_attempt }}
 
 jobs:
-  test:
-    name: Test
+  static-analysis:
```
vvolkgang: 💭 This job could also be consolidated too

```
          - group: static-analysis
            fastlane_method: checkLint
            fastlane_options: ""
```
And if needed skipping the coverage steps when group == `static-analysis`.

.github/workflows/test-sharded.yml
vvolkgang: @SaintPatrck Reverted this, GitHub's default for push / pull_request triggers (using the PR title) works better, won't be ideal for workflow_dispatch but those are few and far between.

vvolkgang:
> @vvolkgang `Test` is still showing as required. Guess we can't avoid updating the required checks. 😞
> 
> Can you do that?

Sharing what we landed on for posterity:

Required checks are setup based on the job name and we can't define them with wildcards, we would need to pick all shards individually and keep maintaining the rule when new ones are added / removed. Instead, added a job with the same name we had which waits for all shards to end and report the status.

| Success test | Fail test |
|--------|--------|
| image | image |

PR #6583, opened 2026-02-25
[PM-32758] ci: Improve CI cache to fix GitHub runners running out of memory

vvolkgang:
## 🎟️ Tracking

PM-32758

## 📔 Objective

Fix GitHub runners running out of memory by replacing our cache approach with `setup-gradle` action. This action comes with a sensible set of default settings following best practices, worth nothing:

* [Cache cleanup](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#configuring-cache-cleanup)
* [Restricted write](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#select-which-branches-should-write-to-the-cache) - defaults to `main` branches only, mitigating cache poisoning and useful caches from `main` being replaced by feature branch caches that aren't useful for others, slowing down all runs.    

While at it:

1. Consolidated everything into our `setup-andoid-build` reusable action
2. Removed local build-cache path override from settings.gradle.kts. Gradle's default cache location is used instead.
3. Increased Gradle heap size from 4g to 6g based on the Post `setup-gradle` action logs, mentioning "cache misses" due to low memory. Test runs - [before](https://github.com/bitwarden/android/actions/runs/22367538659/job/64737475122)  | [after](https://github.com/bitwarden/android/actions/runs/22405671990/job/64864713540)
4. Test workflow was updated to run on release branches in addition to main.

Test Runs (forced to write cache):
* [build-testharness.yml](https://github.com/bitwarden/android/actions/runs/22405671990/job/64864713540)
* [build.yml](https://github.com/bitwarden/android/actions/runs/22411262441)
* [build-authenticator.yml](https://github.com/bitwarden/android/actions/runs/22411270020) 
* [test.yml](https://github.com/bitwarden/android/actions/runs/22412907340/job/64890898693)

### DIFF
--- a/.github/actions/setup-android-build/action.yml
+++ b/.github/actions/setup-android-build/action.yml
@@ -10,8 +10,6 @@ runs:
   steps:
     - name: Setup Gradle
       uses: gradle/actions/setup-gradle@f29f5a9d7b09a7c6b29859002d29d24e1674c884 # v5.0.1
-      with:
-        cache-read-only: false
 
     - name: Configure Ruby
       uses: ruby/setup-ruby@44511735964dcb71245e7e55f72539531f7bc0eb # v1.257.0


settings.gradle.kts
```diff
@@ -39,13 +39,6 @@ dependencyResolutionManagement {
     }
 }
 
-buildCache {
-    local {
-        isEnabled = true
-        directory = File(rootDir, "build-cache")
```
SaintPatrck, line 45: 🗒️ Just want to note that the first local builds after this merges will take a bit longer since the default cache location will be used going forward. 

I also recommend checking your `build-cache` folder and deleting it once this merges so the existing caches are not orphaned.

CC @bitwarden/team-android @david-livefront

.github/workflows/test.yml
```diff
@@ -3,17 +3,15 @@ name: Test
 on:
   push:
     branches:
-      - "main"
-      - "rc"
-      - "hotfix-rc"
+      - main
+      - release/**/*
```
david-livefront, line 7: Do we not need `hotfix` support anymore?

gradle.properties
```diff
@@ -5,7 +5,7 @@ kotlin.code.style=official
 kotlin.compilerArgs=-Xjspecify-annotations=strict, -Xtype-enhancement-improvements-strict-mode
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx6g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -Dfile.encoding=UTF-8
```
david-livefront, line 8: 😄

.github/workflows/build-authenticator.yml
```diff
@@ -31,7 +31,6 @@ on:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  JAVA_VERSION: 21
```
david-livefront, line 34: So we don't need to specify the java version anymore?

.github/workflows/test.yml
vvolkgang: Hotfix branches are created under `release/` too

.github/workflows/build-authenticator.yml
vvolkgang: It's centralised in the `setup-android-build` action input. 

https://github.com/bitwarden/android/blob/39d483674b55618eb49399c25b2e9837106ad904/.github/actions/setup-android-build/action.yml#L7

PR #6589, opened 2026-02-26
[PM-32714] Add cookie domain-suffix resolution and fix cloud config path exclusion

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32714

## 📔 Objective

Cookies stored under a parent cookieDomain (e.g., "bitwarden.com") were not found when requesting via subdomains (e.g., "api.bitwarden.com"). Additionally, cloud servers use "/config" instead of "/api/config", causing the config endpoint to hit cookie handling and read stale state.

This adds a `resolveHostname` helper for domain-suffix fallback, updates `storeCookies` to key on `cookieDomain`, fixes excluded path matching to cover both cloud and self-hosted config paths, and adds debug logging along the cookie management path.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/network/NetworkCookieManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/network/NetworkCookieManagerImpl.kt
@@ -7,6 +7,7 @@ import com.x8bit.bitwarden.data.platform.datasource.disk.model.CookieConfigurati
 import com.x8bit.bitwarden.data.platform.manager.CookieAcquisitionRequestManager
 import com.x8bit.bitwarden.data.platform.manager.model.CookieAcquisitionRequest
 import com.x8bit.bitwarden.data.platform.manager.util.toNetworkCookieList
+import timber.log.Timber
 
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/network/NetworkCookieManagerImpl.kt
```diff
@@ -19,35 +20,55 @@ class NetworkCookieManagerImpl(
...
+                    }
 
-                else -> false
+                    else -> false
+                }
             }
-        }
-        ?: false
+            ?: false
+        Timber.d("needsBootstrap($hostname): $result (cookieDomain=$cookieDomain)")
```
david-livefront, line 56: I assume we are OK to log this stuff

SaintPatrck: Yea. I was hesitant at first, but we already log instance domain details in HTTP calls when logging is enabled, so this is not exposing anything new. If you feel otherwise I don't mind replacing it with a more generic message.

PR #6592, opened 2026-02-26
[PM-32802] fix: 400 error when archiving/unarchiving org-owned ciphers

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32802

## 📔 Objective

Fix a 400 error when users with restricted permissions ("can view, hidden passwords" or "can edit, hidden passwords") attempt to archive or unarchive org-owned ciphers.

**Root cause:** `archiveCipher()` and `unarchiveCipher()` in `CipherManagerImpl` unnecessarily called `encryptCipherAndCheckForMigration()` before the archive/unarchive API call. When the SDK generated a new cipher key (migration), this triggered a full `PUT /api/ciphers/{id}` update that the server rejected with HTTP 400 for users lacking edit permissions — aborting the archive before it was ever attempted.

**Fix:** Remove the encryption/migration check from archive and unarchive flows entirely. These operations now call the archive/unarchive API directly and upsert the server-returned cipher to disk, following the existing `restoreCipher` pattern. The server handles all state changes; no client-side encryption is needed.

**Changes:**
- **CiphersApi** — Archive/unarchive return `SyncResponseJson.Cipher` instead of `Unit`
- **CiphersService/Impl** — Return typed `ArchiveCipherResponseJson` / `UnarchiveCipherResponseJson` with 400 error recovery
- **CipherManagerImpl** — Remove `encryptCipherAndCheckForMigration` from archive/unarchive; upsert server response directly
- **Response models** — Add `ArchiveCipherResponseJson` and `UnarchiveCipherResponseJson` sealed classes
- **Tests** — Updated CipherManagerTest and CiphersServiceTest to match new behavior

### DIFF
--- a/network/src/test/kotlin/com/bitwarden/network/service/CiphersServiceTest.kt
+++ b/network/src/test/kotlin/com/bitwarden/network/service/CiphersServiceTest.kt
@@ -66,7 +66,6 @@ class CiphersServiceTest : BaseServiceTest() {
         unmockkStatic(Uri::class)
     }
 
-    @Suppress("MaxLineLength")
     @Test
     fun `archiveCipher with success response should return a Success with the correct cipher`() =
         runTest {
@@ -82,7 +81,6 @@ class CiphersServiceTest : BaseServiceTest() {
             )
         }
 
-    @Suppress("MaxLineLength")
     @Test
     fun `archiveCipher with an invalid response should return an Invalid with the correct data`() =
         runTest {
@@ -101,7 +99,6 @@ class CiphersServiceTest : BaseServiceTest() {
             )
         }
 
-    @Suppress("MaxLineLength")
     @Test
     fun `unarchiveCipher with success response should return a Success with the correct cipher`() =
         runTest {


network/src/test/kotlin/com/bitwarden/network/service/CiphersServiceTest.kt
```diff
@@ -64,21 +66,75 @@ class CiphersServiceTest : BaseServiceTest() {
...
+            assertEquals(
+                ArchiveCipherResponseJson.Invalid(
+                    message = "You do not have permission to edit this.",
+                    validationErrors = null,
+                ),
+                result.getOrThrow(),
+            )
+        }
+
+    @Suppress("MaxLineLength")
```
david-livefront: I don't think we need this suppression

network/src/test/kotlin/com/bitwarden/network/service/CiphersServiceTest.kt
```diff
@@ -64,21 +66,75 @@ class CiphersServiceTest : BaseServiceTest() {
...
+            val result = ciphersService.archiveCipher(cipherId = "cipherId")
+            assertEquals(
+                ArchiveCipherResponseJson.Success(
+                    cipher = createMockCipher(number = 1),
+                ),
+                result.getOrThrow(),
+            )
+        }
 
+    @Suppress("MaxLineLength")
```
david-livefront: Same here

network/src/test/kotlin/com/bitwarden/network/service/CiphersServiceTest.kt
```diff
@@ -64,21 +66,75 @@ class CiphersServiceTest : BaseServiceTest() {
         unmockkStatic(Uri::class)
     }
 
+    @Suppress("MaxLineLength")
```
david-livefront: And here

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CipherManagerImpl.kt
```diff
@@ -207,29 +205,25 @@ class CipherManagerImpl(
         cipherView: CipherView,
     ): UnarchiveCipherResult {
         val userId = activeUserId ?: return UnarchiveCipherResult.Error(NoActiveUserException())
-        return cipherView
-            .encryptCipherAndCheckForMigration(userId = userId, cipherId = cipherId)
```
david-livefront, line 211: So this was completely unnecessary?

SaintPatrck: At some point it was because the archive endpoints were not returning the entire updated cipher, it was only returning the "mini" version. This was never updated when that changed.

PR #6596, opened 2026-02-27
🍒 [PM-32802] fix: 400 error when archiving/unarchiving org-owned ciphers

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32802

🍒Cherry-picked to `2026.2-rc47` from #6592

## 📔 Objective

Fix a 400 error when users with restricted permissions ("can view, hidden passwords" or "can edit, hidden passwords") attempt to archive or unarchive org-owned ciphers.

**Root cause:** `archiveCipher()` and `unarchiveCipher()` in `CipherManagerImpl` unnecessarily called `encryptCipherAndCheckForMigration()` before the archive/unarchive API call. When the SDK generated a new cipher key (migration), this triggered a full `PUT /api/ciphers/{id}` update that the server rejected with HTTP 400 for users lacking edit permissions — aborting the archive before it was ever attempted.

**Fix:** Remove the encryption/migration check from archive and unarchive flows entirely. These operations now call the archive/unarchive API directly and upsert the server-returned cipher to disk, following the existing `restoreCipher` pattern. The server handles all state changes; no client-side encryption is needed.

**Changes:**
- **CiphersApi** — Archive/unarchive return `SyncResponseJson.Cipher` instead of `Unit`
- **CiphersService/Impl** — Return typed `ArchiveCipherResponseJson` / `UnarchiveCipherResponseJson` with 400 error recovery
- **CipherManagerImpl** — Remove `encryptCipherAndCheckForMigration` from archive/unarchive; upsert server response directly
- **Response models** — Add `ArchiveCipherResponseJson` and `UnarchiveCipherResponseJson` sealed classes
- **Tests** — Updated CipherManagerTest and CiphersServiceTest to match new behavior

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CipherManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CipherManagerImpl.kt
@@ -8,10 +8,12 @@ import com.bitwarden.core.data.util.asSuccess
 import com.bitwarden.core.data.util.flatMap
 import com.bitwarden.data.manager.file.FileManager
 import com.bitwarden.data.manager.model.DownloadResult
+import com.bitwarden.network.model.ArchiveCipherResponseJson
 import com.bitwarden.network.model.AttachmentJsonResponse
 import com.bitwarden.network.model.CreateCipherInOrganizationJsonRequest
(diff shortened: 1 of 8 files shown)


david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #6608, opened 2026-03-02
Move FakeLifecycle to core module

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR moves the `FakeLifecycle` and `FakeLifecycleOwner` to the `core` module so it can be shared across all tests that need it.


david-livefront:
Thanks @SaintPatrck

PR #6627, opened 2026-03-09
[PM-33262] feat: Add cookie support to Glide image requests

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33262

## 📔 Objective

Enterprise environments using SSO cookie vendors (e.g., Cloudflare Access) require cookies on image requests for icons to load through cookie-gated proxies. Without cookies, Glide requests fail silently and icons don't display.

This adds a lenient `GlideCookieInterceptor` to Glide's OkHttpClient that:
- Attaches cookies from `NetworkCookieManager` to image requests by hostname
- Never throws exceptions or triggers cookie bootstrap (unlike the API `CookieInterceptor`)
- Degrades gracefully if no cookies are available — images simply won't load rather than crashing

### Changes
- **`GlideCookieInterceptor`** (new): Lenient OkHttp interceptor that reads cookies via `CookieProvider` and attaches them to Glide requests
- **`BitwardenAppGlideModule`**: Wires `NetworkCookieManager` into the Glide OkHttpClient via Hilt entry point
- **Tests**: 5 interceptor tests + 1 entry point test covering cookie attachment, empty cookies, 302 handling, and no-bootstrap verification

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/glide/BitwardenAppGlideModule.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/glide/BitwardenAppGlideModule.kt
@@ -60,7 +60,7 @@ class BitwardenAppGlideModule : AppGlideModule() {
         val client = certificateManager
             .createMtlsOkHttpClient()
             .newBuilder()
-            .addInterceptor(GlideCookieInterceptor(networkCookieManager))
+            .addNetworkInterceptor(GlideCookieInterceptor(networkCookieManager))
             .build()
 
         // Register OkHttpUrlLoader that uses our mTLS + cookie OkHttpClient


app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/glide/GlideCookieInterceptor.kt
```diff
@@ -0,0 +1,38 @@
...
+            .getCookies(hostname)
+            .joinToString("; ") { "${it.name}=${it.value}" }
+
+        if (cookieHeader.isEmpty()) return chain.proceed(originalRequest)
+
+        val request = originalRequest
+            .newBuilder()
+            .header(HEADER_COOKIE, cookieHeader)
+            .build()
+        return chain.proceed(request)
```
david-livefront: So for images, if a 302 occurs, we just let it fail like normal?

SaintPatrck: Yea, I think that's the best approach for a couple of reasons. 

If we trigger acquisition from Glide it's going to trigger for every icon that fails to load. That could get weird.

If we let Glide cache the redirect it ends up caching the login page, then Glide crashes when trying to decode it from cache because it's not an image. 

This approach avoids any timing issues/overhead from `n` icon load failure, and prevents corrupting the cache when cookies are need.

david-livefront: Good points, this seems fine to me

david-livefront: Do we want to catch the 302 and fail explicitly?

Or are e OK with letting the 302 happen?

SaintPatrck: Updated to actually catch and throw our internal exception. Forgot to stage that part. Thanks!

PR #6616, opened 2026-03-05
[PM-29673] feat: Improved pre-polutated data on the FlightRecorder logs

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29673

## 📔 Objective

Added more info to the header of the FlightRecorder logs, making it more similar to the info displayed on the About's copy version.
Moved `BuildInfoManagerExtensions.kt` from UI to Core

PR #6633, opened 2026-03-10
[PM-24380] fix: Correct and redact flight recorder hostname on logs

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24380

## 📔 Objective

This aims two address different scenarios.

1. On the network exception we are logging the cloud api when the user is self-hosted. This is caused because we are looking at the instance without the `BaseUrlInterceptor` having updated the url. So now we are checking on the `NetworkResultCall` if we should update the BaseUrl using the existing `BaseUrlsProvider`
2. Replace the hostname from self-hosted instance on flight recorder logs in order to protect PII information from the logs.


network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
```diff
@@ -67,8 +71,32 @@ internal class NetworkResultCall<T>(
...
+            }
+        } else {
+            null
+        }
+
+        // Rebuild the URL without query params, using actual host if available
+        val url = if (actualHost != null) {
+            "${originalUrl.protocol}://$actualHost${originalUrl.path}"
+        } else {
+            "${originalUrl.protocol}://${originalUrl.authority}${originalUrl.path}"
```
vvolkgang, line 79: Do we need to rebuild here, can't we return `originalUrl`?

SaintPatrck (commented):
❓ I recently added logging of `hostname` to `CookieIntercepter`. Do those need to be checked/redacted also?

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
```diff
@@ -103,16 +107,33 @@ internal class FlightRecorderWriterImpl(
                         bw.append(it)
                     }
                     bw.append(" – ")
-                    bw.append(message)
+                    bw.append(message.redactUrls()) // Apply hostname redaction
                     throwable?.let {
                         bw.append(" – ")
-                        bw.append(it.getStackTraceString())
+                        bw.append(it.getStackTraceString().redactUrls()) // Also redact stack traces
```
david-livefront: Can we merge these changes first?
https://github.com/bitwarden/android/pull/6616

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
```diff
@@ -67,8 +71,32 @@ internal class NetworkResultCall<T>(
     fun executeForResult(): NetworkResult<T> = requireNotNull(execute().body())
 
     private fun Throwable.toFailure(): NetworkResult<T> {
-        // We rebuild the URL without query params, we do not want to log those
-        val url = backingCall.request().url.toUrl().run { "$protocol://$authority$path" }
+        val originalUrl = backingCall.request().url.toUrl()
+
+        // Check if this is a hardcoded default URL that will be replaced by BaseUrlInterceptor
+        // Match against the defaults from RetrofitsImpl.kt line 111 and EnvironmentUrlDataJson
+        val actualHost = if (baseUrlsProvider != null) {
```
david-livefront: Can we use a `let`

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
```diff
@@ -67,8 +71,32 @@ internal class NetworkResultCall<T>(
...
+            when (originalUrl.host) {
+                "api.bitwarden.com" -> baseUrlsProvider.getBaseApiUrl().toHttpUrlOrNull()?.host
```
david-livefront: We can avoid all this complexity with an Interceptor that handles the error logging.

```kotlin
class NetworkErrorLogInterceptor : Interceptor {
    override fun intercept(chain: Interceptor.Chain): Response {
        return chain
            .proceed(chain.request())
            .also {
                if (!it.isSuccessful) {
                    val url = it.request.url.toUrl().run { "$protocol://$authority$path" }
                    Timber.e("Network Error: $url")
                }
            }
    }
}
```

aj-rosado:
> ❓ I recently added logging of `hostname` to `CookieIntercepter`. Do those need to be checked/redacted also?

This change is redacting them, so we are fine

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
aj-rosado: on it

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
aj-rosado: We are hiding the query params on the rebuild

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
aj-rosado: 🤔 The kind of logs that are showing the wrong url are not being catched by the interceptors.
The `UnknownHostException` is only being returned by the `enqueue`'s `callback` from `Retrofit`

david-livefront: I see, that is pretty annoying. I just don't like the idea of having the `baseUrlsProvider` in this class, as it breaks our separation of concerns.

I'd be curious if @SaintPatrck has any thoughts about this?

SaintPatrck: Apologies for not seeing this earlier.

🤔 I agree that injecting `baseUrlsProvider` here feels like the wrong layer. Since we're already comparing against hardcoded cloud hostnames, we could skip the provider entirely and just categorize them. If we map `*.bitwarden.com` → "US", `*.bitwarden.eu` → "EU", anything else → "SH", self-hosted hostnames never touch the log, the wrong-URL bug disappears, and the class stays clean.

Thoughts?

david-livefront: That would resolve my worries. Unless there is a desire to get specific base urls?

SaintPatrck: I re-read comments on the Jira ticket and it seems like there's a desire to keep the cloud hostnames. If that's still the case, we can just check and map the self-hosted hostnames to `[REDACTED]` and leave the known ones alone.

aj-rosado: Initial approach was to have a regex that would identify URL patterns and [REDACT] them.
This had one issue, sometimes we are throwing logs with the hostname that is similar to a namespace.
Like on UnknownHostException
`java.net.UnknownHostException: Unable to resolve host "vault.qa.bitwarden.pw": No address associated with hostname`

and NetworkCookieManager
`2026-03-02 13:05:31:871 – DEBUG – NetworkCookieManagerImpl – getCookies(vault.qa.bitwarden.pw): resolved=vault.qa.bitwarden.pw, count=0`

We can have a more specific approach to these scenarios, we might have unidentified scenarios that are leaking the hostname and we need to be aware of this on future implementations.

I am not sure if we feel that need to not redact the hostname on cloud or if it was simply a nice to have @vvolkgang

vvolkgang: Knowing if the actual log was a call to our servers or not will help us troubleshoot issues.

> we can just check and map the self-hosted hostnames to [REDACTED] and leave the known ones alone.

works although @aj-rosado will know better.

aj-rosado: If we want to keep the cloud domains without the `BaseUrlProvide` we end up with the wrong URL on the UnknownHostException, as it will always return `https://api.bitwarden.com/`, I am not sure if we have an alternative to get the correct url on this scenario 🤔 

> 2026-03-04 17:39:47:781 – WARNING – NetworkResultCall – Network Error: https://api.bitwarden.com/accounts/prelogin
> java.net.UnknownHostException: Unable to resolve host "vault.qa.bitwarden.pw": No address associated with hostname

david-livefront: We have a concept of `EnvironmentRegion` in the app already.

```kotlin
enum class EnvironmentRegion {
    UNITED_STATES,
    EUROPEAN_UNION,
    INTERNAL,
    SELF_HOSTED,
}
```

What if we replaced the hostname with `[UNITED_STATES]` or `[SELF_HOST]`?

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
```diff
@@ -34,6 +36,19 @@ internal class FlightRecorderWriterImpl(
...
+        get() {
+            val environment = serverConfigRepository.serverConfigStateFlow.value
+                ?.serverData?.environment ?: return emptySet()
+            return listOfNotNull(
+                environment.vaultUrl,
+                environment.apiUrl,
+                environment.identityUrl,
+                environment.notificationsUrl,
+                environment.ssoUrl,
+            ).mapNotNull { runCatching { URI(it).host }.getOrNull() }.toSet()
```
david-livefront: Can we use the chain formatting here:

```kotlin
return listOfNotNull(
    environment.vaultUrl,
    environment.apiUrl,
    environment.identityUrl,
    environment.notificationsUrl,
    environment.ssoUrl,
)
    .mapNotNull { runCatching { URI(it).host }.getOrNull() }
    .toSet()
```

david-livefront: Additionally, can we drop the `runCatching` and use one of the helper methods for this:

```kotlin
toUriOrNull().host
```

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
```diff
@@ -26,7 +28,8 @@ internal class NetworkResultCall<T>(
 ) : Call<NetworkResult<T>> {
     override fun cancel(): Unit = backingCall.cancel()
 
-    override fun clone(): Call<NetworkResult<T>> = NetworkResultCall(backingCall, successType)
+    override fun clone(): Call<NetworkResult<T>> =
+        NetworkResultCall(backingCall, successType)
```
david-livefront: Why did this get changed?

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCallAdapter.kt
```diff
@@ -13,5 +13,6 @@ internal class NetworkResultCallAdapter<T>(
 ) : CallAdapter<T, Call<NetworkResult<T>>> {
 
     override fun responseType(): Type = successType
-    override fun adapt(call: Call<T>): Call<NetworkResult<T>> = NetworkResultCall(call, successType)
+    override fun adapt(call: Call<T>): Call<NetworkResult<T>> =
+        NetworkResultCall(call, successType)
```
david-livefront: Why did this change?

aj-rosado: On a previous commit it had more properties and was not correctly reverted to original state

network/src/main/kotlin/com/bitwarden/network/core/NetworkResultCall.kt
aj-rosado: On a previous commit it had more properties and was not correctly reverted to original state

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
aj-rosado: using `toUriOrNull()` would imply to move it and other methods from `:app` into `:core`. 
Can make it just wanted to confirm we want it before proceeding

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
```diff
@@ -34,6 +36,21 @@ internal class FlightRecorderWriterImpl(
...
+            val environment = serverConfigRepository.serverConfigStateFlow.value
+                ?.serverData?.environment ?: return emptySet()
+            return listOfNotNull(
+                environment.vaultUrl,
+                environment.apiUrl,
+                environment.identityUrl,
+                environment.notificationsUrl,
+                environment.ssoUrl,
+            )
+                .mapNotNull { runCatching { URI(it).host }.getOrNull() }
```
david-livefront: Can we use a helper to avoid this `runCatching`

data/src/main/kotlin/com/bitwarden/data/manager/flightrecorder/FlightRecorderWriterImpl.kt
```diff
@@ -47,7 +47,7 @@ internal class FlightRecorderWriterImpl(
                 environment.notificationsUrl,
                 environment.ssoUrl,
             )
-                .mapNotNull { runCatching { URI(it).host }.getOrNull() }
+                .mapNotNull { it.toUri().host }
```
david-livefront, line 50: Should this be `toUriOrNull()`?

aj-rosado: This call the `Uri.parse(string)` and it only throws if the parameter string is null. On this scenario we have the `mapNotNull` so I believe we are safe to use as it is

david-livefront (approved):

david-livefront (approved):

PR #6639, opened 2026-03-11
[PM-33394] fix: Propagate CookieRedirectException error message

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33394

## 📔 Objective

When SSO cookies expire in a load-balanced environment, API actions trigger a `CookieRedirectException`. After the app completes the "Sync with browser" re-authentication flow, a **generic error** was shown ("We were unable to process your request") instead of the exception's descriptive message ("Your request was interrupted because the app needed to re-authenticate. Please try again.").

**Root cause:** `CookieRedirectException` extends `IOException`, so `toBitwardenError()` classified it as `BitwardenError.Network`. The `parseErrorBodyOrNull()` function only handles `BitwardenError.Http`, returning `null` for network errors — causing the message to be lost.

**Fix:** Updated `toBitwardenError()` to intercept `CookieRedirectException` before the generic `IOException` check (since it's a subclass). Synthesizes a `BitwardenError.Http` with a 400 response and JSON body containing the exception message, allowing it to flow through the existing `parseErrorBodyOrNull` pipeline with zero changes to downstream code.

**Scope:** Fixes the 26 service methods using `recoverCatching` + `parseErrorBodyOrNull` (cipher create/update/archive/unarchive, send CRUD, account operations, identity operations, folder update, cipher import). The 80 service methods without `recoverCatching` still show a generic error but preserve the exception via "Share Error Details" — to be addressed in a follow-up.

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/d0628145-699f-4696-9d4e-80f6c078b89c" />

### DIFF
--- a/network/src/main/kotlin/com/bitwarden/network/model/BitwardenError.kt
+++ b/network/src/main/kotlin/com/bitwarden/network/model/BitwardenError.kt
@@ -1,6 +1,9 @@
 package com.bitwarden.network.model
 
+import com.bitwarden.network.exception.CookieRedirectException
+import okhttp3.ResponseBody.Companion.toResponseBody
 import retrofit2.HttpException
+import retrofit2.Response
 import java.io.IOException
 
 /**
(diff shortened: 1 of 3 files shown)


david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #6650, opened 2026-03-12
llm: Add resolving-sdk-updates skill

SaintPatrck:
## 🎟️ Tracking

Internal tooling — no Jira ticket.
Used to resolve breaking changes in #6615 

## 📔 Objective

Adds a new Claude Code skill (`resolving-sdk-updates`) that teaches Claude to diagnose and resolve build failures from Bitwarden SDK (`com.bitwarden:sdk-android`) version updates.

**The skill provides a five-phase workflow:**
1. **Identify** — extract SDK version diff and parse PR body for `sdk-internal` PR references and Jira tickets
2. **Diagnose** — categorize compiler errors (exhaustive `when`, removed API, renamed type, new required parameter, new error type)
3. **Investigate** — read `sdk-internal` PR diffs to understand the author's intent behind breaking changes
4. **Fix** — apply appropriate patterns from the bundled reference guide, then grep for non-compiler-caught usages
5. **Assess** — classify changes as compile-only vs feature-requiring and trace Jira tickets for follow-up work

**Preprocessed context injection:**
- Skill automatically runs `git branch`, `git diff`, and `gh run list` on load to detect current branch, SDK version delta, and CI failures
- Proactive behavior skips to Phase 2 when CI failures are detected

**Allowed-tools (least-privilege):**
- Read-only git: `git branch --show-current`, `git diff *`
- Read-only GitHub CLI: `gh run list *`, `gh run view *`, `gh pr view *`, `gh pr diff *`
- No write/destructive operations permitted

**Files added:**
- `.claude/skills/resolving-sdk-updates/SKILL.md` — main skill instructions with preprocessed context and allowed-tools frontmatter
- `.claude/skills/resolving-sdk-updates/references/common-fix-patterns.md` — error category to fix strategy quick-reference

### DIFF
--- a/.claude/skills/resolving-sdk-updates/SKILL.md
+++ b/.claude/skills/resolving-sdk-updates/SKILL.md
@@ -9,6 +9,7 @@ description: >
   update", "resolve SDK breaking changes", "check the SDK PR", "SDK version bump",
   "sdk-android", "sdk-internal". Do NOT use for general dependency updates unrelated to the
   Bitwarden SDK.
+allowed-tools: Bash(git branch --show-current), Bash(git diff *), Bash(gh run list *), Bash(gh run view *), Bash(gh pr view *), Bash(gh pr diff *)
 ---
 
 # Resolving SDK Updates
@@ -28,7 +29,7 @@ CRITICAL: Before applying any fix, always read the `sdk-internal` PR diff to und
 
 - **Current branch**: !`git branch --show-current`
 - **SDK version diff vs main**: !`git diff main -- gradle/libs.versions.toml | grep sdk-android || echo "No SDK version change detected"`
-- **Latest CI failure (current branch)**: !`gh run list --branch $(git branch --show-current) --status failure --limit 1 --json databaseId,event,conclusion -q '.[0]' 2>/dev/null || echo "No failures found"`
+- **Latest CI failure (sdlc/sdk-update)**: !`gh run list --branch sdlc/sdk-update --status failure --limit 1 --json databaseId,event,conclusion -q '.[0]' 2>/dev/null`
 
 ## Proactive Behavior
 
--- a/.claude/skills/resolving-sdk-updates/SKILL.md
+++ b/.claude/skills/resolving-sdk-updates/SKILL.md
@@ -24,19 +24,25 @@ Sequential five-phase workflow for diagnosing and resolving build failures cause
 
 CRITICAL: Before applying any fix, always read the `sdk-internal` PR diff to understand the author's intent. A compile fix alone may be insufficient if the SDK change introduces new behavior that requires a dedicated code path.
 
+## Current State (preprocessed)
+
+- **Current branch**: !`git branch --show-current`
+- **SDK version diff vs main**: !`git diff main -- gradle/libs.versions.toml | grep sdk-android || echo "No SDK version change detected"`
+- **Latest CI failure (current branch)**: !`gh run list --branch $(git branch --show-current) --status failure --limit 1 --json databaseId,event,conclusion -q '.[0]' 2>/dev/null || echo "No failures found"`
+
 ## Proactive Behavior
 
-- If reviewing an SDK update PR and CI checks are failing, skip Phase 1 and jump directly to Phase 2.
+- If the preprocessed state above shows CI failures, skip Phase 1 and jump directly to Phase 2.
 - If no CI failures exist, focus on Phase 1 (changelog review) and Phase 5 (impact assessment).
-- If invoked from a worktree checkout of an SDK update PR, check CI status first and begin diagnosis if failing.
+- If the SDK version diff above shows no change, confirm the branch and version catalog before proceeding.
 
 ---
 
 ## Phase 1: Identify the Update
 
-Determine what changed in the SDK version bump.
+Determine what changed in the SDK version bump. The preprocessed SDK version diff above may already provide this; verify and continue.
 
-1. **Extract version diff** from `gradle/libs.versions.toml`:
+1. **Extract version diff** (if not already shown above) from `gradle/libs.versions.toml`:
    ```bash
    git diff main -- gradle/libs.versions.toml | grep sdk-android
    ```
--- a/.claude/skills/resolving-sdk-updates/SKILL.md
+++ b/.claude/skills/resolving-sdk-updates/SKILL.md
@@ -18,7 +18,7 @@ Sequential five-phase workflow for diagnosing and resolving build failures cause
 
 ## Important
 
-- **SDK repo**: `bitwarden/sdk-internal` (private)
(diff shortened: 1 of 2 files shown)


SaintPatrck:
@claude `version` is not an officially documented frontmatter property for Skills. existing ones are using it incorrectly. There _are_ newlines. Your diff may be hiding the whitespace.

.claude/skills/resolving-sdk-updates/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+allowed-tools: Bash(git branch --show-current), Bash(git diff *), Bash(gh run list *), Bash(gh run view *), Bash(gh pr view *), Bash(gh pr diff *)
+---
+
+# Resolving SDK Updates
+
+Sequential five-phase workflow for diagnosing and resolving build failures caused by Bitwarden SDK (`com.bitwarden:sdk-android`) version updates.
+
+## Important
+
+- **SDK repo**: `bitwarden/sdk-internal` (private)
```
vvolkgang: ⛏️ public repo. 
```suggestion
- **SDK repo**: `bitwarden/sdk-internal`
```

.claude/skills/resolving-sdk-updates/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+- **Artifact**: `com.bitwarden:sdk-android` published to GitHub Packages
+- **Version catalog**: `gradle/libs.versions.toml` (key: `sdk-android`)
+- **Update branch convention**: `sdlc/sdk-update`
+
+CRITICAL: Before applying any fix, always read the `sdk-internal` PR diff to understand the author's intent. A compile fix alone may be insufficient if the SDK change introduces new behavior that requires a dedicated code path.
+
+## Current State (preprocessed)
```
vvolkgang, line 28: 💭 We can cut most of these setup steps if we assume this runs when `sdlc-sdk-update.yml` fails, SDK Update PR description will have the changelog as a list of sdk-internal PRs and we have the build run failure logs / know which run failed. We can then focus tokens and runtime on (1) preliminary diagnose based on common errors and if that isn't enough, (2) analysing the PRs in the changelog and how they may have influenced the build failure.

SaintPatrck: I don't want to assume that this is run only when the update fails. Since we mock the SDK behavior in tests there's a real risk of behavior changes in the SDK going unnoticed even when our tests pass. The sdk update workflow only catches breaking _code_ changes so, imo, it's important that the SDK pull requests are always reviewed. Just like a Human contributor/maintainer is expected to do.

SaintPatrck: Also, side note: the `!<command>` syntax I'm using below doesn't have a significant impact on token usage. The preceding `!` indicates the `<command>` is executed as a raw bash command and the result is injected into the content. My understanding is it's **not** executed by the LLM so there's not overhead of invoking tools and parsing responses. More about `!` in the [Advanced Patterns](https://code.claude.com/docs/en/slash-commands#advanced-patterns) docs.

vvolkgang: 💭 breaking it down:

1. Reviewing SDK PR code to identify behaviour breaking changes - followup from our recent discussions we would have a Skill running on each SDK PR (in the SDK repo), as an additional measure to raise visibility of breaking changes in advance, so PR authors can adapt or update the PR description / label the PR with `breaking-changes`.
2. Integration testing - reckon this would be better solved by integrating our QA automation testing suit in the process. 
3. Android / iOS side, reviewing SDK Update PRs - we're already planning on highlighting the PRs that have a `breaking-change` label but, we could also pull up their description and then have a Skill focused on analysing that (and build failure logs if they happen) and proposing additional implementation details? 

Regardless of the above we can move forward with merging this now and circle back when the other pieces are set in place.

SaintPatrck:
Converting to draft. After more brainstorming, we may have a better approach that may or may not use this skill.

.claude/skills/resolving-sdk-updates/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+description: >
+  Diagnose and resolve build failures from Bitwarden SDK updates (com.bitwarden:sdk-android).
+  Investigates sdk-internal PRs, fixes exhaustive when expressions, and assesses behavioral
+  impact. Use when reviewing SDK update PRs, fixing SDK build errors, encountering "when
+  expression must be exhaustive" after SDK bump, updating sdk-android in libs.versions.toml,
+  or when any PR from the sdlc/sdk-update branch has failing CI. Triggered by "fix the SDK
+  update", "resolve SDK breaking changes", "check the SDK PR", "SDK version bump",
+  "sdk-android", "sdk-internal". Do NOT use for general dependency updates unrelated to the
+  Bitwarden SDK.
+allowed-tools: Bash(git branch --show-current), Bash(git diff *), Bash(gh run list *), Bash(gh run view *), Bash(gh pr view *), Bash(gh pr diff *)
```
theMickster, line 12: ⚠️ The `allowed-tools` block is missing `Bash(grep *)` — grep commands will be blocked at runtime. I found that Phase 4 Step 2 in `common-fix-patterns.md` both contain `grep -r`/`grep -rn` commands that require `Bash(grep ...)`

I think you even call out Claude pretty clearly below with a "Do not skip the codebase-wide grep in Phase 4 step 2. The compiler only catches exhaustive `when` expressions, not `when` statements."

.claude/skills/resolving-sdk-updates/references/common-fix-patterns.md
```diff
@@ -0,0 +1,195 @@
...
+   ```bash
+   gh pr diff <N> --repo bitwarden/sdk-internal | grep "<OldTypeName>"
+   ```
+2. Update all imports and usages:
+   ```bash
+   grep -rn "import.*<OldTypeName>" --include="*.kt" app/src/main/kotlin/
+   grep -rn "<OldTypeName>" --include="*.kt" app/src/main/kotlin/
+   ```
+3. Use find-and-replace across affected files
+
```
theMickster, line 79: 🤔 The other fix strategies have a before and after helper code block. Would something simple like the following example help or hurt Claude?

```kotlin
// Before (unresolved broken reference)
import com.bitwarden.sdk.OldTypeName

// After (renamed type)
import com.bitwarden.sdk.NewTypeName
```

theMickster (commented):
One frontmatter tool to fix and one idea. Otherwise, I think that this is at least worth a try. Given that this is all Claude just working in a new workflow; I don't think there's much hard to at least trying it out to see if it help us. I think it could be the least intrusive attempt with the highest potential upsite.

SaintPatrck:
Superseded by #7251

PR #6662, opened 2026-03-13
[PM-33511] feat: Add creationDate to UserState.Account

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33511

## 📔 Objective

Adds `creationDate: Instant?` to `UserState.Account`, mapped from `profile.creationDate` in `UserStateJsonExtensions`. This exposes the account creation date for downstream age-based eligibility checks (e.g., premium upgrade banner requires account age >= 7 days).

- Adds property to `UserState.Account` data class
- Maps from `ProfileJson.creationDate` in the extension function
- Updates all test fixtures with `creationDate` parameter
- Adds dedicated unit test for `creationDate` mapping


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/model/UserState.kt
```diff
@@ -77,6 +79,7 @@ data class UserState(
         val onboardingStatus: OnboardingStatus,
         val firstTimeState: FirstTimeState,
         val isExportable: Boolean,
+        val creationDate: Instant?,
```
david-livefront, line 82: Why is it nullable?

SaintPatrck: Because it's nullable in AccountJson

david-livefront: Ahh, I was looking at the sync response. It's nonnull there

PR #6664, opened 2026-03-13
[PM-33356] feat: Sync when push notification policy changed is received

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33356
## 📔 Objective

This pull request adds support for handling a new type of push notification indicating that an organization policy has changed. The implementation ensures that when a "policy changed" notification is received, a full sync is triggered for the active user. Corresponding tests have been added to verify this behavior.

**Push notification handling:**

* Added a new notification type, `POLICY_CHANGED`, to the `NotificationType` enum to represent policy change events.
* Updated `PushManagerImpl` to emit the active user ID to the `fullSyncSharedFlow` when a `POLICY_CHANGED` notification is received, triggering a full sync.

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/data/platform/manager/PushManagerTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/data/platform/manager/PushManagerTest.kt
@@ -281,7 +281,7 @@ class PushManagerTest {
             }
 
             @Test
-            fun `onMessageReceived with policy changed to fullSyncFlow`() = runTest {
+            fun `onMessageReceived with policy changed emit to fullSyncFlow`() = runTest {
                 val activeUserId = authDiskSource.userState?.activeUserId
                 pushManager.fullSyncFlow.test {
                     pushManager.onMessageReceived(POLICY_CHANGED_NOTIFICATION_MAP)


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/PushManagerImpl.kt
```diff
@@ -201,6 +195,12 @@ class PushManagerImpl @Inject constructor(
                     }
             }
 
+            NotificationType.POLICY_CHANGED -> {
+                activeUserId?.let {
+                    mutableFullSyncSharedFlow.tryEmit(it)
```
david-livefront, line 206: How do we know the user ID is correct, can we pull that information from the payload?

andrebispo5: Unfortunately no, but if it is not the active user, the user will have to unlock the vault and sync eventually since the revision changes.

david-livefront: Anyway they could add the user ID?

Literally all other notifications have the userId present in the payload

andrebispo5: Indeed I agree with you and I already asked for the change, but Imo this should be in first because the server update can take a while.

app/src/test/kotlin/com/x8bit/bitwarden/data/platform/manager/PushManagerTest.kt
```diff
@@ -228,10 +228,22 @@ class PushManagerTest {
...
                     assertEquals(
-                        "078966a2-93c2-4618-ae2a-0a2394c88d37",
+                        activeUserId,
+                        awaitItem(),
+                    )
+                }
+            }
+
+            @Test
+            fun `onMessageReceived with policy changed to fullSyncFlow`() = runTest {
```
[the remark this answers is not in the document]
david-livefront: @andrebispo5 Could we make this change?

PR #6680, opened 2026-03-16
[PM-33510] feat: Add Play Billing Library dependency and PlayBillingManager

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33510

## 📔 Objective

Introduces a flavor-specific `PlayBillingManager` wrapping the Google Play Billing Library to determine whether in-app billing is available on the current device.

- Adds `com.android.billingclient:billing` v8.3.0 as a `standardImplementation` dependency
- Defines `PlayBillingManager` interface with `isInAppBillingSupportedFlow: StateFlow<Boolean>`
- Standard flavor implementation uses `BillingClient.getBillingConfigAsync()` with a connect-per-call lifecycle to probe billing availability
- F-Droid flavor always emits `true` (F-Droid users are eligible for the upgrade flow)
- Wires `PlayBillingManager` into `BillingRepositoryImpl`, delegating `isInAppBillingSupportedFlow` through the repository layer
- Updates `BillingModule` to provide `PlayBillingManager` as a singleton

### DIFF
--- a/app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
+++ b/app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
@@ -1,0 +1,133 @@
+package com.x8bit.bitwarden.data.billing.manager
+
+import android.content.Context
+import com.android.billingclient.api.BillingClient
+import com.android.billingclient.api.BillingClientStateListener
+import com.android.billingclient.api.BillingConfig
+import com.android.billingclient.api.BillingResult
+import com.android.billingclient.api.GetBillingConfigParams
+import com.android.billingclient.api.PendingPurchasesParams
+import com.bitwarden.annotation.OmitFromCoverage
+import com.bitwarden.core.data.manager.dispatcher.DispatcherManager
+import com.bitwarden.core.data.util.asFailure
+import com.bitwarden.core.data.util.asSuccess
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+
+private const val SUPPORTED_BILLING_COUNTRY = "US"
+
+/**
+ * Standard implementation of [PlayBillingManager] using the Google Play Billing Library.
+ *
+ * Uses a connect-per-call lifecycle: a new [BillingClient] is created, connected, queried,
+ * and disconnected for each call.
+ */
+@OmitFromCoverage
+class PlayBillingManagerImpl(
+    private val context: Context,
+    dispatcherManager: DispatcherManager,
+) : PlayBillingManager {
+
+    private val unconfinedScope =
+        CoroutineScope(dispatcherManager.unconfined)
+
+    private val mutableIsInAppBillingSupportedFlow = MutableStateFlow(false)
+
+    override val isInAppBillingSupportedFlow: StateFlow<Boolean> =
+        mutableIsInAppBillingSupportedFlow.asStateFlow()
+
+    init {
+        unconfinedScope.launch {
+            mutableIsInAppBillingSupportedFlow.value =
+                queryBillingCountry().getOrNull() == SUPPORTED_BILLING_COUNTRY
+        }
+    }
+
+    private suspend fun queryBillingCountry(): Result<String> {
+        val billingClient = BillingClient
+            .newBuilder(context)
+            .setListener { _, _ ->
+                // No-op: we don't handle purchases.
+            }
+            .enablePendingPurchases(
+                PendingPurchasesParams
+                    .newBuilder()
+                    .enableOneTimeProducts()
+                    .build(),
+            )
+            .build()
+
+        return billingClient.useConnection {
+            if (responseCode != BillingClient.BillingResponseCode.OK) {
+                return@useConnection BillingException("Connection failed: $debugMessage")
+                    .asFailure()
+            }
+            val (configResult, billingConfig) =
+                billingClient.getBillingConfig()
+            if (configResult.responseCode != BillingClient.BillingResponseCode.OK ||
+                billingConfig == null
+            ) {
+                BillingException("Config query failed: ${configResult.debugMessage}").asFailure()
+            } else {
+                billingConfig.countryCode.asSuccess()
+            }
+        }
+    }
+}
+
+/**
+ * Connects to the [BillingClient], executes [block] with the [BillingResult], and guarantees
+ * [BillingClient.endConnection] is called when finished. Catches [IllegalStateException] thrown
+ * by [BillingClient.startConnection] if the client is already connected or mid-connection.
+ */
+private suspend fun <T> BillingClient.useConnection(
+    block: suspend BillingResult.() -> Result<T>,
+): Result<T> = try {
+    block(
+        suspendCancellableCoroutine { continuation ->
+            startConnection(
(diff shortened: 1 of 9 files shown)


app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+            .setListener { _, _ ->
+                // No-op: we don't handle purchases.
+            }
+            .enablePendingPurchases(
+                PendingPurchasesParams
+                    .newBuilder()
+                    .enableOneTimeProducts()
+                    .build(),
+            )
+            .build()
```
david-livefront, line 63: Would it make sense to inject this as a singleton or do we want a new `BillingClient` for each request?

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+                billingConfig == null
+            ) {
+                BillingException("Config query failed: ${configResult.debugMessage}").asFailure()
+            } else {
+                billingConfig.countryCode.asSuccess()
+            }
+        } catch (e: IllegalStateException) {
+            e.asFailure()
+        } finally {
+            billingClient.endConnection()
```
david-livefront: The `startConnection` and `endConnection` calls make me feel like we don't want to inject it 😄

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+
+        return try {
+            val connectionResult = billingClient.connect()
+            if (connectionResult.responseCode !=
+                BillingClient.BillingResponseCode.OK
+            ) {
+                return BillingException(
+                    "Connection failed: ${connectionResult.debugMessage}",
+                )
+                    .asFailure()
```
david-livefront, line 68: Is there a specific part of this code that throws?

It would be nice if we could pull some of this stuff out of the `try-catch` block.

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+                override fun onBillingSetupFinished(
+                    billingResult: BillingResult,
+                ) {
+                    if (continuation.isActive) {
+                        continuation.resume(billingResult)
+                    }
+                }
+
+                override fun onBillingServiceDisconnected() {
+                    // No-op: connect-per-call lifecycle, no reconnection.
```
david-livefront: Should we call `continuation.cancel()` here?

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+                }
+            },
+        )
+    }
+
+/**
+ * Wraps [BillingClient.getBillingConfigAsync] in a suspend function using
+ * [suspendCancellableCoroutine].
+ */
+@Suppress("MaxLineLength")
```
david-livefront: I don't think we need this suppression, right?

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
SaintPatrck: Keeping it around is not needed. Maybe if we add more billing related functionality in the future, but right now we only need this one-off check that can be done on init.

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
SaintPatrck: A few calls do. I can isolate them.

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
SaintPatrck: Yes. Good catch

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
SaintPatrck: Agreed. That's another reason I chose to go with creating a new instance every time.

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+            .enablePendingPurchases(
+                PendingPurchasesParams
+                    .newBuilder()
+                    .enableOneTimeProducts()
+                    .build(),
+            )
+            .build()
+
+        return try {
+            val connectionResult = billingClient.connect()
```
david-livefront: What do you think of a reusable function like this?

```kotlin
    private suspend fun <T> BillingClient.useConnection(
        block: suspend BillingResult.() -> Result<T>,
    ): Result<T> = try {
        block(
            suspendCancellableCoroutine { continuation ->
                startConnection(
                    object : BillingClientStateListener {
                        override fun onBillingSetupFinished(
                            billingResult: BillingResult,
                        ) {
                            if (continuation.isActive) {
                                continuation.resume(billingResult)
                            }
                        }

                        override fun onBillingServiceDisconnected() {
                            // No-op: connect-per-call lifecycle, no reconnection.
                        }
                    },
                )
            },
        )
    } catch (e: IllegalStateException) {
        e.asFailure()
    } finally {
        endConnection()
    }
}
```

Then we can do this:
```kotlin
        return billingClient.useConnection {
            if (responseCode != BillingClient.BillingResponseCode.OK) {
                return@useConnection BillingException("Connection failed: $debugMessage")
                    .asFailure()
            }
            val (configResult, billingConfig) = billingClient.getBillingConfig()
            if (configResult.responseCode != BillingClient.BillingResponseCode.OK ||
                billingConfig == null
            ) {
                BillingException("Config query failed: ${configResult.debugMessage}").asFailure()
            } else {
                billingConfig.countryCode.asSuccess()
            }
        }
    }
```

You might even be able to move the `responseCode` check into the helper method

app/src/standard/kotlin/com/x8bit/bitwarden/data/billing/manager/PlayBillingManagerImpl.kt
```diff
@@ -0,0 +1,133 @@
...
+                    override fun onBillingSetupFinished(
+                        billingResult: BillingResult,
+                    ) {
+                        if (continuation.isActive) {
+                            continuation.resume(billingResult)
+                        }
+                    }
+
+                    override fun onBillingServiceDisconnected() {
+                        // No-op: connect-per-call lifecycle, no reconnection.
```
david-livefront, line 104: So we don't want to cancel the continuation here?

SaintPatrck: No, we don't need it now that we have `useConnection`. By the time this would fire, we would have already resumed/completed, or have already called `endConnection()` and it firing is expected.

PR #6687, opened 2026-03-18
[PM-23560] bug: Added guard to ensure duplicate scan events are not fired

aj-rosado:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-23560

## 📔 Objective
When scanning QR code totp items, multiples are being added, this is caused by `QrCodeAnalyzer` firing `onQrCodeScanned` every time it gets a result.
In order to avoid adding multiple items, a check on ViewModel was added in order to only allow the first result to be returned.
On some cases, if the image is analyzed right after opening the screen, the `EventsEffect` have not been initiated and the user is not being returned to be previous screen. To workaround that scenario, when the code has been scanned, we do not emit the code again but fire a NavigateBack event.

Same scenario was happening on PasswordManager, despite not creating multiple items, multiple events were fired.

### DIFF
--- a/authenticator/src/test/kotlin/com/bitwarden/authenticator/ui/authenticator/feature/qrcodescan/QrCodeScanViewModelTest.kt
+++ b/authenticator/src/test/kotlin/com/bitwarden/authenticator/ui/authenticator/feature/qrcodescan/QrCodeScanViewModelTest.kt
@@ -173,7 +173,6 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
         }
 
     @Test
-    @Suppress("MaxLineLength")
     fun `on SaveToBitwardenErrorDismiss receive should clear dialog state`() {
         val viewModel = createViewModel()
         every {
@@ -259,7 +258,6 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
         }
 
     @Test
-    @Suppress("MaxLineLength")
     fun `on QrCodeScanReceive when code is invalid should emit result and navigate back`() =
         runTest {
             val viewModel = createViewModel()
@@ -276,7 +274,6 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
         }
 
     @Test
-    @Suppress("MaxLineLength")
     fun `on QrCodeScanReceive second scan after valid TOTP scan is ignored`() = runTest {
         val viewModel = createViewModel()
         every { settingsRepository.defaultSaveOption } returns DefaultSaveOption.LOCAL
@@ -300,7 +297,6 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
     }
 
     @Test
-    @Suppress("MaxLineLength")
     fun `on QrCodeScanReceive second scan after error scan is ignored`() = runTest {
         val viewModel = createViewModel()
         val invalidQrCode = "otpauth://totp/secret=SECRET"
@@ -325,7 +321,7 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
     }
 
     private fun createViewModel(
-        initialState: QrCodeScanState? = DEFAULT_STATE,
+        initialState: QrCodeScanState? = null,
     ) = QrCodeScanViewModel(
         savedStateHandle = SavedStateHandle().apply { set("state", initialState) },
         authenticatorBridgeManager = authenticatorBridgeManager,
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModelTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModelTest.kt
@@ -127,7 +127,7 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
         }
 
     private fun createViewModel(
-        initialState: QrCodeScanState? = DEFAULT_STATE,
+        initialState: QrCodeScanState? = null,
     ): QrCodeScanViewModel =
         QrCodeScanViewModel(
             savedStateHandle = SavedStateHandle().apply { set("state", initialState) },


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModel.kt
```diff
@@ -35,6 +41,14 @@ class QrCodeScanViewModel @Inject constructor(
 
     // For more information: https://bitwarden.com/help/authenticator-keys/#support-for-more-parameters
     private fun handleQrCodeScanReceive(action: QrCodeScanAction.QrCodeScanReceive) {
+        if (hasHandledScan) {
+            // There is a possible race condition where the QrCodeScan is received before the
+            // EventsEffect has been initiated, this ensures the user navigates back when
+            // the scan has been handled.
```
david-livefront: Can we fix that by hooking up the scan callback before the `EventsEffect`?

We could even do it via a disposable effect, right?

david-livefront: Something like this after the `EventsEffect`?

```kotlin
    DisposableEffect(Unit) {
        qrCodeAnalyzer.onQrCodeScanned = {
            viewModel.trySendAction(QrCodeScanAction.QrCodeScanReceive(it))
        }
        onDispose { qrCodeAnalyzer.onQrCodeScanned = null }
    }
```

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanScreen.kt
```diff
@@ -69,6 +69,7 @@ fun QrCodeScanScreen(
             }
         }
     }
+
```
david-livefront: Can we remove this

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModel.kt
```diff
@@ -55,8 +66,9 @@ sealed class QrCodeScanEvent {
 
     /**
      * Navigate back.
+     * Added BackgroundEvent as QrCodeScan might be fired before events are consumed
      */
-    data object NavigateBack : QrCodeScanEvent()
+    data object NavigateBack : QrCodeScanEvent(), BackgroundEvent
```
david-livefront: We should not have navigation events as BackgroundEvents, that can cause crashes.

We need to stop the events from occurring before the screen is resumed. Did the `DisposableEffect` not work as we discussed?

We have other tools for this as well, like the `LifecycleEventEffect`, which can guarantee that the listener is applied after we resume.

aj-rosado: `DisposableEffect` did not work. `LifecycleEventEffect` , when using the `ON_RESUME` works. 

❓ Would the `DeferredBackgroundEvent` also cause issues?

david-livefront: `DeferredBackgroundEvent` would also solve this issue

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModel.kt
```diff
@@ -16,6 +17,12 @@ class QrCodeScanViewModel @Inject constructor(
 ) : BaseViewModel<Unit, QrCodeScanEvent, QrCodeScanAction>(
     initialState = Unit,
 ) {
+    /**
+     * Guards against duplicate QR code scans being processed. Once a scan has been handled,
+     * subsequent [QrCodeScanAction.QrCodeScanReceive] actions are ignored.
+     */
+    private var hasHandledScan: Boolean = false
```
david-livefront: Can we move this into the state

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/qrcodescan/QrCodeScanViewModel.kt
```diff
@@ -55,8 +66,9 @@ sealed class QrCodeScanEvent {
 
     /**
      * Navigate back.
+     * Added BackgroundEvent as QrCodeScan might be fired before events are consumed
```
david-livefront: Can we update the comment to reference the `DeferredBackgroundEvent`

authenticator/src/test/kotlin/com/bitwarden/authenticator/ui/authenticator/feature/qrcodescan/QrCodeScanViewModelTest.kt
```diff
@@ -259,14 +275,67 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
...
+            expectNoEvents()
+        }
+        verify(exactly = 1) {
+            authenticatorRepository.emitTotpCodeResult(TotpCodeResult.CodeScanningError)
+        }
+        verify(exactly = 0) { authenticatorRepository.emitTotpCodeResult(VALID_TOTP_RESULT) }
+    }
+
+    private fun createViewModel(
+        initialState: QrCodeScanState? = DEFAULT_STATE,
```
david-livefront: Can we make the default state null

authenticator/src/test/kotlin/com/bitwarden/authenticator/ui/authenticator/feature/qrcodescan/QrCodeScanViewModelTest.kt
```diff
@@ -259,14 +275,67 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
...
+        }
+        verify(exactly = 1) { authenticatorRepository.emitTotpCodeResult(VALID_TOTP_RESULT) }
+        assertEquals(
+            DEFAULT_STATE.copy(hasHandledScan = true),
+            viewModel.stateFlow.value,
+        )
+    }
+
+    @Test
+    @Suppress("MaxLineLength")
```
david-livefront: This suppression is not needed

authenticator/src/test/kotlin/com/bitwarden/authenticator/ui/authenticator/feature/qrcodescan/QrCodeScanViewModelTest.kt
```diff
@@ -170,7 +174,7 @@ class QrCodeScanViewModelTest : BaseViewModelTest() {
 
     @Test
     @Suppress("MaxLineLength")
```
david-livefront: I know you didn't write this but can we drop this suppression while you are here?

PR #6689, opened 2026-03-18
misc: Rename the VaultDiskSource Flows

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR updates the function names that return a flow in the `VaultDiskSource` to include the word `Flow`. This will be useful as we add new functions that retrieve similar data as a suspending call instead of a Flow.


david-livefront:
Thanks @SaintPatrck

PR #6692, opened 2026-03-18
[PM-33513] feat: Add checkout callback deep link handling

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33513

## 📔 Objective

Add deep link handling for Stripe checkout callbacks so the app can detect when a user returns from a Stripe payment flow.

- Add `isPremiumCheckoutCallback` Intent extension property in `ShortcutUtils.kt` for the `bitwarden://premium-upgrade-callback` deep link
- Add `PremiumCheckoutResult` data object to the `SpecialCircumstance` sealed class
- Handle checkout callback URI in `MainViewModel.handleIntent()`, setting the special circumstance
- Add exhaustiveness branches in `RootNavViewModel.kt` for the new sealed class member
- Unit tests for intent parsing (`ShortcutUtilsTest`) and ViewModel handling (`MainViewModelTest`)

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/util/ShortcutUtilsTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/util/ShortcutUtilsTest.kt
@@ -56,4 +56,28 @@ class ShortcutUtilsTest {
         }
         assertFalse(mockIntent.isPasswordGeneratorShortcut)
     }
+
+    @Test
+    fun `isPremiumCheckoutCallback should return true when dataString is checkout callback`() {
+        val mockIntent = mockk<Intent> {
+            every { dataString } returns "bitwarden://premium-upgrade-callback"
+        }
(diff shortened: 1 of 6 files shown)


app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/util/ShortcutUtilsTest.kt
```diff
@@ -56,4 +56,29 @@ class ShortcutUtilsTest {
...
+
+    @Test
+    fun `isPremiumCheckoutCallback should return true when dataString is checkout callback`() {
+        val mockIntent = mockk<Intent> {
+            every { dataString } returns "bitwarden://premium-upgrade-callback"
+        }
+        assertTrue(mockIntent.isPremiumCheckoutCallback)
+    }
+
+    @Suppress("MaxLineLength")
```
david-livefront: I feel like this one is close. Do we need this suppression?

PR #6693, opened 2026-03-19
PM-33893: bug: Crash caused by empty credential password

david-livefront:
## 🎟️ Tracking

[PM-33893](https://bitwarden.atlassian.net/browse/PM-33893)

## 📔 Objective

This PR addresses a crash that occurs when the password is null or empty.

This is because the `PasswordCredential` requires the password to be populated.

```kotlin
    init {
        require(password.isNotEmpty()) { "password should not be empty" }
    }
```

[PM-33893]: https://bitwarden.atlassian.net/browse/PM-33893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/CredentialProviderCompletionManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/CredentialProviderCompletionManagerImpl.kt
@@ -131,8 +131,8 @@ class CredentialProviderCompletionManagerImpl(
                             intent = intent,
                             response = GetCredentialResponse(
                                 credential = PasswordCredential(
-                                    id = result.credential.username ?: "",
-                                    password = result.credential.password ?: "",
+                                    id = result.username.orEmpty(),
+                                    password = result.password,
                                 ),
                             ),
                         )
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/model/GetPasswordCredentialResult.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/model/GetPasswordCredentialResult.kt
@@ -1,7 +1,6 @@
 package com.x8bit.bitwarden.ui.credentials.manager.model
 
 import com.bitwarden.ui.util.Text
-import com.bitwarden.vault.LoginView
 
 /**
  * Represents the result of Password authentication.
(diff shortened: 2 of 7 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/CredentialProviderCompletionManagerImpl.kt
```diff
@@ -131,8 +131,8 @@ class CredentialProviderCompletionManagerImpl(
                             intent = intent,
                             response = GetCredentialResponse(
                                 credential = PasswordCredential(
-                                    id = result.credential.username ?: "",
-                                    password = result.credential.password ?: "",
+                                    id = result.username.orEmpty(),
```
david-livefront, line 134: There is no requirement for the `id` to be populated like the `password`. But we could do more to populate it.

Leaving this up to @SaintPatrck

SaintPatrck: Hm. `id` is expected to be the actual username/email portion of a basic auth credential. Unlike a regular autofill flow, CredentialManager responses don't populate any UI. Similar to passkey flow, it's all handled silently in the background. That makes me weary of being too presumptuous because the user doesn't get an opportunity to see or correct what is being sent if we populate it with something else. 

TLDR; I think a 1:1 mapping to the Login's `username` is safest.

david-livefront: So if we were to pass in the empty username, it should just fail saying the passkey was incorrect?

david-livefront:
Thanks @SaintPatrck

app/src/main/kotlin/com/x8bit/bitwarden/ui/credentials/manager/CredentialProviderCompletionManagerImpl.kt
SaintPatrck: Correct. If username is required for authentication in their system, I would expect the **relying party** to notify the user that the _username/password_ was incorrect.

david-livefront: Makes sense 👍

PR #6699, opened 2026-03-19
PM-33913: bug: Remove org event to avoid duplicate entry

david-livefront:
## 🎟️ Tracking

[PM-33913](https://bitwarden.atlassian.net/browse/PM-33913)

## 📔 Objective

This PR removes an org event entry from the app to avoid duplicate entries. This is because the cloud logs these events for us.

[PM-33913]: https://bitwarden.atlassian.net/browse/PM-33913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


david-livefront:
Thanks @SaintPatrck

PR #6702, opened 2026-03-20
chore: Remove unused how_to_manage_my_vault string

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR removes the unused `how_to_manage_my_vault` string from the primary resources file.


david-livefront:
Thanks @SaintPatrck

PR #6709, opened 2026-03-23
PM-32721: bug: Sort password history before persisting

david-livefront:
## 🎟️ Tracking

[PM-32721](https://bitwarden.atlassian.net/browse/PM-32721)

## 📔 Objective

This PR updates the logic for how we persist the password history. We now ensure that it is saved in sorted order so that other client that do not sort the list before displaying the data still display everything in the correct order.

[PM-32721]: https://bitwarden.atlassian.net/browse/PM-32721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


david-livefront:
Thanks @andrebispo5 & @SaintPatrck

PR #6714, opened 2026-03-24
BWA-238: bug: Send additional cipher data for Authenticator Sync

david-livefront:
## 🎟️ Tracking

[BWA-238](https://bitwarden.atlassian.net/browse/BWA-238)

## 📔 Objective

This PR adds additional data to the Authenticator Sync payload in order to display information better and handle Steam TOTP codes.

There is additional logic to support backwards compatibility between the Password Manager and Authenticator in case a user updates one app but not the other. It should be pretty seamless overall.

## 📸 Screenshots

| Before | After |
| --- | --- |
|  |  |

[BWA-238]: https://bitwarden.atlassian.net/browse/BWA-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/authenticatorbridge/src/main/kotlin/com/bitwarden/authenticatorbridge/model/SharedAccountData.kt
+++ b/authenticatorbridge/src/main/kotlin/com/bitwarden/authenticatorbridge/model/SharedAccountData.kt
@@ -16,14 +16,33 @@ data class SharedAccountData(
      * @param name name associated with the account.
      * @param email email associated with the account.
      * @param environmentLabel environment associated with the account.
-     * @param totpUris list of totp URIs associated with the account.
-     * @param lastSyncTime the last time the account was synced by the main Bitwarden app.
+     * @param cipherData list of ciphers containing totp URIs associated with the account.
      */
     data class Account(
         val userId: String,
         val name: String?,
         val email: String,
         val environmentLabel: String,
-        val totpUris: List<String>,
+        val cipherData: List<CipherData>,
+    )
+
+    /**
+     * Models a single shared cipher containing a totp.
+     *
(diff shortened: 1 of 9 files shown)


andrebispo5:
🤔 Noticed that the countdown circles are not vertically aligned, is it supposed to be this way? seems odd. 
edit: the purpose of the PR is not to fix this but maybe we can create a tech debt ticket to fix it later.

authenticatorbridge/src/main/kotlin/com/bitwarden/authenticatorbridge/model/SharedAccountData.kt
```diff
@@ -16,14 +16,33 @@ data class SharedAccountData(
      * @param name name associated with the account.
      * @param email email associated with the account.
      * @param environmentLabel environment associated with the account.
-     * @param totpUris list of totp URIs associated with the account.
-     * @param lastSyncTime the last time the account was synced by the main Bitwarden app.
```
andrebispo5, line 20: nice catch👍

andrebispo5 (approved):
✅ Approved ✅ 

:recycle: Consider the creation of the tech debt ticket

david-livefront:
> 🤔 Noticed that the countdown circles are not vertically aligned, is it supposed to be this way? seems odd. edit: the purpose of the PR is not to fix this but maybe we can create a tech debt ticket to fix it later.

I noticed that too. I'll speak to design and see if I can get that fixed next.

david-livefront:
Thanks @andrebispo5

PR #6717, opened 2026-03-25
feat: Add collection management to Android

dasien:
## Summary

- Add full CRUD support for managing collections on Android via Settings > Vault > Collections
- Collections are organization-scoped vault items available on paid plans
- Network layer: CollectionsApi, CollectionService with create/update/delete/get endpoints
- Data layer: CollectionManager with encrypt → API → disk → decrypt flow
- Permission model: expanded with collection-specific permission fields and role-based checks matching web client logic
- UI: CollectionsScreen (list with org subtitles, permission-gated FAB), CollectionAddEditScreen (name field, save, delete)
- Creating user automatically gets manage access on new collections
- Update requests preserve existing group/user access permissions

## Test plan

- [ ] Log in with account that has a paid organization
- [ ] Navigate to Settings > Vault > Collections
- [ ] Verify FAB is visible for Owner/Admin roles
- [ ] Create a new collection and verify snackbar confirmation
- [ ] Move an item to the new collection (verify no permission error)
- [ ] Edit a collection name and verify save works
- [ ] Delete a collection and verify confirmation dialog and removal
- [ ] Verify FAB is hidden for users without create permission
- [ ] Verify `/` character is rejected in collection name

## Notes

- SDK `encryptCollection` requires a local SDK build from `sdk-internal` with collection encryption support (not yet in published SDK)
- Commit `27eab557` adapts to local SDK API changes for development; will need adjustment when official SDK is bumped
- Requirements doc at `docs/COLLECTIONS_REQUIREMENTS.md`
- Implementation plan at `.claude/outputs/plans/COLLECTION-MANAGEMENT-PLAN.md`


app/src/main/kotlin/com/x8bit/bitwarden/data/vault/manager/CollectionManagerImpl.kt
```diff
@@ -0,0 +1,163 @@
...
+            listOf(
+                CollectionAccessSelectionJson(
+                    id = it,
+                    readOnly = false,
+                    hidePasswords = false,
+                    manage = true,
+                ),
+            )
+        }
+        return vaultSdkSource
```
quexten, line 47: Thought: Encrypt-only in SDK was really only meant as a transition phase while API, state were not mature in the SDK, and the expectation is to move state and API to the SDK too. In this case, moving at least API one layer down into the SDK would already reduce this PR by a large margin, and remove redundant work from the corresponding iOS PR and would allow dropping TS clients code.

Given that this is a net new feature, can we consider just moving this functionality into the SDK in the first place instead of duplicating logic? Moving this down later will cause more effort overall compared to doing it here.

SaintPatrck: While that is all true, I think introducing SDK managed state and network calls are out of scope for this PR. The Mobile apps have not migrated to using the SDK for state or network calls anywhere, so this would require a non-trivial change in architecture and introduces a second pattern that we have to actively maintain.

fnc314: Sorry for being an outsider weighing in... but is the _end goal_ to use the SDK for state/network calls in a larger sense?  By this I ask, in accordance with @SaintPatrck, to what scale are these two patterns going to co-exist in the "ideal end state"?

I think, perhaps, this perspective can help drive a better choice for this (and the `iOS` PR).  Maybe the "things-moved-to-sdk-tomorrow" can be authored as a separate throw-away Gradle module, further isolating the "temp changes" and the "second pattern" should it need to be repeated elsewhere until such a time the apps migrate to the newer SDK state?

quexten: https://contributing.bitwarden.com/architecture/sdk

> It serves as the single source of truth for core business logic. Written in Rust, the SDK is versatile and provides bindings for a variety of platforms, including mobile clients (Kotlin and Swift) and web clients (JavaScript/TypeScript). The general aspiration is to write as much code as possible once in the SDK and have it consumed by all clients, ensuring feature parity and reducing duplication.

In the ideal end state, the SDK exposes high level operations. The clients are mere presentation layers, and no relevant business logic lives in the application code, aside from platform specific features (biometric unlock), and UI wrapping.

That is to say that in the "ideal end state" there are no network request, and little-to-no state management in the application code.

PR #6720, opened 2026-03-26
[PM-34125] feat: Add card text analysis pipeline

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34125

## 📔 Objective

Add the complete text analysis pipeline for credit card scanning:
- **CardNumberUtils**: Shared Luhn validation and card number sanitization (ui module)
- **CardDataParser**: Interface and regex-based implementation for extracting card details from OCR text (number, expiry, CVV)
- **CardTextAnalyzer**: ML Kit Text Recognition integration for camera frame analysis
- **CardScanOverlay**: Camera overlay composable for scan UI
- **CardScanData**: Data class for parsed card fields
- **CardNumberUtils in app**: Brand detection (`detectCardBrand`) using `VaultCardBrand`

Includes full test coverage for parser logic, Luhn validation, and sanitization.

### DIFF
--- a/ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
+++ b/ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
@@ -7,8 +7,6 @@ private val EXPIRY_REGEX = Regex("""\b(0[1-9]|1[0-2])\s?[/\-]\s?(\d{2}|\d{4})\b"
 private val CVV3_REGEX = Regex("""\b\d{3}\b""")
 private val CVV4_REGEX = Regex("""\b\d{4}\b""")
 
-private val NAME_REGEX = Regex("""^[A-Z][A-Z .'-]+$""")
-
 /**
  * Default [CardDataParser] implementation that uses regex patterns
  * and Luhn validation to extract card details from OCR text.
@@ -49,13 +47,10 @@ class CardDataParserImpl : CardDataParser {
             }
             ?.value
 
-        val cardholderName = extractCardholderName(text)
-
         return listOfNotNull(
             number,
             expirationMonth,
             expirationYear,
-            cardholderName,
             securityCode,
         )
             .takeIf { it.isNotEmpty() }
@@ -64,16 +59,8 @@ class CardDataParserImpl : CardDataParser {
                     number = number,
                     expirationMonth = expirationMonth,
                     expirationYear = expirationYear,
-                    cardholderName = cardholderName,
                     securityCode = securityCode,
                 )
             }
     }
 }
-
-@Suppress("MagicNumber")
-private fun extractCardholderName(text: String): String? =
-    text.lines()
-        .map { it.trim() }
-        .filter { it.length > 3 }
-        .firstOrNull { NAME_REGEX.matches(it) }
--- a/ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardScanData.kt
+++ b/ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardScanData.kt
@@ -6,19 +6,16 @@ package com.bitwarden.ui.platform.feature.cardscanner.util
  * @property number The detected card number.
  * @property expirationMonth The detected expiration month (01-12).
  * @property expirationYear The detected expiration year.
- * @property cardholderName The detected cardholder name.
  * @property securityCode The detected security code.
  */
 data class CardScanData(
     val number: String?,
     val expirationMonth: String?,
     val expirationYear: String?,
-    val cardholderName: String?,
     val securityCode: String?,
 ) {
     override fun toString(): String = "CardScanData(number=****," +
         " expirationMonth=$expirationMonth," +
         " expirationYear=$expirationYear," +
-        " cardholderName=****," +
         " securityCode=****)"
 }
--- a/ui/src/test/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImplTest.kt
+++ b/ui/src/test/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImplTest.kt
@@ -16,7 +16,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -31,7 +30,6 @@ class CardDataParserImplTest {
                 number = null,
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -46,7 +44,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "06",
                 expirationYear = "2028",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -61,7 +58,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = "789",
             ),
             parser.parseCardData(text),
@@ -76,7 +72,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -95,7 +90,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -110,7 +104,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = null,
             ),
             parser.parseCardData(text),
@@ -130,7 +123,6 @@ class CardDataParserImplTest {
                 number = "4111111111111111",
                 expirationMonth = "12",
                 expirationYear = "2025",
-                cardholderName = null,
                 securityCode = "456",
             ),
             parser.parseCardData(text),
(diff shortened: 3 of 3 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/util/CardNumberUtils.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Detects the card brand based on the card number prefix.
+ *
+ * @return The detected [VaultCardBrand], or [VaultCardBrand.OTHER] if no match is found.
+ */
+@Suppress("CyclomaticComplexMethod", "MagicNumber")
+fun String.detectCardBrand(): VaultCardBrand {
+    val digits = sanitizeCardNumber()
+    if (digits.isEmpty()) return VaultCardBrand.OTHER
```
david-livefront: Can this just be the first case in the `when`:

```kotlin
return when {
    digits.isEmpty() -> VaultCardBrand.OTHER
    ...
```

ui/src/main/kotlin/com/bitwarden/ui/platform/components/camera/CardScanOverlay.kt
```diff
@@ -0,0 +1,73 @@
...
+@Suppress("MagicNumber")
+@Composable
+private fun CardScanOverlayCanvas(
+    color: Color,
+    strokeWidth: Dp,
+    modifier: Modifier = Modifier,
+) {
+    Canvas(modifier = modifier) {
+        val strokeWidthPx = strokeWidth.toPx()
+        val cornerRadiusPx = 12.dp.toPx()
```
david-livefront, line 61: Should this be passed in as a param too?

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
```diff
@@ -0,0 +1,68 @@
...
+
+private val NAME_REGEX = Regex("""^[A-Z][A-Z .'-]+$""")
+
+/**
+ * Default [CardDataParser] implementation that uses regex patterns
+ * and Luhn validation to extract card details from OCR text.
+ */
+class CardDataParserImpl : CardDataParser {
+
+    @Suppress("MagicNumber")
```
david-livefront: I don't think you need this suppression.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
```diff
@@ -0,0 +1,68 @@
...
+    override fun parseCardData(text: String): CardScanData {
+        val panMatch = PAN_REGEX.find(text)
+        val number = panMatch
+            ?.value
+            ?.filter { it.isDigit() }
+            ?.takeIf { it.isValidLuhn() }
```
david-livefront, line 21: Do we want to return null if there is no number.

I noticed that later on you have a check for the number [here](https://github.com/bitwarden/android/pull/6720/changes#diff-b14f5d9a418b377f1b93f902f148fc8487964d7997958b1e23dfeef4c7bb4f14R54) anyways.

So maybe just return null if it's not valid

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardScanData.kt
```diff
@@ -0,0 +1,18 @@
...
+ * @property expirationYear The detected expiration year.
+ * @property cardholderName The detected cardholder name.
+ * @property securityCode The detected security code.
+ */
+data class CardScanData(
+    val number: String? = null,
+    val expirationMonth: String? = null,
+    val expirationYear: String? = null,
+    val cardholderName: String? = null,
+    val securityCode: String? = null,
```
david-livefront: Can we get rid of the defaults?

I don't think they are ever used.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzerImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private val isInAnalysis = AtomicBoolean(false)
+
+    private val recognizer = TextRecognition.getClient(
+        TextRecognizerOptions.DEFAULT_OPTIONS,
+    )
+
+    override lateinit var onCardScanned: (CardScanData) -> Unit
+
+    @Suppress("UnsafeOptInUsageError")
```
david-livefront: Can we opt in instead:

```kotlin
    @OptIn(ExperimentalGetImage::class)
```

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzerImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * [CardTextAnalyzer] implementation that uses ML Kit Text Recognition
+ * to detect credit card details from camera frames.
+ *
+ * @property cardDataParser The parser used to extract card data from
+ * recognized text.
+ */
+@OmitFromCoverage
+class CardTextAnalyzerImpl(
+    private val cardDataParser: CardDataParser = CardDataParserImpl(),
```
david-livefront: Can this be passed in where the `CardTextAnalyzerImpl` is created, it seems odd to have the `impl` here.

ui/src/test/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImplTest.kt
```diff
@@ -0,0 +1,176 @@
...
+
+class CardDataParserImplTest {
+
+    private val parser = CardDataParserImpl()
+
+    @Test
+    fun `parseCardData extracts valid card number with Luhn check`() {
+        val text = "4111 1111 1111 1111 12/25"
+        val result = parser.parseCardData(text)
+        assertEquals("4111111111111111", result.number)
```
david-livefront: Why not just assert the whole result

ui/src/main/kotlin/com/bitwarden/ui/platform/components/camera/CardScanOverlay.kt
SaintPatrck: I don't think so. The corner radius is a fixed visual property of the card shape, not something the callers should modify.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
SaintPatrck: I think that couples the parser logic too tightly to the consumers use case. The purpose of this parser is to extract all of the card data it can. It's up to the consumer to decide whether the extracted data is sufficient for its needs. This feel like the correct separation of concerns boundary to me.

ui/src/test/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImplTest.kt
SaintPatrck: This keeps the test scope focused and less likely to break if a change to parsing other fields is made. I.e., a change to name parsing shouldn't break a Luhn validation test.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/camera/CardScanOverlay.kt
david-livefront: Fair enough

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardDataParserImpl.kt
david-livefront: It just feels very weird to have a data class that has all nullable properties.

But I see your point, maybe that is OK in this case.

SaintPatrck: I agree. I'm not against a check to ensure at least _some_ data is parsed, otherwise we return `null`. That would at least ensure callers don't receive an empty object.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzerImpl.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val isInAnalysis = AtomicBoolean(false)
+
+    private val recognizer = TextRecognition.getClient(
+        TextRecognizerOptions.DEFAULT_OPTIONS,
+    )
+
+    override lateinit var onCardScanned: (CardScanData) -> Unit
+
+    override fun close() {
+        recognizer.close()
```
david-livefront: I am unsure how we will call this when it is provided via a `LocalComposition`

SaintPatrck: Good point. I don't actually think we need it, anyways. At least not until we see actual memory issues. If we need to, we might be able to leverage `remember` and `DisposableEffect` to make sure it's closed properly. I'll revert for now and we can come back if issues start showing up.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzer.kt
```diff
@@ -0,0 +1,16 @@
+package com.bitwarden.ui.platform.feature.cardscanner.util
+
+import androidx.camera.core.ImageAnalysis
+import androidx.compose.runtime.Stable
+
+/**
+ * An interface used to analyze camera frames for credit card text.
+ */
+@Stable
+interface CardTextAnalyzer : ImageAnalysis.Analyzer {
```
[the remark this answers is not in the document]
SaintPatrck: Removed closeable implementation. Will address if/when memory issues become apparent.

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzerImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+@OmitFromCoverage
+class CardTextAnalyzerImpl(
+    private val cardDataParser: CardDataParser,
+) : CardTextAnalyzer {
+
+    private val isInAnalysis = AtomicBoolean(false)
+
+    private val recognizer = TextRecognition.getClient(
+        TextRecognizerOptions.DEFAULT_OPTIONS,
+    )
```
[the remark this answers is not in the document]
SaintPatrck: MLKit manages its own resources reasonably well. We will address this if memory issues arise.

SaintPatrck:
After discussion with Design team, cardholder name extraction will be omitted. It is too error prone, and comparable features in other applications also omit cardholder name when scanning with a camera.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/camera/CardScanOverlay.kt
```diff
@@ -0,0 +1,73 @@
...
+            strokeWidth = strokeWidth,
+            modifier = Modifier
+                .padding(all = 8.dp)
+                .width(overlayWidth)
+                .aspectRatio(CARD_ASPECT_RATIO),
+        )
+    }
+}
+
+@Suppress("MagicNumber")
```
david-livefront, line 52: Is this suppression needed?

PR #6723, opened 2026-03-26
[PM-34168] Add future CalyxOS Chromium key to FIDO2 privilege community list

lucasmz-dev:
We should now be migrating to org.calyxos.chromium, instead of using the default Chromium package name, we're also using HSMs for signing our builds now, including this key, see https://calyxos.org/news/2026/02/10/calyxos-hsm-signing/

## 📔 Objective

Prepare Bitwarden for compatibility with our Chromium builds as soon as they're out

### DIFF
--- a/app/src/main/assets/fido2_privileged_community.json
+++ b/app/src/main/assets/fido2_privileged_community.json
@@ -43,6 +43,18 @@
     {
       "type": "android",
       "info": {
+        "package_name": "org.calyxos.chromium",
+        "signatures": [
+          {
+            "build": "release",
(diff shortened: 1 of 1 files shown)


lucasmz-dev:
Kind of a continuation of #6297 

We don't have any Chromium builds out yet, neither have them have been previously installable on non-CalyxOS devices, sorry about that

I can get other members of the team to vouch for this key if it is required or would at all help.

(You could consider Aysha's like one :P)

SaintPatrck (approved):

lucasmz-dev:
BTW: this PR worked as I was able to get passkeys to work with our Chromium builds with 2026.4.0.

PR #6725, opened 2026-03-27
Revert "PM-34193: Rollback SDK update for Vault lockout bug"

david-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34193

## 📔 Objective

Reverts bitwarden/android#6724

This is being reverted because we have found a better way to address PM-34193 without rolling back the entire SDK.


david-livefront:
Thanks @SaintPatrck

PR #6743, opened 2026-03-30
BWA-228: bug: Update identity custom field keys to use index

david-livefront:
## 🎟️ Tracking

[BWA-228](https://bitwarden.atlassian.net/browse/BWA-228)

## 📔 Objective

This PR updates the identity screen to use the custom field index as part of it's unique key. This is the same as the other cipher types and ensure that they are unique even if their contents are identical.

[BWA-228]: https://bitwarden.atlassian.net/browse/BWA-228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


david-livefront:
Thanks @aj-rosado

PR #6747, opened 2026-03-31
[PM-34487] llm: Add Android device interaction MCP server with ADB tooling

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34487

## 📔 Objective

Add a TypeScript MCP server and companion skill for structured Android device interaction via ADB. Replaces the initial shell script approach with proper XML parsing, structured `dumpsys window` parsing, and native geometry for obstruction detection.

## 🏗️ Architecture

```
Claude Code ──stdio──> MCP Server (TypeScript)
                         ├── tools/        (6 tool handlers)
                         ├── adb/          (execFile wrapper, device discovery)
                         ├── parsers/      (UIAutomator XML, dumpsys window)
                         └── geometry/     (bounds, obstruction, visible-region)
```

## 🔧 MCP Tools

| Tool | Description |
|------|-------------|
| `capture` | Dump UI hierarchy XML and/or screenshot |
| `find_element` | Find element by text/content-desc with two-layer obstruction detection |
| `tap_element` | Find + tap + capture (auto-adjusts for obstructions) |
| `tap_at` | Tap specific coordinates + capture |
| `navigate` | Home, back, app-drawer navigation + capture |
| `input_text` | Type text into focused field with shell-safe escaping |

## 🛡️ Obstruction Detection

Two-layer system for detecting when UI elements are blocked:
- **Layer 1 — System overlays**: Parses `dumpsys window windows` for TalkBack, PiP, accessibility overlays using `touchable region` bounds
- **Layer 2 — In-app elements**: Finds topmost clickable element at tap point via XML tree traversal (catches FABs, dialogs, bottom sheets)

When obstructed, computes the largest visible strip (top/bottom/left/right of obstructor) and returns adjusted tap coordinates.

## 📁 Key Files

- `.mcp.json` — MCP server config (stdio, auto-build)
- `.claude/mcp/android-device-server/` — TypeScript MCP server (self-contained for future plugin extraction)
- `.claude/skills/interacting-with-android-device/SKILL.md` — Companion skill with tool docs and `allowed-tools`

## ✅ Testing

- 57 unit tests across geometry, parsers, ADB discovery (fixture-based, no device required)
- `fast-xml-parser` for UIAutomator XML → typed tree
- Zod validation on all tool inputs
- `execFile` (not `exec`) prevents host-side command injection


.claude/skills/interacting-with-android-device/scripts/adb-find-element.sh
```diff
@@ -0,0 +1,63 @@
...
+                found = True
+                break
+
+    if not found:
+        print("ERROR: Element not found", file=sys.stderr)
+        sys.exit(1)
+except Exception as e:
+    print(f"ERROR: {e}", file=sys.stderr)
+    sys.exit(1)
+EOF
```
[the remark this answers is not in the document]
SaintPatrck: Removed all python usage.

.claude/skills/interacting-with-android-device/scripts/adb-tap-and-capture.sh
```diff
@@ -0,0 +1,30 @@
...
+# Usage: ./adb-tap-and-capture.sh <x> <y> [wait_seconds]
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo "Usage: $0 <x> <y> [wait_seconds]"
+    exit 1
+fi
+
+X=$1
+Y=$2
+WAIT=${3:-2}
```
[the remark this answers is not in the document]
SaintPatrck: Validation added

.claude/skills/interacting-with-android-device/scripts/adb-navigate.sh
```diff
@@ -0,0 +1,53 @@
...
+        echo "Going to home screen..."
+        $ADB shell input keyevent 3
+        ;;
+    back)
+        echo "Pressing back button..."
+        $ADB shell input keyevent 4
+        ;;
+    app-drawer)
+        echo "Opening app drawer..."
+        $ADB shell input swipe 672 2800 672 500 1000
```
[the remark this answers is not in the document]
SaintPatrck: Script now calculates values correctly.

.claude/skills/interacting-with-android-device/SKILL.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Quick Start: Using Helper Scripts
+
+Helper scripts in the `.claude/skills/interacting-with-android-device/scripts/` directory automate repetitive UI testing tasks and reduce token overhead.
+
+**Available scripts:**
+- `adb-capture.sh [--xml] [--screenshot] [--all]` - Capture current device state. Default (no flags): both screenshot and XML hierarchy.
+- `adb-find-element.sh <text>` - Find element by `text` or `content-desc`, return center coordinates (`X Y`). Dumps UI hierarchy, parses XML, calculates center from bounds.
+- `adb-tap-and-capture.sh <x> <y> [wait_seconds=2]` - Tap at coordinates, wait, capture and pull screenshot.
+- `adb-tap-element.sh <text> [wait_seconds=2]` - Find, tap, and capture in one command (recommended). Combines `adb-find-element.sh` + `adb-tap-and-capture.sh`.
```
david-livefront: Maybe this is getting too far into the weeds but are we at all concerned about z-order issues.

I know we have seen that in our own tests, where the FAB will be on top of another button, so when we click those coordinates, we expect the button clicked but we actually click the FAB.

SaintPatrck: Yes. I just ran into that. Claude figured out how to manually adjust the touch coordinates pretty quickly given the other instructions and commands that are present. I'll see if there's a relatively easy way to mitigate it.

SaintPatrck: Pivoted to an MCP server because the bash was getting _crrraaaazy_. This made handling z-order easier, and Claude is more likely to use them without direct instruction thanks to built-in MCP discovery. Converted the skill into an "instruction manual" for how to use the MCP server and perform common interactions. 🚀

theMickster:
❓♻️  We have many places where the mcp tools write to `view.xml` / `screen.png`. Any concurrent calls will clobber each other (e.g. a parallel `capture` from the LLM). Do we have any thoughts on something like `os.tmpdir()`or with a unique per-invocation filename (UUID or pid+counter) to avoid the problem?

Files that I found with this include

1. capture.ts
2. find-element-pipeline.ts
3. navigate.ts
4. tap-at.ts
5. tap-element.ts

.claude/mcp/android-device-server/src/geometry/bounds.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function boundsEqual(a: Rect, b: Rect): boolean {
+  return a.left === b.left && a.top === b.top && a.right === b.right && a.bottom === b.bottom;
+}
+
+/**
+ * Parse Android bounds string "[left,top][right,bottom]" into a Rect.
+ */
+export function parseBounds(bounds: string): Rect | null {
+  const match = bounds.match(/\[(\d+),(\d+)\]\[(\d+),(\d+)\]/);
```
theMickster: ♻️ A local multi-agent review came up with the following finding. ❓Thoughts❓

<details>
<summary>`parseBounds` regex rejects negative coordinates UIAutomator emits for partially-offscreen elements</summary>

`.claude/mcp/android-device-server/src/geometry/bounds.ts:46`

  ```ts
  const match = bounds.match(/\[(\d+),(\d+)\]\[(\d+),(\d+)\]/);
  if (!match) return null;d
  ```

UIAutomator can emit bounds with negative `left`/`top` values for elements that are partially scrolled/translated off-screen (common during swipe animations or when using RTL layouts). This regex returns `null` for those bounds, and `findElementWithObstruction` then surfaces `"Element found but has no bounds"` for a legitimately-visible element. Fix: `\[(-?\d+),(-?\d+)\]\[(-?\d+),(-?\d+)\]`.

</details>

theMickster:
❓ So we have unit tests for this MCP server just like [bitwarden-atlassian-tools](https://github.com/bitwarden/ai-plugins/tree/main/plugins/bitwarden-atlassian-tools) and we have a `vitest run` in `package.json`, but I think we have some gaps to address in follow-up PRs. 

- The next engineer doesn't easily know that these tests must be run manually on their local workstation prior to raising a PR. 
- The next code reviewer doesn't easily know that there's not CI+CD wired up to run these new unit tests. 

🤔

SaintPatrck:
> ❓♻️ We have many places where the mcp tools write to `view.xml` / `screen.png`. Any concurrent calls will clobber each other (e.g. a parallel `capture` from the LLM). Do we have any thoughts on something like `os.tmpdir()`or with a unique per-invocation filename (UUID or pid+counter) to avoid the problem?
> 
> Files that I found with this include
> 
> 1. capture.ts
> 2. find-element-pipeline.ts
> 3. navigate.ts
> 4. tap-at.ts
> 5. tap-element.ts

In theory, this is technically correct. In practice, however, this is unlikely to manifest any real issue because "parallel `capture`" isn't a valid use-case. `capture` contains the entire screen contents. If the LLM is looking for multiple elements, a single `capture` will contain them all.

.claude/mcp/android-device-server/src/geometry/bounds.ts
theMickster: Thank you for addressing this in 8a9ce1755713f29b26f50a3f23fa2d63b3edff5b 
@SaintPatrck

SaintPatrck:
@theMickster valid point, regarding tests. I'll follow-up adding test execution to the build pipeline.

theMickster (approved):

david-livefront (approved):

PR #6754, opened 2026-04-01
[PM-33982] feat: Add device management screen

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33982

## 📔 Objective

Add a new screen for device management on Android.
This new screen should display the current device, followed by the pending authorization requests and finally all devices with active sessions ordered by "last active date".

## 📸 Screenshots
Manage_devices_screen
Account_security_options_screen


app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+    this ?: return null
+    val nowDate = clock.instant().atZone(clock.zone).toLocalDate()
+    val activityDate = this.atZone(clock.zone).toLocalDate()
+    val daysAgo = ChronoUnit.DAYS.between(activityDate, nowDate)
+    val resId = when {
+        daysAgo <= 0 -> BitwardenString.active_today
+        daysAgo < 7 -> BitwardenString.active_this_week
+        daysAgo < 14 -> BitwardenString.active_last_week
+        daysAgo < 30 -> BitwardenString.active_this_month
+        else -> BitwardenString.active_over_thirty_days_ago
```
david-livefront: So this is all based on number of days in a week/month. Should it be based on the day of the week?

If today is Monday, 3 days ago is last week, not this week, right?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Returns a localized string describing how recently this device was active,
+ * or null if no activity date is available.
+ *
+ * Buckets are based on calendar days in the device's local timezone, matching
+ * the web client behaviour. Using [java.time.LocalDate] comparison makes this DST-safe without
+ * requiring rounding (unlike the JavaScript equivalent).
+ */
+@Suppress("MagicNumber")
+fun Instant?.toLastActivityLabel(clock: Clock, resourceManager: ResourceManager): String? {
```
david-livefront: Since this is a UI layer component, can we just return a `Text`?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+        18 -> DeviceTypeEntry(BitwardenString.web, "Vivaldi")
+        19 -> DeviceTypeEntry(BitwardenString.extension, "Vivaldi")
+        20 -> DeviceTypeEntry(BitwardenString.extension, "Safari")
+        21 -> DeviceTypeEntry(BitwardenString.sdk, "")
+        22 -> DeviceTypeEntry(BitwardenString.server, "")
+        23 -> DeviceTypeEntry(BitwardenString.cli, "Windows")
+        24 -> DeviceTypeEntry(BitwardenString.cli, "MacOs")
+        25 -> DeviceTypeEntry(BitwardenString.cli, "Linux")
+        26 -> DeviceTypeEntry(BitwardenString.extension, "DuckDuckGo")
+        else -> return resourceManager.getString(BitwardenString.unknown_device)
```
david-livefront: Same thing here, can we just return a `Text`

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+        13 -> DeviceTypeEntry(BitwardenString.web, "IE")
+        14 -> DeviceTypeEntry(BitwardenString.web, "Unknown")
+        15 -> DeviceTypeEntry(BitwardenString.mobile, "Amazon")
+        16 -> DeviceTypeEntry(BitwardenString.desktop, "Windows UWP")
+        17 -> DeviceTypeEntry(BitwardenString.web, "Safari")
+        18 -> DeviceTypeEntry(BitwardenString.web, "Vivaldi")
+        19 -> DeviceTypeEntry(BitwardenString.extension, "Vivaldi")
+        20 -> DeviceTypeEntry(BitwardenString.extension, "Safari")
+        21 -> DeviceTypeEntry(BitwardenString.sdk, "")
+        22 -> DeviceTypeEntry(BitwardenString.server, "")
```
david-livefront: Should these say something?

Also, should these platforms be translatable?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt
```diff
@@ -0,0 +1,475 @@
...
+                    state = viewState,
+                    onNavigateToLoginApproval = {
+                        viewModel.trySendAction(ManageDevicesAction.PendingRequestRowClick(it))
+                    },
+                )
+            }
+
+            ManageDevicesState.ViewState.Error -> BitwardenErrorContent(
+                message = stringResource(
+                    id = BitwardenString.generic_error_message,
```
david-livefront, line 162: Do we have nothing more specific than this?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt
```diff
@@ -0,0 +1,475 @@
...
+                            dividerPadding = 0.dp,
+                        ),
+                        modifier = Modifier
+                            .testTag("LoginRequestCell")
+                            .fillMaxWidth()
+                            .standardHorizontalMargin(),
+                    )
+                }
+
+                else -> SessionItem(
```
david-livefront: Can we make this exhaustive

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
aj-rosado: I have mirrored the other clients logic but that makes sense to me 🤔

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
aj-rosado: Other clients are returning the same data.
Regarding translations, nothing is translatable on other clients 🤔 

It makes sense to me to be able to translate words like "server" or "extension"

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt
aj-rosado: We do not, mobile was based roughly on existing Pending Auth screen and Manage Devices extension screen. 
Probably will see some more design work before being approved

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
aj-rosado: We have updated the texts so that it is clearer 🙌

david-livefront: The language does match the functionality now. But don't we want it to say `this week`, not `last seven days`?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
```diff
@@ -0,0 +1,33 @@
...
+    val activityDate = this.atZone(clock.zone).toLocalDate()
+    val daysAgo = ChronoUnit.DAYS.between(activityDate, nowDate)
+    val resId = when {
+        daysAgo <= 0 -> BitwardenString.today
+        daysAgo < 7 -> BitwardenString.past_seven_days
+        daysAgo < 14 -> BitwardenString.past_fourteen_days
+        daysAgo < 30 -> BitwardenString.past_thirty_days
+        else -> BitwardenString.over_thirty_days_ago
+    }
+    return resourceManager.getString(resId).asText()
```
david-livefront: Now that you are returning a `Text` you do not need the `ResourceManager` here anymore.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.bitwarden.ui.util.Text
+import com.bitwarden.ui.util.asText
+import com.x8bit.bitwarden.ui.platform.manager.resource.ResourceManager
+
+/**
+ * Converts a device type integer to a human-readable display name.
+ * Returns e.g. "Mobile - Android", "Extension - Chrome", "Desktop - Windows".
+ */
+@Suppress("CyclomaticComplexMethod", "MagicNumber")
+fun Int.toReadableDeviceTypeName(resourceManager: ResourceManager): Text {
```
david-livefront: Same here, the `ResourceManager` is unnecessary

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
aj-rosado: It was decided to keep it simpler as this change will affect all other clients. So we are keeping the logic and adapting the texts.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
```diff
@@ -0,0 +1,57 @@
...
+            24 -> DeviceTypeEntry(BitwardenString.cli, "MacOs")
+            25 -> DeviceTypeEntry(BitwardenString.cli, "Linux")
+            26 -> DeviceTypeEntry(BitwardenString.extension, "DuckDuckGo")
+            else -> return BitwardenString.unknown_device.asText()
+        }
+
+        return if (entry.platform.isNotEmpty()) {
+            entry.categoryResId.asText()
+                .concat(
+                    " - ${entry.platform}".asText(),
```
david-livefront: Instead of concat, can we make these format args?

ui/src/main/res/values/strings.xml
```diff
@@ -1232,6 +1232,25 @@ Do you want to switch to this account?</string>
     <string name="continue_without_syncing">Continue without syncing</string>
     <string name="external_link">External link</string>
     <string name="external_link_format" comment="Used for accessibility to indicate that tapping this item will leave the app">%1$s, External link</string>
+    <string name="manage_devices">Manage devices</string>
```
david-livefront: This managed devices string is being used in both prod code and for feature flags.

The feature flag name should be in the unlocalized file.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -0,0 +1,490 @@
...
+    private fun handleCloseClicked() {
+        sendEvent(ManageDevicesEvent.NavigateBack)
+    }
+
+    private fun handleHideBottomSheet() {
+        mutableStateFlow.update { it.copy(internalHideBottomSheet = true) }
+    }
+
+    private fun handleOnLifecycleResumed() {
+        updateAuthRequestList()
```
david-livefront, line 110: The `updateAuthRequestList` is a flow, why reset it on resume?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -0,0 +1,490 @@
...
+
+    private fun handleRefreshPull() {
+        mutableStateFlow.update {
+            it.copy(
+                isRefreshing = true,
+                devicesLoaded = false,
+                authRequestsLoaded = false,
+            )
+        }
+        updateAuthRequestList()
```
david-livefront, line 121: Same here, why restart the flow here?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -0,0 +1,490 @@
...
+            .getAuthRequestsWithUpdates()
+            .map { ManageDevicesAction.Internal.AuthRequestsResultReceive(it) }
+            .onEach(::sendAction)
+            .launchIn(viewModelScope)
+    }
+
+    private fun fetchAllDevices() {
+        devicesJob.cancel()
+        devicesJob = viewModelScope.launch {
+            coroutineScope {
```
david-livefront: What is the purpose of the `coroutineScope` here?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -0,0 +1,490 @@
...
+    }
+}
+
+/**
+ * Models state for the Manage Devices screen.
+ */
+@Parcelize
+data class ManageDevicesState(
+    val authRequests: List<AuthRequest>,
+    val devices: List<DeviceInfo>,
```
david-livefront: Can we make all these lists into `ImmuableList`s

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
aj-rosado: Was to couple both calls together (one needs the other). Although the `GetDeviceByIdentifier` is not necessary 🤔

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
aj-rosado: Want to force to get updated data `onResume` as this is an operation that has a time limit

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
aj-rosado: Want to force to get updated data when user pulls to refresh as this is an operation that has a time limit

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensions.kt
```diff
@@ -0,0 +1,53 @@
...
+            14 -> DeviceTypeEntry(BitwardenString.web_platform, "Unknown")
+            15 -> DeviceTypeEntry(BitwardenString.mobile_platform, "Amazon")
+            16 -> DeviceTypeEntry(BitwardenString.desktop_platform, "Windows UWP")
+            17 -> DeviceTypeEntry(BitwardenString.web_platform, "Safari")
+            18 -> DeviceTypeEntry(BitwardenString.web_platform, "Vivaldi")
+            19 -> DeviceTypeEntry(BitwardenString.extension_platform, "Vivaldi")
+            20 -> DeviceTypeEntry(BitwardenString.extension_platform, "Safari")
+            21 -> DeviceTypeEntry(BitwardenString.sdk, "")
+            22 -> DeviceTypeEntry(BitwardenString.server, "")
+            23 -> DeviceTypeEntry(BitwardenString.cli_platform, "Windows")
```
david-livefront: Do we need this intermediary step?

Can we just return the correct value here?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
david-livefront: I'm confused, this will continue to request updates every 5 seconds indefinitely what time-limit are you concerned about?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -0,0 +1,471 @@
...
+            )
+        }
+    }
+
+    private fun handleAuthRequestsResultReceived(
+        action: ManageDevicesAction.Internal.AuthRequestsResultReceive,
+    ) {
+        val filteredRequests = when (val result = action.authRequestsUpdatesResult) {
+            is AuthRequestsUpdatesResult.Update ->
+                result.authRequests.filterRespondedAndExpired(clock = clock)
```
david-livefront, line 191: Can we wrap this in curly braces.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
aj-rosado: This behaviour is copied from the `PendingRequestsViewModel`. 
As the Auth requests have a duration of 5 minutes, we do not want to show to the user expired requests. This is what I understood from this.

david-livefront: Ahh, 5 minutes makes way more sense. Sorry about that!

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModelTest.kt
```diff
@@ -0,0 +1,289 @@
...
+        mutableAuthRequestsWithUpdatesFlow.tryEmit(
+            AuthRequestsUpdatesResult.Update(authRequests = listOf(validAuthRequest)),
+        )
+
+        val content = viewModel.stateFlow.value.viewState as ManageDevicesState.ViewState.Content
+        assertEquals(DeviceSessionStatus.Current, content.items[0].status)
+        assertEquals(DeviceSessionStatus.Pending, content.items[1].status)
+        assertEquals(DeviceSessionStatus.None, content.items[2].status)
+        assertEquals(currentDevice.id, content.items[0].id)
+        assertEquals(pendingDevice.id, content.items[1].id)
```
david-livefront: You should be comparing again the entire `ManageDevicesState`, not individual ids and statuses

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModelTest.kt
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    @Test
+    fun `when getDevices returns error should show error state`() {
+        coEvery { authRepository.getDevices() } returns GetDevicesResult.Error
+        val viewModel = createViewModel()
+        assertEquals(
+            ManageDevicesState.ViewState.Error,
+            viewModel.stateFlow.value.viewState,
```
david-livefront: You should compare the entire thing, not just the viewSTate

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModelTest.kt
```diff
@@ -0,0 +1,289 @@
...
+        mockkStatic(TemporalAccessor::toFormattedDateTimeStyle)
+        mockkStatic(::isBuildVersionAtLeast)
+        every { isBuildVersionAtLeast(any()) } returns true
+        every {
+            fixedClock.instant().toFormattedDateTimeStyle(
+                dateStyle = FormatStyle.MEDIUM,
+                timeStyle = FormatStyle.MEDIUM,
+                clock = fixedClock,
+            )
+        } returns "Oct 27, 2023, 12:00:00 PM"
```
david-livefront: Why is this needed? The instant is fixed, so it should always return the same string

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensionsTest.kt
```diff
@@ -0,0 +1,226 @@
...
+            "Mobile - Android",
+            0.readableDeviceTypeName.toString(resources),
+        )
+    }
+
+    @Test
+    fun `type 1 should return Mobile - iOS`() {
+        assertEquals(
+            "Mobile - iOS",
+            1.readableDeviceTypeName.toString(resources),
```
david-livefront: None of resource mocking is needed, you can just compare the outputs:

```kotlin
assertEquals(
    BitwardenString.mobile_platform.asText("iOS"),
    1.readableDeviceTypeName,
)
```

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceLastActivityExtensions.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Returns a localized string describing how recently this device was active,
+ * or null if no activity date is available.
+ *
+ * Buckets are based on calendar days in the device's local timezone, matching
+ * the web client behaviour. Using [java.time.LocalDate] comparison makes this DST-safe without
+ * requiring rounding (unlike the JavaScript equivalent).
+ */
+@Suppress("MagicNumber")
+fun Instant?.toLastActivityLabel(clock: Clock): Text? {
+    this ?: return null
```
david-livefront: I can't seem to find where this is used?

Do we need to handle null here, could it just be?

```kotlin
fun Instant.toLastActivityLabel(clock: Clock): Text {
```

aj-rosado: This transforms data returned from the API and it can be nullable.

It is used [here](https://github.com/bitwarden/android/blob/2e7148368ef5ceb2b62869f25bbd9b5258888dd7/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt#L263) and transforms the Instant from [DeviceInfo](https://github.com/bitwarden/android/blob/2e7148368ef5ceb2b62869f25bbd9b5258888dd7/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/model/DeviceInfo.kt) into that string.

From the [server model](https://github.com/bitwarden/server/blob/a18b60014765ef717621c628c4868c900604872c/src/Core/Entities/Device.cs#L50) we can understand that older devices will return null

> Null if the device has not authenticated since activity tracking was introduced.

david-livefront: Can you not just the call the function like this:

```kotlin
nullableInstant?.toLastActivityLabel(clock = clock)
```

That will improve the API by allowing the return value to be a known nonull type when the instant is also nonnull.

app/src/test/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryTest.kt
```diff
@@ -6874,6 +6877,45 @@ class AuthRepositoryTest {
...
+
+        val result = repository.getDevices()
+
+        coVerify(exactly = 1) { devicesService.getDevices() }
+        assertTrue(result is GetDevicesResult.Success)
+        assertEquals(1, (result as GetDevicesResult.Success).devices.size)
+        val device = result.devices.first()
+        assertEquals("deviceId", device.id)
+        // identifier "deviceIdentifier" != uniqueAppId "testUniqueAppId", so not current device
+        assertFalse(device.isCurrentDevice)
```
david-livefront: We should be comparing the entire result

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/util/DeviceTypeExtensionsTest.kt
```diff
@@ -0,0 +1,161 @@
+package com.x8bit.bitwarden.ui.platform.feature.settings.accountsecurity.managedevices.util
+
+import com.bitwarden.ui.platform.resource.BitwardenString
+import com.bitwarden.ui.util.asText
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class DeviceTypeExtensionsTest {
```
SaintPatrck, line 10: 🎨 Non-blocking suggestion from Claude.

>Consider consolidating these per-type tests with `@TestFactory` + `dynamicTest`. Beyond reducing boilerplate, a data-driven approach makes it obvious when the `when` branch in `DeviceTypeExtensions.kt` and the test table drift apart — right now types `4, 8, 10, 11, 12, 13, 14, 17, 18, 19` are defined in the extension but have no test coverage.

```
@TestFactory
fun `readableDeviceTypeName returns correct value for each known device type`() =
    listOf(
        0 to BitwardenString.mobile_platform.asText("Android"),
        1 to BitwardenString.mobile_platform.asText("iOS"),
        2 to BitwardenString.extension_platform.asText("Chrome"),
        3 to BitwardenString.extension_platform.asText("Firefox"),
        4 to BitwardenString.extension_platform.asText("Opera"),
        5 to BitwardenString.extension_platform.asText("Edge"),
        6 to BitwardenString.desktop_platform.asText("Windows"),
        7 to BitwardenString.desktop_platform.asText("MacOS"),
        8 to BitwardenString.desktop_platform.asText("Linux"),
        9 to BitwardenString.web_platform.asText("Chrome"),
        10 to BitwardenString.web_platform.asText("Firefox"),
        11 to BitwardenString.web_platform.asText("Opera"),
        12 to BitwardenString.web_platform.asText("Edge"),
        13 to BitwardenString.web_platform.asText("IE"),
        14 to BitwardenString.web_platform.asText("Unknown"),
        15 to BitwardenString.mobile_platform.asText("Amazon"),
        16 to BitwardenString.desktop_platform.asText("Windows UWP"),
        17 to BitwardenString.web_platform.asText("Safari"),
        18 to BitwardenString.web_platform.asText("Vivaldi"),
        19 to BitwardenString.extension_platform.asText("Vivaldi"),
        20 to BitwardenString.extension_platform.asText("Safari"),
        21 to BitwardenString.sdk.asText(),
        22 to BitwardenString.server.asText(),
        23 to BitwardenString.cli_platform.asText("Windows"),
        24 to BitwardenString.cli_platform.asText("MacOS"),
    listOf(
        0 to BitwardenString.mobile_platform.asText("Android"),
        1 to BitwardenString.mobile_platform.asText("iOS"),
        2 to BitwardenString.extension_platform.asText("Chrome"),
        3 to BitwardenString.extension_platform.asText("Firefox"),
        4 to BitwardenString.extension_platform.asText("Opera"),
        5 to BitwardenString.extension_platform.asText("Edge"),
        6 to BitwardenString.desktop_platform.asText("Windows"),
        7 to BitwardenString.desktop_platform.asText("MacOS"),
        8 to BitwardenString.desktop_platform.asText("Linux"),
        9 to BitwardenString.web_platform.asText("Chrome"),
        10 to BitwardenString.web_platform.asText("Firefox"),
        11 to BitwardenString.web_platform.asText("Opera"),
        12 to BitwardenString.web_platform.asText("Edge"),
        13 to BitwardenString.web_platform.asText("IE"),
        14 to BitwardenString.web_platform.asText("Unknown"),
        15 to BitwardenString.mobile_platform.asText("Amazon"),
        16 to BitwardenString.desktop_platform.asText("Windows UWP"),
        17 to BitwardenString.web_platform.asText("Safari"),
        18 to BitwardenString.web_platform.asText("Vivaldi"),
        19 to BitwardenString.extension_platform.asText("Vivaldi"),
        20 to BitwardenString.extension_platform.asText("Safari"),
        21 to BitwardenString.sdk.asText(),
        22 to BitwardenString.server.asText(),
        23 to BitwardenString.cli_platform.asText("Windows"),
        24 to BitwardenString.cli_platform.asText("MacOS"),
        25 to BitwardenString.cli_platform.asText("Linux"),
        26 to BitwardenString.extension_platform.asText("DuckDuckGo"),
    ).map { (type, expected) ->
        dynamicTest("type $type should return $expected") {
            assertEquals(expected, type.readableDeviceTypeName)
        }
    }

@Test
fun `unknown type should return unknown device string`() {
    assertEquals(BitwardenString.unknown_device.asText(), 999.readableDeviceTypeName)
}

@Test
fun `negative type should return unknown device string`() {
    assertEquals(BitwardenString.unknown_device.asText(), (-1).readableDeviceTypeName)
}
```

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/model/DeviceInfo.kt
```diff
@@ -0,0 +1,31 @@
...
+ * @property lastActivityDate The date and time of the device's last activity, if available.
+ * @property pendingAuthRequest The pending auth request for this device, if any.
+ * @property isCurrentDevice If this is the current device being used.
+ */
+@Parcelize
+data class DeviceInfo(
+    val id: String,
+    val name: String,
+    val identifier: String,
+    val type: Int,
```
david-livefront: Can this `type` be enumerated?

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/util/DeviceResponseJsonExtensions.kt
```diff
@@ -12,7 +13,7 @@ fun DeviceResponseJson.toDeviceInfo(currentDeviceIdentifier: String): DeviceInfo
         id = id,
         name = name,
         identifier = identifier,
-        type = type,
+        type = DeviceType.entries.firstOrNull { it.value == type } ?: DeviceType.UNKNOWN,
```
david-livefront: Can we add the type for `DeviceResponseJson` too

david-livefront (dismissed):

theMickster:
Howdy @aj-rosado et. al 👋🏼

I am field testing [the new multi-agent code review](https://github.com/bitwarden/ai-plugins/pull/96), I gave it a local run against the PR. The findings are below for your analysis. Please incorporate the findings that are true signals.

We are also looking closely into the difference in the way that Claude Code runs the review given different models for pros, cons, false-positive rates, instructions not being followed, etc.  so please don't hesitate to let me know your thoughts.

Many Thanks! 

---

<details>
<summary> Opus Code Review: [PM-33982] feat: Add device management screen (#6754)</summary>
**Date:** 2026-05-07 | **Reviewed by:** Claude Code | **Model:** Opus

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 0     |
| ♻️ Refactor  | 0     |

This PR adds a well-structured device management screen that follows the established architectural patterns of the codebase. The multi-source async loading pattern (`devicesLoaded` + `authRequestsLoaded` coordination), navigation placement inside `vaultUnlockedGraph`, the `hideBottomSheet` F-Droid/notification logic, and pull-to-refresh behavior were all validated as consistent with existing sibling screens (`PendingRequestsScreen`/`PendingRequestsViewModel`). No security issues were found: sensitive fields (`encryptedUserKey`, `encryptedPublicKey`) from the network response are correctly dropped at the domain mapping layer, and the feature is properly gated behind a feature flag. Test coverage is thorough.

## Findings

No findings found.

## Reviewed and Dismissed

<details><summary>🔍 3 initial findings dismissed after validation</summary>

#### SessionItem test tag is hardcoded "CurrentItemCell" but renders for both Current and non-current devices

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt:742`
**Caught by:** Code quality agent
**Original severity:** ⚠️ Important
**Original confidence:** 90/100
**Dismissed at:** Step 4 validation
**Dismissed because:** The "test-tag uniqueness" concern is incorrect — the established codebase pattern applies a single shared `testTag` to all rows in a list (e.g. `SendCell`, `LoginRequestCell`, `CipherCell`). The remaining naming concern (non-current devices bearing the `CurrentItemCell` tag) is a nitpick a senior engineer would not flag in code review. Additionally, line 742 in the finding is a diff position rather than a file line number.

---

#### `hideBottomSheet` logic inverts F-Droid and pre-Tiramisu guards, ignoring user dismissal

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt:296-300`
**Caught by:** Bug analysis agent
**Original severity:** ⚠️ Important
**Original confidence:** 85/100
**Dismissed at:** Step 4 validation
**Dismissed because:** Pre-existing pattern, not introduced by this PR. The identical `hideBottomSheet` getter (`internalHideBottomSheet && !isFdroid && isBuildVersionAtLeast(TIRAMISU)`) with the same screen-level OR composition exists verbatim in `PendingRequestsViewModel.kt:238-241` and `PendingRequestsScreen.kt:106-110`. This PR copies the established pattern.

---

#### Pull-to-refresh never refetches devices after the first successful load

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt:113-125`
**Caught by:** Security & logic agent
**Original severity:** ⚠️ Important
**Original confidence:** 85/100
**Dismissed at:** Step 4 validation
**Dismissed because:** Intentional design verified by the test `RefreshPull when devices loaded should re-fetch auth requests only` (`ManageDevicesViewModelTest.kt`) which explicitly asserts `coVerify(exactly = 1) { authRepository.getDevices() }`. A companion test (`RefreshPull when devices failed should re-fetch both devices and auth requests`) shows devices are re-fetched only when the prior load failed. No P01–P06 / VD/EK/AT/SC/TC implication.

</details>

</details>

---

<details>
<summary>Sonnet Code Review: [PM-33982] feat: Add device management screen (#6754) </summary>

**Date:** 2026-05-07 | **Reviewed by:** Claude Code | **Model:** Sonnet

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 3     |
| ♻️ Refactor  | 2     |

This PR adds a well-structured Device Management screen following Bitwarden's MVVM/UDF and State/Action/Event patterns, with appropriate feature-flag gating and broad test coverage. The primary bug is an inverted boolean condition in `ManageDevicesState.hideBottomSheet` that permanently breaks the "Skip for now" dismissal for F-Droid and pre-Android-13 users — a scenario with no test coverage. The success-path test for `AuthRepositoryImpl.getDevices` does not exercise the `isCurrentDevice=true` branch despite that branch being the sole reason `currentDeviceIdentifier` is threaded through. One auth-request filter function is duplicated verbatim from a sibling ViewModel.

## Findings

### ⚠️ Important

#### `hideBottomSheet` dismiss is permanently ineffective on F-Droid and pre-Android-13 builds

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt:293`
**Caught by:** Bug analysis agent

  <details><summary>Details</summary>

  The `hideBottomSheet` computed property is:

  ```kotlin
  val hideBottomSheet: Boolean
      get() = internalHideBottomSheet &&
          !isFdroid &&
          isBuildVersionAtLeast(Build.VERSION_CODES.TIRAMISU)
  ```

  Because all three conditions are joined with `&&`, the property can only return `true` when the user dismissed it AND the build is not F-Droid AND the OS is Android 13+.

  **F-Droid** (`isFdroid = true`): `!isFdroid` is `false`, so `hideBottomSheet` is permanently `false` regardless of `internalHideBottomSheet`. Tapping "Skip for now" sets `internalHideBottomSheet = true` but the sheet reappears on every visit. F-Droid users also cannot receive FCM push notifications, so the notification-permission prompt is irrelevant to them.

  **Android < 13**: `isBuildVersionAtLeast(TIRAMISU)` is `false`, so `hideBottomSheet` is permanently `false` — dismiss has no effect here either.

  Correct semantics (hide when on F-Droid, OR pre-Tiramisu, OR user dismissed):

  ```kotlin
  val hideBottomSheet: Boolean
      get() = isFdroid || !isBuildVersionAtLeast(Build.VERSION_CODES.TIRAMISU) || internalHideBottomSheet
  ```

  </details>

#### `getDevices` success test only covers `isCurrentDevice=false`; `isCurrentDevice=true` case is untested

`app/src/test/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryTest.kt:6888`
**Caught by:** Code quality agent

  <details><summary>Details</summary>

  The success test constructs a `DeviceResponseJson` with `identifier = "deviceIdentifier"`, which differs from the stub `uniqueAppId = "testUniqueAppId"`, so only the `isCurrentDevice = false` branch of `toDeviceInfo(currentDeviceIdentifier)` is exercised.

  The `isCurrentDevice = true` path — which determines which device is sorted to the top of the list and labeled "Current session" — has no coverage at the repository or extension layer. The sole reason `currentDeviceIdentifier` is threaded through `AuthRepositoryImpl → toDeviceInfo` is to produce this flag, yet the `true` branch is never asserted.

  A second test case where `identifier == authDiskSource.uniqueAppId` should verify that `isCurrentDevice = true` is produced.

  </details>

#### `HideBottomSheet` test does not cover `isFdroid=true`; the dismissal bug goes undetected

`app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModelTest.kt:113`
**Caught by:** Validation agent (collateral)

  <details><summary>Details</summary>

  The `HideBottomSheet should set hideBottomSheet to true` test hardcodes `buildInfoManager.isFdroid = false` in the class-level mock and mocks `isBuildVersionAtLeast(any())` to `true`. Under those conditions `hideBottomSheet` evaluates correctly and the test passes.

  There is no test exercising `isFdroid = true`, which is the exact configuration where `hideBottomSheet` is permanently `false` regardless of `internalHideBottomSheet`. A complementary test such as:

  ```kotlin
  @Test
  fun `HideBottomSheet has no effect on F-Droid builds`() {
      every { buildInfoManager.isFdroid } returns true
      val viewModel = createViewModel()
      viewModel.trySendAction(ManageDevicesAction.HideBottomSheet)
      assertFalse(viewModel.stateFlow.value.hideBottomSheet)
  }
  ```

  would have caught the logic inversion before this PR reached review.

  </details>

### ♻️ Refactor

#### `filterRespondedAndExpired` duplicated verbatim from `PendingRequestsViewModel`

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt:467`
**Caught by:** Architecture agent

  <details><summary>Details</summary>

  The private extension function `filterRespondedAndExpired(clock: Clock)` at line 467 is character-for-character identical to the same function in `PendingRequestsViewModel.kt`. Both live in the `accountsecurity` feature tree and share the same auth-request filtering contract.

  The project anti-pattern list explicitly lists "Create new patterns when established ones exist" as a DON'T. If the 5-minute expiry threshold changes or the filter predicate evolves, both copies must be updated in lockstep or they silently diverge.

  Suggested fix: move the function to a shared utility such as `app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/util/AuthRequestFilterExtensions.kt` and import it at both call sites.

  </details>

#### `testTag("CurrentItemCell")` applied to both `None` and `Current` status branches

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt:229`
**Caught by:** Code quality agent

  <details><summary>Details</summary>

  `.testTag("CurrentItemCell")` is applied to the `SessionItem` composable in the `when` branch that covers both `DeviceSessionStatus.None` and `DeviceSessionStatus.Current`. The tag name implies it identifies only the current-session device, but every inactive session also receives it.

  Any future test querying `onNodeWithTag("CurrentItemCell")` in a list with mixed statuses will get multiple nodes and fail with an ambiguous-node error. Renaming to `"SessionItemCell"` (with a separate tag for `Current`-only if needed) would align with the project convention where tags describe content rather than selection state.

  </details>



## Reviewed and Dismissed

   <details><summary>🔍 2 initial findings dismissed after validation</summary>

#### Inside `fingerprintPhrase?.let` block, passes `item.fingerprintPhrase` instead of `it`

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt:201`
**Caught by:** Code quality agent
**Original severity:** ♻️ Refactor
**Original confidence:** 85/100
**Dismissed at:** Step 4 validation
**Dismissed because:** Inside a `?.let` block, Kotlin smart-casts the nullable receiver to non-null in the block body, making `item.fingerprintPhrase` and `it` semantically identical; a senior engineer would not flag this in review.

#### `hideBottomSheet` logic inverted for F-Droid builds (security framing)

`app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt:293`
**Caught by:** Security & logic agent
**Original severity:** ⚠️ Important
**Original confidence:** 82/100
**Dismissed at:** Step 4 validation
**Dismissed because:** Exact duplicate of bug-1 — same property, same line, same root cause; the security-angle framing adds no new finding beyond what bug-1 already covers.

   </details>

</details>

---


[PM-33982]: https://bitwarden.atlassian.net/browse/PM-33982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-33982]: https://bitwarden.atlassian.net/browse/PM-33982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesScreen.kt
```diff
@@ -0,0 +1,448 @@
...
+    val snackbarHostState = rememberBitwardenSnackbarHostState()
+    EventsEffect(viewModel = viewModel) { event ->
+        when (event) {
+            ManageDevicesEvent.NavigateBack -> onNavigateBack()
+            is ManageDevicesEvent.NavigateToLoginApproval -> {
+                onNavigateToLoginApproval(event.fingerprint)
+            }
+
+            is ManageDevicesEvent.ShowSnackbar ->
+                snackbarHostState.showSnackbar(event.data)
```
david-livefront, line 95: This should have curly braces if it wraps to the second line

david-livefront (approved):

PR #6776, opened 2026-04-09
[PM-34833] bug: Search improvements

gabrbrand:
## 📔 Objective

Improvements to the search experience:
- Show search icon only when content is displayed. Hidden when content is loading, there is an error or there are no items.
- Hide clear icon when search term is empty. (Animated)
- Use a formatted search string instead of hard coding the order. e.g. in German the words are in a different order

## 📸 Screenshots
https://github.com/user-attachments/assets/adffdbf1-e950-41e6-a135-8b53b8a88305

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultScreen.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultScreen.kt
@@ -236,10 +236,12 @@ private fun VaultScreenScaffold(
                             updateAccountMenuVisibility(!accountMenuVisible)
                         },
                     )
-                    BitwardenSearchActionItem(
-                        contentDescription = stringResource(id = BitwardenString.search_vault),
-                        onClick = vaultHandlers.searchIconClickAction,
-                    )
+                    if (state.shouldShowSearchIcon) {
+                        BitwardenSearchActionItem(
+                            contentDescription = stringResource(id = BitwardenString.search_vault),
+                            onClick = vaultHandlers.searchIconClickAction,
+                        )
+                    }
                     BitwardenOverflowActionItem(
                         menuItemDataList = persistentListOf(
                             OverflowMenuItemData(
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultScreen.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultScreen.kt
@@ -236,12 +236,11 @@ private fun VaultScreenScaffold(
                             updateAccountMenuVisibility(!accountMenuVisible)
                         },
                     )
-                    if (state.shouldShowSearchIcon) {
-                        BitwardenSearchActionItem(
-                            contentDescription = stringResource(id = BitwardenString.search_vault),
-                            onClick = vaultHandlers.searchIconClickAction,
-                        )
-                    }
+                    BitwardenSearchActionItem(
+                        contentDescription = stringResource(id = BitwardenString.search_vault),
+                        isDisplayed = state.shouldShowSearchIcon,
+                        onClick = vaultHandlers.searchIconClickAction,
+                    )
                     BitwardenOverflowActionItem(
                         menuItemDataList = persistentListOf(
                             OverflowMenuItemData(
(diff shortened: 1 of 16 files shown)


ui/src/main/res/values-de-rDE/strings.xml
```diff
@@ -171,7 +171,7 @@
     <string name="autofill_noun">Automatisches Ausfüllen</string>
     <string name="autofill_verb">Automatisches Ausfüllen</string>
     <string name="autofill_or_view">Möchtest du diese Zugangsdaten automatisch ausfüllen oder ansehen?</string>
-    <string name="search">Durchsuchen</string>
+    <string name="search">%1$s durchsuchen</string>
```
david-livefront: We usually leave this up to translators in Crowdin.

Our automation will overwrite this when it runs in a few days.

ui/src/main/res/values/strings.xml
```diff
@@ -171,7 +171,7 @@
     <string name="autofill_noun">Autofill</string>
     <string name="autofill_verb">Autofill</string>
     <string name="autofill_or_view">Do you want to autofill or view this item?</string>
-    <string name="search">Search</string>
+    <string name="search">Search %1$s</string>
```
david-livefront: Might be worth just deleting this string and creating a new one like:

```xml
    <string name="search_format">Search %1$s</string>
```

@SaintPatrck Any thoughts on this?

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultScreen.kt
```diff
@@ -236,10 +236,12 @@ private fun VaultScreenScaffold(
                             updateAccountMenuVisibility(!accountMenuVisible)
                         },
                     )
-                    BitwardenSearchActionItem(
-                        contentDescription = stringResource(id = BitwardenString.search_vault),
-                        onClick = vaultHandlers.searchIconClickAction,
-                    )
+                    if (state.shouldShowSearchIcon) {
+                        BitwardenSearchActionItem(
```
david-livefront: What do you think about adding a parameter to `BitwardenSearchActionItem` for this:

```kotlin
fun BitwardenSearchActionItem(
    contentDescription: String,
    isDisplayed: Boolean = true,
    onClick: () -> Unit,
) {
    if (!isDisplayed) return
    BitwardenStandardIconButton(
        ...
``` 

```kotlin
BitwardenSearchActionItem(
    contentDescription = stringResource(id = BitwardenString.search_vault),
    isDisplayed = state.shouldShowSearchIcon,
    onClick = vaultHandlers.searchIconClickAction,
)
```

ui/src/main/res/values/strings.xml
SaintPatrck: I agree. I think we use `x`, so `"search_x"`. Would be helpful to add comment/context for translators too.

gabrbrand: Where do I need to add the comment for translators?

david-livefront: Something like this:
```kotlin
    <string name="search_format" comment="Used to indicate what type of items are being searched">Search %1$s</string>
```

ui/src/main/res/values/strings.xml
```diff
@@ -171,7 +171,7 @@
     <string name="autofill_noun">Autofill</string>
     <string name="autofill_verb">Autofill</string>
     <string name="autofill_or_view">Do you want to autofill or view this item?</string>
-    <string name="search">Search %1$s</string>
+    <string name="search_x" comment="%1$s is a placeholder for the type of item being searched for (e.g. Logins).">Search %1$s</string>
```
david-livefront, line 174: 👍 

Thanks for that

ui/src/main/res/values-de-rDE/strings.xml
```diff
@@ -171,7 +171,7 @@
     <string name="autofill_noun">Automatisches Ausfüllen</string>
     <string name="autofill_verb">Automatisches Ausfüllen</string>
     <string name="autofill_or_view">Möchtest du diese Zugangsdaten automatisch ausfüllen oder ansehen?</string>
-    <string name="search">%1$s durchsuchen</string>
+    <string name="search">Durchsuchen</string>
```
david-livefront, line 174: Can we drop this change?

gabrbrand: Will it be overwritten or should I add back the originally updated translation?

`<string name="search_x">%1$s durchsuchen</string>`

david-livefront: Ohh, sorry, I was looking at a partial diff.

Looks great as is.

david-livefront (approved):

david-livefront:
Thanks @gabrbrand

PR #6777, opened 2026-04-09
PM-34840: bug: Allow related-origin passkey creation

david-livefront:
## 🎟️ Tracking

[PM-34840](https://bitwarden.atlassian.net/browse/PM-34840)

## 📔 Objective

This PR pulls the requested origin for privileged apps (browsers) since they have already had their origin validated. This allows us to create passkeys from websites that use related-origins.

[PM-34840]: https://bitwarden.atlassian.net/browse/PM-34840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt
@@ -20,6 +20,7 @@ import com.bitwarden.fido.Fido2CredentialAutofillView
 import com.bitwarden.fido.Origin
 import com.bitwarden.fido.UnverifiedAssetLink
 import com.bitwarden.sdk.Fido2CredentialStore
+import com.bitwarden.ui.platform.base.util.prefixHttpsIfNecessary
 import com.bitwarden.ui.platform.base.util.prefixHttpsIfNecessaryOrNull
 import com.bitwarden.ui.platform.base.util.toAndroidAppUriString
 import com.bitwarden.vault.CipherListView
@@ -343,7 +344,16 @@ class BitwardenCredentialManagerImpl(
             ?.let { ClientData.DefaultWithCustomHash(hash = it) }
             ?: return Fido2RegisterCredentialResult.Error.InvalidAppSignature
 
-        val sdkOrigin = createPublicKeyCredentialRequest.origin
+        val requestedOrigin = this
+            .getPasskeyAttestationOptionsOrNull(createPublicKeyCredentialRequest.requestJson)
+            ?.relyingParty
+            ?.id
+            ?.prefixHttpsIfNecessary()
+
+        // PM-35130: We use the requested relying party for the basis of the origin for privileged
+        // apps to ensure that related-origin requests are processed successfully. In the future,
+        // the SDK should handle this for us and we will be able to send in the real origin.
+        val sdkOrigin = (requestedOrigin ?: createPublicKeyCredentialRequest.origin)
             ?.let { Origin.Web(it) }
             ?: return Fido2RegisterCredentialResult.Error.MissingHostUrl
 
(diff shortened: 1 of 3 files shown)


SaintPatrck (approved):

SaintPatrck (dismissed):

app/src/test/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerTest.kt
```diff
@@ -512,9 +512,10 @@ class BitwardenCredentialManagerTest {
 
     @Suppress("MaxLineLength")
     @Test
-    fun `registerFido2Credential should return MissingHostUrl when calling app if privileged and origin is missing`() =
+    fun `registerFido2Credential should return MissingHostUrl when calling app if privileged and relaying party and origin are missing`() =
```
iinuwa: nit: sp
```suggestion
    fun `registerFido2Credential should return MissingHostUrl when calling app if privileged and relying party and origin are missing`() =
```

app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt
```diff
@@ -343,7 +344,13 @@ class BitwardenCredentialManagerImpl(
...
             ?: return Fido2RegisterCredentialResult.Error.InvalidAppSignature
 
-        val sdkOrigin = createPublicKeyCredentialRequest.origin
+        val requestedOrigin = this
+            .getPasskeyAttestationOptionsOrNull(createPublicKeyCredentialRequest.requestJson)
+            ?.relyingParty
+            ?.id
+            ?.prefixHttpsIfNecessary()
+
+        val sdkOrigin = (requestedOrigin ?: createPublicKeyCredentialRequest.origin)
```
iinuwa, line 356: I thought about this some more from my previous suggestion: While this "works" to trick the SDK into not doing extra validation, it means that the generated `clientDataJSON` will look like `{"origin":"https://rpId.com", ...}` rather than `{"origin":"https://relatedorigin.com", ...}`, which means that the relying party loses information.

The correct thing to do is to fix this in the SDK. I think it's OK to move forward with this workaround, but maybe we should add a TODO to address that later?

iinuwa (dismissed):
LGTM, but this should be reviewed by someone else with more experience with the Android implementation than me to be sure.

app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt
```diff
@@ -343,7 +344,16 @@ class BitwardenCredentialManagerImpl(
...
 
-        val sdkOrigin = createPublicKeyCredentialRequest.origin
+        val requestedOrigin = this
+            .getPasskeyAttestationOptionsOrNull(createPublicKeyCredentialRequest.requestJson)
+            ?.relyingParty
+            ?.id
+            ?.prefixHttpsIfNecessary()
+
+        // We use the requested relying party for the basis of the origin for privileged apps to
+        // ensure that related-origin requests are process successfully. In the future, the SDK
```
SaintPatrck: ⛏️ 

```suggestion
        // ensure that related-origin requests are processed successfully. In the future, the SDK
```

app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt
SaintPatrck: Agreed. Is there a ticket we can reference here?

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck & @iinuwa

PR #6792, opened 2026-04-13
Chore: Remove UI decisions based on portrait vs landscape

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR removes the `isPortrait` checks from the Authenticator app and brings it inline with the Password Manager by using the window size to make the same determinations.


SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6805, opened 2026-04-16
PM-35273: feat: Add support for SDK API calls by providing base urls

david-livefront:
## 🎟️ Tracking

[PM-35273](https://bitwarden.atlassian.net/browse/PM-35273)

## 📔 Objective

This PR adds support for instantiating the Bitwarden SDK Client with `ClientSettings` allowing the SDK to make appropriate network requests.

[PM-35273]: https://bitwarden.atlassian.net/browse/PM-35273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManager.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManager.kt
@@ -11,7 +11,20 @@ interface SdkClientManager {
      * Returns the cached [Client] instance for the given [userId], otherwise creates and caches
      * a new one and returns it.
      */
-    suspend fun getOrCreateClient(userId: String?): Client
+    suspend fun getOrCreateClient(userId: String): Client
+
+    /**
+     * Helper function to retrieve a new instance of the [Client] and use it in the given [block].
+     * This client is never persisted after the [block] completes.
+     *
+     * @param userId The used to create the [Client]. If null, the SDK is unassociated with a user.
+     * @param accessToken The access token used in network requests.
+     */
+    suspend fun <T> singleUseClient(
+        userId: String? = null,
+        accessToken: String? = null,
+        block: suspend Client.() -> T,
+    ): T
 
     /**
      * Clears any resources from the [Client] associated with the given [userId] and removes it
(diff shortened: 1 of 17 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManagerImpl.kt
```diff
@@ -32,7 +32,7 @@ class SdkClientManagerImpl(
             }
     },
 ) : SdkClientManager {
-    private val userIdToClientMap = mutableMapOf<String?, Client>()
+    private val userIdToClientMap = mutableMapOf<String, Client>()
```
david-livefront, line 41: Previously, we would create a single client that was unassociated with a user and reuse it. This will no longer work because we need to make sure the base URLs are appropriately set based on the current settings.

So now we never store the unassociated client.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManagerImpl.kt
```diff
@@ -44,9 +44,13 @@ class SdkClientManagerImpl(
     }
 
     override suspend fun getOrCreateClient(
-        userId: String?,
+        userId: String,
     ): Client = userIdToClientMap.getOrPut(key = userId) { clientProvider(userId) }
 
+    override suspend fun <T> singleUseClient(
+        block: suspend Client.() -> T,
+    ): T = clientProvider(null).use { it.block() }
```
david-livefront: I have completely removed the ability to fetch a client for a null UserID and added this function to ensure that a short lived client is always closed when it is done with its task.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManager.kt
```diff
@@ -11,7 +11,20 @@ interface SdkClientManager {
...
+    /**
+     * Helper function to retrieve a new instance of the [Client] and use it in the given [block].
+     * This client is never persisted after the [block] completes.
+     *
+     * @param userId The used to create the [Client]. If null, the SDK is unassociated with a user.
+     * @param accessToken The access token used in network requests.
+     */
+    suspend fun <T> singleUseClient(
+        userId: String? = null,
+        accessToken: String? = null,
```
david-livefront, line 25: The need for this access token is specifically for intermediate states during the login flow.

The login process has not yet completed, so we have not persisted the access token. The SDK still requires the access token to do it's job properly, so we provide it manually here.

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6806, opened 2026-04-16
Fix to set latest on PWM release notes only

mpbw2:
## 📔 Objective

Occasionally Authenticator is marked as 'Latest' when the release notes are taken out of draft, which shouldn't happen (we only set `Latest` on PWM). This fixes that.

Reasoning per claude:

`gh release edit` without `--latest` may send a `make_latest:true` default to the GitHub API, which would override the `--latest=false` set during draft creation. By explicitly forwarding the same `is_latest_release` value, the Authenticator draft release will consistently remain `make_latest=false`.


mpbw2:
This is only an issue when setting `--draft=false` which doesn't happen in this flow. Abandoning to change [elsewhere](https://github.com/bitwarden/gh-actions/pull/717).

PR #6808, opened 2026-04-17
chore: Use private key accessor

david-livefront:
## 🎟️ Tracking

NA

## 📔 Objective

This PR updates the way we access the private key from the login response to ensure we are always attempting to retrieve it from the `accountKeys` before falling back to the legacy `privateKey` property. Additionally, a new `AccountKeysJson.toAccountCryptographicState` extension was added to simplify the transformation to a `WrappedAccountCryptographicState`.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/util/AccountKeysJsonExtensions.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/util/AccountKeysJsonExtensions.kt
@@ -1,0 +1,23 @@
+package com.x8bit.bitwarden.data.auth.repository.util
+
+import com.bitwarden.core.WrappedAccountCryptographicState
+import com.bitwarden.network.model.AccountKeysJson
+import com.x8bit.bitwarden.data.vault.repository.util.createWrappedAccountCryptographicState
+
+/**
+ * Creates a [WrappedAccountCryptographicState] based on the available cryptographic parameters.
+ *
+ * Returns [WrappedAccountCryptographicState.V2] if signing key, signed public key, and security
+ * state are all present, otherwise returns [WrappedAccountCryptographicState.V1].
+ *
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/util/AccountKeysJsonExtensions.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Returns [WrappedAccountCryptographicState.V2] if signing key, signed public key, and security
+ * state are all present, otherwise returns [WrappedAccountCryptographicState.V1].
+ *
+ * @receiver The users account keys.
+ * @param privateKey The user's wrapped private key.
+ */
+fun AccountKeysJson?.toAccountCryptographicState(
+    privateKey: String,
+): WrappedAccountCryptographicState = createWrappedAccountCryptographicState(
```
david-livefront, line 18: This function is gonna get more use in future v2 encryption PRs

app/src/test/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryTest.kt
```diff
@@ -1977,7 +1977,7 @@ class AuthRepositoryTest {
             coVerify { identityService.preLogin(email = EMAIL) }
             fakeAuthDiskSource.assertPrivateKey(
                 userId = USER_ID_1,
-                privateKey = "privateKey",
+                privateKey = "mockWrappedPrivateKey-1",
```
david-livefront, line 1980: These test updates are because we now favor the private key from the `AccountKeysJson` instead of the private key directly on the login response.

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6810, opened 2026-04-19
[PM-35352] [PM-21264] Group card numbers in vault item display

### LINKED ISSUE #5138: Inconsistency: Card numbers are not in groups (while in the other apps they are in groups of four)
### Steps To Reproduce

1. Go to an existing Card item with a Card number - or create a new Card item with a Card number.
2. Open the "View item" view for that Card.
3. Toggle visibility of the Card number.

### Expected Result

Card numbers are grouped - like in the browser extension, in the desktop app and in the web vault.

### Actual Result

Card numbers are not grouped. An inconsistency, as they **are** grouped in the other BW apps.

### Screenshots or Videos

Android mobile app (2025.4.0) - the numbers are NOT grouped:

Image

&nbsp;

Browser extension (2025.4.0) - the numbers ARE grouped:

Image

&nbsp;

(Windows) Desktop app (2025.3.0) - the numbers ARE grouped:

Image

&nbsp;

Web vault (2025.4.1) - the numbers ARE grouped:

Image

### Additional Context

_No response_

### Build Version

2025.4.0

EU

### Self-host Server Version

_No response_

### Environment Details

- Fairphone 5
- Android 14

### Issue Tracking Info

Cooperr06:
## 🎟️ Tracking

https://github.com/bitwarden/android/issues/5138

## 📔 Objective

This PR implements brand-aware card number formatting in the vault item detail view to resolve the inconsistency between the Android app and other Bitwarden platforms.

**Changes**:

- **Formatting Logic**: Added formatCardNumber utility to handle brand-specific spacing:
  - **Amex**: 4-6-5 grouping.
  - **Diners Club (14-digit)**: 4-6-4 grouping.
  - **Standard**: Default 4-4-4-4 grouping for other brands.
- **UI Integration**: Applied the formatting in CipherViewExtensions.kt so users see the grouped numbers in the item detail screen.
- **Clipboard Compatibility**: Updated VaultItemViewModel.kt to ensure that clicking "Copy Number" still copies only the raw digits (no spaces). This maintains compatibility with web forms that may not handle spaces correctly.
- **Unit Tests**: Added comprehensive tests in CardNumberUtilsTest.kt to verify formatting across multiple card brands.

### Before

Screenshot_20260419_185755

#### Default Card Number

Screenshot_20260419_185628

#### American Express Card Number

Screenshot_20260419_185603

#### Diners Club Card Number

Screenshot_20260419_185644


Cooperr06:
Fixed the Codecov findings

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/util/CardNumberUtils.kt
```diff
@@ -1,8 +1,56 @@
...
+
+@Suppress("MagicNumber")
+private fun String.formatDinersClub(): String {
+    val result = StringBuilder()
+    for (i in indices) {
+        if (i == 4 || i == 10) result.append(" ")
+        result.append(this[i])
+    }
+    return result.toString()
+}
```
SaintPatrck: 🎨 These are 1:1 copies. Can we combine them into something like `formatWithSpacesAt(vararg positions: Int): String`? That could replace `formatDefault()` too.

Cooperr06: Per your suggestion, I replaced the duplicated formatting logic with a shared `formatWithSpacesAt(vararg positions: Int)` helper and removed the `formatDefault()` wrapper. `formatCardNumber()` now computes the default 4-digit grouping once and reuses it for the non-AMEX / non-14-digit-Diners paths.

SaintPatrck (dismissed):
LGTM.

@david-livefront any concerns?

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/util/CardNumberUtils.kt
```diff
@@ -1,8 +1,53 @@
...
+ * All other brands are formatted in groups of 4 digits.
+ *
+ * @return The formatted card number.
+ */
+@Suppress("MagicNumber")
+fun String.formatCardNumber(): String {
+    val digits = sanitizeCardNumber()
+    val brand = digits.detectCardBrand()
+    val defaultFormatted =
+        digits.formatWithSpacesAt(*(4 until digits.length step 4).toList().toIntArray())
```
david-livefront: Small nit but can we format this with a chain? Makes it easier to read.

```kotlin
val defaultFormatted = digits
    .formatWithSpacesAt(*(4 until digits.length step 4)
    .toList()
    .toIntArray())
```

Cooperr06: Since the `.toList().toIntArray()` calls are made on the `*(...)` statement, I would rather choose one of these formats instead:

```kt
val defaultFormatted = digits.formatWithSpacesAt(
    *(4 until digits.length step 4).toList().toIntArray()
)
```

or

```kt
val defaultFormatted = digits.formatWithSpacesAt(
    *(4 until digits.length step 4)
        .toList()
        .toIntArray()
)
```

david-livefront: Ha, I definitely misread (and miswrote) it the first time.

Both of the options you provided look good to me, take your pick. Be sure to add a trailing comma too, otherwise CI will yell at you 😄

Cooperr06: I went with the second option, thanks for the tip with the trailing comma 👍

SaintPatrck (changes requested):
⚠️ I checked grouping logic in `ios` and `clients`. They handle Maestro, UnionPay, and Diners Club differently. We need to follow their implementation so the grouping is consistent everywhere.

<details>
<summary>Recommended fix</summary>

Replace `formatWithSpaces((int)` with `formattingBlocks(digitCount)` and update `formatCardNumber()` to mirror the other platforms.

```kotlin
  @Suppress("MagicNumber")
  private fun VaultCardBrand.formattingBlocks(digitCount: Int): List<Int> {
      val default = listOf(4, 4, 4, 4)
      return when (this) {
          VaultCardBrand.AMEX -> listOf(4, 6, 5)
          VaultCardBrand.DINERS_CLUB -> listOf(4, 6, 4)
          VaultCardBrand.MAESTRO -> when (digitCount) {
              13 -> listOf(4, 4, 5)
              15 -> listOf(4, 6, 5)
              19 -> listOf(4, 4, 4, 4, 3)
              else -> default
          }
          VaultCardBrand.UNIONPAY -> if (digitCount == 19) listOf(6, 13) else default
          else -> default
      }
  }

  fun String.formatCardNumber(): String {
      val digits = sanitizeCardNumber()
      if (digits.isEmpty()) return this
      val blocks = digits.detectCardBrand().formattingBlocks(digitCount = digits.length)
      return digits.chunkByBlocks(blocks).joinToString(separator = " ")
  }

  private fun String.chunkByBlocks(blocks: List<Int>): List<String> = buildList {
      var remaining = this@chunkByBlocks
      for (size in blocks) {
          if (remaining.isEmpty()) return@buildList
          add(remaining.take(size))
          remaining = remaining.drop(size)
      }
      if (remaining.isNotEmpty()) add(remaining)
  }
```

</details>

Edit: Updated fix to be more idiomatic.

Cooperr06:
I implemented your proposed changes and added tests accordingly.

I changed the Diners Club formatting blocks so that the 4-6-4 format is selected only when the card number has 14 digits. According to [this blog](https://groundlabs.com/blog/anatomy-of-a-credit-card#:~:text=Diners%20Club%20US,16) from May 2022, Diners Club US and Canada card numbers have 16 digits, so the default formatting should be applied:

```kotlin
private fun VaultCardBrand.formattingBlocks(digitCount: Int): List<Int> {
    val default = listOf(4, 4, 4, 4)
    return when (this) {
        [...] -> [...]
        VaultCardBrand.DINERS_CLUB -> if (digitCount == 14) listOf(4, 6, 4) else default
        [...] -> [...]
        else -> default
    }
}
```

This is also done in `clients` [here](https://github.com/bitwarden/clients/blob/db9f14601f5173d95ad5454f71603d9b4663a2df/libs/vault/src/pipes/credit-card-number.pipe.ts#L21), where the format is applied only if the number of digits is 14.

SaintPatrck (approved):
Thanks for making those last minute changes. This looks great. @david-livefront any final thoughts?

david-livefront:
> Thanks for making those last minute changes. This looks great. @david-livefront any final thoughts?

@SaintPatrck this all looks good to me!

david-livefront (approved):

PR #6828, opened 2026-04-23
[PM-32009] feat: Add infrastructure for new vault item types

SaintPatrck:
## 🎟️ Tracking

[PM-32009](https://bitwarden.atlassian.net/browse/PM-32009)

## 📔 Objective

First of three stacked PRs introducing Bank Account, Driver's License, and Passport cipher types. This PR lands the foundation only — network/model plumbing, enums, feature-flag registration, and exhaustive-when branches — so the UI and vault-integration PRs can be reviewed in isolation on top of it.

The entire effort ships behind the `pm-32009-new-item-types` feature flag so infrastructure can merge well ahead of user-visible surface area.

**Why it looks the way it does**

- **SDK decryption gracefully skips unsupported types.** The Bitwarden SDK does not yet expose types for these ciphers, so sync could otherwise fail open on accounts that already hold server-side rows with the new `CipherType` values. The mapping layer defensively drops unknown types instead of throwing, keeping sync resilient while SDK variants are being added.
- **Exhaustive `when` branches are updated across every affected ViewModel.** Adding the enum values forces compile-time coverage everywhere — reviewers can trust that no call site silently falls into an `else`.
- **`LinkedIdTypeJson` entries (600–812) mirror the server's contract** so linked-field resolution works the moment data arrives, without a second round-trip PR.

**Stacked on:** `main`
**Followed by:** `new-item-types/phase-05-07_cipher-type-ui` (UI screens)

[PM-32009]: https://bitwarden.atlassian.net/browse/PM-32009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6829, opened 2026-04-23
[PM-32009] feat: Add add/edit and view UI for new cipher types

SaintPatrck:
## 🎟️ Tracking

[PM-32009](https://bitwarden.atlassian.net/browse/PM-32009)

## 📔 Objective

Second of three stacked PRs. Builds the Compose add/edit forms, view screens, and associated handlers for Bank Account, Driver's License, and Passport on top of the infrastructure landed in #6828.

Kept separate from the vault-integration PR so reviewers can focus on screen-level concerns — field layouts, masking, handlers, and ViewModel state — without also triaging listing/search/navigation changes.

**Why it looks the way it does**

- **Bank Account account number and PIN default to hidden with reveal controls**, matching the existing Card type's treatment of sensitive fields. Copy actions are wired through the same handler pattern so future auditing and analytics stay consistent.
- **Driver's License and Passport use month/year pickers** (rather than a free-form date field) because issuance/expiration at the document level is typically known only to month precision, and this avoids introducing a day-level schema the server doesn't track.
- **Per-type handler classes (`Vault…TypeHandlers`)** isolate each new type's action plumbing from `VaultAddEditViewModel` and `VaultItemViewModel`, which already carry significant complexity for Login/Card/Identity/SecureNote. Keeping new surface area out of those files reduces merge risk with parallel cipher-type work.
- **Nothing is user-visible yet:** the feature flag from #6828 still gates entry points. This PR can merge independently of the vault-integration PR.

**Stacked on:** #6828
**Followed by:** `new-item-types/phase-08-11_vault-integration` (vault listing, search, item-type selection)

## 📸 Screenshots

_To be added before marking ready for review._

[PM-32009]: https://bitwarden.atlassian.net/browse/PM-32009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SaintPatrck:
Split into smaller chunks.

PR #6830, opened 2026-04-23
[PM-32009] feat: Integrate new cipher types into vault listing and search

SaintPatrck:
## 🎟️ Tracking

[PM-32009](https://bitwarden.atlassian.net/browse/PM-32009)

## 📔 Objective

Third of three stacked PRs. Wires the Bank Account, Driver's License, and Passport types built in #6829 into the vault listing, search, and new item-type selection flow. After this PR merges (and the `pm-32009-new-item-types` flag is enabled), users can discover, filter, and create the new cipher types end-to-end.

**Why it looks the way it does**

- **Item-type selection becomes a screen, not a dialog, when the flag is on.** Six choices (Login, Card, Identity, Secure Note, plus the three new types) no longer fit a bottom-sheet affordance without crowding, and a dedicated screen gives the new types room for iconography and descriptions. The dialog path is preserved unchanged for flag-off users.
- **SDK filter predicates currently return `false` for the new types.** The SDK's `CipherListViewType` does not yet expose variants for these ciphers, and returning `false` is a safer default than a speculative mapping that might mis-categorize items once the SDK lands. Tracked under PM-34060, PM-32691, and PM-32694.
- **"Note" renamed to "Secure note" in empty-state strings** so the empty states read consistently alongside the new types in the selection screen.
- **Feature-flag gating is applied at the data-extension layer** rather than individual call sites so the zero-group-counts behavior is uniform across the vault surface and can be removed in a single step when the flag is retired.

**Stacked on:** #6829
**Follow-up:** SDK mapping work tracked under PM-34060 / PM-32691 / PM-32694

## 📸 Screenshots

_To be added before marking ready for review._

[PM-32009]: https://bitwarden.atlassian.net/browse/PM-32009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SaintPatrck:
Split into smaller chunks

PR #6834, opened 2026-04-24
[PM-30625] fix: Filtering empty totp from count on vault screen

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30625

## 📔 Objective
SDK is now returning a `MissingSecret` error when the TOTP value is empty. We are dealing with errors properly although we are not validating on Vault screen if the TOTP values are valid so the count does not match the number of items if some of them are invalid. This checks them before in order to display the correct count.

This PR is also not sending the `MissingSecret` error to crashlytics but keeping them on FlightRecorder


app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/VaultSdkSourceImpl.kt
```diff
@@ -432,13 +434,21 @@ class VaultSdkSourceImpl(
...
             .vault()
             .generateTotpCipherView(
                 view = cipherListView,
                 time = time,
             )
+    }.onFailure { throwable ->
+        val isMissingSecret = throwable is BitwardenException.Totp &&
+            throwable.v1 is TotpException.MissingSecret
+        if (isMissingSecret) {
+            Timber.w("TOTP generation skipped: missing secret")
```
david-livefront: Do we need this here?

Do the regular `runCatchingWithLogs` functionality not cover this?

aj-rosado: This is to avoid throwing to the Firebase's crashlytics to not fill it with this error. This way it will be presented on FlightRecorder but not on Crashlytics.

iOS applied the same logic for this scenario.

david-livefront: Gotcha

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/datasource/sdk/VaultSdkSourceImpl.kt
```diff
@@ -432,13 +434,21 @@ class VaultSdkSourceImpl(
...
         time: Instant?,
-    ): Result<TotpResponse> = runCatchingWithLogs {
+    ): Result<TotpResponse> = runCatching {
         getClient(userId = userId)
             .vault()
             .generateTotpCipherView(
                 view = cipherListView,
                 time = time,
             )
+    }.onFailure { throwable ->
```
david-livefront: Can we move the `.failure {` down a line.

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -279,6 +283,12 @@ class VaultViewModel @Inject constructor(
             .onEach(::sendAction)
             .launchIn(viewModelScope)
 
+        authCodesFlow
+            .mapNotNull { (it as? DataState.Loaded)?.data }
+            .map { VaultAction.Internal.AuthCodesReceive(it) }
```
david-livefront: Can you elaborate on why this is needed?

It seems very inefficient to observe all the realtime updates of TOTP codes just to get a count.

aj-rosado: We cannot rely on directly on the encrypted vault data because a TOTP that is empty after decrypting will appear as not empty when encrypted.

Although you are correct, we should not and definitely do not need to listen to all the updates.

david-livefront: Can we go back to design and get an error state or something.

We should not add such a heavy operation to this screen for an edge-case scenario like this.

david-livefront: I believe iOS validates the TOTP codes on the VaultScreen by generating a single code. We could do that in the `VaultDataExtensions` and that would be significantly more efficient.

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -1248,9 +1248,28 @@ class VaultViewModel @Inject constructor(
...
-        updateViewState(
-            vaultData = action.vaultData,
-        )
+        updateViewState(vaultData = action.vaultData)
+
+        if (action.vaultData.data != null) {
+            viewModelScope.launch {
+                updateTotpCodesCount(
+                    vaultRepository.countValidTotpCiphers(
+                        isPremium = state.isPremium,
```
david-livefront: Can we just flatmap this request in with the `vaultDataStateFlow`. 

```kotlin
        vaultRepo
            .vaultDataStateFlow
            .map { vaultRepo.countValidTotpCiphers() to it }
            .map { (count, vaultData) -> VaultAction.Internal.VaultDataReceive(count, vaultData) }
            .onEach {
                // When the vault data is received, the current activity is about to
                // be recreated. Adding this delay prevents the dialogs from disappearing.
                delay(VAULT_DATA_RECEIVED_DELAY)
                trySendAction(it)
            }
            .launchIn(viewModelScope)
```

Then the value is never nullable.

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -438,6 +438,31 @@ class VaultRepositoryImpl(
...
+            ?: return 0
+
+        return decryptedCipherList
+            .count { cipher ->
+                (isPremium || cipher.organizationUseTotp) &&
+                    vaultSdkSource.generateTotpForCipherListView(
+                        userId = userId,
+                        cipherListView = cipher,
+                        time = time,
+                    ).isSuccess
```
david-livefront: Can you use proper chain formatting:

```kotlin
vaultSdkSource
    .generateTotpForCipherListView(
        userId = userId,
        cipherListView = cipher,
        time = time,
    )
    .isSuccess
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -438,6 +438,32 @@ class VaultRepositoryImpl(
...
+
+        return decryptedCipherList
+            .count { cipher ->
+                (isPremium || cipher.organizationUseTotp) &&
+                    vaultSdkSource.generateTotpForCipherListView(
+                        userId = userId,
+                        cipherListView = cipher,
+                        time = time,
+                    )
+                        .isSuccess
```
david-livefront, line 463: This is still not right:

```kotlin
vaultSdkSource
    .generateTotpForCipherListView(
        userId = userId,
        cipherListView = cipher,
        time = time,
    )
    .isSuccess
```

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModelTest.kt
```diff
@@ -4006,6 +4111,7 @@ private val DEFAULT_USER_STATE = UserState(
 private fun createMockVaultState(
     viewState: VaultState.ViewState,
     dialog: VaultState.DialogState? = null,
+    totpCodesCount: Int? = null,
```
david-livefront: This should not be here, the value should just be set on the passed in `viewState`

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultDataExtensions.kt
```diff
@@ -47,6 +47,7 @@ fun VaultData.toViewState(
     vaultFilterType: VaultFilterType,
     restrictItemTypesPolicyOrgIds: List<String>,
     isArchiveEnabled: Boolean,
+    totpItemsCount: Int? = null,
```
david-livefront: Can we make this nonnull, Once it's part of the state, there shouldn't be a reason to not have it

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -1556,6 +1573,7 @@ data class VaultState(
     val isIntroducingArchiveActionCardDismissed: Boolean,
     val isPremiumUpgradeBannerEligible: Boolean = false,
     val isAwaitingKdfSync: Boolean = false,
+    val totpCodesCount: Int? = null,
```
david-livefront: Can we make this nonnull. It will be populated once the data arrives

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -1248,9 +1254,18 @@ class VaultViewModel @Inject constructor(
...
+        updateViewState(vaultData = action.vaultData)
+        updateTotpCodesCount(count = action.count)
+    }
+
+    private fun updateTotpCodesCount(count: Int) {
+        mutableStateFlow.update { state ->
+            val content = state.viewState as? VaultState.ViewState.Content
+            state.copy(
+                totpCodesCount = count,
+                viewState = content?.copy(totpItemsCount = count) ?: state.viewState,
```
david-livefront: Can we just combine this with `updateViewState`

```kotloin
updateViewState(
    vaultData = action.vaultData,
    totpItemsCount = action.count,
)
```

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -2369,6 +2387,7 @@ sealed class VaultAction {
          * Indicates a vault data was received.
          */
         data class VaultDataReceive(
+            val count: Int,
```
david-livefront: Can we be more specific, maybe call this `totpItemsCount`

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -624,6 +632,7 @@ class VaultViewModel @Inject constructor(
         // Re-process the current vault data with the new filter
         updateViewState(
             vaultData = vaultRepository.vaultDataStateFlow.value,
+            totpItemsCount = state.totpItemsCount,
         )
```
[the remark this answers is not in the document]
david-livefront: Claude made a good point here, instead of returning a count and passing that around, we could pass back a list of valid ids and use that to compare with the active number of ciphers inside the `toViewState` function.

```kotlin
 val totpItemsCount = activeCipherViews.filter { it.id in validTotpIds }
```

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
```diff
@@ -449,19 +449,21 @@ class VaultRepositoryImpl(
...
+                    vaultSdkSource
+                        .generateTotpForCipherListView(
+                            userId = userId,
+                            cipherListView = cipher,
+                            time = time,
+                        )
+                        .isSuccess
+            }
+            .mapNotNull { it.id }
+            .toSet()
```
david-livefront, line 466: Why a set?

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultDataExtensions.kt
```diff
@@ -135,7 +135,7 @@ fun VaultData.toViewState(
 
         VaultState.ViewState.Content(
             itemTypesCount = itemTypesCount,
-            totpItemsCount = totpItemsCount,
+            totpItemsCount = activeCipherViews.count { it.id in validTotpIds },
```
david-livefront, line 138: 👍

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultDataExtensions.kt
```diff
@@ -46,7 +46,7 @@ fun VaultData.toViewState(
     baseIconUrl: String,
     vaultFilterType: VaultFilterType,
     restrictItemTypesPolicyOrgIds: List<String>,
-    totpItemsCount: Int,
+    validTotpIds: Set<String>,
```
david-livefront, line 49: Can you make sure this is an `ImmutableSet` or `ImmutableList`

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/VaultRepositoryImpl.kt
aj-rosado: As we are comparing if the id is present and not "listing" the data, a set would make sense here.

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/util/VaultDataExtensions.kt
david-livefront: Sorry, this should apply to the state not the function.

https://github.com/bitwarden/android/pull/6834#diff-be10566d2d8b0351b08882dce4d1c23de9e09ee543d67f21fd72c92642219706R1590

david-livefront (approved):

PR #6779, opened 2026-04-09
[BWA-99] feat: Add next TOTP code preview in Authenticator

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BWA-99

## 📔 Objective

Add a "next TOTP code" preview to the Authenticator app. When a verification code is about to expire (< 10 seconds remaining), the upcoming code is displayed below the current one, allowing users to pre-read the next code without waiting for rotation.

**Key behaviors:**
- Next code appears only when `timeLeftSeconds < 10`
- Displayed in smaller, secondary-color text below the current code
- Controlled by a "Show next code" toggle in Settings > Data (default: off)
- Tapping an item still copies the **current** code

**Implementation details:**
- Data layer: Second `generateTotp()` SDK call with `currentTime + periodSeconds`
- Settings: Full `DiskSource → Repository → ViewModel → Screen` chain for the toggle
- UI: Conditional rendering in `VaultVerificationCodeItem` composable
- Wired through both `ItemListingViewModel` and `ItemSearchViewModel`

### Settings

| Figma | Actual |
|--------|--------|
|  |  | 

*Disclaimer: Figma includes the new Switch row, but other settings are outdated.*

### Upcoming code

<video src="https://github.com/user-attachments/assets/b8337cce-0b52-4c17-9ffb-623f7a460397" width="365" />

PR #6842, opened 2026-04-27
[BWA-99] feat: Add next TOTP code preview in Authenticator

mpbw2:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BWA-99

## 📔 Objective

Shows the upcoming TOTP code inline below the current code during the final 5 seconds of each period, so users can copy the next code before it becomes active without waiting for the rollover.

**Changes:**
- `VerificationCodeItem` — added `nextCode: String?` field
- `TotpCodeManagerImpl` — computes next code via a second `generateTotp(uri, time + period)` call at each period boundary; silently `null` on SDK failure
- `VerificationCodeDisplayItem` — added `nextAuthCode: String?` field
- `VerificationCodeItemExtensions` — maps `nextCode` → `nextAuthCode` in `toDisplayItem()`
- `VaultVerificationCodeItem` — renders next code below current code when `timeLeftSeconds <= 5`; applies to both item listing and search screens
- New `next_verification_code` string resource (`"Next: %1$s"`) in `:ui` module

### DIFF
--- a/authenticator/src/main/kotlin/com/bitwarden/authenticator/data/authenticator/manager/TotpCodeManagerImpl.kt
+++ b/authenticator/src/main/kotlin/com/bitwarden/authenticator/data/authenticator/manager/TotpCodeManagerImpl.kt
@@ -106,16 +106,21 @@ class TotpCodeManagerImpl @Inject constructor(
                 authenticatorSdkSource
                     .generateTotp(item.otpUri, dateTime)
                     .onSuccess { response ->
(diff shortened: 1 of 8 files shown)


gabrbrand:
I don't know about any internal matters, but isn't this feature already implemented with the merged PR #6779.

mpbw2:
@gabrbrand That's correct, this was only for internal testing purposes.

PR #6856, opened 2026-04-30
llm: Expand string resource naming convention in CLAUDE.md

aj-rosado:
## 🎟️ Tracking

Internal tooling improvement — no Jira ticket.

## 📔 Objective

Expands the string resource guidance in `CLAUDE.md` to include the naming convention: each resource name should be derived from its own text content in `snake_case`, rather than using generic suffixes like `_message` or `_title`.

For example:
- `one_or_more_email_addresses_are_incorrect` ✅
- `invalid_email_addresses_message` ❌

This helps Claude (and contributors) produce consistent, self-documenting string resource names from the start, reducing review churn.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -99,7 +99,7 @@ User Request (UI Action)
 - **Formatter**: Android Studio with `bitwarden-style.xml` | **Line Limit**: 100 chars | **Detekt**: Enabled
 - **Naming**: `camelCase` (vars/fns), `PascalCase` (classes), `SCREAMING_SNAKE_CASE` (constants), `...Impl` (implementations)
 - **KDoc**: Required for all public APIs
-- **String Resources**: Add new strings to `:ui` module (`ui/src/main/res/values/strings.xml`). Use typographic quotes/apostrophes (`"` `"` `'`) not escaped ASCII (`\"` `\'`)
+- **String Resources**: Add new strings to `:ui` module (`ui/src/main/res/values/strings.xml`). Use typographic quotes/apostrophes (`"` `"` `'`) not escaped ASCII (`\"` `\'`). Name each resource from its own text content in `snake_case` — not with generic suffixes (`_message`, `_title`). E.g., `one_or_more_email_addresses_are_incorrect`, not `invalid_email_addresses_message`.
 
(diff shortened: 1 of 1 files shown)


.claude/CLAUDE.md
```diff
@@ -99,7 +99,7 @@ User Request (UI Action)
 - **Formatter**: Android Studio with `bitwarden-style.xml` | **Line Limit**: 100 chars | **Detekt**: Enabled
 - **Naming**: `camelCase` (vars/fns), `PascalCase` (classes), `SCREAMING_SNAKE_CASE` (constants), `...Impl` (implementations)
 - **KDoc**: Required for all public APIs
-- **String Resources**: Add new strings to `:ui` module (`ui/src/main/res/values/strings.xml`). Use typographic quotes/apostrophes (`"` `"` `'`) not escaped ASCII (`\"` `\'`)
+- **String Resources**: Add new strings to `:ui` module (`ui/src/main/res/values/strings.xml`). Use typographic quotes/apostrophes (`"` `"` `'`) not escaped ASCII (`\"` `\'`). Name each resource from its own text content in `snake_case` — not with generic suffixes (`_message`, `_title`). E.g., `one_or_more_email_addresses_are_incorrect`, not `invalid_email_addresses_message`.
```
SaintPatrck, line 102: 🌱 Not requesting changes now, but let's keep an eye on this section. If it gets any larger we should consider extracting the rules into an appropriate skill (`implementing-android-code` or a new `modifying-string-resources` skill) and/or moving them to `STYLE_AND_BEST_PRACTICES.md`. There are no String Resource rules in `STYLE_AND_BEST_PRACTICES.md`, so that may be why it's not following our convention in the first place.

SaintPatrck (approved):

.claude/CLAUDE.md
aj-rosado: Claude actually wanted to add it to the `STYLE_AND_BEST_PRACTICES.md`.

SaintPatrck: In that case, let go ahead and move it. If we see Claude still "misbehaving" we can move it back into CLAUDE.md or a skill.

SaintPatrck: Ha. I was too slow. 😆

aj-rosado: Will open a new PR with it

PR #6863, opened 2026-05-01
[PM-36057] feat: Add Upgraded to Premium acknowledgment

SaintPatrck:
## 🎟️ Tracking

[PM-36075](https://bitwarden.atlassian.net/browse/PM-36075)

## 📔 Objective

Acknowledge a Free → Premium transition with two surfaces depending on where the user is when the upgrade is detected:

- **On-Plan**: full-screen celebration when the upgrade completes while the user is on the Plan screen. Dismiss is driven by `PlanMode` so a Modal entry (CTA host like Send file dialog) returns to the CTA host (pop celebration + Plan), and a Standard entry (Settings → Plan) leaves the user on Plan with the now-up-to-date Premium content.
- **Off-Plan**: per-screen action card on the four root-nav destinations (Vault, Send, Generator, root Settings) for upgrades detected via push while the user is elsewhere. The card is queued via `pending=true` in `SettingsDiskSource` and renders on the next visit to one of the four surfaces.

Both surfaces converge on `PremiumStateManager.dismissUpgradedToPremiumCard()` to preserve the single-acknowledgment invariant: a user who saw the celebration will not also see an action card later, and vice versa.

`PlanViewModel.onPremiumUpgradeSuccess()` transitions the viewstate to `Premium` and refreshes the subscription so the Standard dismiss lands on the up-to-date Plan UI rather than the stale Free state.

The celebration registration lives in `vaultUnlockedGraph` alongside `PlanRoute.Modal`. Standard reaches it through the existing settings → vaultUnlockedNavBar → vaultUnlockedGraph callback chain — the same pattern used for `exportVault`, `folders`, etc. `MainActivity` is uninvolved.

### Upgraded to Premium screen

https://github.com/user-attachments/assets/7dd61ab9-4c4f-4317-9770-a98b0ef5b258

### Upgraded to Premium Action Card

[PM-36057]: https://bitwarden.atlassian.net/browse/PM-36057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-36075]: https://bitwarden.atlassian.net/browse/PM-36075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -88,6 +95,67 @@ class PremiumStateManagerImpl(
...
+            }
+            .distinctUntilChanged()
+            .scanPairs()
+            .onEach { (previous, current) ->
+                if (current == null) return@onEach
+                val (currentUserId, currentIsPremium) = current
+                if (!currentIsPremium) return@onEach
+                // Same user transitioning from non-premium to premium counts as an upgrade.
+                if (previous?.first == currentUserId && previous.second == false) {
+                    markUpgradedToPremiumCardPending(userId = currentUserId)
```
david-livefront, line 152: Will this cause any issue if a user becomes premium in the background?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanNavigation.kt
```diff
@@ -88,9 +92,13 @@ fun NavGraphBuilder.planDestination(
  */
 fun NavGraphBuilder.planModalDestination(
     onNavigateBack: () -> Unit,
+    onNavigateToCelebration: () -> Unit,
```
david-livefront: There is a mismatch between theses function that navigate to `celebration` and the actual name of the screen: `UpgradedToPremiumScreen`. Can we make sure the language is consistent throughout

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/SettingsViewModel.kt
```diff
@@ -161,9 +198,16 @@ data class SettingsState(
...
     private val isMobilePremiumUpgradeEnabled: Boolean = false,
+    private val isUpgradedToPremiumCardEligible: Boolean = false,
 ) {
     val shouldShowCloseButton: Boolean = isPreAuth
 
+    /**
+     * Whether the "Upgraded to Premium" action card should be shown. The card is only visible
+     * post-authentication.
+     */
+    val shouldShowUpgradedToPremiumCard: Boolean = !isPreAuth && isUpgradedToPremiumCardEligible
```
david-livefront, line 209: Should we enumerate this for future action cards?

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/SendScreen.kt
```diff
@@ -179,32 +189,61 @@ fun SendScreen(
...
-            is SendState.ViewState.Content -> SendContent(
-                policyDisablesSend = state.policyDisablesSend,
-                state = viewState,
-                sendHandlers = sendHandlers,
-                modifier = modifier,
-            )
+        Column(modifier = Modifier.fillMaxSize()) {
+            if (state.isUpgradedToPremiumCardEligible) {
+                Spacer(modifier = Modifier.height(height = 12.dp))
+                BitwardenActionCard(
```
david-livefront: So this does not scroll with the content?

This seems different from the others, like the one on the vault screen

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
SaintPatrck: No. Background handling is intended since upgrade signals can be received at any time from a push notification or background sync. Is there something specific that stands out as a possible issue?

david-livefront: I was thinking about how this would work if you were switching between users and the new current user was upgraded while they were not the active user. Since this code requires the newly active user to also be the previous user (only the premium status changes).

But I think we are OK here since you cannot sync a background user. So switching to the new active user will cause the pending sync (from the background notification) to go through and we will find out the user is premium at that time.

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/SendScreen.kt
```diff
@@ -56,7 +56,7 @@ import kotlinx.collections.immutable.persistentListOf
 /**
  * UI for the send screen.
  */
-@Suppress("LongMethod")
+@Suppress("LongMethod", "CyclomaticComplexMethod")
```
david-livefront, line 59: Is this still needed after your recent changes?

SaintPatrck: It is

david-livefront: 😞

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -123,5 +214,17 @@ private fun DataState<VaultData>.activeVaultItemCount(): Int =
         ?.count { it.isActive }
         ?: 0
 
+/**
+ * Emits successive (previous, current) pairs from the upstream flow, starting with
+ * (`null`, first-emission).
+ */
+private fun <T : Any> Flow<T?>.scanPairs(): Flow<Pair<T?, T?>> = flow {
```
david-livefront: So weird that this doesn't just exist.

What do you think about moving it to the `FlowExtensions` and making it public?

david-livefront: Also, does something like this make sense for a generic function?

```kotlin
fun <T> Flow<T>.scanPairs(
    initial: T? = null,
): Flow<Pair<T?, T>> = flow {
    var previous: T? = initial
    collect { current ->
        emit(previous to current)
        previous = current
    }
}
```

david-livefront (approved):

PR #6867, opened 2026-05-04
[PM-34038] fix: Address card scanner QA findings

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34038

## 📔 Objective

- **PM-36412** — Atomic gate in `VaultAddEditViewModel.handleCardScanResultReceive`: require number AND month before applying, then unconditionally overwrite number / brand / month / year. Replaces per-`?:` field merging that left stale data from a prior scan. Drops CVV from the apply path so manually entered codes survive a re-scan.
- **PM-36382** — Three layered defenses in `CardTextAnalyzerImpl` before parsing: frame gating against the on-screen scan rect, orientation gating (>10° baselines rejected from `Line.cornerPoints`), and 2-of-3-frame temporal voting. Replaces the previous behavior of flattening the entire viewport and emitting the first Luhn-valid PAN found — ML Kit's `rotationDegrees`-driven auto-rotation made off-axis cards parseable. New framework-free `RecognizedText` abstraction enables plain JVM tests.
- **PM-36360 + PM-36381** — "Card scanned" snackbar on AddEdit after the scanner sheet dismisses, in lieu of a fullscreen "Fetching card details" loading screen (OCR completes immediately; the delay would be fake).
- **PM-36380** — Instruction text repositioned above the scan frame. Regression-guard test asserts `instructionBottom <= scanFrameTop` against a tagged scan-frame `Box`.
- **PM-36383** — After 5s with no successful scan, surfaces a "Hold steady and ensure all card details are visible" banner with `Modifier.semantics { liveRegion = LiveRegionMode.Polite }` for TalkBack. Partially covers PM-36384.

## 📸 Screenshots

https://github.com/user-attachments/assets/bc1acc05-58f4-4096-8901-dcd0204b6ab4

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -1940,6 +1940,9 @@ class VaultAddEditViewModel @Inject constructor(
                         expirationYear = data.expirationYear ?: cardType.expirationYear,
                     )
                 }
+                sendEvent(
+                    VaultAddEditEvent.ShowSnackbar(BitwardenString.card_scanned.asText()),
+                )
                 sendEvent(VaultAddEditEvent.FocusCardHolderName)
             }
 
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -1930,19 +1930,14 @@ class VaultAddEditViewModel @Inject constructor(
         when (val result = action.cardScanResult) {
             is CardScanResult.Success -> {
                 val data = result.cardScanData
+                val number = data.number ?: return
+                val expirationMonth = data.expirationMonth?.toExpirationMonth() ?: return
                 updateCardContent { cardType ->
                     cardType.copy(
-                        number = data.number ?: cardType.number,
-                        expirationYear = data.expirationYear
-                            ?: cardType.expirationYear,
-                        expirationMonth = data.expirationMonth
-                            ?.toExpirationMonth()
-                            ?: cardType.expirationMonth,
-                        securityCode = data.securityCode
-                            ?: cardType.securityCode,
-                        brand = data.number
-                            ?.detectCardBrand()
-                            ?: cardType.brand,
+                        number = number,
+                        brand = number.detectCardBrand(),
+                        expirationMonth = expirationMonth,
+                        expirationYear = data.expirationYear ?: cardType.expirationYear,
                     )
                 }
                 sendEvent(VaultAddEditEvent.FocusCardHolderName)
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -1931,7 +1931,7 @@ class VaultAddEditViewModel @Inject constructor(
             is CardScanResult.Success -> {
                 val data = result.cardScanData
                 val number = data.number ?: return
-                val expirationMonth = data.expirationMonth?.toExpirationMonth() ?: return
+                val expirationMonth = requireNotNull(data.expirationMonth).toExpirationMonth()
                 updateCardContent { cardType ->
                     cardType.copy(
                         number = number,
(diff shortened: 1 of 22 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
```diff
@@ -1930,21 +1930,19 @@ class VaultAddEditViewModel @Inject constructor(
         when (val result = action.cardScanResult) {
             is CardScanResult.Success -> {
                 val data = result.cardScanData
+                val number = data.number ?: return
+                val expirationMonth = data.expirationMonth?.toExpirationMonth() ?: return
```
david-livefront: I wonder if this should constitute a failure?

Should we display something to the user?

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/cardscanner/CardScanScreen.kt
```diff
@@ -87,97 +93,79 @@ fun CardScanScreen(
...
+                        .testTag("CardScanInstruction")
+                        .fillMaxWidth()
+                        .padding(all = 12.dp),
+                )
+                Box(
+                    contentAlignment = Alignment.Center,
+                    modifier = Modifier
+                        .testTag("CardScanFrame")
+                        .fillMaxSize()
+                        .navigationBarsPadding(),
```
david-livefront: Don't we want the camera viewfinder to go edge-to-edge?

This will block that

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/cardscanner/CardScanScreen.kt
```diff
@@ -87,97 +93,79 @@ fun CardScanScreen(
...
+                        modifier = Modifier.fillMaxSize(),
+                    )
+                    CardScanOverlay(
+                        overlayWidth = when (rememberWindowSize()) {
+                            WindowSize.Compact -> 300.dp
+                            WindowSize.Medium -> 250.dp
+                        },
+                        modifier = Modifier.fillMaxSize(),
+                    )
+                    if (state.showHint) {
```
david-livefront: Can we add an fade here

app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
SaintPatrck: Hmm. I was following the iOS logic and there is no error shown in this case. However, iOS does not dismiss the scanner if these are null, so this does introduce a gap between the two.

I think in order to more closely mirror iOS we need to leave the scanner open and track the expiry like we're tracking the PAN across captures. Good catch!

ui/src/test/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardScanPipelineTest.kt
```diff
@@ -0,0 +1,252 @@
...
+                        bounds = bounds,
+                        corners = listOf(
+                            ImagePoint(bounds.left, bounds.top),
+                            ImagePoint(bounds.right, bounds.top),
+                            ImagePoint(bounds.right, bounds.bottom),
+                            ImagePoint(bounds.left, bounds.bottom),
+                        ),
+                    ),
+                )
+            }.toTypedArray(),
```
david-livefront: Can we format this:

```kotlin
*lines
    .map { (text, bounds) ->
        block(
            bounds = bounds,
            line(
                text = text,
                bounds = bounds,
                corners = listOf(
                    ImagePoint(bounds.left, bounds.top),
                    ImagePoint(bounds.right, bounds.top),
                    ImagePoint(bounds.right, bounds.bottom),
                    ImagePoint(bounds.left, bounds.bottom),
                ),
            ),
        )
    }
    .toTypedArray(),
```

david-livefront (approved):

ui/src/main/kotlin/com/bitwarden/ui/platform/feature/cardscanner/util/CardTextAnalyzerImpl.kt
```diff
@@ -27,6 +53,10 @@ class CardTextAnalyzerImpl(
         TextRecognizerOptions.DEFAULT_OPTIONS,
     )
 
+    private val voteBuffer = PanVoteBuffer()
+
+    private val expiryBuffer = ExpiryBuffer()
```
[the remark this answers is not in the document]
SaintPatrck: Will be addressed in a follow-up.

PR #6870, opened 2026-05-04
Chore: Remove the unused register API

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR removes the `identityService.register` API from the app, it it no longer supported and should have been removed when the email verification feature flag was removed.


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -947,12 +946,11 @@ class AuthRepositoryImpl(
         return SwitchAccountResult.AccountSwitched
     }
 
-    @Suppress("LongMethod")
     override suspend fun register(
         email: String,
         masterPassword: String,
         masterPasswordHint: String?,
-        emailVerificationToken: String?,
+        emailVerificationToken: String,
```
david-livefront, line 953: There is no location in the app that passes in a null `emailVerificationToken` so the call to `identityService.register` never actual occurs.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/repository/AuthRepositoryImpl.kt
```diff
@@ -980,39 +978,21 @@ class AuthRepositoryImpl(
                 kdf = kdf,
             )
             .flatMap { registerKeyResponse ->
-                if (emailVerificationToken == null) {
-                    // TODO PM-6675: Remove register call and service implementation
```
david-livefront, line 984: This ticket was already marked as done

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6874, opened 2026-05-05
build: Allow logging and access to the debug menu for beta builds

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR updates the beta outputs so they have access to the debug menu. This will be useful for QA testing since we no longer deploy beta builds to the store.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -121,7 +121,7 @@ configure<ApplicationExtension> {
                 "proguard-rules.pro",
             )
 
-            buildConfigField(type = "boolean", name = "HAS_DEBUG_MENU", value = "false")
+            buildConfigField(type = "boolean", name = "HAS_DEBUG_MENU", value = "true")
             buildConfigField(type = "boolean", name = "HAS_LOGS_ENABLED", value = "false")
         }
         release {
--- a/app/src/beta/res/xml/network_security_config.xml
+++ b/app/src/beta/res/xml/network_security_config.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config xmlns:tools="http://schemas.android.com/tools">
+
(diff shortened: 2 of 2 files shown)


SaintPatrck (dismissed):

aj-rosado (dismissed):

app/src/beta/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,28 @@
...
+            <!-- Trust pre-installed CAs -->
+            <certificates src="system" />
+            <!-- Additionally trust user added CAs -->
+            <certificates
+                src="user"
+                tools:ignore="AcceptsUserCertificates" />
+        </trust-anchors>
+    </base-config>
+
+    <domain-config cleartextTrafficPermitted="false">
```
vvolkgang, line 27: Prod app has a domain config for CRL / OCSP checking, we should match it so both apps are as similar as possible.

app/build.gradle.kts
```diff
@@ -121,7 +121,7 @@ configure<ApplicationExtension> {
                 "proguard-rules.pro",
             )
 
-            buildConfigField(type = "boolean", name = "HAS_DEBUG_MENU", value = "false")
+            buildConfigField(type = "boolean", name = "HAS_DEBUG_MENU", value = "true")
             buildConfigField(type = "boolean", name = "HAS_LOGS_ENABLED", value = "false")
```
vvolkgang, line 125: @david-livefront about debug logging, would rather have it off by default as we already have dev builds for that, we can do ad-hoc builds with it on and if it becomes a regular need, wire it to our workflows.

david-livefront: Sounds good, I'll leave it disabled

vvolkgang (approved):

david-livefront:
Thanks @SaintPatrck, @aj-rosado, and @vvolkgang

PR #6879, opened 2026-05-06
Chore: Add parsing support for LocalDate

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR adds support for parsing the `LocalDate` from a network request.


SaintPatrck:
@claude your findings are acceptable. Creation points are flag-gated in all clients so it is not possible to receive them. Proper handling of the new types will be implemented in a later PR.

SaintPatrck (dismissed):

PR #2619, opened 2023-07-14
[PM-3021] Add support for Browser in /e/ OS

althafvly:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [x] Other

## Objective

Add support for [/e/ OS browser ](https://gitlab.e.foundation/e/os/browser).

## Code changes

Add foundation.e.browser as supported package.

* **file.ext:** Description of what was changed and why

## Before you submit
- Please check for formatting errors (`dotnet format --verify-no-changes`) (required)
- Please add **unit tests** where it makes sense to do so (encouraged but not required)
- If this change requires a **documentation update** - notify the documentation team
- If this change has particular **deployment requirements** - notify the DevOps team

PR #6885, opened 2026-05-08
Add support for Browser in /e/ OS

althafvly:
## 📔 Objective

Add support for [/e/ OS browser ](https://gitlab.e.foundation/e/os/browser).

### DIFF
--- a/app/src/main/assets/fido2_privileged_community.json
+++ b/app/src/main/assets/fido2_privileged_community.json
@@ -47,6 +47,22 @@
     {
       "type": "android",
       "info": {
+        "package_name": "foundation.e.browser",
+        "signatures": [
+          {
+            "build": "release",
+            "cert_fingerprint_sha256": "FD:8B:1E:55:5D:40:2A:47:ED:3B:72:F3:C4:73:82:5E:A7:D2:FD:9E:C7:AD:36:60:7C:2C:E0:04:C2:10:08:06"
+          },
+          {
+            "build": "release",
+            "cert_fingerprint_sha256": "FE:36:B0:0A:C2:F0:5F:2E:29:50:33:C6:DE:C2:4E:A4:52:0E:D4:39:24:59:39:74:8D:8F:03:A5:E0:C1:90:A9"
+          }
+        ]
+      }
+    },
+    {
+      "type": "android",
+      "info": {
         "package_name": "io.github.forkmaintainers.iceraven",
         "signatures": [
           {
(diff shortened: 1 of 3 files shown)


althafvly:
This was raised around 3 years ago and it was closed by devs without merging.

https://github.com/bitwarden/android/pull/2619

app/src/main/assets/fido2_privileged_google.json
```diff
@@ -863,6 +863,22 @@
...
+      "info": {
+        "package_name": "foundation.e.browser",
+        "signatures": [
+          {
+            "build": "release",
+            "cert_fingerprint_sha256": "FD:8B:1E:55:5D:40:2A:47:ED:3B:72:F3:C4:73:82:5E:A7:D2:FD:9E:C7:AD:36:60:7C:2C:E0:04:C2:10:08:06"
+          },
+          {
+            "build": "release",
+            "cert_fingerprint_sha256": "FE:36:B0:0A:C2:F0:5F:2E:29:50:33:C6:DE:C2:4E:A4:52:0E:D4:39:24:59:39:74:8D:8F:03:A5:E0:C1:90:A9"
```
lucasmz-dev: why are there two keys

althafvly: @lucasmz-dev /e/OS has both official and community builds, and they use different signing keys for browser.

lucasmz-dev: Is it the same for all community builds...? Because I could see this being resigned by every maintainer and it would be crazy to add every unofficial build here

lucasmz-dev: It's kind of weird that it is different at all to me. There's like prebuilts too

althafvly: @lucasmz-dev Both official and community builds are made by /e/OS team.  So it's same for all.

althafvly: Community builds are not officially tested by the team for each devices. That's why it's seperated. Unofficial builds type is usually called 'test'. Which is not included in this list as it uses public keys.

app/src/main/assets/fido2_privileged_google.json
lucasmz-dev, line 1: this is the wrong file

lucasmz-dev: also make sure to put the data in alphabetical order, so foundation.e.browser would be one of the first on the list, after com. maybe

althafvly: @lucasmz-dev its in alphabetical order now.

PR #6888, opened 2026-05-08
[PM-36867] fix: Disable card scanner on F-Droid builds

SaintPatrck:
## 🎟️ Tracking

- https://bitwarden.atlassian.net/browse/PM-36867
- https://github.com/bitwarden/android/issues/6884

## 📔 Objective

F-Droid builds were crashing on launch. `CardTextAnalyzerImpl` was constructed eagerly as a default parameter on `LocalManagerProvider` — its property initializer called ML Kit's `TextRecognition.getClient(...)` during composition setup at app startup, which fails on devices without functional Google Play Services (GrapheneOS owner profile, restricted GMS, etc.).

F-Droid policy disallows Google ML Kit, so the card scanner is disabled there entirely:

- ML Kit dependency moves from `:ui` to `:app` `standardImplementation`
- `CardTextAnalyzerImpl` is split by flavor: real ML Kit impl in `app/src/standard`, no-op stub in `app/src/fdroid` (matching the existing flavor-stub pattern used by `LogsManagerImpl`, `PlayBillingManagerImpl`, `AppReviewManagerImpl`)
- The standard impl's `recognizer` is now `by lazy` so ML Kit is never touched in `<init>`, hardening against future composition-default mistakes
- `VaultAddEditViewModel.isCardScannerEnabled` is gated on `!buildInfoManager.isFdroid` for both initial state and the runtime feature-flag-update path, so the Scan Card button never renders on F-Droid builds

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -5,6 +5,7 @@ import androidx.credentials.CreatePublicKeyCredentialRequest
 import androidx.credentials.provider.CallingAppInfo
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
+import com.bitwarden.core.data.manager.BuildInfoManager
 import com.bitwarden.core.data.manager.model.FlagKey
 import com.bitwarden.core.data.manager.toast.ToastManager
 import com.bitwarden.core.data.repository.model.DataState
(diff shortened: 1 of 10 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
```diff
@@ -185,7 +187,8 @@ class VaultAddEditViewModel @Inject constructor(
             }
 
             VaultAddEditState(
-                isCardScannerEnabled = featureFlagManager.getFeatureFlag(FlagKey.CardScanner),
+                isCardScannerEnabled = featureFlagManager
+                    .getFeatureFlag(FlagKey.CardScanner) && !buildInfoManager.isFdroid,
```
david-livefront, line 191: 👍

david-livefront (dismissed):

david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #6897, opened 2026-05-09
Use string resource for button label

gabrbrand:
## 📔 Objective

Currently a hard coded string is used for the primary button label on tutorial screens in the authenticator app, which makes it impossible to translate. A string resource in now used.


ui/src/main/res/values/strings.xml
```diff
@@ -812,7 +812,7 @@ Do you want to switch to this account?</string>
     <string name="error_connecting_with_the_duo_service_use_a_different_two_step_login_method_or_contact_duo_for_assistance">Error connecting with the Duo service. Use a different two-step login method or contact Duo for assistance.</string>
     <string name="master_password_hint_not_specified">Master password hint</string>
     <string name="new_master_password_hint">New master password hint</string>
-    <string name="get_started">Get started</string>
+    <string name="get_started">Get Started</string>
```
lucasmz-dev: Material 3 guidelines tell you to not use Pascal Case like this

gabrbrand: Thanks for pointing that out. I copied it directly from the original string.

david-livefront (approved):

david-livefront:
Thanks @gabrbrand 

Great catch!

PR #6940, opened 2026-05-18
PM-37690: feat: Update copy for adding a new item

david-livefront:
## 🎟️ Tracking

[PM-37690](https://bitwarden.atlassian.net/browse/PM-37690)

## 📔 Objective

This PR updates the copy when adding a new item. The copy will always using the pattern `Add [type]` instead of the previous pattern of `New [type]`.

## 📸 Screenshots

| Before | After |
| --- | --- |
|  |  |

[PM-37690]: https://bitwarden.atlassian.net/browse/PM-37690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/itemlisting/VaultItemListingScreenTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/itemlisting/VaultItemListingScreenTest.kt
@@ -500,7 +500,7 @@ class VaultItemListingScreenTest : BitwardenComposeTest() {
     @Test
     fun `floating action button click should send AddItemClick action`() {
         composeTestRule
-            .onNodeWithContentDescription("Add Item")
+            .onNodeWithContentDescription("Add item")
             .performClick()
         verify { viewModel.trySendAction(VaultItemListingsAction.AddVaultItemClick) }
     }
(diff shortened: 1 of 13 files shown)


app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/itemlisting/VaultItemListingScreenTest.kt
```diff
@@ -777,7 +777,7 @@ class VaultItemListingScreenTest : BitwardenComposeTest() {
 
         composeTestRule
             .onNodeWithContentDescription("Add item")
-            .assertDoesNotExist()
+            .assertIsDisplayed()
```
david-livefront, line 780: That was a bug in the test, it should have been present and was passing because the content description used was incorrect.

SaintPatrck:
Can we make sure the New Item Type designs in Figma are updated? I don't want QA to kick those tickets back because it says "New" there but we're using "Add" now.

SaintPatrck (approved):

david-livefront:
> Can we make sure the New Item Type designs in Figma are updated? I don't want QA to kick those tickets back because it says "New" there but we're using "Add" now.

I'll talk to design about it.

david-livefront:
Thanks @SaintPatrck

PR #6946, opened 2026-05-19
[PM-37335] fix: Route attachments upgrade CTA through in-app plan modal

SaintPatrck:
## 🎟️ Tracking

[PM-37335](https://bitwarden.atlassian.net/browse/PM-37335)

## 📔 Objective

The Attachments "Upgrade to Premium" dialog was missed by the PM-33519 rewire and continued sending users to the legacy web URL even when the in-app upgrade path was available. Users on cloud accounts with in-app upgrade enabled would leave the app to complete what should be an in-app flow — inconsistent with every other upgrade entry point already routed through the plan modal.

`AttachmentsViewModel.handleUpgradeToPremiumClick` now consults `PremiumStateManager.isInAppUpgradeAvailable()` and emits `NavigateToPlanModal` when in-app upgrade is supported, falling back to the existing web URL for environments where it is not. Mirrors the `VaultItemViewModel.handleUpgradeToPremiumClick` shape already in place for the item-detail upgrade CTA.

[PM-37335]: https://bitwarden.atlassian.net/browse/PM-37335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/attachments/AttachmentsViewModelTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/attachments/AttachmentsViewModelTest.kt
@@ -15,6 +15,7 @@ import com.bitwarden.vault.CipherView
 import com.x8bit.bitwarden.data.auth.datasource.disk.model.OnboardingStatus
 import com.x8bit.bitwarden.data.auth.repository.AuthRepository
 import com.x8bit.bitwarden.data.auth.repository.model.UserState
+import com.x8bit.bitwarden.data.billing.manager.PremiumStateManager
 import com.x8bit.bitwarden.data.platform.error.NoActiveUserException
 import com.x8bit.bitwarden.data.platform.manager.FeatureFlagManager
 import com.x8bit.bitwarden.data.platform.manager.model.FirstTimeState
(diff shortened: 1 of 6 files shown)


app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/attachments/AttachmentsViewModelTest.kt
```diff
@@ -145,23 +149,37 @@ class AttachmentsViewModelTest : BaseViewModelTest() {
...
+                )
+            }
+            verify(exactly = 1) {
+                environmentRepository.environment
+            }
         }
-        verify(exactly = 1) {
-            environmentRepository.environment
+
+    @Suppress("MaxLineLength")
```
david-livefront: This suppression isn't needed, is it?

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/attachments/AttachmentsViewModelTest.kt
```diff
@@ -145,23 +149,37 @@ class AttachmentsViewModelTest : BaseViewModelTest() {
         }
     }
 
+    @Suppress("MaxLineLength")
```
david-livefront: This isn't needed either, right?

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/attachments/AttachmentsViewModelTest.kt
SaintPatrck: Nope

david-livefront (approved):

PR #6949, opened 2026-05-20
[PM-37289] fix: Refresh archive row after premium upgrade

SaintPatrck:
## 🎟️ Tracking

[PM-37289](https://bitwarden.atlassian.net/browse/PM-37289)

## 📔 Objective

The Archive row on the Vault screen kept showing its locked state — zero count, lock icon, "Premium required" subtext — after a user upgraded to Premium. The row updated correctly only after a manual pull-to-refresh or app restart, making freshly-paid customers believe the upgrade had not taken effect.

`VaultViewModel.handleUserStateUpdateReceive` writes the new `isPremium` value into state on profile updates, but the Archive row's UI fields are precomputed in `VaultDataExtensions.kt` when the viewState is first built and are not re-derived from state on each emission. When `isPremium` flips, the vault data flow does not re-emit, so the stale precomputed fields stay on screen.

Triggering a `updateViewState` call on the `isPremium` transition re-derives those fields against the cached `vaultDataStateFlow.value`, surfacing the new entitlement immediately and keeping the rest of the screen stable for unrelated profile updates.

[PM-37289]: https://bitwarden.atlassian.net/browse/PM-37289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModelTest.kt
+++ b/app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModelTest.kt
@@ -49,17 +49,17 @@ import com.x8bit.bitwarden.data.platform.manager.network.NetworkConnectionManage
 import com.x8bit.bitwarden.data.platform.repository.SettingsRepository
 import com.x8bit.bitwarden.data.platform.repository.util.FakeEnvironmentRepository
 import com.x8bit.bitwarden.data.vault.datasource.sdk.model.createMockBankAccountView
-import com.x8bit.bitwarden.data.vault.datasource.sdk.model.createMockDriversLicenseView
-import com.x8bit.bitwarden.data.vault.datasource.sdk.model.createMockPassportView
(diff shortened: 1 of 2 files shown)


app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModelTest.kt
```diff
@@ -2429,6 +2429,102 @@ class VaultViewModelTest : BaseViewModelTest() {
             )
         }
 
+    @Suppress("MaxLineLength")
```
david-livefront: Is this suppression needed?

SaintPatrck: Nope

david-livefront (approved):

PR #6954, opened 2026-05-20
[PM-37814] feat: Add debug flag to disable self-host premium check

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37814

## 📔 Objective

Three Bitwarden internal dev/QA environments — `vault.qa.bitwarden.pw`, `qa-team.sh.bitwarden.pw`, and `vault.usdev.bitwarden.pw` — are classified as self-hosted, which causes the existing Premium Upgrade self-host gating to hide the flow and prevent QA from exercising it on those environments.

This adds a debug-only feature flag (`debug-disable-self-host-premium-check`) that, when enabled in the debug menu, treats self-hosted environments as cloud for premium upgrade purposes. The flag defaults off, only surfaces in builds where the debug menu is available, and leaves cloud environments completely unaffected.

Ports the iOS approach in [bitwarden/ios#2657](https://github.com/bitwarden/ios/pull/2657) for cross-platform parity. Like iOS, the mechanism is pure flag-based — no URL whitelist — so QA can use it against any self-hosted setup they need to test.

The gating decision is exposed as a `StateFlow` on `PremiumStateManager` and observed by Settings, so the Plan row reacts the moment the toggle flips (no environment-switch or screen re-entry required).

## 📸 Screenshots

https://github.com/user-attachments/assets/327afeba-db3b-45cf-a292-ab103fe8811a

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
@@ -259,11 +259,11 @@ class PremiumStateManagerImpl(
             .launchIn(unconfinedScope)
     }
 
+    override val isSelfHosted: Boolean get() = isSelfHostedFlow.value
+
     override fun isInAppUpgradeAvailable(): Boolean =
         billingRepository.isInAppBillingSupportedFlow.value &&
             featureFlagManager.getFeatureFlag(FlagKey.MobilePremiumUpgrade)
-
-    override fun isSelfHosted(): Boolean = isSelfHostedFlow.value
 
     override fun dismissPremiumUpgradeBanner() {
         val activeUserId = authDiskSource.userState?.activeUserId ?: return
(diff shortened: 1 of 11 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -246,6 +263,8 @@ class PremiumStateManagerImpl(
         billingRepository.isInAppBillingSupportedFlow.value &&
             featureFlagManager.getFeatureFlag(FlagKey.MobilePremiumUpgrade)
 
+    override fun isSelfHosted(): Boolean = isSelfHostedFlow.value
```
david-livefront: Could make this a `val`

SaintPatrck: Sure

david-livefront (approved):

PR #6955, opened 2026-05-20
PM-37568: Feat: Remove feature flags

david-livefront:
## 🎟️ Tracking

[PM-37568](https://bitwarden.atlassian.net/browse/PM-37568)

## 📔 Objective

This PR removes the `CredentialExchangeProtocolImport`, `CredentialExchangeProtocolExport`, and `SendEmailVerification` feature flags from the app.

[PM-37568]: https://bitwarden.atlassian.net/browse/PM-37568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
@@ -121,7 +121,7 @@ class VaultViewModel @Inject constructor(
     private val browserAutofillDialogManager: BrowserAutofillDialogManager,
     private val credentialExchangeRegistryManager: CredentialExchangeRegistryManager,
     private val buildInfoManager: BuildInfoManager,
-    private val featureFlagManager: FeatureFlagManager,
+    featureFlagManager: FeatureFlagManager,
     snackbarRelayManager: SnackbarRelayManager<SnackbarRelay>,
 ) : BaseViewModel<VaultState, VaultEvent, VaultAction>(
     initialState = run {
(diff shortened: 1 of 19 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/vault/VaultViewModel.kt
```diff
@@ -292,16 +292,16 @@ class VaultViewModel @Inject constructor(
...
-
         featureFlagManager.getFeatureFlagFlow(FlagKey.NewItemTypes)
             .map { VaultAction.Internal.NewItemTypesFlagUpdateReceive(isEnabled = it) }
             .onEach(::sendAction)
             .launchIn(viewModelScope)
 
+        if (!buildInfoManager.isFdroid) {
+            viewModelScope.launch {
+                credentialExchangeRegistryManager.register()
+            }
```
david-livefront, line 303: This is here because we would previously `register` or `unregister` after the first emission from the feature flag flow above (assuming we are not Fdroid). Since that flow no longer exists, we just `register` when not Fdroid.

I assume we do not need to `unregister`, right?

SaintPatrck: Correct. Thanks for calling this out.

app/src/test/kotlin/com/x8bit/bitwarden/data/platform/repository/DebugMenuRepositoryTest.kt
```diff
@@ -105,8 +105,8 @@ class DebugMenuRepositoryTest {
             debugMenuRepository.resetFeatureFlagOverrides()
             verify(exactly = 1) {
                 mockFeatureFlagOverrideDiskSource.saveFeatureFlag(
-                    FlagKey.CredentialExchangeProtocolImport,
-                    FlagKey.CredentialExchangeProtocolImport.defaultValue,
+                    FlagKey.AttachmentUpdates,
```
SaintPatrck, line 108: ❓ Was this missed before?

david-livefront: Is is a test for the DebugMenu repo, and for this particular test, it needs a _real_ flag, it can't be one of the Dummy flags we use for the other tests.

SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6957, opened 2026-05-20
[PM-37691] feat: Surface pending cancellation status on Premium plan view

SaintPatrck:
## 🎟️ Tracking

- https://bitwarden.atlassian.net/browse/PM-37691

## 📔 Objective

After canceling a subscription via the Stripe customer portal, the badge on the Premium plan screen stayed on "Active" until the user manually refreshed. The `/account/billing/vnext/subscription` payload at that point reports `status: active` with `cancelAt: <future-instant>`, and the app never sees a push notification for the transition. Two gaps caused this:

- The portal launch went through `intentManager.launchUri(...)` (plain Chrome Custom Tab) with no return signal.
- The status mapping only looked at the raw `status` field, so `active + cancelAt` collapsed to `PremiumSubscriptionStatus.ACTIVE`.

This PR routes the portal launch through `startAuthTab` with a dedicated `stripePortalLauncher`, so closing the tab fires `MainAction.StripePortalResult` → `SpecialCircumstance.StripePortal` → a fresh `/subscription` fetch. A new `PremiumSubscriptionStatus.PENDING_CANCELLATION` is derived when `status ∈ {active, trialing}` and `cancelAt` is non-null, rendering a warning badge plus "Your subscription is scheduled to cancel on **{date}**. You can reinstate it anytime before then."

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |  |


app/src/main/kotlin/com/x8bit/bitwarden/MainActivity.kt
```diff
@@ -95,6 +95,21 @@ class MainActivity : AppCompatActivity() {
         mainViewModel.trySendAction(MainAction.PremiumCheckoutResult(it))
     }
 
+    private val stripePortalLauncher = AuthTabIntent.registerActivityResultLauncher(this) {
+        mainViewModel.trySendAction(MainAction.StripePortalResult(it))
+    }
+
+    private val authTabLaunchers by lazy {
+        AuthTabLaunchers(
```
david-livefront, line 103: I assume this is to ensure that we only build them once?

david-livefront (approved):

PR #6963, opened 2026-05-22
llm: Add interface KDoc rule to implementing-android-code skill

SaintPatrck:
## 🎟️ Tracking

No Jira ticket — meta change to the `implementing-android-code` skill itself, originating from repeated review feedback.

## 📔 Objective

Reviewers have repeatedly flagged interface member KDoc that drifts into describing **how** the implementation fulfills the contract — naming concrete collaborators, describing call ordering, mentioning caching or fallback behavior. The contract belongs on the interface; the *why* (when non-obvious) belongs on the `...Impl` override; the *what* belongs nowhere because well-named identifiers already convey it.

Encoding the rule in the skill turns a per-PR correction into a guardrail that applies to every future change that triggers the `implementing-android-code` skill.

### What's in the skill update (`v0.1.3` → `v0.1.4`)

- **Section D — Data Layer Implementation** gains a "KDoc on Interfaces vs. Implementations" subsection covering:
  - The contract-vs-implementation distinction, with wrong/right examples on a `Repository` interface.
  - Red flags that an interface KDoc has drifted into implementation territory (naming concrete collaborators, describing ordering, mentioning caching/retries/logging/threading, restating the signature in prose).
  - The "would a different valid implementation be free to change this behavior?" test for separating contract from implementation detail.
  - The carve-out for behaviors callers genuinely depend on (e.g., "safe to call before vault unlock", "cached for the session") — those *are* contract.
- **Anti-pattern list** gains a `NEVER put implementation details in interface KDoc` entry that points back to section D.

The rule applies to every interface in the codebase — repositories, managers, data sources, validators, UI-layer interfaces alike — and aligns interface KDoc with the project-wide comment rule from `CLAUDE.md` (WHY not WHAT, only when non-obvious).

### DIFF
--- a/.claude/skills/implementing-android-code/SKILL.md
+++ b/.claude/skills/implementing-android-code/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: implementing-android-code
-version: 0.1.3
+version: 0.1.4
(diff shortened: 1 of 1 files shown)


withinfocus (approved):

SaintPatrck:
Thanks @withinfocus

PR #6975, opened 2026-05-25
[PM-37255] feat: Consume fill-assist targeting rules from server config

aj-rosado:
## 🎟️ Tracking

## 📔 Objective

Implements the data pipeline for Fill-Assist Targeting Rules (Task 2 of 3).
  This PR fetches, parses, and caches server-provided autofill field rules per
  account environment. The rules are not yet applied to autofill (Task 3); this
  PR establishes the infrastructure that Task 3 will consume.

  ### What's included

  **Feature flag** (`fill-assist-targeting-rules`):
  - Added to `FlagKey`, `FeatureFlagListItems`, and `strings_non_localized.xml`

  **Network layer (`:network` module)**:
  - `FillAssistManifestJson` and `FillAssistFormsV1Json` models
  - `FillAssistService` / `FillAssistServiceImpl` using `createStaticRetrofit()`
    to bypass `BaseUrlInterceptor` (the fill-assist CDN is external to the
    Bitwarden API base URL)
  - `fillAssistRulesUrl` added to `ConfigResponseJson.EnvironmentJson`

  **Data layer (`:app` module)**:
  - `FillAssistRules` domain model — CSS selectors parsed at sync time into
    `SelectorClause(tag, id, name, type, role)` for efficient matching at
    autofill time
  - `FillAssistDiskSource` — per-server SharedPreferences storage (keyed by the
    server's fill-assist CDN URL so multiple accounts on the same server share
    one cached copy)
  - `FillAssistManager` — self-managing singleton that syncs on app start,
    config refresh, and account switch. Uses a 6-hour timestamp throttle before
    checking the manifest CID, matching the browser implementation

  **Sync behaviour**:
  - Timestamp check first (skip all network if < 6 hours since last fetch)
  - Manifest fetch to compare CID (skip forms download if unchanged)
  - Forms download and parse on CID change only
  - `syncIfIdle()` drop-if-busy pattern matches `VaultSyncManagerImpl`
  - `CURRENT_CACHE_VERSION` migration clears all cached rules on breaking parser
   changes

  **CSS parsing**:
  - Compound selectors (`input#id[name='x']`) fully supported
  - Shadow DOM selectors (`>>>`) excluded (unreachable via Android autofill)
  - Forms from all pathnames pooled per host and merged by category (one
    `HostRule` per form category per host, deduplicating identical entries)
  - `FileEntryJson.deprecated` — logs a warning when the current schema version
    is marked end-of-life

  Currently targets `CURRENT_FORMS_VERSION = "v0"` matching the prerelease
  schema version of the map-the-web project.

  ### Not in this PR
  - Applying rules in `AutofillParserImpl` 
  - Opt-in user setting toggle

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/datasource/disk/FillAssistDiskSourceImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/datasource/disk/FillAssistDiskSourceImpl.kt
@@ -1,0 +1,72 @@
+package com.x8bit.bitwarden.data.autofill.datasource.disk
+
+import android.content.SharedPreferences
+import com.bitwarden.core.data.util.decodeFromStringOrNull
+import com.bitwarden.data.datasource.disk.BaseDiskSource
+import com.x8bit.bitwarden.data.autofill.model.FillAssistRules
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+// Bump this constant in two cases:
+// 1. The parsing logic changes in a way that invalidates previously cached results.
+// 2. EXPECTED_SCHEMA_MAJOR in FillAssistManagerImpl is updated to support a new schema major.
+//    Without bumping this, the staleness check would skip re-downloading data that was previously
+//    rejected for an unsupported schema — meaning the app would never pick up the new rules.
+// On the next app launch after a bump, all stored fill-assist data is cleared and re-downloaded.
+private const val CURRENT_CACHE_VERSION = 0
+
+private const val FILL_ASSIST_CACHE_VERSION_KEY = "fillAssistCacheVersion"
+private const val FILL_ASSIST_RULES_KEY = "fillAssistRules"
+private const val FILL_ASSIST_CID_KEY = "fillAssistLastCid"
+private const val FILL_ASSIST_TIMESTAMP_KEY = "fillAssistLastFetchTimestamp"
+
+/**
+ * Primary implementation of [FillAssistDiskSource].
+ */
+class FillAssistDiskSourceImpl(
+    sharedPreferences: SharedPreferences,
+    private val json: Json,
+) : BaseDiskSource(sharedPreferences),
+    FillAssistDiskSource {
+
+    init {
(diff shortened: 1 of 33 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/datasource/disk/FillAssistDiskSourceImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Primary implementation of [FillAssistDiskSource].
+ */
+class FillAssistDiskSourceImpl(
+    sharedPreferences: SharedPreferences,
+    private val json: Json,
+) : BaseDiskSource(sharedPreferences),
+    FillAssistDiskSource {
+
```
david-livefront, line 31: So nothing here is tied to a user? They are all global settings?

aj-rosado: They are provided by the config so they are tied to a server

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,226 @@
...
+        serverConfigRepository.serverConfigStateFlow
+            .filterNotNull()
+            .onEach { syncIfNecessary() }
+            .launchIn(unconfinedScope)
+    }
+
+    override fun syncIfNecessary() {
+        if (!featureFlagManager.getFeatureFlag(FlagKey.FillAssistTargetingRules)) return
+        val serverUrl = serverConfigRepository.serverConfigStateFlow.value
+            ?.serverData?.environment?.fillAssistRulesUrl ?: return
```
david-livefront, line 67: Can we chain this?

```kotlin
        val serverUrl = serverConfigRepository
            .serverConfigStateFlow
            .value
            ?.serverData
            ?.environment
            ?.fillAssistRulesUrl
            ?: return
```

PR #6978, opened 2026-05-26
[PM-37232] fix: Hide upgrade CTAs while a Premium upgrade is pending

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37232

## 📔 Objective

When a user completes Stripe checkout and returns to the app, the server-side subscription confirmation can lag the client. During that window the Vault home upgrade banner and the Plan screen Upgrade Now button kept advertising the upgrade we were already processing, which nudged users to start the checkout flow a second time.

Promote the "PendingUpgrade" signal from transient `PlanViewModel` state to a per-user disk-backed flag exposed via `PremiumStateManager`. The banner-eligibility flow now gates on it, the Plan screen hides its Upgrade Now CTA while it is set, and the flag clears when the server flips the active user to Premium. Dismissing the PendingUpgrade dialog is a UI-only action and leaves the pending state intact, so CTAs stay suppressed until the upgrade actually resolves.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
@@ -102,7 +102,10 @@ class PremiumStateManagerImpl(
             .stateIn(
                 scope = unconfinedScope,
                 started = SharingStarted.Eagerly,
-                initialValue = false,
+                initialValue = authDiskSource.userState
+                    ?.activeUserId
+                    ?.let { settingsDiskSource.getPremiumUpgradePending(userId = it) }
+                    ?: false,
             )
 
     @OptIn(ExperimentalCoroutinesApi::class)
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
@@ -85,6 +85,33 @@ class PremiumStateManagerImpl(
                 initialValue = SubscriptionStatusState.Loading,
             )
 
+    /**
+     * Reactive read of the per-user "Premium upgrade pending" disk flag. There is no premium /
+     * consumed gating — this is just the raw flag, set by [markPremiumUpgradePending] when sync
+     * confirms post-checkout that `isPremium` has not yet flipped, and cleared either explicitly
+     * (the Continue button on the PendingUpgrade dialog) or implicitly when
+     * `hasPremiumPersonally` transitions `false → true` in the init block below.
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    override val isPremiumUpgradePendingFlow: StateFlow<Boolean> =
+        authDiskSource
+            .activeUserIdChangesFlow
+            .flatMapLatest { userId ->
+                userId
+                    ?.let { id ->
+                        settingsDiskSource
+                            .getPremiumUpgradePendingFlow(id)
+                            .map { it ?: false }
+                    }
+                    ?: flowOf(false)
+            }
+            .distinctUntilChanged()
+            .stateIn(
+                scope = unconfinedScope,
+                started = SharingStarted.Eagerly,
+                initialValue = false,
+            )
+
     @OptIn(ExperimentalCoroutinesApi::class)
     override val isPremiumUpgradeBannerEligibleFlow: StateFlow<Boolean> =
         combine(
@@ -118,6 +145,9 @@ class PremiumStateManagerImpl(
             )
         }
             .combine(subscriptionStatusStateFlow) { inputs, subscriptionStatus ->
+                inputs to subscriptionStatus
+            }
+            .combine(isPremiumUpgradePendingFlow) { (inputs, subscriptionStatus), isPending ->
                 val profile = inputs.userState?.activeAccount?.profile
                     ?: return@combine false
                 val isAccountOldEnough = profile.creationDate.isOlderThanDays(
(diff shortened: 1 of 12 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -85,6 +85,26 @@ class PremiumStateManagerImpl(
...
+                            .getPremiumUpgradePendingFlow(id)
+                            .map { it ?: false }
+                    }
+                    ?: flowOf(false)
+            }
+            .distinctUntilChanged()
+            .stateIn(
+                scope = unconfinedScope,
+                started = SharingStarted.Eagerly,
+                initialValue = false,
```
david-livefront: Could we have an `initialValue` here?

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -285,6 +310,20 @@ class PremiumStateManagerImpl(
         )
     }
 
+    override fun markPremiumUpgradePending(userId: String) {
+        settingsDiskSource.storePremiumUpgradePending(
+            userId = userId,
+            isPending = true,
```
david-livefront, line 330: Is this state independent of some of the other states we already have stored?

I am wondering if an enum would server better or if this all needs to be tracked independently?

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceImpl.kt
```diff
@@ -264,6 +269,7 @@ class SettingsDiskSourceImpl(
         // - Premium upgrade banner dismissed
         // - Upgraded to Premium action card consumed
         // - Upgraded to Premium action card pending
+        // - Premium upgrade pending
```
david-livefront, line 272: 👍

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
SaintPatrck: Good question. There's a short lifecycle window of `Free -> PendingUpgrade -> Premium`, but that is independent of other signals being tracked like, `UpgradeToPremiumCardPending`. Technically we could enumerate a `UpgradeLifecycleState`. Feels like that adds scope creep to this PR though. 🤔

david-livefront: OK, might be worth investigating. Ideally we would not need to add a whole migration flow as well

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
```diff
@@ -285,6 +313,20 @@ class PremiumStateManagerImpl(
...
+        )
+    }
+
+    override fun clearPremiumUpgradePending(userId: String) {
+        settingsDiskSource.storePremiumUpgradePending(
+            userId = userId,
+            isPending = false,
```
david-livefront: Does `false` mean something different than `null` in this case?

app/src/test/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceTest.kt
```diff
@@ -955,6 +962,72 @@ class SettingsDiskSourceTest {
                 }
         }
 
+    @Suppress("MaxLineLength")
```
david-livefront: I don't think this suppression is needed, right?

app/src/test/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/SettingsDiskSourceTest.kt
```diff
@@ -955,6 +962,72 @@ class SettingsDiskSourceTest {
...
+        val key = "${baseKey}_$mockUserId"
+        fakeSharedPreferences.edit { putBoolean(key, true) }
+        settingsDiskSource.storePremiumUpgradePending(
+            userId = mockUserId,
+            isPending = null,
+        )
+        assertFalse(fakeSharedPreferences.contains(key))
+    }
+
+    @Suppress("MaxLineLength")
```
david-livefront: Ditto

app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanViewModelTest.kt
```diff
@@ -540,12 +549,67 @@ class PlanViewModelTest : BaseViewModelTest() {
...
+                viewModel.trySendAction(PlanAction.ContinueClick)
+                assertEquals(PlanEvent.NavigateBack, awaitItem())
+            }
+
+            verify(exactly = 0) {
+                mockPremiumStateManager.clearPremiumUpgradePending(any())
+            }
+        }
+
+    @Suppress("MaxLineLength")
```
david-livefront: Ditto

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
SaintPatrck: True. Didn't think about migration. Should be able to get it in without too many extra lines. 👀

app/src/main/kotlin/com/x8bit/bitwarden/data/billing/manager/PremiumStateManagerImpl.kt
SaintPatrck: Nope. Looks like surrounding handling was copied-forward. I'll open another PR to correct the surrounding ones that should be using null too.

david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #6981, opened 2026-05-26
[PM-37287] feat: Make Premium subscription required text tappable under authenticator key

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37287

## 📔 Objective

Non-Premium users see "Premium subscription required" under the Authenticator key on the View and Edit login screens, but the text is inert — there is no path from the gated state to the upgrade flow on either screen. This makes the placeholder tappable. Tapping opens an upsell dialog explaining that TOTP is a Premium feature, with **Cancel** and **Upgrade to Premium** buttons.

The Upgrade button routes through the existing `handleUpgradeToPremiumClick` logic: when in-app upgrade is available (`MobilePremiumUpgrade` flag + in-app billing supported) it opens the in-app Plan modal; otherwise it falls back to the web vault premium URI used by other upgrade CTAs.

**Incidental behavior change:** `VaultAddEditViewModel.handleUpgradeToPremiumClick` now also dismisses the active dialog before emitting the navigation event. This aligns Edit with the View screen's existing behavior and incidentally affects the existing `ArchiveRequiresPremium` upsell on Edit.

#### In-app Upgrade **DISABLED**

| Figma | Actual |
|--------|--------|
|  | <video src="https://github.com/user-attachments/assets/220147c2-e331-4814-a7c4-7d7d13b2c083" /> |

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditLoginItems.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditLoginItems.kt
@@ -134,6 +134,7 @@ fun LazyListScope.vaultAddEditLoginItems(
         TotpRow(
             totpKey = loginState.totp,
             canViewTotp = loginState.canViewPassword,
+            isAuthenticatorKeyPremiumGated = loginState.isAuthenticatorKeyPremiumGated,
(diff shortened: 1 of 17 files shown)


SaintPatrck:
Implemented by https://github.com/bitwarden/android/pull/7090

PR #6986, opened 2026-05-28
[PM-38118] fix: Support Firefox updated toolbar in accessibility autofill

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38118

## 📔 Objective

 Firefox's toolbar now exposes the URL bar under a Compose test tag (`ADDRESSBAR_URL_BOX`) via  `testTagsAsResourceId`, with the URL stored in `contentDescription` rather than `text`. The standard `findAccessibilityNodeInfosByViewId` API cannot locate these nodes, causing the accessibility tile to silently do nothing on affected pages.

  Changes:
  - Adds `possibleUrlSemanticIds` to `Browser` for URL bar IDs that require manual tree traversal instead of the system API lookup
  - Updates all Mozilla browser entries to include `ADDRESSBAR_URL_BOX` in `possibleUrlSemanticIds` with a `mozillaUrlExtractor` that reads `contentDescription` when `text` is empty (toolbar format: `" $url. Search or enter address"`)
  - `AccessibilityParserImpl` tries `possibleUrlFieldIds` via the fast system API first; falls back to `possibleUrlSemanticIds` tree traversal only when the standard lookup returns empty — no performance regression for non-Mozilla browsers
  - Adds unit test coverage for the semantic ID + `contentDescription` extraction path

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
@@ -40,17 +40,17 @@ class AccessibilityParserImpl(
         val browser = packageName
             .getSupportedBrowserOrNull()
             ?: return "androidapp://$packageName".toUri()
-        return (browser.possibleUrlFieldIds
+        return browser.possibleUrlFieldIds
             .flatMap { viewId ->
                 rootNode.findAccessibilityNodeInfosByViewId("$packageName:id/$viewId")
             }
             .ifEmpty {
-                browser.possibleUrlSemanticIds.flatMap { viewId ->
+                browser.possibleUrlSemanticIds.flatMap { semanticId ->
                     accessibilityNodeInfoManager.findAccessibilityNodeInfoList(rootNode) {
-                        it.viewIdResourceName == viewId
+                        it.viewIdResourceName == semanticId
                     }
                 }
-            })
+            }
             .firstNotNullOfOrNull { node ->
                 val urlText = node.text?.toString()?.takeIf { it.isNotEmpty() }
                     ?: node.contentDescription?.toString()?.takeIf { it.isNotEmpty() }
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
@@ -46,6 +46,8 @@ class AccessibilityParserImpl(
             }
             .ifEmpty {
                 browser.possibleUrlSemanticIds.flatMap { semanticId ->
+                    // Semantic IDs are exposed as viewIdResourceName via testTagsAsResourceId
+                    // and cannot be found via findAccessibilityNodeInfosByViewId on Firefox.
                     accessibilityNodeInfoManager.findAccessibilityNodeInfoList(rootNode) {
                         it.viewIdResourceName == semanticId
                     }
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
@@ -40,25 +40,32 @@ class AccessibilityParserImpl(
         val browser = packageName
             .getSupportedBrowserOrNull()
             ?: return "androidapp://$packageName".toUri()
-        return browser
-            .possibleUrlFieldIds
+        return (browser.possibleUrlFieldIds
             .flatMap { viewId ->
-                rootNode
-                    .findAccessibilityNodeInfosByViewId("$packageName:id/$viewId")
-                    .map { accessibilityNodeInfo ->
-                        browser
-                            .urlExtractor(accessibilityNodeInfo.text.toString())
-                            ?.trim()
-                            ?.let { rawUrl ->
-                                if (rawUrl.contains(other = ".") && !rawUrl.hasHttpProtocol()) {
-                                    "https://$rawUrl"
-                                } else {
-                                    rawUrl
-                                }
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/util/BrowserUtil.kt
```diff
@@ -67,7 +83,12 @@ private val ACCESSIBILITY_SUPPORTED_BROWSERS = listOf(
     Browser(
         packageName = "com.qwant.liberty",
         // 2nd = Legacy (before v4)
-        possibleUrlFieldIds = listOf("mozac_browser_toolbar_url_view", "url_bar_title"),
+        possibleUrlFieldIds = listOf(
+            "mozac_browser_toolbar_url_view",
+            "url_bar_title",
+        ),
+        possibleUrlSemanticIds = listOf("ADDRESSBAR_URL_BOX"),
```
david-livefront, line 90: Did you test all these browsers?

aj-rosado: I did not, as this is a fallback measure I have applied it on all browsers that look for that id

david-livefront: Is there any indication that these other browsers would use the same semantic id?

aj-rosado: Simply basing on the use of `mozac_browser_toolbar_url_view`

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
```diff
@@ -40,25 +40,32 @@ class AccessibilityParserImpl(
         val browser = packageName
             .getSupportedBrowserOrNull()
             ?: return "androidapp://$packageName".toUri()
-        return browser
-            .possibleUrlFieldIds
+        return (browser.possibleUrlFieldIds
```
david-livefront: This parenthesis is not needed, right?

Can it just be:
```kotlin
return browser
    .possibleUrlFieldIds
    ...
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
```diff
@@ -40,25 +40,32 @@ class AccessibilityParserImpl(
...
-                                if (rawUrl.contains(other = ".") && !rawUrl.hasHttpProtocol()) {
-                                    "https://$rawUrl"
-                                } else {
-                                    rawUrl
-                                }
-                            }
+                rootNode.findAccessibilityNodeInfosByViewId("$packageName:id/$viewId")
+            }
+            .ifEmpty {
+                browser.possibleUrlSemanticIds.flatMap { viewId ->
```
david-livefront: Instead of `viewId` can we use `semanticId`. That might mitigate some confusion later on

david-livefront: Hmmm, or is it a `viewId`?

Now I am confused 😄

aj-rosado: or the `possibleUrlSemanticIds` should change? Because the field we are looking at from the node is the `viewIdResName`

aj-rosado: maybe `possibleUrlSemanticViewIds`? I am awful naming stuff

david-livefront: Maybe call it `semanticId` and then drop a comment in the code indicating that we map the `semanticId` to the `viewId`

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
```diff
@@ -40,25 +40,34 @@ class AccessibilityParserImpl(
         val browser = packageName
             .getSupportedBrowserOrNull()
             ?: return "androidapp://$packageName".toUri()
-        return browser
-            .possibleUrlFieldIds
+        return browser.possibleUrlFieldIds
             .flatMap { viewId ->
```
david-livefront, line 45: Can you move the `.possibleUrlFieldIds` down onto it's own line too

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/parser/AccessibilityParserImpl.kt
```diff
@@ -40,25 +40,34 @@ class AccessibilityParserImpl(
...
-                                } else {
-                                    rawUrl
-                                }
-                            }
+                rootNode.findAccessibilityNodeInfosByViewId("$packageName:id/$viewId")
+            }
+            .ifEmpty {
+                browser.possibleUrlSemanticIds.flatMap { semanticId ->
+                    // Semantic IDs are exposed as viewIdResourceName via testTagsAsResourceId
+                    // and cannot be found via findAccessibilityNodeInfosByViewId on Firefox.
```
david-livefront: 👍

david-livefront (approved):

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/util/BrowserUtil.kt
david-livefront: I approved, I'm still a little unsure about adding this to all these different browsers

PR #6990, opened 2026-05-29
[PM-37255] feat: Consume fill-assist targeting rules data

aj-rosado:
## 🎟️ Tracking

## 📔 Objective

**Changes:**
  - `ConfigResponseJson.EnvironmentJson` — adds `fillAssistRulesUrl: String?` (`@SerialName("fillAssistRules")`) from the server config response
  - `FillAssistManifestJson` — model for the CDN manifest file, with a version-keyed map (`Map<String, FileEntryJson?>`) so new schema versions appear automatically without model changes
  - `FillAssistFormsJson` — model for the forms rules file; uses `JsonElement` for the composite selector array to handle both string and array-of-string alternatives
  - `FillAssistApi` / `FillAssistService` / `FillAssistServiceImpl` — Retrofit service using `@Url` annotation with `createStaticRetrofit()` to bypass `BaseUrlInterceptor` (the fill-assist CDN is external to the Bitwarden API)
  - `BitwardenServiceClient` — exposes `fillAssistService`
  - `PlatformNetworkModule` — provides `FillAssistService`
  - All `EnvironmentJson` test fixtures updated with `fillAssistRulesUrl = null`


network/src/main/kotlin/com/bitwarden/network/model/FillAssistFormsJson.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @property hosts Map of hostname (optionally with port) to [HostEntryJson], or null if the host
+ * is explicitly excluded from fill-assist.
+ */
+@Serializable
+data class FillAssistFormsJson(
+    @SerialName("schemaVersion")
+    val schemaVersion: String? = null,
+
+    @SerialName("hosts")
+    val hosts: Map<String, HostEntryJson?>? = null,
```
david-livefront: Why do we have all the default values here?

Also, is everything in this model really nullable?

network/src/main/kotlin/com/bitwarden/network/model/FillAssistManifestJson.kt
```diff
@@ -0,0 +1,64 @@
...
+    val buildId: String? = null,
+
+    @SerialName("timestamp")
+    val timestamp: String? = null,
+
+    @SerialName("gitSha")
+    val gitSha: String? = null,
+
+    @SerialName("maps")
+    val maps: MapsJson? = null,
```
david-livefront: Same questions as above

network/src/test/kotlin/com/bitwarden/network/model/FillAssistFormsJsonTest.kt
```diff
@@ -0,0 +1,305 @@
...
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonPrimitive
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Test
+
+class FillAssistFormsJsonTest {
+
+    private val json = Json { ignoreUnknownKeys = true }
```
david-livefront: Can we use the real Json instance to insure the tests match reality?

```kotlin
private val json = CoreModule.providesJson(buildInfoManager = mockk(relaxed = true))
```

david-livefront: Hmmm, looking closer, do these tests just verify that the objects serialize properly?

We normally do not write tests for that as the `FillAssistServiceTest` should cover that for us

network/src/main/kotlin/com/bitwarden/network/model/FillAssistFormsJson.kt
aj-rosado: Should not be nullable

network/src/main/kotlin/com/bitwarden/network/model/FillAssistManifestJson.kt
aj-rosado: Should not be nullable

network/src/main/kotlin/com/bitwarden/network/model/FillAssistFormsJson.kt
```diff
@@ -0,0 +1,68 @@
...
+     * @property forms Site-wide fallback form descriptions.
+     * @property pathnames Pathname-specific overrides; a null value means that path is excluded.
+     */
+    @Serializable
+    data class HostEntryJson(
+        @SerialName("forms")
+        val forms: List<FormJson>? = null,
+
+        @SerialName("pathnames")
+        val pathnames: Map<String, PathnameEntryJson?>? = null,
```
david-livefront: Can we ditch the default params on these parameters

network/src/main/kotlin/com/bitwarden/network/model/FillAssistFormsJson.kt
```diff
@@ -0,0 +1,68 @@
...
+     * Each array element is either a CSS selector string or an array of strings for composite
+     * multi-input fields. Unknown fields are gracefully ignored via [ignoreUnknownKeys].
+     */
+    @Serializable
+    data class FormJson(
+        @SerialName("category")
+        val category: String,
+
+        @SerialName("container")
+        val container: List<String>? = null,
```
david-livefront: Same here

network/src/main/kotlin/com/bitwarden/network/model/FillAssistManifestJson.kt
```diff
@@ -0,0 +1,64 @@
...
+        val filename: String,
+
+        @SerialName("cid")
+        val cid: String,
+
+        @SerialName("schema")
+        val schema: String,
+
+        @SerialName("deprecated")
+        val deprecated: Boolean? = null,
```
david-livefront: Same here

network/src/main/kotlin/com/bitwarden/network/service/FillAssistServiceImpl.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.bitwarden.network.util.toResult
+
+/**
+ * Default implementation of [FillAssistService].
+ */
+internal class FillAssistServiceImpl(
+    private val api: FillAssistApi,
+) : FillAssistService {
+
+    override suspend fun getManifest(url: String): Result<FillAssistManifestJson> =
```
david-livefront: So this has to be done with a passed in url?

There is no good way to manage this via a `BaseUrlInterceptor`?

aj-rosado: Added it to the `BaseUrlInterceptor`, we will need afterwards to write it from config into the environment, this was the best option I have found.

network/src/main/kotlin/com/bitwarden/network/api/FillAssistApi.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Defines endpoints for retrieving fill-assist targeting rules. The base URL is set dynamically
+ * at runtime via [com.bitwarden.network.interceptor.BaseUrlInterceptors.fillAssistInterceptor].
+ */
+internal interface FillAssistApi {
+    /**
+     * Fetches the fill-assist manifest.
+     */
+    @GET("manifest.json")
```
david-livefront, line 17: 👍

david-livefront (approved):

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/datasource/disk/EnvironmentDiskSource.kt
```diff
@@ -27,4 +27,10 @@ interface EnvironmentDiskSource {
      * Stores the [urls] for the given [userEmail].
      */
     fun storePreAuthEnvironmentUrlDataForEmail(userEmail: String, urls: EnvironmentUrlDataJson)
+
+    /**
+     * The fill-assist URL provided by the server config, or `null` if the server does not
+     * configure fill-assist targeting rules.
+     */
+    var fillAssistRulesUrl: String?
```
david-livefront, line 35: Should this be per user?

david-livefront: Or are you trying to make this the realtime value for the current user similar to the `preAuthEnvironmentUrlData`?

PR #6991, opened 2026-05-29
[PM-37255] feat: Add fill-assist targeting rules data layer

aj-rosado:
## 🎟️ Tracking

## 📔 Objective

Adds the data layer for caching and managing fill-assist targeting rules, including CSS selector parsing and the self-managing sync manager.

  **Changes:**
  - `FillAssistRules` — domain model with `HostRule` and `SelectorClause(tag, id, name, type, role)`; CSS selectors are parsed at sync time, not at autofill time
  - `FillAssistDiskSource` / `FillAssistDiskSourceImpl` — per-server SharedPreferences storage (keyed by `fillAssistRulesUrl`); includes `CURRENT_CACHE_VERSION` migration to invalidate cached data when parsing logic changes
  - `FillAssistManager` / `FillAssistManagerImpl` — self-managing singletonthat syncs on `serverConfigStateFlow` changes; uses a 6-hour timestamp throttle before checking the manifest CID; CSS forms parsed from all pathnames, pooled and merged by category (one `HostRule` per form type per host); shadow DOM selectors (`>>>`) use the last segment for matching
  - `FillAssistModule` — Hilt DI bindings


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    private val unconfinedScope = CoroutineScope(dispatcherManager.unconfined)
+    private val ioScope = CoroutineScope(dispatcherManager.io)
+    private var syncJob: Job = Job().apply { complete() }
+
+    init {
+        serverConfigRepository.serverConfigStateFlow
+            .onEach { config ->
+                environmentDiskSource.fillAssistRulesUrl =
+                    config?.serverData?.environment?.fillAssistRulesUrl
```
david-livefront, line 68: Why store the same data in two places?

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+        val rules = parseForms(forms)
+        fillAssistDiskSource.storeFillAssistRules(serverUrl = serverUrl, rules = rules)
+        fillAssistDiskSource.storeLastKnownCid(serverUrl = serverUrl, cid = versionEntry.cid)
+        fillAssistDiskSource.storeLastFetchTimestamp(
+            serverUrl = serverUrl,
+            timestamp = clock.millis(),
+        )
+    }.also { result ->
+        result.onFailure { Timber.w(it, "Fill-assist sync failed") }
+    }
```
david-livefront, line 132: The `also` is redundant, you can just do this:
```kotlin
    ...
    fillAssistDiskSource.storeLastFetchTimestamp(
        serverUrl = serverUrl,
        timestamp = clock.millis(),
    )
}
    .onFailure { Timber.w(it, "Fill-assist sync failed") }
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+
+private fun buildFieldsByCategory(
+    forms: List<FillAssistFormsJson.FormJson>,
+): Map<String, MutableMap<String, MutableList<SelectorClause>>> {
+    val result = mutableMapOf<String, MutableMap<String, MutableList<SelectorClause>>>()
+    forms.forEach { form ->
+        val parsedFields = form.fields
+            .mapValues { (_, elem) -> parseCompositeSelectorArray(elem) }
+            .filterValues { it.isNotEmpty() }
+            .takeIf { it.isNotEmpty() } ?: return@forEach
```
david-livefront: The `?: return@forEach` should be on it's own line

david-livefront: Also, should the `return` be a `continue`?

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+            categoryFields.getOrPut(fieldKey) { mutableListOf() }.addAll(selectors)
+        }
+    }
+    return result
+}
+
+private fun parseCompositeSelectorArray(element: JsonElement): List<SelectorClause> {
+    if (element !is JsonArray) return emptyList()
+    val result = mutableListOf<SelectorClause>()
+    for (item in element) {
```
david-livefront: Can we make this more functional?

```kotlin
private fun parseCompositeSelectorArray(element: JsonElement): List<SelectorClause> {
    if (element !is JsonArray) return emptyList()
    return element.flatMap { item ->
        when (item) {
            is JsonPrimitive -> listOfNotNull(parseSingleSelector(item.content))
            is JsonArray -> {
                item
                    .filterIsInstance<JsonPrimitive>()
                    .mapNotNull { parseSingleSelector(it.content) }
            }

            else -> emptyList()
        }
    }
}
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+            fields = fields.mapValues { (_, selectors) -> selectors.distinct() },
+        )
+    }
+}
+
+private fun buildFieldsByCategory(
+    forms: List<FillAssistFormsJson.FormJson>,
+): Map<String, MutableMap<String, MutableList<SelectorClause>>> {
+    val result = mutableMapOf<String, MutableMap<String, MutableList<SelectorClause>>>()
+    forms.forEach { form ->
```
david-livefront: This one looks a bit more complicated but could this also be made more functional... Using `groupBy` might do the trick

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+    var role: String? = null
+
+    ATTRIBUTE_REGEX.findAll(effective).forEach { match ->
+        val attrName = match.groupValues[1]
+        val attrValue = match.groupValues[2]
+        when (attrName) {
+            "id" -> id = attrValue
+            "name" -> name = attrValue
+            "type" -> type = attrValue
+            "role" -> role = attrValue
```
david-livefront: Can we make these constants?

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+    return FillAssistRules(hostRules = hostRules)
+}
+
+private fun parseHostEntry(
+    hostEntry: FillAssistFormsJson.HostEntryJson,
+): List<FillAssistRules.HostRule> {
+    val allForms = buildList {
+        addAll(hostEntry.forms.orEmpty())
+        hostEntry.pathnames?.values?.filterNotNull()?.forEach { addAll(it.forms) }
+    }.distinct()
```
david-livefront: What do you think of this?
```kotlin
    val allForms = buildList {
        addAll(hostEntry.forms.orEmpty())
        addAll(hostEntry.pathnames?.values?.filterNotNull()?.flatMap { it.forms }.orEmpty())
    }
        .distinct()
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+    val tag = TAG_REGEX.find(effective)?.groupValues?.get(1)
+
+    var id: String? = null
+    var name: String? = null
+    var type: String? = null
+    var role: String? = null
+
+    ATTRIBUTE_REGEX.findAll(effective).forEach { match ->
+        val attrName = match.groupValues[1]
+        val attrValue = match.groupValues[2]
```
david-livefront, line 230: Can we add some additional comments here to explain what is actually happening.

This all seems rather odd (Autofill usually is that way lol)

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
aj-rosado: We cannot use `ServerConfigRepository` on the `BaseUrlInterceptor` as it would cause a circular dependency. Stored on `environmentDiskSource` to avoid it.

I think that despite storing same data in two places, it is cleaner and more consistent with our code than the other approach of passing the url

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
aj-rosado: Had to confirm it but it acts like a `continue` on this scenario, `continue` here returns an error
`'break' and 'continue' are only allowed inside loops.`

https://kotlinlang.org/docs/returns.html#return-to-labels

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
aj-rosado: We are still processing the data returned from the server `forms`. Basically this PR is processing the server data into something that can be useful to use when Autofilling.
Added a comment with an example that should make it easier to understand.

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,244 @@
...
+            fillAssistDiskSource.storeLastFetchTimestamp(
+                serverUrl = serverUrl,
+                timestamp = clock.millis(),
+            )
+            return@runCatching
+        }
+
+        val forms = fillAssistService
+            .getForms(filename = versionEntry.filename)
+            .getOrThrow()
```
david-livefront: Can we not throw here.

```kotlin
val forms = fillAssistService
    .getForms(filename = versionEntry.filename)
    .getOrNull()
    ?: return@runCatching
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,244 @@
...
+            ?.fillAssistRulesUrl
+            ?: return
+        val lastFetch = fillAssistDiskSource.getLastFetchTimestamp(serverUrl) ?: 0L
+        if (clock.millis() - lastFetch < UPDATE_INTERVAL_MS) return
+        if (!syncJob.isCompleted) return
+        syncJob = ioScope.launch { sync(serverUrl) }
+    }
+
+    private suspend fun sync(serverUrl: String) = runCatching {
+        val manifest = fillAssistService.getManifest().getOrThrow()
```
david-livefront: Same, let's handle all these error cases more gracefully, we shouldn't be throwing everything

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,244 @@
...
+
+private fun buildFieldsByCategory(
+    forms: List<FillAssistFormsJson.FormJson>,
+): Map<String, Map<String, List<SelectorClause>>> =
+    forms
+        .mapNotNull { form ->
+            val parsedFields = form.fields
+                .mapValues { (_, elem) -> parseCompositeSelectorArray(elem) }
+                .filterValues { it.isNotEmpty() }
+                .takeIf { it.isNotEmpty() } ?: return@mapNotNull null
```
david-livefront: Formatting:

```kotlin
.takeIf { it.isNotEmpty() }
?: return@mapNotNull null
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private suspend fun sync(serverUrl: String) = runCatching {
+        val manifest = fillAssistService
+            .getManifest()
+            .getOrNull()
+            ?: return@runCatching
+
+        val versionEntry = manifest.maps.forms[CURRENT_FORMS_VERSION]
+            ?: error("Version $CURRENT_FORMS_VERSION not found in manifest")
```
david-livefront: Could we just return here?

```koltin
 val versionEntry = manifest.maps.forms[CURRENT_FORMS_VERSION] ?: run {
     Timber.w("Fill-assist sync failed: Fill-assist forms $CURRENT_FORMS_VERSION is deprecated")
     return
}
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
```diff
@@ -0,0 +1,249 @@
...
+
+        val rules = parseForms(forms)
+        fillAssistDiskSource.storeFillAssistRules(serverUrl = serverUrl, rules = rules)
+        fillAssistDiskSource.storeLastKnownCid(serverUrl = serverUrl, cid = versionEntry.cid)
+        fillAssistDiskSource.storeLastFetchTimestamp(
+            serverUrl = serverUrl,
+            timestamp = clock.millis(),
+        )
+    }
+        .onFailure { Timber.w(it, "Fill-assist sync failed") }
```
david-livefront: If we remove that last spot that throws, we can drop the `runCatching` and this error handler, right?

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/manager/FillAssistManagerImpl.kt
aj-rosado: I have completely removed the deprecated. Is something that only concern us, no need to throw logs on clients for that.

david-livefront (approved):

PR #6992, opened 2026-05-29
[PM-37255] feat: Wire fill-assist manager into vault sync and app startup

aj-rosado:
## 🎟️ Tracking

## 📔 Objective

Connects the fill-assist manager to the rest of the app so rules are fetched and kept fresh automatically.

  **Changes:**
  - `VaultSyncManagerImpl` — calls `fillAssistManager.syncIfNecessary()` after vault data is stored; fill-assist failures are silently dropped and never affect vault sync
  - `VaultManagerModule` — injects `FillAssistManager` into `VaultSyncManager`
  - `BitwardenApplication` — injects `FillAssistManager` to trigger eager initialization; the `init {}` block subscribes to `serverConfigStateFlow` for config-change-triggered syncs


app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/VaultSyncManagerTest.kt
```diff
@@ -143,7 +144,12 @@ class VaultSyncManagerTest {
         every { databaseSchemeChangeFlow } returns mutableDatabaseSchemeChangeFlow
     }
 
+    private val fillAssistManager: FillAssistManager = mockk {
+        every { syncIfNecessary() } just runs
+    }
+
     private val vaultSyncManager: VaultSyncManager = VaultSyncManagerImpl(
+        fillAssistManager = fillAssistManager,
```
david-livefront, line 153: Can you make sure the appropriate `verify` calls happen

david-livefront (approved):

PR #6995, opened 2026-05-29
Chore: Move dispatcher for sdk functions into SDK sources

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR moves the usages of the io dispatcher for SDK operations in the the `AuthSdkSource` and the `GeneratorSdkSource`. This ensures that it will always be applied appropriately no matter where it is called from and simplifies the higher level classes that are currently do this manually.


SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #6997, opened 2026-05-29
[PM-37920] fix: Show Storage cost row when additional storage is present

SaintPatrck:
## 🎟️ Tracking

PM-37920

## 📔 Objective

The Plan screen's cost breakdown hid the "Storage cost" row whenever the
storage cost was `null` or zero, wrongly suppressing a legitimate
present-but-free $0.00 storage line — additional storage that exists but
currently costs nothing.

This gates the row on the *presence* of additional storage (`storageCost`
non-null) rather than its value, so a $0.00 line still renders while a truly
absent line stays hidden. A dedicated `toPresentMoneyText()` helper handles
the present-when-non-null case; the Discount line keeps its existing
hide-at-zero behavior.

Covered by unit tests: `null` storage hides the row, zero storage renders
`$0.00`, and zero discount still hides — no device screenshots needed.


app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/util/BigDecimalExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Formats this amount as a negative money string for the Discount line item (e.g. "-$5.00"),
+ * returning `null` when the amount is `null` or non-positive so the row is omitted when there is
+ * no discount.
+ */
+fun BigDecimal?.toDiscountMoneyText(currencyFormatter: NumberFormat): String? =
+    this
+        ?.takeIf { it.signum() > 0 }
+        ?.let { "-${currencyFormatter.format(it)}" }
```
david-livefront: I know this is technically unchanged from the previous code but should this minus be `\u2212`?

SaintPatrck: Good call. Yes, since we're dealing with dollhairs

david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #7009, opened 2026-06-02
[PM-37571] feat: Map Passport and License to SDK types

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37571

## 📔 Objective

Wire Passport and Driver's License through create, save, and the vault list now that their SDK mapping is available:

- Map `driversLicense`/`passport` in the cipher upload request so the server accepts creation.
- Allow Driver's License to save by marking the type SDK-supported and mapping it to its cipher type.
- Render both types in the main vault list.


app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModelTest.kt
```diff
@@ -2521,13 +2420,13 @@ class VaultAddEditViewModelTest : BaseViewModelTest() {
...
     @Test
-    fun `ItemType License should expose DRIVERS_LICENSE itemTypeOption and not be SDK supported`() {
+    fun `ItemType License should expose DRIVERS_LICENSE itemTypeOption and be SDK supported`() {
         val itemType = VaultAddEditState.ViewState.Content.ItemType.License()
         assertEquals(
             VaultAddEditState.ItemTypeOption.LICENSE,
             itemType.itemTypeOption,
         )
-        assertFalse(itemType.isSdkSupported)
+        assertTrue(itemType.isSdkSupported)
```
david-livefront, line 2429: Didn't you remove this property completely?

SaintPatrck: Not yet, but we can now that all the types are present and supported. Shouldn't be too many references if you're okay with removing it in this PR.

david-livefront (approved):

SaintPatrck:
Thanks @david-livefront

PR #7013, opened 2026-06-03
🍒 [PM-38364] fix: Multiply subscription line-item cost by quantity

SaintPatrck:
## 🎟️ Tracking

Cherry-pick of #7012 onto `release/2026.5-rc55`.

https://bitwarden.atlassian.net/browse/PM-38364

## 📔 Objective

Subscription cart totals were computed from each line item's per-unit `cost` without applying its `quantity`, so multi-unit additional storage (e.g. 3 GB at $4/GB) was presented as a single unit — undercharging the displayed storage cost and next-charge total.

This corrects the mapping so each line item contributes `cost × quantity`. It also brings two discount behaviors into parity with the web client's `cart-summary` component, verified against the `server` contract:

- The Password Manager seats item-level discount is now applied (against the seats line total) in addition to the cart-level discount; both are combined into the surfaced discount amount.
- Percent-off discounts treat a value below 1 as an already-decimal fraction and round half-up, matching the web client's arithmetic.

Item-level discounts on other line items remain intentionally unapplied, mirroring the web client's current scope. The single-object discount model was verified against the server source and required no change.

## 📸 Screenshots

| Before | After |
|--------|--------|
|  |  |


andrebispo5 (approved):
go go go

PR #7048, opened 2026-06-09
[PM-38808] Support optional encrypted Cipher.name

shane-melton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38808

## 📔 Objective

The SDK is making the encrypted `Cipher.name` field optional (`String` → `String?`) as part of the blob encryption migration ([sdk-internal#1122](https://github.com/bitwarden/sdk-internal/pull/1122)) — blob-encrypted ciphers carry the name inside the sealed `data` blob rather than as a top-level field.

`Cipher.toFailureCipherListView()` builds a `CipherListView` (whose `name` is non-null) directly from the encrypted `Cipher`, so the now-nullable `name` no longer compiles. This falls back to an empty string when the name is absent.

The decrypted `CipherView` / `CipherListView` `name` fields are unchanged, so all UI and sorting is unaffected.

> Pairs with the SDK version bump.


david-livefront:
This has been addressed in [this PR](https://github.com/bitwarden/android/pull/7096).

PR #7055, opened 2026-06-12
PM-38998: Feat: Apply optional auth token to config endpoint

david-livefront:
## 🎟️ Tracking

[PM-38998](https://bitwarden.atlassian.net/browse/PM-38998)

## 📔 Objective

This PR applies the `Authorization` header when an access token is available for the config endpoint. This will allow us to have per-user feature flags apply when authenticated.

[PM-38998]: https://bitwarden.atlassian.net/browse/PM-38998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/network/src/main/kotlin/com/bitwarden/network/interceptor/AuthTokenManager.kt
+++ b/network/src/main/kotlin/com/bitwarden/network/interceptor/AuthTokenManager.kt
@@ -19,7 +19,6 @@ import java.time.Clock
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 
-private const val MISSING_TOKEN_MESSAGE: String = "Auth token is missing!"
 private const val MISSING_PROVIDER_MESSAGE: String = "Refresh token provider is missing!"
 private const val EXPIRATION_OFFSET_MINUTES: Long = 5L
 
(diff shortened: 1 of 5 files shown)


network/src/main/kotlin/com/bitwarden/network/interceptor/AuthTokenManager.kt
```diff
@@ -83,8 +82,10 @@ internal class AuthTokenManager(
 
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
-        val token = getAccessToken()
-            ?: throw IOException(IllegalStateException(MISSING_TOKEN_MESSAGE))
+        val token = getAccessToken() ?: run {
+            Timber.w("Auth token is missing! Proceeding without token.")
```
aj-rosado, line 86: Wondering if we would want to make a new call after the user authenticates and has a token

david-livefront: Good call, I have update it to make the call when the environment or user changes

aj-rosado (approved):

david-livefront:
Thanks @aj-rosado

PR #7070, opened 2026-06-17
[PM-39185] deps: Update kotlinx-collections-immutable to 0.5.0

DmitryNekrasov:
Update kotlinx.collections.immutable to 0.5.0, the first stable release in the 0.5.x line. 0.5.x renames the copy-returning `Persistent*` methods ([KEEP-0459]) and deprecates the old names, but the app uses only the read-only `ImmutableList` / `ImmutableMap` / `ImmutableSet` interfaces, so no call sites change.

[KEEP-0459]: https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0459-naming-conventions-for-copy-returning-operations.md


david-livefront (approved):

david-livefront:
This was updated as part of the greater Kotlin update [PR found here](https://github.com/bitwarden/android/pull/7086)

PR #7102, opened 2026-06-24
Bug: Handle token refreshes when token is not present

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

Stumbled upon a crash that can occur now that we allow optionally authenticated calls to be made. This means the access token might not be present when the authenticator is invoked (and that's OK)


SaintPatrck (approved):

david-livefront:
Thanks @SaintPatrck

PR #7115, opened 2026-06-29
[PM-39762] bug: Fix RTL divider padding

helpingstar:
## 🎟️ Tracking

No linked issue. While reviewing RTL layout behavior, I noticed that `topDivider` and `bottomDivider` were resolving `paddingStart` and `paddingEnd` to the wrong physical sides when `LayoutDirection.Rtl` is active.

## 📔 Objective

Fix RTL divider padding in `ModifierExtensions`.

`paddingStart` should resolve to the right side in RTL layouts, and `paddingEnd` should resolve to the left side. This updates `topDivider` and `bottomDivider` so dividers with start padding, such as card/list item dividers, are drawn with the correct inset in RTL.

## 📸 Screenshots

Force RTL layout direction was enabled from Android Developer options to isolate RTL behavior while keeping the app text readable.

| Before - English with Force RTL | After - English with Force RTL |
| --- | --- |
| before-force-rtl | after-force-rtl |

| After - LTR |
| --- |
| after-ltr |

The visible divider inset moves to the correct leading side in the Force RTL after screenshot, while LTR remains unchanged.

### DIFF
--- a/ui/src/main/kotlin/com/bitwarden/ui/platform/base/util/ModifierExtensions.kt
+++ b/ui/src/main/kotlin/com/bitwarden/ui/platform/base/util/ModifierExtensions.kt
@@ -186,7 +186,7 @@ fun Modifier.topDivider(
         if (enabled) {
             val (startX, endX) = when (layoutDirection) {
                 LayoutDirection.Ltr -> paddingStart.toPx() to (size.width - paddingEnd.toPx())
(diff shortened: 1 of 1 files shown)


david-livefront (approved):

david-livefront:
Thanks for your contribution @helpingstar

PR #7120, opened 2026-06-30
[PM-39432] chore: pass null Send repository in SdkRepositoryFactoryImpl

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39432

## 📔 Objective

Paired with [bitwarden/sdk-internal#1217](https://github.com/bitwarden/sdk-internal/pull/1217), which adds a `Send` slot to the SDK's client-managed `Repositories` struct (via the `create_client_managed_repositories!` macro in `bitwarden-pm/src/migrations.rs`). The new field surfaces through the UniFFI Kotlin binding as a required `send` constructor parameter on `com.bitwarden.sdk.Repositories`.

This PR adds `send = null` to the call site in `SdkRepositoryFactoryImpl` so the Kotlin code compiles against the new SDK. No behavioral change — Android doesn't persist Sends locally today, so passing `null` matches the existing pattern used for `folder`, `userKeyState`, `ephemeralPinEnvelopeState`, and `organizationSharedKey`.

Without this paired change, the SDK PR's breaking-change detection job for Android fails with:

```
e: SdkRepositoryFactoryImpl.kt:37:13 No value passed for parameter 'send'.
```

When Android starts persisting Sends locally, this `null` can be replaced with a real `SdkSendRepository` implementation, similar to how `cipher` and `localUserDataKeyState` work today.

## 📸 Screenshots

N/A — no UI changes.


vvolkgang:
On hold until https://github.com/bitwarden/sdk-internal/pull/1217 is merged.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/sdk/SdkRepositoryFactoryImpl.kt
```diff
@@ -35,6 +35,7 @@ class SdkRepositoryFactoryImpl(
             ),
             ephemeralPinEnvelopeState = null,
             organizationSharedKey = null,
+            send = null,
```
david-livefront, line 38: Do we need to implement this any time soon?

adudek-bw: Not unless it's already on your roadmap :)

david-livefront (approved):

PR #7123, opened 2026-06-30
[PM-39807] fix: Gate fill-assist rules sync on user settings

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39807

## 📔 Objective

`FillAssistManagerImpl.syncIfNecessary()` was fetching fill-assist targeting rules from the server as soon as the server config became available and the `FillAssistTargetingRules` feature flag was enabled — regardless of whether the user had opted into fill assist in their settings.

This PR gates the sync behind `settingsRepository.isFillAssistEnabled`, so rules are only fetched once the user explicitly enables fill assist. The check is short-circuited alongside the existing feature-flag guard in `syncIfNecessary()`.

Changes:
- Added `var isFillAssistEnabled: Boolean` to `SettingsRepository` / `SettingsRepositoryImpl`, backed by a new `getFillAssistEnabled` / `storeFillAssistEnabled` pair in `SettingsDiskSource` / `SettingsDiskSourceImpl`
- Added `settingsRepository: SettingsRepository` to `FillAssistManagerImpl` and `FillAssistModule`
- Updated `syncIfNecessary()` to return early when `isFillAssistEnabled` is `false`
- Added test coverage for the new early-return path


david-livefront (approved):

aj-rosado:
thanks @david-livefront

PR #7131, opened 2026-07-01
[PM-39840] fix: Use dedicated OkHttpClient for fill-assist to prevent load-balancer prompt

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39840

## 📔 Objective

The fill-assist service fetches targeting rules from `https://github.com/bitwarden/map-the-web/releases/latest/download/manifest.json`. GitHub's CDN responds with an HTTP 302 redirect for content delivery.

The `CookieInterceptor`, present on the shared unauthenticated `OkHttpClient`, treats any 302 response as a Bitwarden load-balancer auth redirect. It calls `cookieProvider.acquireCookies(hostname)` as a side effect before throwing `CookieRedirectException`, which navigates the app to the "Sync with Browser" load-balancer screen — even though the 302 was a legitimate CDN redirect unrelated to Bitwarden authentication.

This affected users on self-hosted environments without a load balancer (USQA, USDEV, and similar), blocking sign-in and vault unlock.

**Fix:** Introduce a dedicated `fillAssistOkHttpClient` in `RetrofitsImpl` that keeps headers and SSL certificate validation but excludes `CookieInterceptor` and `PermissionInterceptor`. `fillAssistRetrofit` now uses this client so OkHttp's default redirect handling processes the CDN 302 correctly.

### DIFF
--- a/network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
+++ b/network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
@@ -113,7 +113,7 @@ internal class RetrofitsImpl(
 
     // For requests to external (non-Bitwarden) URLs. CookieInterceptor must be excluded because
     // it treats all 302s as Bitwarden load-balancer auth redirects, which is only correct for
-    // Bitwarden own infrastructure.
+    // Bitwarden's own infrastructure.
     private val externalOkHttpClient: OkHttpClient = OkHttpClient.Builder()
         .addInterceptor(headersInterceptor)
         .configureSsl(certificateProvider = certificateProvider)
--- a/network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
+++ b/network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
@@ -111,9 +111,10 @@ internal class RetrofitsImpl(
         .configureSsl(certificateProvider = certificateProvider)
         .build()
 
-    // Fill-assist might use HTTP 302 redirects for content delivery.
-    // CookieInterceptor treats all 302s as auth redirects.
-    private val fillAssistOkHttpClient: OkHttpClient = OkHttpClient.Builder()
+    // For requests to external (non-Bitwarden) URLs. CookieInterceptor must be excluded because
+    // it treats all 302s as Bitwarden load-balancer auth redirects, which is only correct for
+    // Bitwarden's own infrastructure.
+    private val externalOkHttpClient: OkHttpClient = OkHttpClient.Builder()
         .addInterceptor(headersInterceptor)
         .configureSsl(certificateProvider = certificateProvider)
         .build()
@@ -175,7 +176,7 @@ internal class RetrofitsImpl(
         baseRetrofit
             .newBuilder()
             .client(
-                fillAssistOkHttpClient
+                externalOkHttpClient
                     .newBuilder()
                     .addInterceptor(baseUrlInterceptor)
                     .addInterceptor(loggingInterceptor)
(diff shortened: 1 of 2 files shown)


network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
```diff
@@ -111,6 +111,14 @@ internal class RetrofitsImpl(
         .configureSsl(certificateProvider = certificateProvider)
         .build()
 
+    // Fill-assist might use HTTP 302 redirects for content delivery.
+    // CookieInterceptor treats all 302s as auth redirects and PermissionInterceptor does a DNS
+    // preflight that can fail spuriously — both must be excluded from this client.
+    private val fillAssistOkHttpClient: OkHttpClient = OkHttpClient.Builder()
+        .addInterceptor(headersInterceptor)
+        .configureSsl(certificateProvider = certificateProvider)
+        .build()
```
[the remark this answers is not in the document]
aj-rosado: Re-added

network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
```diff
@@ -162,5 +170,19 @@ internal class RetrofitsImpl(
...
+        baseRetrofit
+            .newBuilder()
+            .client(
+                fillAssistOkHttpClient
+                    .newBuilder()
+                    .addInterceptor(baseUrlInterceptor)
+                    .addInterceptor(loggingInterceptor)
+                    .build(),
+            )
+            .build()
```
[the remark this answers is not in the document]
aj-rosado: Added missing tests

network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
```diff
@@ -111,6 +111,13 @@ internal class RetrofitsImpl(
         .configureSsl(certificateProvider = certificateProvider)
         .build()
 
+    // Fill-assist might use HTTP 302 redirects for content delivery.
+    // CookieInterceptor treats all 302s as auth redirects.
+    private val fillAssistOkHttpClient: OkHttpClient = OkHttpClient.Builder()
```
david-livefront: Can we make this more generic? I can envision a future API that also need to avoid the Cookie Interceptor.

Something that makes it more reusable and not explicitly tied to the FillAssist feature.

network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
```diff
@@ -162,5 +170,20 @@ internal class RetrofitsImpl(
             )
             .build()
 
+    private fun createFillAssistRetrofit(
```
david-livefront: Can we rename this function to `createExternalRetrofit` to match the OkHttpClient too.

network/src/main/kotlin/com/bitwarden/network/retrofit/RetrofitsImpl.kt
```diff
@@ -113,7 +113,7 @@ internal class RetrofitsImpl(
 
     // For requests to external (non-Bitwarden) URLs. CookieInterceptor must be excluded because
     // it treats all 302s as Bitwarden load-balancer auth redirects, which is only correct for
-    // Bitwarden's own infrastructure.
+    // Bitwarden own infrastructure.
```
david-livefront: This should have an `s`, right?

aj-rosado: yes

david-livefront (approved):

PR #7146, opened 2026-07-07
[PM-39612] fix: resolve autofill container-hint redirect to autofillable child

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39612

## 📔 Objective

Some Android apps wrap their login `EditText` fields inside a container view (e.g., Material `TextInputLayout`). The container carries the semantic autofill hint (`USERNAME`, `PASSWORD`) but has `autofillType = NONE (0)`, meaning it cannot receive a fill value. The inner `EditText` has `autofillType = TEXT (1)` but no hint, so heuristics never classify it. As a result, autofill silently failed — the fill dataset was built against the container's `autofillId` with `autofillType = 0`, which the framework dropped at fill time.

**Changes:**
- `ViewNodeExtensions.kt`: When a container node carries a semantic hint but has `autofillType = NONE`, `findFirstAutofillableChild()` locates the first descendant with `autofillType = TEXT`. The child's `autofillId` and `autofillType` are used in `AutofillView.Data` so the fill reaches the actual input field.
- `AutofillParserImpl.kt`: Added `claimedAutofillIds` tracking during traversal to prevent a child node from being added a second time after the container has already redirected to it.


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/util/ViewNodeExtensions.kt
```diff
@@ -53,25 +54,57 @@ fun AssistStructure.ViewNode.toAutofillView(
...
         website = this.website ?: parentWebsite,
     )
-    return buildAutofillView(
+    val view = buildAutofillView(
         autofillOptions = autofillOptions,
         autofillViewData = autofillViewData,
-        autofillHint = this.supportedAutofillHint,
+        autofillHint = hint,
     )
+    return view
```
david-livefront: Why did we add the `view` property, can we leave it as a return?

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -302,8 +310,22 @@ private fun AssistStructure.ViewNode.traverse(
...
+                    // filter out existing AutofillIds to avoid duplicates
+                    .filter { view ->
+                        val id = view.data.autofillId
+                        if (id in claimedAutofillIds) {
+                            return@filter false
+                        }
+                        if (view !is AutofillView.Unused) {
+                            claimedAutofillIds.add(id)
+                        }
+                        true
```
david-livefront: What do you think of this?

```kotlin
val id = view.data.autofillId
if (id in claimedAutofillIds) {
    false
} else if (view !is AutofillView.Unused) {
    claimedAutofillIds.add(id)
    true
} else {
    true
}
```

david-livefront (approved):

aj-rosado:
Thank you @david-livefront

PR #7152, opened 2026-07-09
Remove Credential Exchange dependency on f-droid builds

vvolkgang:
## 🎟️ Tracking

PM-27903

## 📔 Objective

For f-droid builds Credential Exchange is disabled and the underline dependency fully removed (`libs.androidx.credentials.providerevents.play.services`).


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/importitems/ImportItemsScreen.kt
```diff
@@ -98,6 +98,7 @@ fun ImportItemsScreen(
         onNavigateBack = handler.onNavigateBack,
         onImportFromComputerClick = handler.onImportFromComputerClick,
         onImportFromAnotherAppClick = handler.onImportFromAnotherAppClick,
+        isImportFromAnotherAppSupported = credentialExchangeImporter.isSupported(),
```
david-livefront, line 101: Thoughts on making `isSupported()` into a `StateFlow`?

Since this is used in the UI, we can observe it as state and it'll make tests way easier. Plus it makes it easier if we ever need to change it dynamically in the future.

```kotlin
    val isSupported: StateFlow<Boolean>
```

```kotlin
     val isImportFromAnotherAppSupported by credentialExchangeImporter
            .isSupported
            .collectAsStateWithLifecycle()
    ImportItemsScaffold(
        onNavigateBack = handler.onNavigateBack,
        onImportFromComputerClick = handler.onImportFromComputerClick,
        onImportFromAnotherAppClick = handler.onImportFromAnotherAppClick,
        isImportFromAnotherAppSupported = isImportFromAnotherAppSupported,
        snackbarHostState = snackbarHostState,
    )
```

PR #7158, opened 2026-07-10
PM-40154: feat: Migrate user data to KeystoreEncryptedSharedPreferences

david-livefront:
## 🎟️ Tracking

[PM-40154](https://bitwarden.atlassian.net/browse/PM-40154)

## 📔 Objective

This PR creates a new `KeystoreEncryptedSharedPreferences` that utilizes the `AndroidKeystore` to encrypt the data stored.

Notable changes:
* This migrates us away from `EncryptedSharedPreferences` which is deprecated and is not FIPS compliant.
* The keys are not stored as plaintext.
* We currently only support the storing of `Strings` (this is the only type we need at the moment).

[PM-40154]: https://bitwarden.atlassian.net/browse/PM-40154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
@@ -21,6 +21,7 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.onSubscription
 import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.json.Json
+import timber.log.Timber
 import java.time.Instant
 import java.util.UUID
 
@@ -33,6 +34,7 @@ private const val BIOMETRICS_UNLOCK_KEY = "userKeyBiometricUnlock"
 private const val USER_AUTO_UNLOCK_KEY_KEY = "userKeyAutoUnlock"
 private const val DEVICE_KEY_KEY = "deviceKey"
 private const val PENDING_ADMIN_AUTH_REQUEST_KEY = "pendingAdminAuthRequest"
+private const val ACCOUNT_CRYPTOGRAPHIC_STATE_KEY = "accountCryptographicState"
 
 // These keys should not be encrypted
 private const val UNIQUE_APP_ID_KEY = "appId"
@@ -58,7 +60,6 @@ private const val ONBOARDING_STATUS_KEY = "onboardingStatus"
 private const val SHOW_IMPORT_LOGINS_KEY = "showImportLogins"
 private const val LAST_LOCK_TIMESTAMP = "lastLockTimestamp"
 private const val PROFILE_ACCOUNT_KEYS_KEY = "profileAccountKeys"
-private const val ACCOUNT_CRYPTOGRAPHIC_STATE_KEY = "accountCryptographicState"
 private const val V2_UPGRADE_TOKEN = "v2UpgradeToken"
 
 /**
@@ -67,11 +68,13 @@ private const val V2_UPGRADE_TOKEN = "v2UpgradeToken"
 @Suppress("TooManyFunctions")
 class AuthDiskSourceImpl(
     encryptedSharedPreferences: SharedPreferences,
+    keystoreEncryptedPreferences: SharedPreferences,
     sharedPreferences: SharedPreferences,
     legacySecureStorageMigrator: LegacySecureStorageMigrator,
     private val json: Json,
 ) : BaseEncryptedDiskSource(
     encryptedSharedPreferences = encryptedSharedPreferences,
+    keystoreEncryptedPreferences = keystoreEncryptedPreferences,
     sharedPreferences = sharedPreferences,
 ),
     AuthDiskSource {
@@ -132,6 +135,9 @@ class AuthDiskSourceImpl(
         // We must migrate the Private Key and Account Keys to use the Account Cryptographic state
         // from now on.
         migrateAccountKeys()
+
+        // Migrate to the Keystore Encrypted SharedPreferences.
+        migrateToKeystoreEncryption()
     }
 
     override var authenticatorSyncSymmetricKey: ByteArray?
@@ -285,10 +291,7 @@ class AuthDiskSourceImpl(
     }
 
     override fun getUserAutoUnlockKey(userId: String): String? =
-        getEncryptedString(
-            key = USER_AUTO_UNLOCK_KEY_KEY.appendIdentifier(userId),
-            default = null,
-        )
+        getEncryptedString(key = USER_AUTO_UNLOCK_KEY_KEY.appendIdentifier(userId))
 
     override fun storeUserAutoUnlockKey(
         userId: String,
@@ -739,5 +742,21 @@ class AuthDiskSourceImpl(
                     putString(key = privateKeyKey, value = null)
                 }
             }
+    }
+
+    private fun migrateToKeystoreEncryption() {
+        var isMigrated = false
+        isMigrated = migrateKeyByPrefix(keyPrefix = AUTHENTICATOR_SYNC_SYMMETRIC_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = AUTHENTICATOR_SYNC_UNLOCK_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = ACCOUNT_CRYPTOGRAPHIC_STATE_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = USER_AUTO_UNLOCK_KEY_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = DEVICE_KEY_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = PENDING_ADMIN_AUTH_REQUEST_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = BIOMETRICS_INIT_VECTOR_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = BIOMETRICS_UNLOCK_KEY) || isMigrated
+        isMigrated = migrateKeyByPrefix(keyPrefix = ACCOUNT_TOKENS_KEY) || isMigrated
+        if (isMigrated) {
+            Timber.d("AuthDiskSource has been migrated to keystore encrypted shared preferences")
+        }
     }
 }
--- a/data/src/main/kotlin/com/bitwarden/data/datasource/disk/BaseEncryptedDiskSource.kt
+++ b/data/src/main/kotlin/com/bitwarden/data/datasource/disk/BaseEncryptedDiskSource.kt
@@ -2,31 +2,49 @@ package com.bitwarden.data.datasource.disk
 
 import android.content.SharedPreferences
 import androidx.core.content.edit
-import androidx.security.crypto.EncryptedSharedPreferences
+import timber.log.Timber
+
+private const val LEGACY_PREFIX: String = "bwSecureStorage:"
 
 /**
- * Base class for simplifying interactions with [SharedPreferences] and
- * [EncryptedSharedPreferences].
+ * Base class for simplifying interactions with [SharedPreferences], this includes both regular
+ * and encrypted storage.
  */
 @Suppress("UnnecessaryAbstractClass")
 abstract class BaseEncryptedDiskSource(
     sharedPreferences: SharedPreferences,
     private val encryptedSharedPreferences: SharedPreferences,
-) : BaseDiskSource(
-    sharedPreferences = sharedPreferences,
-) {
+    private val keystoreEncryptedPreferences: SharedPreferences,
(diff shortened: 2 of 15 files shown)


data/src/main/kotlin/com/bitwarden/data/datasource/disk/BaseEncryptedDiskSource.kt
```diff
@@ -12,18 +12,33 @@ import androidx.security.crypto.EncryptedSharedPreferences
...
-        default: String? = null,
-    ): String? = encryptedSharedPreferences.getString(key.withBase(), default)
+    ): String? = keystoreEncryptedPreferences.getString(key.withBase(), null)
 
     protected fun putEncryptedString(
         key: String,
         value: String?,
-    ): Unit = encryptedSharedPreferences.edit { putString(key.withBase(), value) }
+    ) {
+        keystoreEncryptedPreferences.edit { putString(key.withBase(), value) }
```
vvolkgang: 💭 Is this prefix still useful as an easy way to identify encrypted data in SharedPrefs or should we take the opportunity and drop it?

david-livefront: Prefix is gone!

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.bitwarden.core.data.util.flatMap
+import java.security.Key
+import javax.crypto.Cipher
+import javax.crypto.Mac
+import javax.crypto.spec.IvParameterSpec
+
+private const val IV_SIZE_BYTES: Int = 16
+private const val CIPHER_TRANSFORMATION: String = KeyProperties.KEY_ALGORITHM_AES + "/" +
+    KeyProperties.BLOCK_MODE_CBC + "/" +
+    KeyProperties.ENCRYPTION_PADDING_PKCS7
```
david-livefront: So this currently uses `AES/CBC/PKCS7Padding` for the transformation.

I have a separate stash of changes that would allow us to use `AES/GCM/NoPadding` instead. Do we have a preference?

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+        .map {
+            val cipherText = bytes.copyOfRange(fromIndex = IV_SIZE_BYTES, toIndex = bytes.size)
+            it.doFinal(cipherText)
+        }
+
+    override fun encrypt(
+        alias: String,
+        bytes: ByteArray,
+    ): Result<ByteArray> = keystoreManager
+        .getOrCreateKey(alias = KeyAlias.Aes(name = alias))
```
quexten: In general, it is dangerous cryptographic design to offer separate encrypt and sign (actually, this should be authenticate, sign usually is used in asymmetric contexts) primitives. You should only ever offer authenticated encryption that does both.

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/KeystoreManagerImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+        alias: KeyAlias,
+    ): KeyGenParameterSpec =
+        when (alias) {
+            is KeyAlias.Aes -> {
+                KeyGenParameterSpec
+                    .Builder(
+                        alias.name.formatAlias(),
+                        KeyProperties.PURPOSE_ENCRYPT or KeyProperties.PURPOSE_DECRYPT,
+                    )
+                    .setBlockModes(KeyProperties.BLOCK_MODE_CBC)
```
quexten: Replace with BLOCK_MODE_GCM. This makes it significantly simpler for you because GCM both encrypts *and* authenticates. Currently, it is very easy to insecurely mis-use the new API. GCM is also a fips approved security function.

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/KeystoreManagerImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_PKCS7)
```
quexten: Remove when using gcm

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/KeystoreManagerImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+                    .setKeySize(KEY_SIZE)
+                    .build()
+            }
+
+            is KeyAlias.Mac -> {
```
quexten: Remove the mac key entirely

quexten (dismissed):
If we had a tech breakdown beforehand, I'd recommend implementing this in the SDK based on our modern encryption primitives instead. Given the stage, I'll comment the minimum to make this reasonably secure. I recommend dropping the separate MAC key, and switching from CBC to GCM. Currently, the API is very easy to misuse, and you generally do not want to expose separate authentication and encryption (without authentication) APIs.

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
quexten: Saw this after my review. Please do not use CBC. Use GCM, and do not MAC separately. CBC by itself is dangerous to use, and needs a mac to go along with it.

data/src/main/kotlin/com/bitwarden/data/datasource/disk/KeystoreEncryptedSharedPreferences.kt
```diff
@@ -0,0 +1,168 @@
...
+            .map { Base64.getEncoder().encodeToString(it) }
+            .onSuccess { editor.remove(it) }
+        return this
+    }
+
+    private fun encryptAndPutByteArray(
+        key: String,
+        value: ByteArray,
+    ): SharedPreferences.Editor {
+        val encryptedKey = encryptionManager
```
quexten: Oh.. are you deliberately not encrypting the keys, but just want to auth them, but the values should be encrypted *and authenticated*?

quexten: @david-livefront Could you clarify what the intent of protecting the "Key" entries here is? The values should be encrypted and authenticated because they contain secret values, and tampering with them may affect the security of the app, so much is clear. Thus we use authenticated encryption for them. But what about the keys? Is there a perceived threat here?

Depending on what the answer to this is I would recommend either:
1. Use AES-GCM for values as authenticated encryption, do not encrypt or authenticate the key entries whatsoever

(if you decide that the keys are worth protecting integrity of)
2. Use AES-GCM for the values, put the key into the authenticated additional data of AES-GCM, when encrypting the value
  - This binds together the value + key and makes it not possible to swap between different value + key pairs.

However, so far I believe 1. to be the correct path forward unless I'm missing context.

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
david-livefront: Changes have been made to use GCM instead.

data/src/main/kotlin/com/bitwarden/data/datasource/disk/KeystoreEncryptedSharedPreferences.kt
david-livefront: @quexten We are encrypting the keys for two main reasons.

The previous implementation encrypted the keys and I was attempting to match behavior . They used `AES256_SIV` for this because it is required for the keys to be deterministically encrypted (otherwise we can never retrieve the data). `AES256_SIV` is not a compliant option, so I had to come up with something else.

The only threat model I can think of if that if someone were to gain access to the file and attempted to brute force the data. The keys provide information on what each value is, and therefore, which values to brute force first. If this is not a concern, then I can easily drop addition logic and simplify everything.

CC @vvolkgang

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/KeystoreManagerImpl.kt
david-livefront: Done

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/KeystoreManagerImpl.kt
david-livefront: Done

data/src/main/kotlin/com/bitwarden/data/datasource/disk/KeystoreEncryptedSharedPreferences.kt
quexten: Right. 

> The only threat model I can think of if that if someone were to gain access to the file and attempted to brute force the data. The keys provide information on what each value is, and therefore, which values to brute force first. If this is not a concern, then I can easily drop addition logic and simplify everything.

1. Since we are encrypting with a high-entropy key stored in the key-chain, we can assume brute-force is not possible for an aes256-gcm key.
2. With the current design, you would still know what key is which, since we are authenticating but not encrypting them

-> I would recommend dropping the mac on the key and leaving it plaintext, and encrypting the values only. Ideally, we would context bind them so they cannot be swapped between keys, but doing so here would further complicate this PR / it is not blocking from my perspective, and would increase the scope, so I would recommend not doing so here.

david-livefront: Thanks @quexten I have updated the PR.
* The `sign` function is completely gone
* We leave the keys as plaintext
* We now use `AES/GCM/NoPadding` for encrypting the values

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
```diff
@@ -0,0 +1,79 @@
...
+        .mapCatching {
+            val cipherText = bytes.copyOfRange(fromIndex = IV_SIZE_BYTES, toIndex = bytes.size)
+            it.doFinal(cipherText)
+        }
+
+    override fun encrypt(
+        alias: String,
+        bytes: ByteArray,
+    ): Result<ByteArray> = keystoreManager
+        .getOrCreateKey(alias = alias)
```
quexten, line 44: I don't understand the kotlin syntax too well here. Is the IV generated by the javax.crypto.cipher api?

quexten: I don't have a kotlin dev environment set up currently. @david-livefront could you confirm that the java crypto API here samples a random fresh IV on every invocation? If so I'm happy to approve the PR.

(IV / in gcm's case "nonce" re-use leads to a catastrophic cryptographic break, so we want to ensure nonces are correctly randomly sampled).

david-livefront: That is correct, by default a new random IV is generated everytime.

core/src/main/kotlin/com/bitwarden/core/data/manager/encryption/EncryptionManagerImpl.kt
```diff
@@ -0,0 +1,79 @@
...
+        when (initCipher) {
+            is InitCipher.Decrypt -> {
+                init(
+                    Cipher.DECRYPT_MODE,
+                    this@generateCipher,
+                    GCMParameterSpec(TAG_LENGTH, initCipher.iv),
+                )
+            }
+
+            is InitCipher.Encrypt -> init(Cipher.ENCRYPT_MODE, this@generateCipher)
```
david-livefront, line 77: @quexten Here where we initialize the cipher to encrypt data, it is creating a new IV since we are not providing one.

Above, on line 73, we provide the stored IV for decryption.

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/datasource/disk/AuthDiskSourceImpl.kt
```diff
@@ -132,6 +135,9 @@ class AuthDiskSourceImpl(
         // We must migrate the Private Key and Account Keys to use the Account Cryptographic state
         // from now on.
         migrateAccountKeys()
+
+        // Migrate to the Keystore Encrypted SharedPreferences.
+        migrateToKeystoreEncryption()
```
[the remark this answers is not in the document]
aj-rosado: @claude this will only affect V2 accounts and that FF is currently off so does not affect any current user. This will be fixed on a follow up task

aj-rosado (approved):

PR #7186, opened 2026-07-21
PM-40657: feat: Add support for disableUserRegistration boolean

david-livefront:
## 🎟️ Tracking

[PM-35124](https://bitwarden.atlassian.net/browse/PM-35124)
[PM-40675](https://bitwarden.atlassian.net/browse/PM-40675)
[PM-40657](https://bitwarden.atlassian.net/browse/PM-40677)

## 📔 Objective

This PR adds the `setting` to the `ConfigResponseJson` and uses that data to hide the `Create account` button on the Landing screen and the Welcome screen.

[PM-40657]: https://bitwarden.atlassian.net/browse/PM-40657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-40675]: https://bitwarden.atlassian.net/browse/PM-40675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-35124]: https://bitwarden.atlassian.net/browse/PM-35124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/landing/LandingViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/landing/LandingViewModel.kt
@@ -4,6 +4,8 @@ import android.os.Parcelable
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
 import com.bitwarden.core.data.manager.model.FlagKey
+import com.bitwarden.data.datasource.disk.model.ServerConfig
+import com.bitwarden.data.repository.ServerConfigRepository
 import com.bitwarden.data.repository.model.Environment
 import com.bitwarden.ui.platform.base.BackgroundEvent
 import com.bitwarden.ui.platform.base.BaseViewModel
@@ -37,12 +39,13 @@ private const val KEY_STATE = "state"
 /**
  * Manages application state for the initial landing screen.
  */
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LongParameterList")
 @HiltViewModel
 class LandingViewModel @Inject constructor(
     private val authRepository: AuthRepository,
     private val vaultRepository: VaultRepository,
     private val environmentRepository: EnvironmentRepository,
+    serverConfigRepository: ServerConfigRepository,
     snackbarRelayManager: SnackbarRelayManager<SnackbarRelay>,
     savedStateHandle: SavedStateHandle,
     featureFlagManager: FeatureFlagManager,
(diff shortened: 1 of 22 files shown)


network/src/main/kotlin/com/bitwarden/network/model/ConfigResponseJson.kt
```diff
@@ -122,4 +126,16 @@ data class ConfigResponseJson(
...
+     * Represents the settings configuration in the configuration response.
+     *
+     * @param disableUserRegistration If `true`, the environment does not allow for new user
+     * registration.
+     */
+    @Serializable
+    data class SettingJson(
+        @SerialName("disableUserRegistration")
+        val disableUserRegistration: Boolean,
+    )
```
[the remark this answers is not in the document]
david-livefront: @claude This boolean cannot be null when the setting object is present.

app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/landing/LandingViewModel.kt
```diff
@@ -317,7 +346,7 @@ data class LandingState(
      * Determines if the user should be allowed to create a new account.
      */
     val allowCreateAccount: Boolean
-        get() = selectedEnvironmentType != Environment.Type.FED_RAMP
+        get() = !disableCreateAccount && selectedEnvironmentType != Environment.Type.FED_RAMP
```
aj-rosado, line 349: 🤔 more of a thought, we are adding the `disableCreateAccount` for the `FED_RAMP` environment but we are keeping the hard check on the env.

As I believe we never want to display this on `FED_RAMP` env, also feel that it is more of a server responsability 🤔 also on other places we are simply relying on the `disableUserRegistration` value

aj-rosado (approved):
LGTM, just left a little non blocking comment

app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/landing/LandingViewModel.kt
david-livefront: This is the only screen that allows a full FedRamp type to be selected. Since we know the type directly here we can block account creation. Whereas on the StartRegistrationScreen, we completely filter out the FedRamp option. but the individual environments might still have the `disableUserRegistration` boolean set.

Keeping the check for the FED_RAMP type just means we do not need to wait for the config request go through. It's a minor optimization that was added to the requirements.

david-livefront:
Thanks @aj-rosado

PR #7194, opened 2026-07-22
[PM-40747] feat: Add vault folder delete and duplicate name validation

Ashishchidhava:
## 🎟️ Tracking

Local enhancement: folders on My vault had no delete action, and creating folders with the same name was allowed (e.g. two folders both named “Himalya”).

## 📔 Objective

Improve folder management from **My vault** and prevent duplicate folder names.

### 1. Delete / edit folder from My vault
- Folder rows on My vault now show a **⋯ more options** menu.
- Options:
  - **Edit** → opens the existing Edit folder screen
  - **Delete** → confirmation dialog, then deletes the folder and shows a snackbar
- Synthetic “No Folder” rows are unchanged (no overflow menu).
- Reuses existing `vaultRepository.deleteFolder` and edit-folder navigation (no new delete API).

### 2. Duplicate folder name validation
- Creating or renaming a folder to a name that already exists is blocked (**case-insensitive**).
- Nested names compare the full resolved path (e.g. `Parent/Child`).
- Edit mode still allows saving when the name is unchanged for that same folder.
- Applied in:
  - **Add/Edit folder** screen (`FolderAddEditViewModel`)
  - **Inline “add folder”** from item add/edit (`VaultAddEditViewModel`)
- The Add folder screen **subscribes to `foldersStateFlow`** so validation uses a loaded folder list (fixes a race where `.value` was often empty and duplicates were allowed).
- New string: `a_folder_with_this_name_already_exists` → “A folder with this name already exists.”

### 3. Supporting UI / navigation
- `BitwardenGroupItem` supports an optional overflow selection list (used by vault folder rows).
- Navigation wired for `NavigateToEditFolder` through vault → nav bar → unlocked graph.

### Tests
- `FolderAddEditViewModelTest`: duplicate name blocked; edit with unchanged name still saves.
- `VaultViewModelTest`: folder edit navigation, delete confirmation, successful delete snackbar.
- Screen/nav tests updated for the new `onNavigateToEditFolderScreen` callback.

## 📸 Screenshots

Verified on device:
- My vault folder rows show **⋯** with Edit / Delete
Screenshot 2026-07-22 at 16 46 14

- Saving a duplicate folder name shows the error dialog
Screenshot 2026-07-22 at 16 46 06
Screenshot 2026-07-22 at 16 45 52

## Test plan
- [ ] Open My vault → confirm each real folder row has a **⋯** menu
- [ ] **Edit** opens Edit folder and saving works
- [ ] **Delete** shows confirmation → folder is removed → snackbar appears
- [ ] Add folder with an existing name (same casing and different casing) → error: “A folder with this name already exists.”
- [ ] Rename a folder to another existing name → same error
- [ ] Rename a folder without changing the name → save succeeds
- [ ] Add a nested folder name that does not collide with an existing full path → save succeeds


Ashishchidhava:
Small cleanup note for reviewers: the example duplicate folder name in the description was just a local test string — happy to adjust copy or follow any Product/Engineering guidance on UX (e.g. overflow placement, delete confirmation wording).

Happy to iterate quickly on any review feedback.

gabrbrand:
I'm not from the Bitwarden team, but here are some thoughts (should be discussed with the team before implementing):
**Delete / edit folder from My vault**
These options should also be added to the overflow menu on the VaultItemListingScreen. The "Folders" option in "Settings" -> "Vault" wouldn't be necessary anymore. The "Delete" overflow menu item on the FolderAddEditScreen could also be removed because the user can do this using the added dialog options/overflow menu items.

**Duplicate folder name validation**
I would rather use the error state (isError parameter) of the BitwardenTextField composable instead of the current error dialog. ([M3 TextField Guidelines](https://m3.material.io/components/text-fields/guidelines#9ad90554-a793-4506-9075-6812fd7b381a))
The TextField value should be validated for each character typed by user (maybe with a certain delay) and if the name already exists an error text should be shown ("A Folder with this name already exists" or shorter). In addition to that, the "Save" button should be disabled.
grafik

When the folder name is blank the "Save" button should be disabled instead of showing an error dialog, when the user wants to save a folder with a blank name. Maybe use "Name (required)" as a label.

- An info should be added to the confirmation dialog explaining what happens to the items in the folder to be deleted.
Something like:
```
Delete Folder

Do you really want to delete? This cannot be undone.
The items and folders it contains will remain.

Cancel

Delete
```

- The text field should automatically receive focus when the FolderAddEditScreen opens, so that the user can start typing right away (use autoFocus parameter from BitwardenTextField).
- A note about nested folders should be added as supporting text of the folder name text field ("Nest a folder by adding the parent folder's name followed by a “/”. Example: Social/Forums")
grafik

or alternatively a dropdown menu for selecting a parent folder combined with the folder type made available in the add item dialog on the VaultItemListingScreen with the currently opened folder preselected. e.g. Tuta
grafik

- Remove deleted folder from any nested folders names. (Delete Test (child: Nested) -> Nested instead of Test/Nested)
- Pop deleted folder from back stack when deleting from another client or through Settings -> Vault -> Folders and navigate back to My vault tab.

PR #7198, opened 2026-07-22
PM-39902: feat: Add data field for updating a cipher with Blob encryption

david-livefront:
## 🎟️ Tracking

[PM-39902](https://bitwarden.atlassian.net/browse/PM-39902)

## 📔 Objective

This PT adds the `data` payload to the network request payloads responsible to for updating ciphers. This is required to ensure continued functionality with Blob encryption.

[PM-39902]: https://bitwarden.atlassian.net/browse/PM-39902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/network/src/testFixtures/kotlin/com/bitwarden/network/model/CipherJsonRequestUtil.kt
+++ b/network/src/testFixtures/kotlin/com/bitwarden/network/model/CipherJsonRequestUtil.kt
@@ -35,6 +35,7 @@ fun createMockCipherJsonRequest(
     lastKnownRevisionDate: Instant? = Instant.parse("2023-10-27T12:00:00Z"),
     key: String? = "mockKey-$number",
     archivedDate: Instant? = Instant.parse("2023-10-27T12:00:00Z"),
+    data: String? = "mockData-$number",
     encryptedFor: String? = "mockEncryptedFor-$number",
 ): CipherJsonRequest =
     CipherJsonRequest(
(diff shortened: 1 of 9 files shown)


network/src/testFixtures/kotlin/com/bitwarden/network/model/CipherJsonRequestUtil.kt
```diff
@@ -35,6 +35,7 @@ fun createMockCipherJsonRequest(
     lastKnownRevisionDate: Instant? = Instant.parse("2023-10-27T12:00:00Z"),
     key: String? = "mockKey-$number",
     archivedDate: Instant? = Instant.parse("2023-10-27T12:00:00Z"),
+    data: String? = null,
```
andrebispo5: Just to check, all other fields have non null mocked values shouldn't this have something also?

david-livefront: Good call, I added some fake data here and I also updated the `SyncResponseCipherUtil` to do the same thing

andrebispo5 (approved):

david-livefront:
Thanks @andrebispo5

PR #7199, opened 2026-07-22
Km state bridge

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR is a first step in supporting V2 encryption.

* The `V2UpgradeToken` is now supported in the SyncResponse and we store that to disk
* The `SdkStateBridge` is now in place to support communicating data between the SDK and the main app

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/sdk/statebridge/SdkStateBridge.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/sdk/statebridge/SdkStateBridge.kt
@@ -1,0 +1,141 @@
+package com.x8bit.bitwarden.data.platform.manager.sdk.statebridge
+
+import com.bitwarden.core.MasterPasswordUnlockData
+import com.bitwarden.core.StateBridgeForeignImpl
+import com.bitwarden.core.V2UpgradeToken
+import com.bitwarden.core.WrappedAccountCryptographicState
+import com.bitwarden.crypto.EncString
+import com.bitwarden.crypto.PasswordProtectedKeyEnvelope
+import com.bitwarden.crypto.SymmetricCryptoKey
+import com.x8bit.bitwarden.data.auth.datasource.disk.AuthDiskSource
+import com.x8bit.bitwarden.data.auth.repository.util.updateMasterPasswordUnlock
+import com.x8bit.bitwarden.data.vault.repository.util.toSdkMasterPasswordUnlock
+import com.x8bit.bitwarden.data.vault.repository.util.toV2UpgradeToken
+import com.x8bit.bitwarden.data.vault.repository.util.toV2UpgradeTokenJson
+
+/**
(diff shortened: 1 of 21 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/sdk/statebridge/SdkStateBridge.kt
```diff
@@ -0,0 +1,146 @@
...
+) : StateBridgeForeignImpl {
+    override suspend fun setUserKey(value: SymmetricCryptoKey) {
+        authDiskSource.storeLocalUserDataKey(userId = userId, wrappedKey = value)
+    }
+
+    override suspend fun getUserKey(): SymmetricCryptoKey? =
+        authDiskSource.getLocalUserDataKey(userId = userId)
+
+    override suspend fun clearUserKey() {
+        authDiskSource.storeLocalUserDataKey(userId = userId, wrappedKey = null)
```
david-livefront: We may be able to move this to a purely in-memory value but for now, this change is 1-to-1 with the previous implementation.

app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/sdk/statebridge/SdkStateBridge.kt
```diff
@@ -0,0 +1,146 @@
...
+internal class SdkStateBridge(
+    private val userId: String,
+    private val authDiskSource: AuthDiskSource,
+) : StateBridgeForeignImpl {
+    override suspend fun setUserKey(value: SymmetricCryptoKey) {
+        authDiskSource.storeLocalUserDataKey(userId = userId, wrappedKey = value)
+    }
+
+    override suspend fun getUserKey(): SymmetricCryptoKey? =
+        authDiskSource.getLocalUserDataKey(userId = userId)
```
quexten: I'm sorry, I think my explanation was inaccurate. Localuserdatakey is different from userkey. Localuserdatakey is a static frozen encrypted copy of the user-key at one point, used for e.g. password generator history. User-key is a in-memory current copy of the *decrypted* copy of the current user-key and equivalent to the lock / unlock state. 

You should be able to just hold this in memory as a member variable of the class.

quexten: Or, i guess like this pattern below:
```      authDiskSource.storePinProtectedUserKeyEnvelope(
            userId = userId,
            pinProtectedUserKeyEnvelope = null,
            inMemoryOnly = true,
        )
```

david-livefront: I see, I'll make that change

david-livefront:
Reopening in new PR.

PR #7214, opened 2026-07-27
[PM-40995] fix: Allow fill-assist to rescue autofill requests when heuristics find no fields

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40995

## 📔 Objective

`AutofillParserImpl.parseInternal` returned `AutofillRequest.Unfillable` immediately whenever the built-in heuristic field-classifier found zero non-`Unused` fields on a page — before the app ever checked whether fill-assist (server-curated CSS-selector-based field targeting) was enabled or had matching rules for that host. This meant fill-assist could never rescue a page the heuristics failed to parse, defeating its purpose on pages designed to evade naive heuristics (e.g. generically-named, hint-less input fields). Confirmed via a live repro against a test page with no autofill suggestion offered at all.

This PR:
- Lets `parseInternal` fall back to an `Unused`-inclusive candidate view when heuristics find nothing, but only when fill-assist is actually enabled (feature flag + setting) — the disabled path is unchanged in both behavior and cost.
- Gives `toEffectiveViews`'s `Unused` case (previously dead code, since a heuristic-classified focused view could never be `Unused`) real behavior: it checks against both login and card fill-assist categories, since heuristics gave no signal for which one applies.
- Adds a conservative guard so fill-assist is only trusted for the `Unused` case when it actually matches a field on the page — an empty match falls back to heuristics instead of being treated as authoritative, so an unrelated host rule (e.g. one covering a different pathname/category on the same host) can't silently claim a page it has no real match for. The pre-existing `Card`/`Login` branches keep their existing "authoritative even if empty" behavior since those already carry real heuristic confirmation.
- Fixes a pre-existing bug in the shared test `setup()`'s default `toAutofillViewData` mock stub, uncovered by the new tests: it used `firstArg()`/`secondArg()` for the `autofillId`/`website` params, but since it mocks an extension function, the receiver occupies argument position 0 at the JVM level, so the stub was assigning the `ViewNode` itself into the `autofillId` field. Fixed by shifting to `secondArg()`/`thirdArg()`.
- Adds test coverage for the rescue path, the conservative empty-match guard, and confirmation the rescue path stays inert when fill-assist is disabled.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
@@ -123,7 +123,12 @@ class AutofillParserImpl(
         val urlBarWebsite = traversalDataList
             .flatMap { it.urlBarWebsites }
             .firstOrNull()
-        val autofillViews = traversalDataList.toAutofillViews(urlBarWebsite = urlBarWebsite)
+        // Heuristic views: the focused node's candidates with unfillable (Unused) fields removed,
+        // falling back to all fillable views when nothing has focus.
+        val autofillViews = traversalDataList
+            .selectCandidateAutofillViews(urlBarWebsite = urlBarWebsite) {
+                it !is AutofillView.Unused
+            }
 
         val isFillAssistEnabled = featureFlagManager
             .getFeatureFlag(FlagKey.FillAssistTargetingRules) &&
@@ -282,32 +287,24 @@ private fun AssistStructure.traverse(): List<ViewNodeTraversalData> =
 
 /**
  * Selects the autofill views from the node that currently has focus, or falls back to all
- * fillable views if nothing has focus. Unlike [toAutofillViews], this does not filter out
- * [AutofillView.Unused] views.
+ * fillable views if nothing has focus. The optional [predicate] filters the views *before* the
+ * emptiness/fallback check, so callers that only want fillable fields (e.g. excluding
+ * [AutofillView.Unused]) still get the multi-window fallback applied to the filtered set. By
+ * default no views are filtered out.
  */
 private fun List<ViewNodeTraversalData>.selectCandidateAutofillViews(
     urlBarWebsite: String?,
+    predicate: (AutofillView) -> Boolean = { true },
 ): List<AutofillView> {
     val viewsLists = map { it.autofillViews }
     val candidates = viewsLists
         .filter { views -> views.any { it.data.isFocused } }
         .flatten()
+        .filter(predicate)
         .takeUnless { it.isEmpty() }
-        ?: viewsLists.flatten()
+        ?: viewsLists.flatten().filter(predicate)
     return candidates.map { it.updateWebsiteIfNecessary(website = urlBarWebsite) }
 }
-
-/**
- * Assembles the [AutofillView] list from this [ViewNodeTraversalData] list.
- * Take only the autofill views from the node that currently has focus.
- * Then remove all the fields that cannot be filled with data.
- * We fall back to taking all the fillable views if nothing has focus.
- */
-private fun List<ViewNodeTraversalData>.toAutofillViews(
-    urlBarWebsite: String?,
-): List<AutofillView> =
-    selectCandidateAutofillViews(urlBarWebsite = urlBarWebsite)
-        .filter { it !is AutofillView.Unused }
 
 /**
  * Returns the focused [AutofillView], or falls back to the first entry if none is focused.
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -233,18 +239,27 @@ class AutofillParserImpl(
             when (focusedView) {
                 is AutofillView.Card -> rule.category in CARD_FILL_ASSIST_CATEGORIES
                 is AutofillView.Login -> rule.category in LOGIN_FILL_ASSIST_CATEGORIES
-                is AutofillView.Unused -> false
+                is AutofillView.Unused -> rule.category in LOGIN_FILL_ASSIST_CATEGORIES ||
+                    rule.category in CARD_FILL_ASSIST_CATEGORIES
```
david-livefront: Since this is a multi-line statement, can we make sure it is wrapped in curly braces:

```kotlin
is AutofillView.Unused -> {
    rule.category in LOGIN_FILL_ASSIST_CATEGORIES ||
        rule.category in CARD_FILL_ASSIST_CATEGORIES
}
```

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -264,26 +281,40 @@ private fun AssistStructure.traverse(): List<ViewNodeTraversalData> =
...
     val viewsLists = map { it.autofillViews }
-    val autofillViewLists = viewsLists
+    val candidates = viewsLists
         .filter { views -> views.any { it.data.isFocused } }
         .flatten()
-        .filter { it !is AutofillView.Unused }
         .takeUnless { it.isEmpty() }
-        ?: viewsLists
-            .flatten()
-            .filter { it !is AutofillView.Unused }
```
david-livefront, line 283: I see the comment Claude left about this filter but I believe that is necessary, right?

aj-rosado: It is actually right, if nothing had focus previously it would fall to all the views and now only when had fill assist so it is a regression in behavior in comparison with existing code.

Added a test for that scenario.

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -264,26 +281,40 @@ private fun AssistStructure.traverse(): List<ViewNodeTraversalData> =
...
-    return autofillViewLists.map { it.updateWebsiteIfNecessary(website = urlBarWebsite) }
+        ?: viewsLists.flatten()
+    return candidates.map { it.updateWebsiteIfNecessary(website = urlBarWebsite) }
```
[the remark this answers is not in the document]
aj-rosado: @claude this is fixed

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -233,18 +244,29 @@ class AutofillParserImpl(
...
+        // something for this page -- otherwise an unrelated host rule (e.g. one covering a
+        // different pathname/category on the same host) would silently claim an unrelated
+        // form. Card/Login branches keep the existing "authoritative even if empty" behavior
+        // since those already carry real heuristic confirmation.
+        return if (focusedView is AutofillView.Unused && fillAssistViews.isEmpty()) {
             this
+        } else {
+            Timber.d("FillAssist invoked for this autofill attempt")
+            fillAssistViews
         }
```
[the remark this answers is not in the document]
aj-rosado: @claude this has been updated

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
```diff
@@ -123,14 +123,29 @@ class AutofillParserImpl(
...
-            .firstOrNull { it.data.isFocused }
-            ?: autofillViews.firstOrNull()
-            // The view is unfillable if there are no focused views.
+        // we at least have something to hook into). If heuristics found nothing at all, only
+        // give fill-assist a chance to rescue the page when it's actually enabled -- otherwise
+        // the view is unfillable since there are no focused views.
+        val focusedView = autofillViews.firstFocusedOrNull()
+            ?: if (isFillAssistEnabled) {
+                traversalDataList
+                    .selectCandidateAutofillViews(urlBarWebsite = urlBarWebsite)
```
david-livefront, line 144: So for this one we do not want to filter out anything?

david-livefront (approved):

app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/parser/AutofillParserImpl.kt
aj-rosado: yes

aj-rosado:
thank you @david-livefront

PR #7222, opened 2026-07-29
[PM-41103] feat: Add identity autofill feature flag

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41103

## 📔 Objective

Adds `FlagKey.IdentityAutofill` (`pm-38138-mobile-identity-autofill`), defaulted off, so the in-progress identity autofill feature (PM-38138) can be gated before it reaches production. This PR is definition-only — no gating logic is added yet, so behavior is unchanged. Also wires the flag into the debug menu (`FeatureFlagListItems.kt` + `identity_autofill` string) so it can be toggled locally for testing.

A follow-up PR will add the actual gating check in the autofill parser.

## 📸 Screenshots

N/A — debug menu label follows the same list-item pattern as existing flags.


vvolkgang (approved):
:shipit:

PR #7223, opened 2026-07-30
[PM-41240] Exclude archived items from folder item count

gabrbrand:
## 📔 Objective

Currently archived items are included in the folder item count on VaultItemListingScreen but excluded on the VaultScreen. With this PR archived items are excluded on the VaultItemListingScreen as well.

⚠️ WARNING: This is untested because I don't want to play around with my real account. Is there a way to turn on premium features in a development environment?


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/itemlisting/util/VaultItemListingDataExtensions.kt
```diff
@@ -201,7 +201,7 @@ fun VaultData.toViewState(
                     count = allFilteredCipherViewList
                         .count {
                             !it.id.isNullOrBlank() &&
-                                it.deletedDate == null &&
+                                it.isActive &&
```
david-livefront, line 204: This solution seems fine to me, can you make sure a test gets added for these changes

PR #7250, opened 2026-08-07
[PM-41319] fix: Autofill phone-number login credentials via fill-assist

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41319

## 📔 Objective

Fill-assist wasn't autofilling phone-number login credentials on sites where a single field accepts either a phone number or an email address (declared under both `"email"` and `"phone"` keys in the same host rule).

Two separate bugs combined to cause this:
- `toAutofillViewForFieldKey` had no case for the `"phone"` field key, so a matching selector silently produced no view at all.
- When a field matched both `"email"` and `"phone"`, the resolution picked whichever key happened to be listed first in the rules JSON. Since `"email"` was typically listed first, dual-purpose fields resolved to `Login.Email`, whose fill logic gates on `isValidEmail()` — silently rejecting phone-number credentials on a field that would have accepted them.

Fix:
- Map `"phone"` to `Login.Username`, which has no format gate (matches existing heuristic-mode behavior for phone-hinted fields).
- When a field matches multiple keys, explicitly prefer `Login.Username` over other candidate types rather than relying on map ordering.
- Trim the stored username only for the `isValidEmail()` check itself (not the filled value) so incidental leading/trailing whitespace on an otherwise-valid email doesn't silently block the fill.

No UI changes.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/builder/FilledDataBuilderImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/builder/FilledDataBuilderImpl.kt
@@ -154,7 +154,7 @@ class FilledDataBuilderImpl(
                 ) {
                     val value = when (autofillView) {
                         is AutofillView.Login.Email -> {
-                            if (!autofillCipher.username.isValidEmail()) {
+                            if (!autofillCipher.username.trim().isValidEmail()) {
                                 return@mapNotNull null
                             }
                             autofillCipher.username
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/builder/FilledDataBuilderImpl.kt
```diff
@@ -154,7 +154,7 @@ class FilledDataBuilderImpl(
                 ) {
                     val value = when (autofillView) {
                         is AutofillView.Login.Email -> {
-                            if (!autofillCipher.username.isValidEmail()) {
+                            if (!autofillCipher.username.trim().isValidEmail()) {
```
andrebispo5, line 157: having the trim here will create the valid email but will pass along the spaces it removes to the autofill value.

aj-rosado: This was something extra ticket that I noticed while testing (my email for a field had an empty space so it was not being identified as an email by our function)

Did not want to modify the value autofilled to keep it as it is saved. If it is not intentional the user might notice and fix it.

andrebispo5 (approved):
APPROVED :)

PR #7259, opened 2026-08-11
[PM-41870] Allow save prompt and fallback to writable collection when organization data ownership is enabled

### LINKED ISSUE #7256: [PM-41838] "Ask to add item" prompt fails when Personal Vault is disabled
### Steps To Reproduce

1. Enable the "Disable Personal Vault" policy for an organization user.
2. Ensure the user has write access to at least one organization collection.
3. Log into a new website and click Yes, Save Now when it prompts to save credentials.

### Expected Result

The new item should automatically save to the user's first available writable organization collection.

### Actual Result

The save prompt fails silently because the personal vault is unavailable.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Build Version

version 2026.7.1

N/A

### Self-host Server Version

_No response_

### Environment Details

- Device: Pixel 9
- OS Version: GrapheneOS 2026080501

### Issue Tracking Info

woliver99:
## 🎟️ Tracking

Closes #7256 [PM-41838](https://bitwarden.atlassian.net/browse/PM-41838)

## 📔 Objective

1. Removes an early return in `AutofillProcessorImpl.kt` that was automatically canceling Autofill save requests whenever `ORGANIZATION_DATA_OWNERSHIP` policy was enabled.
2. Updates `getDefaultCollectionViewOrNull()` in `CipherViewExtensions.kt` so that when `DEFAULT_USER_COLLECTION` is null, the app automatically selects the first writable collection as a fallback for much better UX.

### 1. Autofill Save Prompt
| Before | After |
| :---: | :---: |
| old-autofill | new-autofill |

### 2. Collection Default Selection
| Before | After |
| :---: | :---: |
| old-default | new-default |

[PM-41838]: https://bitwarden.atlassian.net/browse/PM-41838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/processor/AutofillProcessorImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/processor/AutofillProcessorImpl.kt
@@ -80,11 +80,6 @@ class AutofillProcessorImpl(
             return
         }
 
-        if (policyManager.getActivePolicies(PolicyType.ORGANIZATION_DATA_OWNERSHIP).any()) {
-            saveCallback.onSuccess()
-            return
-        }
-
         request
             .fillContexts
             .lastOrNull()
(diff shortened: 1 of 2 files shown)


woliver99:
Just realised that auto-selecting the first collection might lead to saving into the wrong collection if a user has access to multiple collections.

What if instead we only auto-select if there’s exactly one collection available? That way if they have multiple collections, they’re still forced to pick one. This would also match how the web extension and desktop app handle it.

Let me know if you prefer that and I can update the code!

PR #7269, opened 2026-08-14
[PM-40533] feat: Validate recipient email domains against SendControls allowed domains

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40533

## 📔 Objective

When an org's SendControls policy sets `allowedDomains`, saving a Send with Specific people
recipients now checks each recipient's domain against that list, and blocks the save with an error
dialog if any of them falls outside it.

### What changed

- The check runs at save time in `handleSaveClick`, right after the existing email format
  validation, so it does not fire on every keystroke
- Error dialog reuses the existing "Invalid email addresses" title, with a new description listing
  the allowed domains
- The policy value arrives as one comma-separated string, so it gets trimmed and emptied of blanks
  before both matching and display. Domain matching is case-insensitive
- A policy value with nothing usable in it, `","` for example, counts as no restriction. Failing
  closed there would block saving with no way for the user to recover

### Scope

- Only applies when Who can view is Specific people, since that is the only case with recipients.
  Other access types are untouched
- Flag off or no domains configured means no domain checking at all, and the existing format
  validation is unaffected either way

### Note for reviewers

This stacks on #7266 (PM-40531), so please review that one first. The diff here is only the last
three commits.

## 📸 Screenshots

https://github.com/user-attachments/assets/68c1770a-5d5a-47c9-94d9-a62b87e1bd7f

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
@@ -80,6 +80,27 @@ private const val KEY_STATE = "state"
 private const val MAX_FILE_SIZE_BYTES: Long = 100 * 1024 * 1024
 
 /**
+ * Whether this email address belongs to one of [allowedDomains], compared case-insensitively.
+ */
+private fun String.hasAllowedDomain(allowedDomains: List<String>): Boolean {
+    val domain = substringAfterLast(delimiter = '@', missingDelimiterValue = "")
+    return allowedDomains.any { it.equals(other = domain, ignoreCase = true) }
+}
+
+/**
+ * Splits this comma-separated policy value into trimmed, non-blank domains.
+ *
+ * Returns an empty list when no domain is configured, which includes a value that holds nothing
+ * usable such as `","`. An empty result is treated as no restriction rather than as a restriction
+ * nothing can satisfy, since the latter would block saving with no way for the user to recover.
+ */
+private fun String?.toAllowedDomains(): List<String> = this
+    ?.split(",")
+    ?.map { it.trim() }
+    ?.filter { it.isNotBlank() }
+    .orEmpty()
+
+/**
  * Returns the [SendAuth] this access type restricts a Send to, preserving [current] when it already
  * matches the restricted type so that any emails already entered are kept.
  */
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
```diff
@@ -79,13 +79,34 @@ private const val KEY_STATE = "state"
...
     get() = this != SendAccessTypeJson.ANY
 
+/**
+ * Splits this comma-separated policy value into trimmed, non-blank domains.
+ *
+ * Returns an empty list when no domain is configured, which includes a value that holds nothing
+ * usable such as `","`. An empty result is treated as no restriction rather than as a restriction
+ * nothing can satisfy, since the latter would block saving with no way for the user to recover.
+ */
+private fun String?.toAllowedDomains(): List<String> = this
```
aj-rosado: I am founding it confusing that `allowedDomains` calls the `toAllowedDomains`. Can we name this method to something that indicates it transforms into a list? Like `toAllowedDomainsList`

david-livefront: Can we make sure all these functions return an `ImmutableList`

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
```diff
@@ -1023,6 +1074,14 @@ data class AddEditSendState(
     val isPremium: Boolean,
 ) : Parcelable {
 
+    /**
+     * Helper to determine the recipient email domains the SendControls policy restricts a Send to,
+     * or an empty list when recipients may use any domain. The legacy send options policy has no
+     * equivalent enforcement, so this is only in effect alongside the SendControls feature flag.
+     */
+    val enforcedAllowedDomains: List<String>
```
david-livefront: Make sure this is an `ImmutableList`

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
andrebispo5: Renamed it to `splitToDomains()`, so the call site now reads `allowedDomains.splitToDomains()`. That drops the echo you flagged.

I left off the `List` suffix on purpose: the extensions in this codebase that return lists use a plural noun without one, for example `UserState.toAccountSummaries()` and `BrowserThirdPartyAutofillStatus.toBrowserAutoFillSettingsOptions()`. I could not find a `to...List()` precedent in the app module, so adding one here would make this the odd one out. `splitToDomains` carries the "transforms into a list" signal you were after through the verb instead, and mirrors stdlib `splitToSequence`.

It also returns `ImmutableList<String>` now, per David's comment on the same thread.

aj-rosado (approved):
LGTM

PR #7272, opened 2026-08-17
[PM-42107] feat: Wire SendControls policy awareness into the View Send screen

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42107
https://bitwarden.atlassian.net/browse/PM-42109

## 📔 Objective

Teaches the View Send screen about the SendControls policy so it can tell when a Send the server
marked `disabled` is non-compliant, and why. Also hides the edit FAB for those Sends, which is
PM-42109.

### What changed

- `ViewSendViewModel` now observes the effective Send policy alongside both feature flags, in one
  `combine` so the derived result cannot go stale when any of them changes
- New `SendPolicyRestriction` model with three cases: file send not compliant, type no longer
  allowed, and copy required. Each says whether a compliant copy is possible
- A Send counts as restricted only when `pm-31885-send-controls-existing-sends` is on,
  `pm-31885-send-controls` is on, and the Send is disabled
- File sends report as not compliant even when their type is also disallowed, since the attachment
  cannot be uploaded again either way
- Edit FAB is hidden whenever a restriction applies (PM-42109)

### Not in this PR

- The banner itself, which is PM-42108. Nothing new is visible on screen yet, this is the state and
  the decision logic only
- "Make a copy", which is PM-42110

### Note for reviewers

This stacks on #7269, so please review that first. The diff here is only the last three commits.

## 📸 Screenshots

No visible change in this PR. The banner arrives with PM-42108.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/viewsend/ViewSendViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/viewsend/ViewSendViewModel.kt
@@ -3,8 +3,10 @@ package com.x8bit.bitwarden.ui.tools.feature.send.viewsend
 import android.os.Parcelable
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
+import com.bitwarden.core.data.manager.model.FlagKey
 import com.bitwarden.core.data.repository.model.DataState
 import com.bitwarden.data.repository.util.baseWebSendUrl
+import com.bitwarden.network.model.SendTypeJson
 import com.bitwarden.send.SendView
 import com.bitwarden.ui.platform.base.BackgroundEvent
 import com.bitwarden.ui.platform.base.BaseViewModel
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/viewsend/ViewSendViewModel.kt
```diff
@@ -258,9 +317,37 @@ data class ViewSendState(
...
+    val policyRestriction: SendPolicyRestriction?
+        get() = when {
+            !isSendControlsExistingSendsEnabled || !isSendControlsEnabled -> null
+            !isSendDisabled -> null
+            // A file send is reported as simply non-compliant even when its type is also
+            // disallowed, since its attachment cannot be uploaded again either way.
+            sendType == SendItemType.FILE -> SendPolicyRestriction.FileNotCompliant
+            !isSendTypeAllowed -> SendPolicyRestriction.TypeNotAllowed
+            else -> SendPolicyRestriction.CopyRequired
+        }
```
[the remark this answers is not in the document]
andrebispo5: The premise here does not hold on Android: `disabled` is server-controlled, and there is no client affordance to set it.

The plumbing that suggests otherwise is dead. `AddEditSendHandlers.onDeactivateSendToggle` is declared at `AddEditSendHandlers.kt:23` and wired to `AddEditSendAction.DeactivateThisSendToggle` at line 60, but no composable consumes that handler, so the add/edit screen renders no "Deactivate this Send" control. `AddEditSendStateExtensions.kt:35` therefore only ever round-trips the value the server already sent.

Since the server sets `disabled` as the SendControls non-compliance signal, attributing a disabled Send to that policy while both flags are on is correct, and the extra `EffectiveSendPolicy` signal is not needed.

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/viewsend/ViewSendViewModel.kt
```diff
@@ -258,9 +317,36 @@ data class ViewSendState(
...
+     * A Send is only treated as policy-restricted while both feature flags are enabled, since
+     * `disabled` is also set when the Send is deactivated by hand on another client.
+     */
+    val policyRestriction: SendPolicyRestriction?
+        get() = when {
+            !isSendControlsExistingSendsEnabled || !isSendControlsEnabled -> null
+            !isSendDisabled -> null
+            // A file send is reported as simply non-compliant even when its type is also
+            // disallowed, since its attachment cannot be uploaded again either way.
+            sendType == SendItemType.FILE -> SendPolicyRestriction.FileNotCompliant
```
aj-rosado, line 332: Is this correct? The attachment cannot be modified but other properties can

andrebispo5: Yes file types become non compliant if they are disabled and one of the flags is enabled. They cannot be changed or made copies of.

aj-rosado (approved):
LGTM

PR #7275, opened 2026-08-17
[PM-42108] feat: Add policy restriction banner to View Send

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42108

## 📔 Objective

Shows a read-only banner on View Send when a Send is restricted by the org's SendControls policy,
explaining why it cannot be brought into compliance.

### What changed

- Banner titled "Organization policy restriction" with one of three bodies, picked from the
  restriction computed in PM-42107: file send not compliant, Text sends not allowed, or copy
  required
- Body text lives on `SendPolicyRestriction` itself, the same way `SendAuth` carries its own
  display text
- Renders above the Send details, absent entirely when no restriction applies

### Reusing BitwardenActionCard

- Uses the existing `BitwardenActionCard` rather than a new component. It already has a title, a
  subtitle, a leading icon slot, and the CTA button PM-42110 needs, and `bitwardenCardColors()`
  already defaults to the same `background.tertiary` token as `BitwardenInfoCalloutCard`
- Made `actionText` and `onActionClick` nullable with defaults so the button can be omitted, since
  two of the three variants have no action
- That component has 24 call sites. The change is additive, parameter order is unchanged, and the
  full app plus authenticator suites pass, so existing callers are unaffected

### Not in this PR

- "Make a copy", which is PM-42110. With this card in place that becomes two arguments rather than
  a component change

### Note for reviewers

Fifth in the stack: main <- #7266 <- #7269 <- #7272 <- this. Please review #7272 first. The diff
here is only the last four commits.

## 📸 Screenshots
image
image
image

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/masterpasswordguidance/MasterPasswordGuidanceScreen.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/masterpasswordguidance/MasterPasswordGuidanceScreen.kt
@@ -26,6 +26,7 @@ import com.bitwarden.ui.platform.base.util.annotatedStringResource
 import com.bitwarden.ui.platform.base.util.standardHorizontalMargin
 import com.bitwarden.ui.platform.base.util.toAnnotatedString
 import com.bitwarden.ui.platform.components.appbar.BitwardenTopAppBar
+import com.bitwarden.ui.platform.components.button.model.BitwardenButtonData
 import com.bitwarden.ui.platform.components.card.BitwardenActionCard
 import com.bitwarden.ui.platform.components.card.BitwardenContentCard
 import com.bitwarden.ui.platform.components.content.model.ContentBlockData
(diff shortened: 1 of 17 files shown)


david-livefront:
@andrebispo5 Can you make sure the appropriate images get added to the PR description

ui/src/main/kotlin/com/bitwarden/ui/platform/components/card/BitwardenActionCard.kt
```diff
@@ -57,8 +59,8 @@ import com.bitwarden.ui.util.asText
 @Composable
 fun BitwardenActionCard(
     cardTitle: String,
-    actionText: String,
-    onActionClick: () -> Unit,
+    actionText: String? = null,
+    onActionClick: (() -> Unit)? = null,
```
david-livefront: Since they are now optional, I suggest binding them together with `BitwardenButtonData`.

```kotlin
fun BitwardenActionCard(
    ...
    actionButton: BitwardenButtonData? = null,
    ...
) {
    ...
    actionButton?.let {
        // show button
    }
    ...
}
```

app/src/main/kotlin/com/x8bit/bitwarden/ui/auth/feature/masterpasswordguidance/MasterPasswordGuidanceScreen.kt
```diff
@@ -165,8 +167,10 @@ private fun NeedSomeInspirationCard(
 ) {
     BitwardenActionCard(
         cardTitle = stringResource(BitwardenString.need_some_inspiration),
-        actionText = stringResource(BitwardenString.check_out_the_passphrase_generator),
-        onActionClick = onActionClicked,
+        actionButton = BitwardenButtonData(
```
david-livefront, line 170: 👍

david-livefront (approved):

PR #7276, opened 2026-08-17
[PM-42110] feat: Make a copy of a policy-restricted Send

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42110

## 📔 Objective

Adds a "Make a copy" action to the policy restriction banner. It opens the Add Send screen prefilled
from the restricted Send, with the fields the policy enforces already set, so the user can create a
compliant replacement.

### Copy mode

- New `ModeType.COPY` and `AddEditSendType.CopyItem`. A copy loads like an edit does, then behaves
  like an add for everything else: it saves with `createSend`, so the original is never touched
- Carried over: name, notes, text content, hide-text-by-default, max access count, recipients
- Reset or dropped: password (never readable by design), deactivate, share URL, access count,
  expiration
- From policy: deletion window and access type, reusing the helpers from PM-40531 and PM-40533
- Recipients outside the policy's allowed domains are carried over anyway. The user gets the
  existing save-time error naming the allowed domains, which beats silently dropping recipients

### The button

- Only shown when a copy can actually help, so never for a file send (its attachment cannot be
  uploaded again) or a Send whose type the policy no longer allows
- Uses the optional `actionText` and `onActionClick` added to `BitwardenActionCard` in #7275

### Worth flagging

- `isAddMode` changed from "is AddItem" to "is not EditItem". I traced all 14 consumers: the
  overflow menu, delete button and current-access-count row all correctly hide for a copy, and both
  policy paths now apply to copies, which is what you want if the policy changes while the form is
  open
- The four `when` blocks over `AddEditSendType` are each handled explicitly rather than with an
  `else`, so a future mode cannot slip through

### Note for reviewers

Sixth in the stack: main <- #7266 <- #7269 <- #7272 <- #7275 <- this. Please review #7275 first.
The diff here is only the last six commits.

## 📸 Screenshots

https://github.com/user-attachments/assets/e9fca4a0-7e2c-4161-820f-89cdc3a640ec


app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
```diff
@@ -540,6 +544,39 @@ class AddEditSendViewModel @Inject constructor(
...
+                toCopyViewState(
+                    deletionDate = clock.instant().plus(
+                        state.enforcedDeletionHours?.toLong() ?: DEFAULT_DELETION_HOURS,
+                        ChronoUnit.HOURS,
+                    ),
+                    isHideEmailAddressEnabled = isHideEmailAddressEnabled,
+                    sendAuth = state
+                        .enforcedWhoCanAccess
+                        ?.toEnforcedSendAuth(current = toSendAuth())
+                        ?: toSendAuth(),
```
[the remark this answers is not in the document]
andrebispo5: That does not work because when the user tries to save it, it will fail. It also has to remain like that because when we edit a password protected item the fill stays empty and needs to be saved empty to keep the same password.

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewExtensionsTest.kt
```diff
@@ -50,6 +52,35 @@ class SendViewExtensionsTest {
...
+            .copy(emails = listOf("one@example.com", "two@example.com"))
+
+        assertEquals(
+            SendAuth.Email(
+                emails = persistentListOf(
+                    AuthEmail(value = "one@example.com"),
+                    AuthEmail(value = "two@example.com"),
+                ),
+            ).emails.map { it.value },
+            (sendView.toSendAuth() as SendAuth.Email).emails.map { it.value },
```
david-livefront: We should assert the entire `SendAuth.Email` not just the emails inside

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
```diff
@@ -623,6 +625,109 @@ class AddEditSendViewModelTest : BaseViewModelTest() {
...
+            addEditSendType = AddEditSendType.CopyItem(sendItemId = COPY_SEND_ID),
+        )
+            .stateFlow
+            .value
+
+        assertEquals(
+            copyModeState(
+                sendAuth = SendAuth.Password,
+                deletionDate = ENFORCED_DELETION_DATE,
+            ).copy(
```
david-livefront: The `.copy(` should go on the next line:

```koltin
            copyModeState(
                sendAuth = SendAuth.Password,
                deletionDate = ENFORCED_DELETION_DATE,
            )
                .copy(
```

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
```diff
@@ -623,6 +625,109 @@ class AddEditSendViewModelTest : BaseViewModelTest() {
...
+                isSendControlsEnabled = true,
+                deletionHours = ENFORCED_DELETION_HOURS,
+            ),
+            state,
+        )
+        // The copy adopts the policy window rather than the original send's deletion date.
+        assertEquals(
+            ENFORCED_DELETION_DATE,
+            (state.viewState as AddEditSendState.ViewState.Content).common.deletionDate,
```
david-livefront: We should be asserting complete state

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
```diff
@@ -623,6 +625,109 @@ class AddEditSendViewModelTest : BaseViewModelTest() {
...
+        assertEquals(
+            copyModeState(sendAuth = SendAuth.Password).copy(
+                isSendControlsEnabled = true,
+                whoCanAccess = SendAccessTypeJson.PASSWORD_PROTECTED,
+            ),
+            state,
+        )
+        assertEquals(
+            SendAuth.Password,
+            (state.viewState as AddEditSendState.ViewState.Content).common.sendAuth,
```
david-livefront: Ditto

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewExtensions.kt
```diff
@@ -32,16 +44,7 @@ fun SendView.toViewState(
...
-                hasPassword -> SendAuth.Password
-                emails.isNotEmpty() -> {
-                    SendAuth.Email(
-                        emails = this.emails.map { AuthEmail(value = it) }.toImmutableList(),
-                    )
-                }
-
-                else -> SendAuth.None
-            },
+            sendAuth = toSendAuth(),
```
david-livefront: Can we add a `this.` here?

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
```diff
@@ -535,7 +539,42 @@ class AddEditSendViewModel @Inject constructor(
...
 
-    @Suppress("LongMethod")
+    /**
+     * Maps a loaded [SendView] into the content for whichever mode this screen is in: the send's
+     * own values when editing it, or the values a compliant copy of it should start from.
+     *
+     * @param currentState The state being updated, which is the in-flight value rather than the
+     * [state] property so that the mapping sees the same policy data as the update it belongs to.
+     */
+    private fun SendView.toAddEditViewState(
```
david-livefront: Maybe move this to an extension file and just pass in the `environmentRepo` and `clock`?

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewCopyExtensionsTest.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        val common = result.common
+        // The copy is a brand new Send, so none of these follow it over.
+        assertNull(common.originalSendView)
+        assertNull(common.sendUrl)
+        assertNull(common.currentAccessCount)
+        assertNull(common.expirationDate)
+        assertEquals("", common.passwordInput)
+        assertFalse(common.hasPassword)
+        assertFalse(common.isDeactivateChecked)
```
david-livefront: We should just assert the entire result

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewCopyExtensionsTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    fun `toCopyViewState should use the deletion date it is given`() {
+        val sendView = createMockSendView(number = 1, type = SendType.TEXT)
+
+        val result = sendView.toCopyViewState(
+            deletionDate = COPY_DELETION_DATE,
+            isHideEmailAddressEnabled = true,
+            sendAuth = SendAuth.None,
+        )
+
+        assertEquals(COPY_DELETION_DATE, result.common.deletionDate)
```
david-livefront: Ditto

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewExtensions.kt
```diff
@@ -1,12 +1,84 @@
...
 import com.x8bit.bitwarden.ui.tools.feature.send.addedit.model.SendAuth
 import com.x8bit.bitwarden.ui.tools.feature.send.util.toSendUrl
 import kotlinx.collections.immutable.toImmutableList
+import java.time.Clock
+import java.time.temporal.ChronoUnit
+
+/**
+ * The deletion window applied to a new Send when no deletion date is enforced by policy (7 days).
+ */
+internal const val DEFAULT_DELETION_HOURS: Long = 7 * 24L
```
david-livefront, line 19: Can we make this private now?

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewCopyExtensionsTest.kt
```diff
@@ -0,0 +1,157 @@
...
+            emails = persistentListOf(AuthEmail(id = "id", value = "test@example.com")),
+        )
+
+        val result = sendView.toCopyViewState(
+            deletionDate = COPY_DELETION_DATE,
+            isHideEmailAddressEnabled = true,
+            sendAuth = sendAuth,
+        )
+
+        assertEquals(sendAuth, result.common.sendAuth)
```
david-livefront: Can we make sure all these tests get updated to compare total state?

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
```diff
@@ -623,6 +625,103 @@ class AddEditSendViewModelTest : BaseViewModelTest() {
...
+            addEditSendType = AddEditSendType.CopyItem(sendItemId = COPY_SEND_ID),
+        )
+            .stateFlow
+            .value
+
+        assertEquals(
+            copyModeState(sendAuth = SendAuth.Password)
+                .copy(
+                    isSendControlsEnabled = true,
+                    whoCanAccess = SendAccessTypeJson.PASSWORD_PROTECTED,
```
david-livefront: Instead of copying it, can we just pass these into the `copyModeState` function?

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
```diff
@@ -623,6 +625,103 @@ class AddEditSendViewModelTest : BaseViewModelTest() {
...
+        mutableSendDataStateFlow.value = DataState.Loaded(mockSendView)
+
+        val state = createViewModel(
+            addEditSendType = AddEditSendType.CopyItem(sendItemId = COPY_SEND_ID),
+        )
+            .stateFlow
+            .value
+
+        assertEquals(copyModeState(sendAuth = SendAuth.Password), state)
+        assertTrue(state.isAddMode)
```
david-livefront: Maybe change the `isAddMode` to `isAddOrCopyMode`?

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewExtensions.kt
andrebispo5: It is being used in the extensions file and in the `AddEditSendViewModel`.

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
andrebispo5: No need, in fact it is add mode prefilled, and if we went that route there would be a lot more to change like the name of the class `AddCopyEditSend` 😅

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewExtensions.kt
david-livefront: Gotcha

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
david-livefront: Makes sense, I am a little confused how this works though:

```kotlin
    val isAddMode: Boolean get() = addEditSendType !is AddEditSendType.EditItem
```

Why would this be true?

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
```diff
@@ -535,7 +523,22 @@ class AddEditSendViewModel @Inject constructor(
...
 
-    @Suppress("LongMethod")
+    /**
+     * Maps a loaded [SendView] into the content for whichever mode this screen is in, supplying the
+     * mapping with the collaborators it needs from this view model.
+     *
+     * @param currentState The state being updated, which is the in-flight value rather than the
+     * [state] property so that the mapping sees the same policy data as the update it belongs to.
+     */
+    private fun SendView.toCurrentModeViewState(
```
david-livefront, line 533: Should this go to the extensions file too?

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/util/SendViewCopyExtensions.kt
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Recipient emails are copied as they are, even when a policy restricts their domains. The user is
+ * told which domains are allowed when they save, which is friendlier than silently dropping
+ * recipients they chose.
+ *
+ * @param deletionDate The deletion date to apply to the copy.
+ * @param isHideEmailAddressEnabled Whether hiding the sender's email is currently allowed.
+ * @param sendAuth The access type to apply to the copy, which the policy may have forced.
+ */
+fun SendView.toCopyViewState(
```
david-livefront: Can't this go in the main `SendViewExtensions` file

app/src/test/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModelTest.kt
andrebispo5: This particular one may be a bit misleading let me rename it.

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendViewModel.kt
andrebispo5: `toAddEditViewState` is already in the extensions file, this is just the bit that hands it the view model's `clock`, `environmentRepo` and `isHideEmailAddressEnabled` so the two call sites don't have to repeat them.

app/src/main/kotlin/com/x8bit/bitwarden/ui/tools/feature/send/addedit/AddEditSendContent.kt
```diff
@@ -391,7 +391,7 @@ private fun AddEditSendOptions(
     state: AddEditSendState.ViewState.Content,
     isSendsRestrictedByPolicy: Boolean,
     shouldHideEmailAddressToggle: Boolean,
-    isAddMode: Boolean,
+    isNewSend: Boolean,
```
david-livefront, line 394: 👍

david-livefront (approved):

PR #7280, opened 2026-08-18
[PM-41929] fix: Update the manage devices screen when a passwordless request push is received

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41929

## 📔 Objective

The Manage Devices screen only learned about pending login requests when it polled or when the
user pulled to refresh, so a request initiated while the screen was already open did not appear
until the next poll cycle.

This wires the push notification stream into the screen:

- `AuthRequestManager.getPasswordlessAuthRequestFlow()` observes `PushManager.passwordlessRequestFlow`,
  filters to the active user (a push for another user would otherwise be hydrated with the active
  user's token), hydrates the request with its fingerprint, and emits only requests that are still
  actionable — not already approved, not declined, and under five minutes old. Requests that fail
  to hydrate are logged and dropped rather than surfaced as errors.
- `ManageDevicesViewModel` collects that flow, re-reads the device list (the only source that
  reports which device owns a pending request), and merges the request into state, replacing any
  earlier copy so it cannot be listed twice. Because the refresh is not user-initiated, a failed
  device fetch leaves the screen untouched instead of replacing it with an error — polling and
  pull-to-refresh reconcile it later.

Also extracts the repeated `AuthRequestsResponseJson.AuthRequest` → `AuthRequest` mapping into a
`toAuthRequest(fingerprint)` extension, replacing seven hand-written copies in
`AuthRequestManagerImpl` with no behavior change.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/AuthRequestManagerImpl.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/AuthRequestManagerImpl.kt
@@ -98,7 +98,12 @@ class AuthRequestManagerImpl(
                     isSso = authRequestType.isSso,
                 )
                 .map { request ->
-                    request.toAuthRequest(fingerprint = authRequest.fingerprint)
+                    request.toAuthRequest(
+                        fingerprint = authRequest.fingerprint,
+                        publicKey = request.publicKey,
+                        responseDate = request.responseDate,
+                        isRequestApproved = request.requestApproved ?: false,
+                    )
                 }
                 .fold(
                     onFailure = { emit(CreateAuthRequestResult.Error(error = it)) },
@@ -178,15 +183,14 @@ class AuthRequestManagerImpl(
                             isRequestApproved = false
                             responseDate = clock.instant()
                         }
-                        request
-                            // The PublicKey and Fingerprint should be frozen in place to
-                            // ensure no funny-business happens between multiple requests.
-                            .toAuthRequest(fingerprint = initialAuthRequest.fingerprint)
-                            .copy(
-                                publicKey = initialAuthRequest.publicKey,
-                                responseDate = responseDate,
-                                requestApproved = isRequestApproved,
-                            )
+                        // The PublicKey and Fingerprint should be frozen in place to ensure no
+                        // funny-business happens between multiple requests.
+                        request.toAuthRequest(
+                            fingerprint = initialAuthRequest.fingerprint,
+                            publicKey = initialAuthRequest.publicKey,
+                            responseDate = responseDate,
+                            isRequestApproved = isRequestApproved,
+                        )
                     }
                 }
                 .fold(
@@ -249,6 +253,9 @@ class AuthRequestManagerImpl(
             .mapCatching { response ->
                 response.toAuthRequest(
                     fingerprint = getFingerprintPhrase(response.publicKey).getOrThrow(),
+                    publicKey = response.publicKey,
+                    responseDate = response.responseDate,
+                    isRequestApproved = response.requestApproved ?: false,
                 )
             }
             .fold(
@@ -267,6 +274,9 @@ class AuthRequestManagerImpl(
                 .mapCatching { response ->
                     response.toAuthRequest(
                         fingerprint = getFingerprintPhrase(response.publicKey).getOrThrow(),
+                        publicKey = response.publicKey,
+                        responseDate = response.responseDate,
+                        isRequestApproved = response.requestApproved ?: false,
                     )
                 }
                 .fold(
@@ -286,7 +296,12 @@ class AuthRequestManagerImpl(
             .flatMap { request ->
                 if (request.requestApproved == true) {
                     getFingerprintPhrase(request.publicKey).map { fingerprint ->
-                        request.toAuthRequest(fingerprint = fingerprint)
+                        request.toAuthRequest(
+                            fingerprint = fingerprint,
+                            publicKey = request.publicKey,
+                            responseDate = request.responseDate,
+                            isRequestApproved = true,
+                        )
                     }
                 } else {
                     IllegalStateException("Request not approved.").asFailure()
@@ -299,7 +314,12 @@ class AuthRequestManagerImpl(
             .map { response ->
                 response.authRequests.mapNotNull { request ->
                     getFingerprintPhrase(request.publicKey).getOrNull()?.let { fingerprint ->
-                        request.toAuthRequest(fingerprint = fingerprint)
+                        request.toAuthRequest(
+                            fingerprint = fingerprint,
+                            publicKey = request.publicKey,
+                            responseDate = request.responseDate,
+                            isRequestApproved = request.requestApproved ?: false,
+                        )
                     }
                 }
             }
@@ -329,7 +349,14 @@ class AuthRequestManagerImpl(
                     isApproved = isApproved,
                 )
             }
-            .map { request -> request.toAuthRequest(fingerprint = "") }
+            .map { request ->
+                request.toAuthRequest(
+                    fingerprint = "",
+                    publicKey = request.publicKey,
+                    responseDate = request.responseDate,
+                    isRequestApproved = request.requestApproved ?: false,
+                )
+            }
             .fold(
                 onFailure = { AuthRequestResult.Error(error = it) },
                 onSuccess = { AuthRequestResult.Success(authRequest = it) },
@@ -351,10 +378,13 @@ class AuthRequestManagerImpl(
                 ?.let { pendingAuthRequest ->
                     authRequestsService
                         .getAuthRequest(pendingAuthRequest.requestId)
-                        .map {
+                        .map { request ->
                             NewAuthRequestData(
-                                authRequest = it.toAuthRequest(
+                                authRequest = request.toAuthRequest(
                                     fingerprint = pendingAuthRequest.requestFingerprint,
+                                    publicKey = request.publicKey,
+                                    responseDate = request.responseDate,
+                                    isRequestApproved = request.requestApproved ?: false,
                                 ),
                                 privateKey = pendingAuthRequest.requestPrivateKey,
                                 accessCode = pendingAuthRequest.requestAccessCode,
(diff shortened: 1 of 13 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/AuthRequestManagerImpl.kt
```diff
@@ -183,21 +178,15 @@ class AuthRequestManagerImpl(
...
-                            originUrl = request.originUrl,
-                            responseDate = responseDate,
-                            requestApproved = isRequestApproved,
+                        request
                             // The PublicKey and Fingerprint should be frozen in place to
                             // ensure no funny-business happens between multiple requests.
-                            publicKey = initialAuthRequest.publicKey,
-                            fingerprint = initialAuthRequest.fingerprint,
-                        )
+                            .toAuthRequest(fingerprint = initialAuthRequest.fingerprint)
```
david-livefront: Can we just pass these values into the extension method instead of copying it.

```kotlin
fun AuthRequestsResponseJson.AuthRequest.toAuthRequest(
    fingerprint: String,
    publicKey: String = this.publicKey,
    responseDate: Instant = this.responseDate,
    isRequestApproved: Boolean = this.requestApproved,
): AuthRequest = AuthRequest(
```

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/util/AuthRequestsResponseJsonExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.x8bit.bitwarden.data.auth.manager.model.AuthRequest
+
+/**
+ * Converts the given [AuthRequestsResponseJson.AuthRequest] to an [AuthRequest], given the
+ * [fingerprint] that the response itself does not carry.
+ */
+fun AuthRequestsResponseJson.AuthRequest.toAuthRequest(
+    fingerprint: String,
+): AuthRequest = AuthRequest(
+    id = id,
```
david-livefront: For clarity, can you dd the explicit `this` to all of these.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -228,6 +241,47 @@ class ManageDevicesViewModel @Inject constructor(
         }
     }
 
+    private fun handlePasswordlessAuthRequestReceive(
+        action: ManageDevicesAction.Internal.PasswordlessAuthRequestReceive,
+    ) {
+        // The device list is the only source that reports which device owns a pending request, so
+        // it is re-read before the new request can be rendered against its device.
+        viewModelScope.launch {
+            sendAction(
```
david-livefront: Can we just call a common method from both here and from `handlePasswordlessAuthRequestDevicesReceive`?

david-livefront: Am I missing something or is the `PasswordlessAuthRequestDevicesReceive` action only ever launched from here?

If so, why do we need to actions like this?

david-livefront: OK, I finally go there, the `authRepository.getDevices()` is suspending.

All of this is fine 😄

aj-rosado: I have simplified it a bit by moving `getDevices` into a map on the `getPasswordlessAuthRequestFlow`

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/pendingrequests/PendingRequestsViewModel.kt
```diff
@@ -399,8 +399,4 @@ sealed class PendingRequestsAction {
  * * The request has expired (it is at least 5 minutes old).
  */
 private fun List<AuthRequest>.filterRespondedAndExpired(clock: Clock) =
```
david-livefront: Looks like this function exists in 2 spots. What do you think about consolidating them in the `AuthRequestExtenstions` file?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -72,6 +72,7 @@ class ManageDevicesViewModel @Inject constructor(
     init {
         updateAuthRequestList()
         fetchAllDevices()
+        observePasswordlessAuthRequests()
```
david-livefront: Instead of observing this separately, should `getAuthRequestsWithUpdates` called in `updateAuthRequestList` just observer the push notifications directly?

Is there any reason not to do this?

aj-rosado: `observePasswordlessAuthRequests` Will only get the `authRequest` returned by the push instead of the whole list

app/src/main/kotlin/com/x8bit/bitwarden/data/auth/manager/util/AuthRequestExtensions.kt
```diff
@@ -0,0 +1,23 @@
...
+        responseDate == null &&
+        !creationDate.isOverFiveMinutesOld(clock)
+
+/**
+ * Filters out [AuthRequest]s that match one of the following criteria:
+ * * The request has been approved.
+ * * The request has been declined (indicated by it not being approved & having a responseDate).
+ * * The request has expired (it is at least 5 minutes old).
+ */
+fun List<AuthRequest>.filterRespondedAndExpired(clock: Clock): List<AuthRequest> =
```
david-livefront, line 22: 👍

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
david-livefront: Is there a reason we would not want to live update the `getAuthRequestsWithUpdates` flow with data from a push?

david-livefront (dismissed):

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
aj-rosado: no reason, was simply my choice to try to minimize the network call, although looking at it now we might benefit from getting the whole list as right now it will not update expired requests (or other new requests) and the impact would be minimal

david-livefront: I have approved and you are free to merge but I do agree there would be benefits to merging the push notifications into the other flow. It would make this ViewModel much simplier and improve the speed at which `getAuthRequestsWithUpdates` gets updates, which is good for this screen and other places in the app.

Something to consider.

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
```diff
@@ -193,16 +186,10 @@ class ManageDevicesViewModel @Inject constructor(
...
-                isRefreshing = if (state.devicesLoaded) false else it.isRefreshing,
-            )
-        }
-        if (state.devicesLoaded) {
-            updateContentWithCurrentData()
-        }
+        mutableStateFlow.update { it.copy(authRequests = filteredRequests.toImmutableList()) }
+        // The device list is the only source that reports which device owns a pending request, so
+        // it is re-read before the new list can be rendered against it.
+        fetchAllDevices()
```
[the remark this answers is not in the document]
aj-rosado: @claude We want to get the most recent screen state and also keep the VM simpler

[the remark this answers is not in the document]
david-livefront: Doesn't this mean we will be re-fetching the devices everytime the auth requests emits. Was that part of you're goal?

app/src/test/kotlin/com/x8bit/bitwarden/data/auth/manager/AuthRequestManagerTest.kt
```diff
@@ -1281,6 +1291,39 @@ class AuthRequestManagerTest {
         }
         assertEquals(expected, result)
     }
+
+    @Suppress("MaxLineLength")
```
david-livefront: Is this suppression needed?

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/settings/accountsecurity/managedevices/ManageDevicesViewModel.kt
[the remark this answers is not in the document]
aj-rosado: We need to do that, AuthRequests do not have enough info to display this list, we need the devices data.

We were already doing it by having a single `deviceById` call

david-livefront (approved):

PR #5131, opened 2025-05-06
PM-21255: Implement type-safe navigation

david-livefront:
## 🎟️ Tracking

[PM-21255](https://bitwarden.atlassian.net/browse/PM-21255)

## 📔 Objective

This PR updates the app to use type-safe navigation throughout the app. This removes the stringly types navigation and allows us to more easily pass through data (like passwords and emails) that might have characters that are unsafe for url routes.

There are still some annoyances in how navigation works with screens that have multiple destinations. I have created a [feature request](https://issuetracker.google.com/issues/415834969) to hopefully address this in the future.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-21255]: https://bitwarden.atlassian.net/browse/PM-21255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7281, opened 2026-08-18
[PM-41896] fix: Default the owner dropdown to the collection's organization

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41896

## 📔 Objective

Creating an item while viewing an organization's collection defaulted the Owner dropdown to the personal vault instead of the collection's organization. 

`navigateToVaultAddEdit` was passing the folder ID into the `selectedCollectionId` slot, so the collection ID never reached the add/edit screen. It arrived as null and Owner fell back to the first available owner. Copy/paste slip from #5131. 

## 📸 Screenshots

https://github.com/user-attachments/assets/73855e2c-63c6-477f-9271-7cfab5ed9004


vvolkgang (approved):
:shipit:

PR #7287, opened 2026-08-20
[PM-42198] feat: Send encryptedByKeyId on cipher requests

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42198

Reference: bitwarden/clients#22444 (wiring only — the enrolment migration in
that PR is out of scope here).

## 📔 Objective

The SDK reports the id of the key it used to encrypt a vault item. Sending
that to the server allows the write to be validated against the expected key.

The value comes from the SDK's `EncryptionContext` and should not be
synthesised from app state.

This mirrors the existing `encryptedFor` plumbing, so every write path that
already carries an `EncryptionContext` picks it up with no call-site change:
create, create in organization, update, cipher key migration, share, bulk
share and CXF import. The new parameter is required with no default, so future
callers have to decide explicitly rather than silently omit the field.

### Notes for reviewers

- **Share and bulk share include the id**, matching the SDK's own
  `From<EncryptionContext>` conversions. The web client's `CipherShareRequest`
  drops it, which looks unintentional — worth confirming with the author of
  clients#22444.
- **Nothing is persisted.** `SyncResponseJson.Cipher` and
  `toEncryptedNetworkCipherResponse` are untouched.
- Verified against US dev: the field is sent and writes are accepted.

## 📸 Screenshots

N/A — no UI changes.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/util/VaultSdkCipherExtensions.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/util/VaultSdkCipherExtensions.kt
@@ -43,12 +43,14 @@ import com.bitwarden.vault.UriMatchType
 
 /**
  * Converts a Bitwarden SDK [Cipher] object to a corresponding
- * [SyncResponseJson.Cipher] object.
+ * [CipherJsonRequest] object.
  *
  * @param encryptedFor The ID of the user who this cipher is encrypted by.
+ * @param encryptedByKeyId The ID of the key this cipher is encrypted by.
  */
 fun Cipher.toEncryptedNetworkCipher(
     encryptedFor: String,
+    encryptedByKeyId: String?,
 ): CipherJsonRequest =
     CipherJsonRequest(
         notes = notes,
@@ -76,6 +78,7 @@ fun Cipher.toEncryptedNetworkCipher(
         archivedDate = archivedDate,
         data = data,
         encryptedFor = encryptedFor,
+        encryptedByKeyId = encryptedByKeyId,
     )
 
 /**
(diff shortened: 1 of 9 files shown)


network/src/test/kotlin/com/bitwarden/network/model/CipherWithIdJsonRequestTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    @Test
+    fun `toCipherWithIdJsonRequest should carry over the encryption metadata`() {
+        val request = createMockCipherJsonRequest(number = 1)
+
+        val result = request.toCipherWithIdJsonRequest(id = "mockId-1")
+
+        assertEquals("mockId-1", result.id)
+        assertEquals(request.encryptedFor, result.encryptedFor)
+        assertEquals(request.encryptedByKeyId, result.encryptedByKeyId)
```
david-livefront: Just assert the full result

app/src/test/kotlin/com/x8bit/bitwarden/data/vault/manager/CredentialExchangeImportManagerTest.kt
```diff
@@ -135,6 +135,31 @@ class CredentialExchangeImportManagerTest {
...
+
+            coEvery {
+                vaultSyncManager.syncForResult(forced = true)
+            } returns SyncVaultDataResult.Success(itemsAvailable = true)
+
+            importManager.importCxfPayload(DEFAULT_USER_ID, DEFAULT_PAYLOAD)
+
+            val cipher = capturedRequest.captured.ciphers.first()
+            assertEquals(DEFAULT_CIPHER.encryptedFor, cipher.encryptedFor)
+            assertEquals(DEFAULT_CIPHER.encryptedByKeyId, cipher.encryptedByKeyId)
```
david-livefront: Can we just assert the entire cipher is correct

app/src/main/kotlin/com/x8bit/bitwarden/data/vault/repository/util/VaultSdkCipherExtensions.kt
```diff
@@ -43,12 +43,14 @@ import com.bitwarden.vault.UriMatchType
 
 /**
  * Converts a Bitwarden SDK [Cipher] object to a corresponding
- * [SyncResponseJson.Cipher] object.
+ * [CipherJsonRequest] object.
```
david-livefront, line 46: 👍

david-livefront (approved):

PR #7293, opened 2026-08-24
[PM-42349] fix: use fixed-length password mask

### LINKED ISSUE #7290: [PM-42337] Password / hidden fields revealing actual data length
### Steps To Reproduce

1. Navigate to any saved login or item containing a custom hidden field.  
2. Click on the 'Reveal Data' icon.  

Note that the number of masked symbols matches the actual number of characters in the password or hidden field data. In the updated Web, Browser, and Desktop clients, this behavior differs from that of the Android app, creating an inconsistency and potential security risks.

### Expected Result

The masked password field and the hidden field data appear to be the same length, regardless of the actual number of characters contained within each field.

### Actual Result

The length of the masked symbols matches the actual hidden data, unintentionally revealing the true length of the concealed information.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Build Version

2026.8.0

US

### Self-host Server Version

_No response_

### Environment Details

Samsung Galaxy S26 Plus
Android 16, One UI 8.5

### Issue Tracking Info

488315:
## Tracking

Fixes #7290 (PM-42337)

## Objective

Use a fixed eight-character mask for hidden login passwords and custom hidden fields so their displayed mask no longer reveals the underlying value length.

The shared password component keeps its existing length-based behavior by default. Only the affected login-password and custom-hidden-field call sites opt into the fixed-length mask, so existing behavior for card numbers, security codes, bank fields, identity fields, SSH keys, and other masked values is unchanged.

Regression coverage verifies:
- fixed-length masking for opted-in read-only password fields;
- the real value is shown after reveal;
- default read-only password masking still preserves value length for non-opted-in fields;
- non-interactable login password fields use the fixed mask.

Validation:
- `./gradlew.bat :ui:testDebugUnitTest`
- `./gradlew.bat :ui:lintDebug`
- `./gradlew.bat detekt -Pprecommit=true --rerun-tasks`
- `git diff --check`

All completed checks passed with JDK 21.

## Screenshots

N/A. There is no layout or styling change; affected hidden login/custom values consistently render eight mask characters.


andrebispo5:
Hi @488315 
When the fixed mask is on, `lastTextValue` is still seeded from the raw value while the field renders `displayedValue`. The two disagree, so the first tap on the field makes the `TextField's` `onValueChange` fire with "********" instead of staying quiet.

It's harmless right now since all three opted-in call sites pass `onValueChange = { }`, but `useFixedLengthMask` is public on a shared `:ui` component. The next caller that wires a real callback would get the mask handed to it on first tap, and if that value ever gets saved the password becomes eight asterisks.

Repro (drop into `BitwardenPasswordFieldTest`, fails on this branch with expected:`<[]>` but was:`<[********]>`):

```
@Test
fun `tapping a masked read only field does not emit the mask`() {
    val emitted = mutableListOf<String>()

    setTestContent {
        BitwardenTheme {
            BitwardenPasswordField(
                label = "Password",
                value = "correct horse battery staple",
                showPassword = false,
                showPasswordChange = { },
                onValueChange = { emitted += it },
                readOnly = true,
                useFixedLengthMask = true,
                cardStyle = CardStyle.Full,
            )
        }
    }

    composeTestRule.onNodeWithText("Password").performClick()
    composeTestRule.waitForIdle()

    assertEquals(emptyList<String>(), emitted)
}
```

Keying the state off the displayed value fixes it, and the existing tests still pass:

`var lastTextValue by remember(displayedValue) { mutableStateOf(value = displayedValue) }`

Could you double check this one?

Thanks for the submission 👍

488315:
Thanks for catching that. I double checked it and you were right.

I reproduced it with the regression test you sent first, and it failed with the mask being emitted on the initial tap. I updated `lastTextValue` to key off `displayedValue`, added that regression test, and pushed the fix in `06742ac`.

I also reran the focused `BitwardenPasswordFieldTest` suite (5/5), the full `:ui:testDebugUnitTest` suite, `:ui:lintDebug`, and staged Detekt precommit checks. All are passing on JDK 21.

Appreciate you pointing this one out.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenHiddenPasswordField.kt
```diff
@@ -42,7 +42,7 @@ fun BitwardenHiddenPasswordField(
                 .nullableTestTag(tag = passwordFieldTestTag),
             textStyle = BitwardenTheme.typography.sensitiveInfoSmall,
             label = label?.let { { Text(text = it) } },
-            value = value,
+            value = if (value.isNotEmpty()) MASKED_PASSWORD_VALUE else value,
             onValueChange = { },
             visualTransformation = PasswordVisualTransformation(),
```
david-livefront: Can we make a custom Visual transformation for this?

I think something like this would do the trick:

```kotlin
private class BitwardenPasswordVisualTransformation(
    val mask: Char = '\u2022',
    val staticCharacterCount: Int? = null,
) : VisualTransformation {
    override fun filter(
        text: AnnotatedString,
    ): TransformedText = TransformedText(
        mask.toString().repeat(n = staticCharacterCount ?: text.text.length).toAnnotatedString(),
        object : OffsetMapping {
            override fun originalToTransformed(
                offset: Int,
            ): Int = staticCharacterCount?.let { if (offset > it) it else offset } ?: offset

            override fun transformedToOriginal(
                offset: Int,
            ): Int = if (offset > text.length) text.length else offset
        },
    )
}
```

Make sure it has a way to access it too:
```kotlin
@Composable
fun passwordVisualTransformation(
    mask: Char = '\u2022',
    staticCharacterCount: Int? = null,
): VisualTransformation =
    remember(mask, staticCharacterCount) {
        BitwardenPasswordVisualTransformation(
            mask = mask,
            staticCharacterCount = staticCharacterCount,
        )
    }
```

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenPasswordField.kt
david-livefront: The same visual transformation could be applied here:

```kotlin
!showPassword -> {
    BitwardenPasswordVisualTransformation(staticCharacterCount = 8.takeIf { readOnly })
}
```

We'd want this applied to any readonly password fields. The editable fields ones get the normal treatment.

david-livefront (changes requested):

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenHiddenPasswordField.kt
488315: Yep, added passwordVisualTransformation() backed by a private BitwardenPasswordVisualTransformation. BitwardenHiddenPasswordField now keeps the real value and uses an 8-character static mask through the transformation. The focused UI tests pass locally.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenPasswordField.kt
488315: Updated this to use the same transformation automatically whenever readOnly is true, while editable hidden fields keep the normal value-length mask. I removed the useFixedLengthMask parameter and the call-site flags. The focused UI tests pass locally.

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenPasswordField.kt
```diff
@@ -190,7 +190,9 @@ fun BitwardenPasswordField(
                     }
                 },
                 visualTransformation = when {
-                    !showPassword -> PasswordVisualTransformation()
+                    !showPassword -> passwordVisualTransformation(
+                        staticCharacterCount = 8.takeIf { readOnly },
+                    )
                     readOnly -> nonLetterColorVisualTransformation()
```
david-livefront, line 197: Wanna run the formatter here, there should be a newline after a multiline case

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemScreenTest.kt
```diff
@@ -1180,7 +1184,7 @@ class VaultItemScreenTest : BitwardenComposeTest() {
                 viewState = EMPTY_LOGIN_VIEW_STATE.copy(
                     type = EMPTY_LOGIN_TYPE.copy(
                         passwordData = VaultItemState.ViewState.Content.ItemType.Login.PasswordData(
-                            password = "p@ssw0rd",
+                            password = "correct horse battery staple",
```
david-livefront: You have some failing tests

ui/src/main/kotlin/com/bitwarden/ui/platform/components/field/BitwardenPasswordField.kt
488315: Ran the Android Studio formatter with docs/bitwarden-style.xml. It added the blank line after the multiline case in 57d6252.

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemScreenTest.kt
488315: I found the login test data changes were not needed, so I restored them to p@ssw0rd in 57d6252. The fixed-length case is still covered by the hidden custom-field assertion using a value longer than eight characters. The six focused UI tests pass locally. I could not run VaultItemScreenTest here because GitHub Packages returns 401 for the private SDK with my current token.

app/src/test/kotlin/com/x8bit/bitwarden/ui/vault/feature/item/VaultItemScreenTest.kt
```diff
@@ -972,6 +972,10 @@ class VaultItemScreenTest : BitwardenComposeTest() {
                     )
                 }
 
+                composeTestRule
+                    .onNodeWithTextAfterScroll(textField.name)
+                    .assertTextEquals(textField.name, "••••••••")
+
                 composeTestRule
                     .onNodeWithTextAfterScroll(textField.name)
                     .onChildren()
```
david-livefront, line 981: You still have some failing tests in this file

david-livefront (approved):

david-livefront:
Thanks for all the hard work @488315

488315:
> Thanks for all the hard work @488315 

No problem!

PR #7299, opened 2026-08-25
[PM-36378] fix: Show the new folder created snackbar

andrebispo5:
## 🎟️ Tracking

## 📔 Objective

Creating a folder gave you no confirmation. Two paths were broken:

- **Folder selector on the add/edit item screen**: the result handler set the folder and cleared the dialog, but never sent any feedback. Nothing showed up.
- **Vault folder list**: the folder add screen does publish the snackbar, but the item listing was not subscribed to the relay, so it landed on the vault screen sitting behind it.

Both now show "New folder created." The string already existed, so nothing new to translate.

Two things for review:

- The error path is untouched on purpose. Failure still shows nothing and still clears the selected folder. Worth its own ticket.
- The relay goes to whichever consumer subscribed last, ordered by ViewModel creation rather than what is on screen. Settings > Folders, then the vault folder list, then back to Settings > Folders can still send the snackbar to the wrong place. Narrow path, and the worst case is a missing snackbar like today. A real fix means passing a relay key through the folder add/edit route.

## 📸 Screenshots

https://github.com/user-attachments/assets/4c4745b5-36a0-46d4-aa41-fe23833e188f

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/VaultAddEditViewModel.kt
@@ -1988,11 +1988,16 @@ class VaultAddEditViewModel @Inject constructor(
                 dialog = null,
             )
         }
+        val createdFolder = (action.result as? CreateFolderResult.Success)?.folderView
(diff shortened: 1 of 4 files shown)


david-livefront (approved):

andrebispo5:
@david-livefront Thanks 🎉

PR #7197, opened 2026-07-22
[PM-40773] bug: Fix accessibility detection on Android 16+

7heMech:
## 🎟️ Tracking

Resolves #6267
Resolves #5278 (not sure about this one, but it solves the quick tile part maybe)
Resolves #4601 

## 📔 Objective

Fixes an issue where Bitwarden doesn't detect accessibility is on, Android 16+ in my experience.

## 📸 Screenshots
Screenshot_20260722-211304
Screenshot_20260722-211312
Screenshot_20260722-211235

Issue #6267, opened 2025-12-15
[PM-29779] Tapping on the Autofill Quick Tile on Origin OS 6 (Android 16) does not launch Bitwarden.


SergeantConfused:
### Steps To Reproduce

1. Install Bitwarden and activate the [Autofill Quick Tile](https://bitwarden.com/help/auto-fill-android/#quick-action-tiles).
2. Open Chrome and navigate to a website.
3. Pull down the notification shade and tap on the Autofill Quick Tile.

### Expected Result

Bitwarden should launch and show vault items with a matching URI to the webpage you're at.

### Actual Result

Nothing happens; Bitwarden does not launch.

### Screenshots or Videos

There is a video captured internally.

### Additional Context

This may have started with the update to Origin OS 6.

### Build Version

2025.11.1

### What server are you connecting to?

US

### Self-host Server Version

N/A

### Environment Details

Origin OS 6; Based on Android 16.

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

akmalabiyoga:
Can confirm that this also happened on my iQOO 13 that recently upgraded to Origin OS 6.

Really hope it will be fixed soon since its an integral part of bitwarden experience on android

SaintPatrck:
From the behavior you're describing, it sounds like the Accessibility service is unable to detect fillable fields. Out of curiosity, do you happen to see a Toast message indicating no fields detected when you tap the quick tile? If not, can you make sure Bitwarden is granted Notification permissions? It's a bit odd, but Notification permission also controls our ability to show Toasts when not in the foreground.

akmalabiyoga:
> From the behavior you're describing, it sounds like the Accessibility service is unable to detect fillable fields. Out of curiosity, do you happen to see a Toast message indicating no fields detected when you tap the quick tile? If not, can you make sure Bitwarden is granted Notification permissions? It's a bit odd, but Notification permission also controls our ability to show Toasts when not in the foreground.

There are no toast message and my bitwarden app already granted full permission including notifications.

To note, bitwarden normal autofill is still working (most of the time), it just the quick tile doesnt work completely, which usually the alternatives when normal autofill doesnt work.

chilled-penguin:
[...]

Issue #4601, opened 2025-01-21
"Use accessibility" no longer detects when Accessibility service is turned ON for Bitwarden


WilliamKimball0:
### Steps To Reproduce

Bug appeared when the app was updated to the new design version.
 
3. Go to 'Settings -> Autofill'
4. Click on 'Use accessibility' (when turned OFF)
5. App navigates to the Android native Accessibility service setting view. Regardless of the intial state of Accessibility service for Bitwarden, and the state of it on user action (turning it on), the "Use accessibility" will not turn ON in the app.

### Expected Result

The app detects the accessibility service to be set ON for Bitwarden in Android setting.

### Actual Result

The app doesn't detect the accessibility service state and the "Use accessibility" stays OFF no matter what the user does.

### Additional Context

Tried re-installing the app, with no change to the issue.

### Build Version

2025.1.0

### What server are you connecting to?

EU

### Environment Details

- Device: Google Pixel 8 Pro
- OS: Graphene OS (2025011500)

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

daniellbw:
Hi there,

Thank you for your report! 

I have managed to spot these two issues(https://github.com/bitwarden/android/issues/4586) and (https://github.com/bitwarden/android/issues/4224) where users are experiencing similar behaviour.

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.

Thanks once again!

WilliamKimball0:
> Hi there,
> 
> Thank you for your report!
> 
> I have managed to spot these two issues([#4586](https://github.com/bitwarden/android/issues/4586)) and ([#4224](https://github.com/bitwarden/android/issues/4224)) where users are experiencing similar behaviour.
> 
> If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.
> 
> Thanks once again!

I added a video recording where it's shown how the "use accessibility" setting is disabled despite the service being turned ON for Bitwarden and then explicitly turning it off and on again with no change.

GeekSquirrel:
same issue here on both my Android 11 and 15 phones.

TurboHsu:
Same issue on Samsung S24 Ultra running OneUI 6.1 (Android 14)

© Bitwarden Inc. 2015-2025

[...]

PR #7300, opened 2026-08-25
[PM-40773] bug: Track accessibility service connection state from the service

7heMech:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40773

New PR under the same ticket, as requested in
https://github.com/bitwarden/android/pull/7197#issuecomment-5411934834.

#7197 shipped in 2026.8.0 but the reported behaviour is unchanged, so #4601 and #6267 are
still reproducible on 2026.8.0 (21819).

## 📔 Objective

`isAccessibilityServiceEnabled` returns false on Android 16 even when our accessibility service
is enabled and bound, so "Use accessibility" stays off and the autofill tile shows the "turn on
accessibility" dialog.

Both branches of that check ask the platform about our own service, and Android 16 does not
answer. Logged from inside `com.x8bit.bitwarden` on a Pixel 6 Pro (SDK 36, `targetSdk` 37),
while `dumpsys accessibility` listed the service under both `Enabled services` and
`Bound services`:

```
packageName=com.x8bit.bitwarden
enabledServices.size=0    // AccessibilityManager.getEnabledAccessibilityServiceList
settingsValue=            // Settings.Secure.ENABLED_ACCESSIBILITY_SERVICES, empty string
RESULT=false
```

At the same moment `adb shell settings get secure enabled_accessibility_services` returned the
full value, and a separate test app also targeting 37 got both services back from
`getEnabledAccessibilityServiceList`. The restriction applies to the app asking about itself.

That is why #7197 did not help. It widened which service names the primary check accepts, but
`getEnabledAccessibilityServiceList` is restricted the same way `Settings.Secure` is, so the
check iterates an empty list no matter which names it would have matched.

So stop asking the platform. `BitwardenAccessibilityService` knows when it is connected, and
`AccessibilityEnabledManager` was already notified from its lifecycle callbacks:

- `refreshAccessibilityEnabledFromSettings()` becomes `var isAccessibilityServiceConnected`, set
  in `onServiceConnected` and `onUnbind`. `onCreate` no longer refreshes, since
  `onServiceConnected` always follows it.
- `BitwardenAutofillTileService` reads the manager instead of calling
  `applicationContext.isAccessibilityServiceEnabled` on each tap. Without this the tile keeps
  showing the dialog.
- `isAccessibilityServiceEnabled` is unchanged and still provides the initial value, so
  behaviour before Android 16 is the same.

Verified on the same device with a minified `standardRelease` build at `applicationId`
`com.x8bit.bitwarden`, since the previous fix passed an unminified local test and did nothing in
the Play build.

## 📸 Screenshots

To follow.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/BitwardenAccessibilityService.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/data/autofill/accessibility/BitwardenAccessibilityService.kt
@@ -32,19 +32,13 @@ class BitwardenAccessibilityService : AccessibilityService() {
 
     override fun onInterrupt() = Unit
 
-    override fun onCreate() {
-        super.onCreate()
-        accessibilityEnabledManager.refreshAccessibilityEnabledFromSettings()
-    }
-
     override fun onUnbind(intent: Intent?): Boolean {
-        return super
-            .onUnbind(intent)
-            .also { accessibilityEnabledManager.refreshAccessibilityEnabledFromSettings() }
+        accessibilityEnabledManager.isAccessibilityServiceConnected = false
+        return super.onUnbind(intent)
     }
 
     override fun onServiceConnected() {
         super.onServiceConnected()
-        accessibilityEnabledManager.refreshAccessibilityEnabledFromSettings()
+        accessibilityEnabledManager.isAccessibilityServiceConnected = true
     }
 }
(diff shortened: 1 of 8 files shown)


7heMech:
Superseded by a new PR against the same ticket (PM-40773), per https://github.com/bitwarden/android/pull/7197#issuecomment-5412037062. Note the bot opened PM-42455 for this one, which can be closed as a duplicate.

PR #7301, opened 2026-08-25
[PM-36231] fix: Show Select without dashes on vault add edit menus

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36231

## 📔 Objective

- The Brand dropdown showed `-- Select --` as its default, but the designs just ask for `Select`.
- `SELECT_TEXT` is shared, so this also cleans up card expiration month, bank account type, and identity title. Fixing only Brand would have left the app disagreeing with itself.
- Safe to change: nothing keys off the dashes, and the placeholder never reaches the SDK or the API.

## 📸 Screenshots
| Before | After |
|--------|-------|
|  |  |

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/util/VaultAddEditExtensions.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/util/VaultAddEditExtensions.kt
@@ -3,7 +3,6 @@ package com.x8bit.bitwarden.ui.vault.feature.addedit.util
 import com.bitwarden.ui.platform.resource.BitwardenString
 import com.bitwarden.ui.util.Text
 import com.bitwarden.ui.util.asText
-import com.bitwarden.ui.util.concat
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/x8bit/bitwarden/ui/vault/feature/addedit/util/VaultAddEditExtensions.kt
```diff
@@ -1,21 +1,8 @@
...
 import com.x8bit.bitwarden.ui.vault.model.VaultItemCipherType
 
-/**
- * Default, "select" Text to show on multi select buttons in the VaultAddEdit package.
- */
-val SELECT_TEXT: Text
-    get() = "-- "
-        .asText()
-        .concat(BitwardenString.select.asText())
-        .concat(" --".asText())
```
david-livefront, line 17: I am not sad to see this go 😄

david-livefront (approved):

PR #7337, opened 2026-08-31
[PM-41836] fix: Promote Plan screen off Free view without requiring an in-flight checkout

aj-rosado:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41836

## 📔 Objective

The Plan screen only promoted a premium account off the Free-upgrade view while `isAwaitingPremiumStatus` was true (immediately after a Stripe checkout round-trip), and `handleSubscriptionStatusUpdateReceive` only reacted to trouble-status subscription updates, never a healthy `ACTIVE` one. A premium account reaching the screen outside that narrow window — via a push notification, a background sync, or simply reopening the screen later — stayed stuck on the Free-upgrade UI despite already being premium.

Both signal handlers (`handleUserStateUpdateReceive`, `handleSubscriptionStatusUpdateReceive`) now promote unconditionally whenever the account is premium. The upgrade-celebration event (`NavigateToUpgradedToPremium`) still only fires when a checkout was genuinely in flight (`isAwaitingPremiumStatus`), so a stale trouble-status update arriving mid-checkout can't fire a false celebration.

**Out of scope, tracked separately:** the `isPremium` vs. org-granted-premium (`isPremiumFromSelf`) distinction, and a client-side wiring gap in the Stripe checkout callback (a missing `AndroidManifest.xml` intent-filter affecting devices without Auth Tab support) — both real findings from this investigation, but independent of this fix and better reviewed on their own.

**Review note:** a local review caught that an early version of this change promoted unconditionally on every `UserStateUpdateReceive`, including for non-premium accounts — a regression, since fixed and verified against the full `PlanViewModelTest` suite.

### DIFF
--- a/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanViewModel.kt
+++ b/app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanViewModel.kt
@@ -605,12 +605,12 @@ class PlanViewModel @Inject constructor(
     // Fires the celebration event only when isConfirmedPremium and a checkout was actually
     // in flight; otherwise this is a silent state recovery.
     private fun promoteFreeCloudToPremiumView(isConfirmedPremium: Boolean) {
-        val isAwaitingPremiumStatus = (state.viewState as? PlanState.ViewState.Content.Free.Cloud)
-            ?.isAwaitingPremiumStatus == true
-        if (isConfirmedPremium && isAwaitingPremiumStatus) {
-            onPremiumUpgradeSuccess()
-        } else {
-            promoteToPremiumView()
+        onFreeCloudContent { freeState ->
+            if (isConfirmedPremium && freeState.isAwaitingPremiumStatus) {
+                onPremiumUpgradeSuccess()
+            } else {
+                promoteToPremiumView()
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/test/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanViewModelTest.kt
```diff
@@ -620,40 +635,48 @@ class PlanViewModelTest : BaseViewModelTest() {
...
+                assertEquals(
+                    DEFAULT_PREMIUM_ACTIVE_VIEW_STATE.copy(
+                        status = PremiumSubscriptionStatus.CANCELED,
+                        canceledDateText = "April 21, 2026",
+                        showCancelButton = false,
                     ),
+                    loadedState.viewState,
                 )
-                expectNoEvents()
+                assertEquals(null, loadedState.dialogState)
```
david-livefront: Can we use `assertNull`

david-livefront: Actually, can we assert the complete state

david-livefront (dismissed):

app/src/main/kotlin/com/x8bit/bitwarden/ui/platform/feature/premium/plan/PlanViewModel.kt
```diff
@@ -615,7 +602,19 @@ class PlanViewModel @Inject constructor(
         }
     }
 
-    private fun onPremiumUpgradeSuccess() {
+    // Fires the celebration event only when isConfirmedPremium and a checkout was actually
+    // in flight; otherwise this is a silent state recovery.
+    private fun promoteFreeCloudToPremiumView(isConfirmedPremium: Boolean) {
+        val isAwaitingPremiumStatus = (state.viewState as? PlanState.ViewState.Content.Free.Cloud)
```
david-livefront: Can we make this cleaner?

```kotlin
onFreeCloudContent { freeState ->
    if (isConfirmedPremium && freeState.isAwaitingPremiumStatus) {
        onPremiumUpgradeSuccess()
    } else {
        promoteToPremiumView()
    }
}
```

david-livefront (approved):

PR #7341, opened 2026-09-01
[PM-40535] fix: Organization policy restriction notice padding and icon

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40535

## 📔 Objective

QA sent this back for two design mismatches on the "Organization policy
restriction" notice on text Sends.

- Text ran to the edge of the card. `BitwardenActionCard` only got its trailing
  padding from the dismiss button, and this card has none, so it now adds the
  16dp inset itself.
- The info icon was near-black at 16dp instead of the blue 24dp in the designs.

The padding fix is in the shared card, so it also covers three other cards with
no dismiss button.

## 📸 Screenshots
| Before | After |
|--------|-------|
|  |  |
|  |  |


david-livefront (approved):

andrebispo5:
@david-livefront

PR #7345, opened 2026-09-02
[PM-36378] fix: Remove "New folder created" snackbar from add/edit cipher flow

andrebispo5:
## 🎟️ Tracking

## 📔 Objective

New requirement change. The "New folder created" snackbar should only show when you create a folder from the vault "+" FAB or from Settings > Vault > Folders. It shouldn't show when you create one inline while adding or editing an item, and right now it does.

- Dropped the snackbar event from `VaultAddEditViewModel.handleAddFolderResult`
- The folder is still created and auto-selected in the dropdown, nothing else changes
- `FolderAddEditViewModel` is untouched, so the FAB and settings paths keep their snackbar
- Updated the matching unit test to expect no event


aj-rosado (approved):
LGTM

PR #7349, opened 2026-09-02
chore: Rollback setpassword API changes

david-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

This PR switches the `SetPasswordRequestJson` model to use the original payload structure because the server updates are not fully ready for use. I have managed to maintain the same constructor signature to make the change very simple,


vvolkgang (approved):

david-livefront:
Thanks @vvolkgang

PR #7353, opened 2026-09-03
[PM-36376] fix: Clear folder and owner dropdown lag after a card scan

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36376

## 📔 Objective

- Scanning a card auto-focuses the cardholder name field, so the keyboard is up. Tapping
  the folder dropdown then made the sheet stutter on its way in.
- The folder and owner selectors now clear focus first, like the "Add field" button in
  this screen already does.
- Added two regression tests. Both fail without the fix.


david-livefront (approved):

andrebispo5:
@david-livefront Thanks!

